PR #623, opened 2015-11-09
remove v1 persistence code

robfletcher:
This removes support for the old Redis schema


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Orchestration.groovy
```diff
@@ -20,7 +20,5 @@ import groovy.transform.CompileStatic
 
 @CompileStatic
 class Orchestration extends Execution<Orchestration> {
-  static final int CURRENT_VERSION = 2
```
ajordens: It might not hurt to keep the `CURRENT_VERSION` around for the next time a migration is necessary? Any harm in that?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/Migrator.groovy
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.context.ApplicationListener
+import org.springframework.context.event.ContextRefreshedEvent
+import org.springframework.stereotype.Component
+import redis.clients.jedis.Jedis
+import redis.clients.util.Pool
+import rx.Observable
+
+@Component
+@CompileStatic
+class Migrator implements ApplicationListener<ContextRefreshedEvent> {
```
ajordens: This looks old, but possible adaptable to a point-in-time migration?

Given the OSS ramifications, we could perhaps get this done quickly and not worry about releasing a migrator.

Alternatively, we have something that runs on startup or something and attempts to perform a migration, and it lives in the repo for some period of time.

ajordens:
I think getting a migration story together is important.

I'd be cool with dropping the version provided the migration removes the key (effectively resetting everything back to 0).

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Orchestration.groovy
robfletcher: Yeah, I was kinda undecided. I'll put it back.

robfletcher:
Agreed. I'd like to remove the key and ideally rename `executionStatus` -> `status`, `executionStartTime` -> `startTime` etc. so they're consistent with stages.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/Migrator.groovy
robfletcher: Yeah, I don't know why that class is in there but it's exactly the way I've been doing migrations

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.groovy
```diff
@@ -92,45 +70,7 @@ abstract class Execution<T> implements Serializable {
       return CANCELED
```
robfletcher: Actually I don't think this is even necessary

robfletcher:
Luckily I thought ahead and there's no data migration to do for changing the names of those properties – they already bind to `status`, `startTime` and `endTime` in the Redis schema.

robfletcher:
Merged with squash on master in d69e410d29a5077bc9240b001b0a9f01d706bf20

PR #645, opened 2015-11-16
Stops inconsistent failures between parallel pipeline branches

robfletcher:
If a branch fails the other branch will now get cancelled which means any cancellation handling logic will kick in. It n longer sets the "terminate only" flag on the job as that will cause the inconsistency previously seen where the job execution listeners will fire immediately if the failing branch is the primary one and not until everything finishes if it's another branch. That meant if it was the "primary" branch that failed the pipeline status would immediately become TERMINAL even though things were still running.

 I also had to make TERMINAL status override CANCELED as we don't want the pipeline to show as CANCELED due to the non-TERMINAL branches.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
@@ -73,19 +73,23 @@ class ExecutionPropagationListener extends JobExecutionListenerSupport implement
     def id = executionId(jobExecution)
 
     def orcaTaskStatus
-    if (executionRepository.isCanceled(id)) {
-      orcaTaskStatus = CANCELED
-    } else if (jobExecution.failureExceptions) {
+    if (jobExecution.failureExceptions) {
       orcaTaskStatus = TERMINAL
     } else {
       def stepExecutions = new ArrayList<StepExecution>(jobExecution.stepExecutions).sort { it.lastUpdated }.reverse()
       def stepExecution = stepExecutions.find { it.status == jobExecution.status } ?: stepExecutions[0]
-      orcaTaskStatus = stepExecution?.executionContext?.get("orcaTaskStatus") as ExecutionStatus ?: TERMINAL
+      orcaTaskStatus = stepExecution?.executionContext?.get("orcaTaskStatus") as ExecutionStatus
+    }
+
+    if (executionRepository.isCanceled(id) && orcaTaskStatus != TERMINAL) {
+      orcaTaskStatus = CANCELED
     }
 
     if (orcaTaskStatus == STOPPED) {
       orcaTaskStatus = SUCCEEDED
     }
+
+    if (!orcaTaskStatus) orcaTaskStatus = TERMINAL
 
     executionRepository.updateStatus(id, orcaTaskStatus)
 
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
@@ -89,7 +89,9 @@ class ExecutionPropagationListener extends JobExecutionListenerSupport implement
       orcaTaskStatus = SUCCEEDED
     }
 
-    if (!orcaTaskStatus) orcaTaskStatus = TERMINAL
+    if (!orcaTaskStatus) {
+      orcaTaskStatus = TERMINAL
+    }
 
     executionRepository.updateStatus(id, orcaTaskStatus)
 
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionNotFoundExceptio
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
+import org.springframework.batch.core.ExitStatus
 import org.springframework.batch.core.JobExecution
 import org.springframework.batch.core.StepExecution
 import org.springframework.batch.core.listener.JobExecutionListenerSupport
@@ -83,6 +84,7 @@ class ExecutionPropagationListener extends JobExecutionListenerSupport implement
 
     if (executionRepository.isCanceled(id) && orcaTaskStatus != TERMINAL) {
       orcaTaskStatus = CANCELED
+      jobExecution.exitStatus = ExitStatus.STOPPED
     }
 
     if (orcaTaskStatus == STOPPED) {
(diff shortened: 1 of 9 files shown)


robfletcher:
Associated Jira issue SPIN-1117

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
```diff
@@ -155,7 +157,7 @@ class TaskTasklet implements Tasklet {
 
   private static void setStopStatus(ChunkContext chunkContext, ExitStatus exitStatus, ExecutionStatus executionStatus) {
     chunkContext.stepContext.stepExecution.with {
-      setTerminateOnly()
+//      setTerminateOnly()
```
ajordens: Should just remove this commented out bit

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
```diff
@@ -99,6 +100,7 @@ class TaskTasklet implements Tasklet {
 
         if (result.status == ExecutionStatus.TERMINAL) {
           setStopStatus(chunkContext, ExitStatus.FAILED, result.status)
+          executionRepository.cancel(stage.execution.id)
```
ajordens: Probably worth a comment as to why we're cancelling

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
```diff
@@ -70,6 +70,7 @@ class TaskTasklet implements Tasklet {
     try {
       if (stage.execution.canceled) {
         setStopStatus(chunkContext, ExitStatus.STOPPED, ExecutionStatus.CANCELED)
+        contribution.exitStatus = ExitStatus.STOPPED
```
ajordens: What's this actually do behind the scenes?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
```diff
@@ -73,20 +73,24 @@ class ExecutionPropagationListener extends JobExecutionListenerSupport implement
...
+
+    if (executionRepository.isCanceled(id) && orcaTaskStatus != TERMINAL) {
+      orcaTaskStatus = CANCELED
     }
 
     if (orcaTaskStatus == STOPPED) {
       orcaTaskStatus = SUCCEEDED
     }
 
+    if (!orcaTaskStatus) orcaTaskStatus = TERMINAL
```
ajordens: Why you cram this all on one line and nothing else?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.groovy
```diff
@@ -88,14 +88,14 @@ abstract class Execution<T> implements Serializable {
...
   ExecutionStatus getStatus() {
-    if (canceled) {
-      return CANCELED
-    }
-
     if (version > 1) {
       return executionStatus
     }
 
+    if (canceled) {
```
ajordens: Hmmm ya. 

Makes sense given that we can now end up with pipelines that are technically "canceled = true" but TERMINAL.

You have another PR pending with the schema changes, right? Just pending a migration.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
```diff
@@ -73,20 +73,24 @@ class ExecutionPropagationListener extends JobExecutionListenerSupport implement
...
+    if (jobExecution.failureExceptions) {
       orcaTaskStatus = TERMINAL
     } else {
       def stepExecutions = new ArrayList<StepExecution>(jobExecution.stepExecutions).sort { it.lastUpdated }.reverse()
       def stepExecution = stepExecutions.find { it.status == jobExecution.status } ?: stepExecutions[0]
-      orcaTaskStatus = stepExecution?.executionContext?.get("orcaTaskStatus") as ExecutionStatus ?: TERMINAL
+      orcaTaskStatus = stepExecution?.executionContext?.get("orcaTaskStatus") as ExecutionStatus
+    }
+
+    if (executionRepository.isCanceled(id) && orcaTaskStatus != TERMINAL) {
```
ajordens: Probably worth adapting one of the `ExecutionPropagationListenerSpec` tests to handle the particular scenario of canceled && TERMINAL == TERMINAL.

Looks like a straight forward test to add.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTaskletSpec.groovy
```diff
@@ -108,7 +107,7 @@ class TaskTaskletSpec extends Specification {
 
     then:
     0 * task.execute(_)
-    chunkContext.stepContext.stepExecution.terminateOnly == taskStatus.halt
+//    chunkContext.stepContext.stepExecution.terminateOnly == taskStatus.halt
```
ajordens: Can remove this now I imagine

ajordens:
Few comments but otherwise looks reasonable to me.

Thanks for chasing all this down!

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
robfletcher: my bad, yeah. I'll do that

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
robfletcher: It's an alternative to doing the `setTerminateOnly()` thing – indicates that the Job has stopped because of an interruption but doesn't cause the problems with early / never firing listeners.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.groovy
robfletcher: Correct.

Yes, without this chance setting the `canceled` flag always causes the status to be `CANCELED`. Now we're using the flag to cause any cancellation routines to run. This logic is going away soon as you point out.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTaskletSpec.groovy
robfletcher: Damn, what's with me leaving commented stuff in the commit that day? Sorry.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
robfletcher: Done https://github.com/spinnaker/orca/blob/5b1c0624122b7c6a2965eccfac3f826c1cb981f7/orca-core/src/test/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListenerSpec.groovy#L98-L121

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
```diff
@@ -99,6 +100,9 @@ class TaskTasklet implements Tasklet {
 
         if (result.status == ExecutionStatus.TERMINAL) {
           setStopStatus(chunkContext, ExitStatus.FAILED, result.status)
+          // cancel execution which will halt any parallel branches and run
+          // cancellation routines hopefully leaving clusters in a good state
```
robfletcher: Fixed

PR #658, opened 2015-11-23
GCE-specific find image tweaks

ttomsu:
1.) Add more helpful error message when image is not found

emdmukgrfjt

2.) GCE is not limited to regional images, so we now fallback to non-matching zone iff 1 deploymentDetail entry is found (this allows cross-zonal "Clone" operations).
  a.) Fail with helpful message if this is the case:

mclozzpaytj

3.) Fail with graceful error message if no image is found. Previously, this bubbled up a cryptic Clouddriver error. 
cvn6rwig5k5
  

@duftler, @spinnaker/google-reviewers PTAL

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/FindImageFromClusterTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/FindImageFromClusterTask.groovy
@@ -102,7 +102,7 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
         return [(location): lookupResults]
       } catch (RetrofitError e) {
         if (e.response.status == 404) {
-          def message = "Could not find cluster '$cluster' in account '$account'"
+          def message = "Could not find cluster '$cluster' for '$account' in '$location.value'"
           try {
             Map reason = objectMapper.readValue(e.response.body.in(), new TypeReference<Map<String, Object>>() {})
             if (reason.error.contains("target.fail.strategy")){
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/FindImageFromClusterTask.groovy
```diff
@@ -102,7 +102,7 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
         return [(location): lookupResults]
       } catch (RetrofitError e) {
         if (e.response.status == 404) {
-          def message = "Could not find cluster '$cluster' in account '$account'"
+          def message = "Could not find cluster '$cluster' for '$account' in '$location.value'"
```
duftler: Missing terminating period.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/gce/GoogleServerGroupCreator.groovy
```diff
@@ -55,6 +55,21 @@ class GoogleServerGroupCreator implements ServerGroupCreator {
...
+                                              "'${stage.context.zone}'. Images found from cluster in " +
+                                              "${deploymentDetails.collect{it.zone}.join(",") } - " +
+                                              "only 1 should be available.")
+        }
+        operation.image = deploymentDetails[0].imageId
+      }
+    }
+
+    if (!operation.image) {
+      throw new IllegalStateException("No image could be found in ${stage.context.zone}")
```
duftler: Missing terminating period.

duftler:
LGTM after addressing comments.

PR #661, opened 2015-11-24
Improve Cloud Foundry "find next sequence" in server group operation

gregturn:
* Also adds support for cache refresh on deployment


gregturn:
Related to https://github.com/spinnaker/clouddriver/pull/184

PR #664, opened 2015-11-25
Verify that affected server groups still exist prior to an instance c…

ajordens:
…heck task completing

- Addresses a situation where a task would incorrectly succeed if the ASG was destroyed by an external process (... and the pipeline would unfortunately continue)


ajordens:
PTAL @spinnaker/netflix-reviewers @spinnaker/google-reviewers 

To replicate the original issue, perform a red/black deployment and destroy the newly created server group while it's still waiting on health. 

In such a situation, there's a _very_ good chance that the `waitForUpInstances` task will succeed and the pipeline will continue to disable all other server groups... effectively leaving the cluster with no enabled server groups. Highlander is even worse, you'll be left without any server groups at all.

I've done some testing with red/black + resize + enable + disable in AWS and everything still seems to hang together.

--

@duftler @ttomsu - I'm using the `OortHelper` to fetch a server group as it looked to handle the differences in region/zone between AWS and GCP.

PR #665, opened 2015-11-26
Check for error messages if the bakery returns a 404

ajordens:


ajordens:
Attn @spinnaker/google-reviewers, I'm not sure if there implications to `rosco` but in talking with @coryb we learned that our bakery will return a legitimate error response in the event of a 404 on `CreateBake`.

With this change, we'll now look for a proper error message and re-throw if found. If not, the original retry behavior is preserved.

The error message will be displayed in `deck`.

duftler:
Great improvement.
As-is, no negative implications for rosco (as if reachable, it wouldn't have returned 404 right away). It was up to the poller in both rosco and rush to detect/propagate errors.
Have updated rosco and rush to have fail-fast behavior (https://github.com/spinnaker/rosco/pull/54, https://github.com/spinnaker/rush/pull/36).
Have tested up through deck to confirm the behavior.
Fastest way to verify behavior is to have `configDir` refer to a non-existent directory.

duftler:
image

PR #669, opened 2015-12-01
Update Igor endpoints

AMeng:
Fixes: https://github.com/spinnaker/spinnaker/issues/605
Depends on: https://github.com/spinnaker/igor/pull/42

### DIFF
--- a/orca-core/src/test/groovy/com/netflix/spinnaker/orca/batch/pipeline/util/PackageInfoSpec.groovy
+++ b/orca-core/src/test/groovy/com/netflix/spinnaker/orca/batch/pipeline/util/PackageInfoSpec.groovy
@@ -26,7 +26,7 @@ import spock.lang.Unroll
 class PackageInfoSpec extends Specification {
 
   @Subject
-  PackageInfo info = new PackageInfo(Mock(Stage), 'deb', '_', false, true, Mock(ObjectMapper))
+  PackageInfo info = new PackageInfo(Mock(Stage), 'deb', '_', true, true, Mock(ObjectMapper))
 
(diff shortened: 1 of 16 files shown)


tomaslin:
@duftler - mind taking a look at the packageinfo changes to see if they make sense?

duftler:
The packageInfo changes LGTM.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/batch/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -26,7 +26,7 @@ import spock.lang.Unroll
 class PackageInfoSpec extends Specification {
 
   @Subject
-  PackageInfo info = new PackageInfo(Mock(Stage), 'deb', '_', false, true, Mock(ObjectMapper))
+  PackageInfo info = new PackageInfo(Mock(Stage), 'deb', '_', true, true, Mock(ObjectMapper))
```
duftler: Is the change to that constructor parameter required for these tests to pass?

AMeng: No, but this forces the `parseBuildInfoUrl` method to be called (the one I've edited). Previously that method was not tested through here.

duftler: Ok. Just wanted to be sure I hadn't missed something (since you've added an explicit test of that method here).

PR #677, opened 2015-12-04
enables more stages for strategies

tomaslin:


tomaslin:
@ttomsu PTAL

tomaslin:
@spinnaker/reviewers  ping

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/support/DeploymentStrategyParameterSupport.groovy
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.pipeline.support
+
+import com.netflix.spinnaker.orca.pipeline.model.Pipeline
+
+class DeploymentStrategyParameterSupport {
```
ajordens: Would this fit better as a method on the interface with a default noop interface implementation for Orchestration (and this implementation for Pipeline)?

`DeploymentStrategyParametersSupport.resolveStrategyParams()` vs. `stage.resolveStrategyParams()`

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartScriptTask.groovy
```diff
@@ -52,6 +53,14 @@ class StartScriptTask implements Task {
     String cluster = stage.context.cluster
     String cmc = stage.context.cmc
 
+    if(stage.execution instanceof Pipeline && stage.execution.trigger.parameters?.strategy == true){
+      Map trigger = ((Pipeline) stage.execution).trigger
+      image = image ?: trigger.parameters.amiName ?: trigger.parameters.imageId ?: ''
+      cluster = cluster ?: trigger.parameters.cluster ?: ''
+      account = account ?: trigger.parameters.credentials ?: ''
+      region = region ?: trigger.parameters.region ?: trigger.parameters.zone ?: ''
```
ajordens: We really need to centralize the region vs. zone somewhere sometime. The task level is almost certainly too granular? (funny enough, @ttomsu and I were just talking about this).

ajordens:
Mind squashing your branch down?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/support/DeploymentStrategyParameterSupport.groovy
tomaslin: Not sure it's the right approach as there are some stages where you don't want this to happen

ajordens: But you'd still have to explicitly call the fn?

PR #681, opened 2015-12-11
Precondition expressions no longer need to be wrapped in ${ and }

ajordens:
- The ExpressionPreconditionTask will explicitly evaluate the expression at task execution time
- Any portions of the expression wrapped in ${ and } will be evaluated at both pipeline start and stage start


ajordens:
PTAL @tomaslin, this should be what we were talking about earlier today.

I'll update `deck` to remove any mention of ${ and }.

tomaslin:
would probably add trigger and execution to the lookup to be consistent, otherwise LGTM https://github.com/spinnaker/orca/blob/1228f97803e0ca79fcb9ad2ce66132d509c27b23/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionContextManager.groovy#L77-L82

ajordens:
I've rebased and added trigger/execution to the context.

tomaslin:
LGTM

PR #683, opened 2015-12-14
ExpressionPrecondition should include the executionResult in it's nes…

ajordens:
…ted context

- Deck already has support for showing nested context values (so this will automatically show the executionResult in the UI)


ajordens:
@duftler -- Started down the path of showing executionResult in `deck` but it's far simpler if I just return a modified context including it from `ExpressionPreconditionTask` as `deck` will pick it up correctly OOTB.

I am queueing up a minor deck PR that tidies up stage details UI a bit.

duftler:
Nice.

PR #684, opened 2015-12-15
allow traffic to be  overriden by a strategy

tomaslin:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/ParallelDeployStage.groovy
```diff
@@ -119,6 +119,17 @@ class ParallelDeployStage extends ParallelStage {
           // GCE uses 'image' as the ID key to clouddriver.
           cluster.image = trigger.parameters.imageId
         }
+        if(stage.context.trafficOptions && stage.context.trafficOptions != 'inherit'){
```
ajordens: can we get this formatted and maybe a comment or two, I have some context around what's going on ... but future me likely won't.

ajordens:
LGTM

PR #685, opened 2015-12-15
CreateBakeTask should only look at a trigger if it's actually a Pipeline

ajordens:


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -102,8 +102,11 @@ class CreateBakeTask implements RetryableTask {
     if (stage.context.rebake) {
       return true
     }
-    Map trigger = ((Pipeline) stage.execution).trigger
-    return trigger?.rebake
+    if (stage.execution instanceof Pipeline) {
+      Map trigger = ((Pipeline) stage.execution).trigger
+      return trigger?.rebake == true ? true : false
```
tomaslin: why not just trigger?.rebake ?

ajordens: Type safety more than anything.

Previously if rebake = '1' or 'asdf', it still would have evaluated to true (according to my tests).

PR #690, opened 2015-12-17
less agressive polling for instance stage changes

cfieber:
Updates WaitForTerminatedInstancesTask to a 10 second interval from a 1 second interval
If available, looks up instances from the serverGroup rather than an instance search
Updates AbstractInstancesCheckTask to a 10 second interval from a 5 second interval

@ajordens ptal


ajordens:
LGTM

PR #694, opened 2015-12-28
DetermineSourceServerGroupTask should set zone and serverGroupName in…

duftler:
… stageOutputs if they are present in StageData.Source as returned by SourceResolver.


duftler:
Experimenting with a different approach. Closing for now.

PR #697, opened 2016-01-04
update dependencies

cfieber:
some eureka configuration that was in orca was removed in favor of eureka config in kork-core


orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/OrcaConfiguration.groovy
```diff
@@ -83,7 +82,7 @@ class OrcaConfiguration {
     Duration.of(3, MINUTES)
   }
 
-  @Bean Scheduler scheduler() {
+  @Bean(destroyMethod = '') Scheduler scheduler() {
```
cfieber: rxjava 1.0.15 introduces a shutdown method on Schedulers that Spring autodetects which does not play nice with closing and refreshing a context within a JVM (as is the case in unit tests)

cfieber:
@robfletcher PTAL - I made a couple changes around the way things react to Eureka status changes

* Worker threads will now shutdown on any status except UP
* Ensured that startPolling only starts up if it is not already running

I'm seeing duplication of events between the event publishing in kork-core and the status polling. I'd like to keep the status polling in place just to ensure we detect changes as quickly as possible so changing the start/stop behaviour to idempotent seemed like the better fix (I could be convinced otherwise however)

PR #698, opened 2016-01-05
Revert "update dependencies"

cfieber:
Reverts spinnaker/orca#697


cfieber:
our automated testing has some issues with this one, going to back it out for now and do some further testing locally

PR #702, opened 2016-01-08
Support for pausing and resuming a Pipeline

ajordens:
- Works by having Retryable tasks enter a no-op loop when paused (no execution will occur)
- Non-Retryable tasks can not be paused (a theoretical drawback but in-practice unlikely to cause concern)

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/BatchStepStatus.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/BatchStepStatus.groovy
@@ -41,6 +41,7 @@ class BatchStepStatus {
       case ExecutionStatus.TERMINAL:
         return new BatchStepStatus(RepeatStatus.FINISHED, result.status.exitStatus, BatchStatus.FAILED)
       case ExecutionStatus.RUNNING:
+      case ExecutionStatus.PAUSED:
         return new BatchStepStatus(RepeatStatus.CONTINUABLE, result.status.exitStatus, BatchStatus.STARTED)
       case ExecutionStatus.CANCELED:
         return new BatchStepStatus(RepeatStatus.FINISHED, result.status.exitStatus, BatchStatus.COMPLETED)
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
@@ -17,11 +17,14 @@
 package com.netflix.spinnaker.orca.batch.adapters
 
 import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.orca.DefaultTaskResult
+import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.RetryableTask
 import com.netflix.spinnaker.orca.TaskResult
 import com.netflix.spinnaker.orca.batch.exceptions.ExceptionHandler
 import com.netflix.spinnaker.orca.batch.exceptions.TimeoutException
 import com.netflix.spinnaker.orca.batch.retry.PollRequiresRetry
+import com.netflix.spinnaker.orca.pipeline.model.Execution.PausedDetails
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
 import groovy.transform.CompileStatic
@@ -32,11 +35,14 @@ import org.springframework.batch.core.scope.context.ChunkContext
 import org.springframework.batch.repeat.RepeatStatus
 
 import java.time.Clock
+import java.util.concurrent.TimeUnit
 
 import static com.netflix.spinnaker.orca.pipeline.model.Stage.STAGE_TIMEOUT_OVERRIDE_KEY
 
 @CompileStatic
 class RetryableTaskTasklet extends TaskTasklet {
+  static final long MAX_PAUSE_TIME_MS = TimeUnit.HOURS.toMillis(12)
+
   private final Clock clock
   private final long timeoutMs
 
@@ -61,15 +67,45 @@ class RetryableTaskTasklet extends TaskTasklet {
 
   @Override
   protected TaskResult doExecuteTask(Stage stage, ChunkContext chunkContext) {
-    def now = clock.millis()
-    def startTime = chunkContext.stepContext.getStepExecution().startTime.time
+    if (stage.execution.paused?.isPaused()) {
+      checkTimeout(clock.millis() - stage.execution.paused.pauseTime, MAX_PAUSE_TIME_MS)
+    }
 
+    if (stage.execution.status == ExecutionStatus.PAUSED) {
+      // If the pipeline is PAUSED, ensure that the current stage and all running tasks are PAUSED
+      stage.self.status = ExecutionStatus.PAUSED
+      stage.self.tasks.findAll { it.status == ExecutionStatus.RUNNING }.each {
+        it.status = ExecutionStatus.PAUSED
+      }
+      return new DefaultTaskResult(ExecutionStatus.PAUSED)
+    } else if (stage.execution.status == ExecutionStatus.RUNNING) {
+      // If the pipeline is RUNNING, ensure that the current stage and all paused tasks are RUNNING
+      stage.self.status = ExecutionStatus.RUNNING
+      stage.self.tasks.findAll { it.status == ExecutionStatus.PAUSED }.each {
+        it.status = ExecutionStatus.RUNNING
+      }
+    }
+
+    def startTime = chunkContext.stepContext.getStepExecution().startTime.time
+    def executionTimeMs = determineCurrentExecutionTime(clock, startTime, stage.execution.paused)
     def timeoutMs = (stage.context[STAGE_TIMEOUT_OVERRIDE_KEY] ?: timeoutMs) as long
-    if (now - startTime > timeoutMs) {
+
+    checkTimeout(executionTimeMs, timeoutMs)
+    return super.doExecuteTask(stage, chunkContext)
+  }
+
+  /**
+   * In order to accurately determine whether this task has timed out, any time spent in a paused state must be discounted.
+   */
+  private static long determineCurrentExecutionTime(Clock clock, long startTime, PausedDetails pausedDetails) {
+    return clock.millis() - ((pausedDetails?.pausedMs ?: 0) as Long) - startTime
+  }
+
+  private static void checkTimeout(long elapsedTimeMs, long timeoutMs) {
+    if (elapsedTimeMs > timeoutMs) {
       def formatter = PeriodFormat.getDefault()
       def dur = formatter.print(new Duration(timeoutMs).toPeriod())
       throw new TimeoutException("Operation timed out after ${dur}")
     }
-    return super.doExecuteTask(stage, chunkContext)
   }
 }
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
@@ -91,13 +91,6 @@ class TaskTasklet implements Tasklet {
 
         logResult(result, stage, chunkContext)
 
-        // we should reload the execution now, in case it has been affected
-        // by a parallel process
-        long scheduledTime = stage.scheduledTime
-        stage = currentStage(chunkContext, true)
-        // Setting the scheduledTime if it has been set by the task
-        stage.scheduledTime = scheduledTime
-
         if (result.status == ExecutionStatus.TERMINAL) {
           setStopStatus(chunkContext, ExitStatus.FAILED, result.status)
           // cancel execution which will halt any parallel branches and run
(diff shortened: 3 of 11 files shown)


ajordens:
PTAL @robfletcher / @spinnaker/netflix-reviewers

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/BatchStepStatus.groovy
```diff
@@ -41,6 +41,7 @@ class BatchStepStatus {
       case ExecutionStatus.TERMINAL:
         return new BatchStepStatus(RepeatStatus.FINISHED, result.status.exitStatus, BatchStatus.FAILED)
       case ExecutionStatus.RUNNING:
+      case ExecutionStatus.PAUSED:
```
ajordens: For all intents and purposes, PAUSED is the same as RUNNING.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
```diff
@@ -61,9 +64,23 @@ class RetryableTaskTasklet extends TaskTasklet {
 
   @Override
   protected TaskResult doExecuteTask(Stage stage, ChunkContext chunkContext) {
-    def now = clock.millis()
-    def startTime = chunkContext.stepContext.getStepExecution().startTime.time
+    if (stage.execution.status == ExecutionStatus.PAUSED) {
+      // If the pipeline is PAUSED, ensure that the current stage and all running tasks are PAUSED
+      stage.self.status = ExecutionStatus.PAUSED
```
ajordens: This is house cleaning to avoid unnecessary transformations in `deck`.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
```diff
@@ -72,4 +89,11 @@ class RetryableTaskTasklet extends TaskTasklet {
     }
     return super.doExecuteTask(stage, chunkContext)
   }
+
+  /**
+   * In order to accurately determine whether this task has timed out, any time spent in a paused state must be discounted.
+   */
+  static long determineNow(Clock clock, PausedDetails pausedDetails) {
+    return clock.millis() - ((pausedDetails?.pausedMs ?: 0) as Long)
```
ajordens: For example, if we were paused for 30 minutes ... I want to subtract 30 minutes from clock.millis() when determining whether or not the stage timeout has been exceeded.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
```diff
@@ -91,13 +91,6 @@ class TaskTasklet implements Tasklet {
 
         logResult(result, stage, chunkContext)
 
-        // we should reload the execution now, in case it has been affected
-        // by a parallel process
-        long scheduledTime = stage.scheduledTime
```
ajordens: I couldn't think of a legitimate reason why this was necessary, there's no such thing as a "parallel" process that would be attempting to save the same stage.

This was possibly an issue awhile back when all stages were saved at the same time.

With this in place, it prevents the RetryableTaskTasklet from doing a `stage.self.status = PAUSED` as the stage will be reloaded.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.groovy
```diff
@@ -45,6 +45,7 @@ abstract class Execution<T> implements Serializable {
   ExecutionStatus status = NOT_STARTED
 
   AuthenticationDetails authentication
+  PausedDetails paused
```
ajordens: @robfletcher I opted to nest the paused details rather than flatten them directly on the `Execution`. Thoughts? This is in contrast to what we have for canceled (_although it might make sense to change_).

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
```diff
@@ -135,30 +120,49 @@ class JedisExecutionRepository implements ExecutionRepository {
...
-      }
       Boolean.valueOf(jedis.hget(key, "canceled"))
     }
   }
 
   @Override
+  void pause(String id, String user) {
+    String key = fetchKey(id)
+    withJedis { Jedis jedis ->
+      def pausedDetails = new Execution.PausedDetails(
```
ajordens: I'm going to add a check here that only RUNNING pipelines can be PAUSED ... and only PAUSED pipelines can be resumed.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
```diff
@@ -61,9 +64,23 @@ class RetryableTaskTasklet extends TaskTasklet {
...
   protected TaskResult doExecuteTask(Stage stage, ChunkContext chunkContext) {
-    def now = clock.millis()
-    def startTime = chunkContext.stepContext.getStepExecution().startTime.time
+    if (stage.execution.status == ExecutionStatus.PAUSED) {
+      // If the pipeline is PAUSED, ensure that the current stage and all running tasks are PAUSED
+      stage.self.status = ExecutionStatus.PAUSED
+      stage.self.tasks.findAll { it.status == ExecutionStatus.RUNNING }.each {
+        it.status = ExecutionStatus.PAUSED
+      }
+      return new DefaultTaskResult(ExecutionStatus.PAUSED)
```
ajordens: I will also be adding a check here to auto-cancel a pipeline in the event that it's been paused for more than XX hours (thinking 8hrs).

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/BatchStepStatus.groovy
tomaslin: need to check if this emits an event.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
tomaslin: What is the opinion around downstream pipelines? if I pause a pipeline, should executing pipelines that I have spawned also be paused?

ajordens: Personally, I'd say `no` it shouldn't auto-pause downstream pipelines. 

Seems like an edge case (right now) and i'm not sure the complexity is warranted. I don't think there's anything that would stop us from implementing it tho.

Happy to entertain feedback?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/BatchStepStatus.groovy
ajordens: EchoNotifyingStageExecutionListener:
    if (stepExecution.status.running) {
      return
    }

I wouldn't expect it to emit an event ... we only send events to echo before/after task whereas `PAUSED` occurs during task execution.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
robfletcher: what about stages running in parallel?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
robfletcher: Small thing but it feels like this method could be better named or just even encapsulate the whole check in a method rather than getting values back & doing a comparison in the caller.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
ajordens: Each stage will pickup the PAUSED state and go into busy-wait mode.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.groovy
```diff
@@ -120,4 +121,21 @@ abstract class Execution<T> implements Serializable {
...
   }
+
+  static class PausedDetails implements Serializable {
+    String pausedBy
+    String resumedBy
+
+    Long pauseTime
+    Long resumeTime
+
+    boolean isPaused() {
```
robfletcher: will this need a `@JsonIgnore` too?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
ajordens: Fair enough. 

My intention for a dedicated method was to simplify unit tests against the calculation.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.groovy
ajordens: It doesn't actually need one but for consistency I'll add one.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
```diff
@@ -561,6 +578,21 @@ class JedisExecutionRepository implements ExecutionRepository {
     "pipeline:executions:$pipelineConfigId"
   }
 
+  private String fetchKey(String id) {
```
robfletcher: That's an improvement

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
```diff
@@ -135,30 +120,59 @@ class JedisExecutionRepository implements ExecutionRepository {
...
+  @Override
+  void resume(String id, String user) {
+    String key = fetchKey(id)
+    withJedis { Jedis jedis ->
+      def currentStatus = ExecutionStatus.valueOf(jedis.hget(key, "status"))
+      if (currentStatus != ExecutionStatus.PAUSED) {
+        throw new IllegalStateException("Unable to resume pipeline that is not PAUSED (executionId: ${id}, currentStatus: ${currentStatus})")
+      }
+
+      def pausedDetails = mapper.readValue(jedis.hget(key, "paused"), Execution.PausedDetails)
```
robfletcher: Worth a `null` check here? I guess it should never happen if the status is `PAUSED`.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTaskletSpec.groovy
```diff
@@ -127,7 +127,7 @@ class TaskTaskletSpec extends Specification {
     then:
     stageArgument.immutable
     stageArgument.type == stage.type
-    stageArgument.context == stage.context
+    stageArgument.context.intersect(stage.context) == stage.context
```
robfletcher: WTF?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
ajordens: Should not be possible to be in PAUSED and not have `pausedDetails`. 

The `pause()` method on `JedisExecutionRepository` is the single source of setting the `PAUSED` status and does an atomic set of both `status` and `paused`.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
robfletcher: Oh, I see it's getting it off the `stage.execution`. Read it too quickly before.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
robfletcher: Yeah, there's a whole bunch of similar things that existed prior to when we remodeled the way the saves happened. That line in particular does look like cruft that probably had a bunch more code around it at one time.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.groovy
robfletcher: I like what you've done here.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.groovy
robfletcher: I wasn't sure because `isXXX` should get picked up as a Java Bean property, I think.

PR #705, opened 2016-01-11
Organize pipeline and tasks packages

ttomsu:
**NOTE** - This is based off two other pending PRs - Focus only on the last commit right now.

@cfieber @ajordens @duftler @tomaslin @robfletcher - This PR intends to organize the `c.n.s.orca.clouddriver.[pipeline|tasks]` packages. The current structure is basically just one big catch-all package, with some smaller subpackages for provider-specific and helper classes. I think we can do better, especially in the `tasks` package, which has **50+** classes.

In it's current form, I've only done the production code packages (no tests) to gather feedback. The proposed structure is:

* c.n.s.o.clouddriver.[pipeline|tasks]
  * providers
    * aws
    * gce
    * etc.
  * servergroup
    * strategies
    * support
  * cluster
  * instance
  * securitygroup
  * loadbalancer

For the more visually inclined, here's how it looks in the IntelliJ package tree:
organized-tasks


anotherchrisberry:
@cfieber @ajordens @duftler @tomaslin @robfletcher ping

ajordens:
I dig the re-org ... unsure about the copyright header changing from Netflix -> Google (cc: @aglover).

ttomsu:
I'll try to redo the changes with the copyright modifier off.

duftler:
LGTM. Would also prefer to leave copyright headers untouched during refactoring.

PR #708, opened 2016-01-13
BakeStage is now restartable

ajordens:
- requires a configuration change in deck


ajordens:
Was already working in `orca` so this was an easy one to bang off.

PR #709, opened 2016-01-13
resolve pipeline executions in triggers if only an id is provided

tomaslin:


robfletcher:
LGTM. Punch it Chewie.

PR #710, opened 2016-01-14
WaitForRequisiteCompletion should be checking synthetic child stages …

ajordens:
…otherwise runs the risk of prematurely completing once an initialization stage is complete

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/parallel/WaitForRequisiteCompletionTask.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/parallel/WaitForRequisiteCompletionTask.groovy
@@ -17,7 +17,6 @@
 package com.netflix.spinnaker.orca.pipeline.parallel
 
 import com.netflix.spinnaker.orca.DefaultTaskResult
-import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.RetryableTask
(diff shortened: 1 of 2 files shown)


ajordens:

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/parallel/WaitForRequisiteCompletionTask.groovy
```diff
@@ -38,27 +37,35 @@ class WaitForRequisiteCompletionTask implements RetryableTask {
...
-      if (requisiteStage?.status == TERMINAL) {
-        termainalStageNames << requisiteStage?.name
+        return
       }
 
-      def tasks = (requisiteStage?.tasks ?: []) as List<Task>
-      if (tasks && tasks[-1].status != SUCCEEDED) {
-        // ensure the last task has completed (heuristic for all tasks being complete)
-        allRequisiteStagesAreComplete = false
+      def requisiteStages = [requisiteStage] + stage.execution.stages.findAll { it.parentStageId == requisiteStage.id }
```
ajordens: This was previously only looking at the requisiteStage itself ... in the case of a Canary this would have been an initialization stage that was almost certainly complete (with synthetic children still running*).

This changes the WaitFor logic to also check that all synthetic children are complete before proceeding. 

* I'd suggest tackling the initialization stage statuses independently from this PR

cfieber:
LGTM

PR #712, opened 2016-01-15
Cluster wide tasks will now ignore any server groups that were deploy…

ajordens:
…ed more recently than any server groups deployed as part of the executing pipeline

- Ancestor lookup vs. parentStage lookup (side-effect being that these stages should work correctly as Pipeline Strategies)
- If two highlanders and run within close proximity, the last one will now win (previously they stepped on each other and the cluster was left empty)

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTask.groovy
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.orca.DefaultTaskResult
 import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.RetryableTask
 import com.netflix.spinnaker.orca.TaskResult
+import com.netflix.spinnaker.orca.batch.StageBuilder
 import com.netflix.spinnaker.orca.clouddriver.KatoService
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.CloneServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.CreateServerGroupStage
@@ -32,11 +33,13 @@ import com.netflix.spinnaker.orca.kato.pipeline.CopyLastAsgStage
 import com.netflix.spinnaker.orca.kato.pipeline.DeployStage
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import groovy.transform.Canonical
+import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 
 /**
  * A task that operates on some subset of the ServerGroups in a cluster.
  */
+@Slf4j
 abstract class AbstractClusterWideClouddriverTask extends AbstractCloudProviderAwareTask implements RetryableTask {
 
   @Override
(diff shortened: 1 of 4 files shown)


ajordens:
PTAL @cfieber 

I need to re-test locally (I refactored a few things while writing unit tests) ... but prior to that this worked when I kicked off multiple highlander pipelines (~30-60s in between). 

I also tested with highlander as a pipeline strategy and it worked as well.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTask.groovy
```diff
@@ -201,6 +176,54 @@ abstract class AbstractClusterWideClouddriverTask extends AbstractCloudProviderA
...
+    //if we are a synthetic stage child of a deploy, don't operate on what we just deployed
+    final Set<String> deployStageTypes = [
+      DeployStage.PIPELINE_CONFIG_TYPE,
+      CopyLastAsgStage.PIPELINE_CONFIG_TYPE,
+      CloneServerGroupStage.PIPELINE_CONFIG_TYPE,
+      CreateServerGroupStage.PIPELINE_CONFIG_TYPE,
+    ]
+    List<TargetServerGroup> deployedServerGroups = []
+
+    stage.ancestors({ Stage ancestorStage, StageBuilder stageBuilder ->
```
ajordens: This will now do a proper ancestor lookup vs. only looking at the immediate parent.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTask.groovy
```diff
@@ -201,6 +176,54 @@ abstract class AbstractClusterWideClouddriverTask extends AbstractCloudProviderA
...
+    }
+
+    return deployedServerGroups
+  }
+
+  static List<TargetServerGroup> filterParentAndNewerThanParentDeploys(List<TargetServerGroup> parentDeployedServerGroups,
+                                                                       List<TargetServerGroup> clusterServerGroups) {
+    def recentlyDeployedServerGroups = []
+    if (parentDeployedServerGroups) {
+      Long minCreatedTime = (parentDeployedServerGroups*.createdTime as Collection<Long>).min()
```
ajordens: In the event of their being multiple upstream deployed server groups, I opt'd to take the earliest creation time and then skipping anything created after it.

This seemed reasonable but I'm open to opinions.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTaskSpec.groovy
```diff
@@ -0,0 +1,101 @@
...
+      ]
+    ])
+    pipeline.stages << new PipelineStage(pipeline, CloneServerGroupStage.PIPELINE_CONFIG_TYPE, [
+      "deploy.account.name" : account,
+      "deploy.server.groups": [
+        "us-west-1": [sg2.name, sg4.name]
+      ]
+    ])
+
+    (pipeline.stages[0] as AbstractStage).stageNavigator = new StageNavigator(Mock(ApplicationContext))
```
ajordens: checks both synthetic and requisiteId navigation.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTask.groovy
cfieber: makes sense to me

cfieber:
LGTM

PR #720, opened 2016-01-30
Server groups should be disabled prior to being destroyed

ajordens:
- Insert a DisableClusterStage prior to any ShrinkClusterStage
- WaitForClusterDisableTask will wait a particular amount of time to allow for any connection draining / discovery out_of_service to complete

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/WaitForClusterDisableTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/WaitForClusterDisableTask.groovy
@@ -16,14 +16,33 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.cluster
 
+import com.netflix.spinnaker.orca.DefaultTaskResult
+import com.netflix.spinnaker.orca.ExecutionStatus
+import com.netflix.spinnaker.orca.TaskResult
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup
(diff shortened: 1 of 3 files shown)


ajordens:
Thoughts on this approach @cfieber?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/WaitForClusterDisableTask.groovy
```diff
@@ -16,14 +16,33 @@
...
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup
 import com.netflix.spinnaker.orca.clouddriver.utils.HealthHelper
+import com.netflix.spinnaker.orca.pipeline.model.Stage
 import org.springframework.stereotype.Component
 
 import java.util.function.Function
 
 @Component
 class WaitForClusterDisableTask extends AbstractWaitForClusterWideClouddriverTask {
+  private static final int MINIMUM_WAIT_TIME_MS = 90000
```
ajordens: Not a huge fan of the static wait ... but the current behavior has isServerGroupOperationInProgress completing as soon as the server group is disabled (_vs. waiting for health providers to go DOWN_). 

This smells like a bug but at the very least a small wait should account for any connection draining / discovery activity to complete.

robfletcher:
So the disable cluster will affect the exact same ASGs that the shrink will? Is there any chance of another operation interfering?

ajordens:
I don't believe another operation should interfere ... these stages do look for and ignore any server groups that are newer than the current pipeline.

robfletcher:
ok, in that case LGTM

PR #723, opened 2016-02-03
add a #fromUrl helper method that will allow fetching content 

tomaslin:
Adds a #fromUrl helper method that will either return a list or map from the url that is sent in. 

This is intended to help in situations where you might want to check against a healthcheck or get more details from another pipeline. 

This is just a building block, we will add more helper functions that will make it easier to get things like a #hostname from a cluster, for example. 

Usage:

```#fromUrl( 'http://myserver/endpoint' )[0].instances```

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
@@ -115,8 +115,12 @@ class ContextParameterProcessor {
       evaluationContext.addPropertyAccessor(MapPropertyAccessor)
       evaluationContext.registerFunction('alphanumerical', ContextStringUtilities.getDeclaredMethod("alphanumerical", String))
       evaluationContext.registerFunction('toJson', ContextStringUtilities.getDeclaredMethod("toJson", Object))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tomaslin** [comment]
@spinnaker/reviewers PTAL

File: orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy [hunk 1]
```diff
@@ -140,15 +141,28 @@ abstract class ContextStringUtilities {
...
     Integer.valueOf(str)
   }
+
   static Float toFloat(String str) {
     Float.valueOf(str)
   }
+
+  static Object fromUrl(String url) {
+    String response = new URL(url).text
+    readJson(response)
```
cfieber: would it make sense to just register the readJson method and enable somebody do do a fromUrl on something that just returned a string? e.g. `https://my.jenkins.net/job/MY-JOB/lastSuccessfulBuild/buildNumber`

cfieber: maybe a convenience jsonFromUrl but the constituent parts exposed?

tomaslin:
broke it down into readJson, fromUrl and jsonFromUrl

PR #724, opened 2016-02-05
Handle more complex scenarios for disable-before-destroy

ajordens:
- Specifically pipelines with multiple deploy stages and DestroyServerGroup stages with multiple target regions

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DestroyServerGroupStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DestroyServerGroupStage.groovy
@@ -20,7 +20,9 @@ import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.Targe
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroupLinearStageSupport
 import com.netflix.spinnaker.orca.clouddriver.tasks.MonitorKatoTask
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.DestroyServerGroupTask
+import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.DisableServerGroupTask
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.ServerGroupCacheForceRefreshTask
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTask.groovy
```diff
@@ -198,12 +198,12 @@ abstract class AbstractClusterWideClouddriverTask extends AbstractCloudProviderA
       switch (location.type) {
         case Location.Type.ZONE:
           deployedServerGroups.addAll(clusterServerGroups.findAll {
-            it.zones.contains(location.value) && dsgs[it.region].contains(it.name)
+            it.zones.contains(location.value) && dsgs[it.region]?.contains(it.name)
```
ajordens: This handles a situation where you have multiple deploy stages in your pipeline.

Specifically a Deploy Stage #1 targeting us-east-1 and us-west-1 and a Deploy Stage #2 targeting us-west-1.

Previously a deploy stage was not looking up the pipeline hierarchy to determine what server groups were deployed and would have never ran into this situation ... rather it only looked at it's immediate parent which would have always had a `dsg`.

cfieber:
LGTM

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DestroyServerGroupStage.groovy
```diff
@@ -40,11 +42,14 @@ class DestroyServerGroupStage extends TargetServerGroupLinearStageSupport {
 
   @Override
   public List<Step> buildSteps(Stage stage) {
-    injectBefore(stage, "disableServerGroup", disableServerGroupStage, stage.context)
```
ajordens: This previous behavior worked when the DestroyStage only targeted a single region.

With multiple regions, the subsequent child destroy stages were unable to find a DetermineTargetServerGroup stage and failed. 

May or may not be related to:
```
if (stage.parentStageId) {
      // We only want to determine the target server groups once per stage, so only inject if this is the root stage,
      // i.e. the one the user configured.
      // This may become a bad assumption, or a limiting one, in that we cannot inject a dynamic stage ourselves
      // as part of some other stage that is not itself injecting a determineTargetReferences stage.
      return
    }
```

Although it's a smidge of duplication, I actually prefer what the resulting stage and task graph looks like:

image

PR #725, opened 2016-02-06
ServerGroupCacheForceRefresh should monitor force cache attempts and …

ajordens:
…poll for completion

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
@@ -15,9 +15,12 @@
  */
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.servergroup
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.orca.DefaultTaskResult
-import com.netflix.spinnaker.orca.ExecutionStatus
-import com.netflix.spinnaker.orca.Task
+import com.netflix.spinnaker.orca.RetryableTask
 import com.netflix.spinnaker.orca.TaskResult
 import com.netflix.spinnaker.orca.clouddriver.OortService
 import com.netflix.spinnaker.orca.clouddriver.tasks.AbstractCloudProviderAwareTask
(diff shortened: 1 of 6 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -26,43 +29,164 @@ import groovy.util.logging.Slf4j
...
+   *
+   * An HTTP 202 response indicates that the force cache operation has been queued and will complete at some point in
+   * the near future.
+   */
+  private Optional<TaskResult> performForceCacheRefresh(String account, String cloudProvider, StageData stageData) {
+    def zone = stageData.zone
+
+    def refreshableServerGroups = stageData.deployServerGroups.collect { region, serverGroups ->
+      serverGroups.findResults { String serverGroup ->
+        def model = [asgName: serverGroup, serverGroupName: serverGroup, region: region, account: account]
```
ajordens: I've got an unreleased change in `clouddriver:ClusterCachingAgent` that'll let us get rid of `asgName` here.

ajordens:
I need to write a few unit tests to cover the polling behavior but this has been working well locally via deck.

The gist is that `forceCacheRefresh` will wait until:
(1) it see's a pending refresh request
(2) this pending refresh request is processed (ie. goes through the ClusterCachingAgent.loadData() cycle)

I'll also update deck such that `forceCacheRefresh` tasks are no longer hidden given that they'll no longer be no-ops.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -26,43 +29,164 @@ import groovy.util.logging.Slf4j
...
+          stageData.removeRefreshedServerGroup(model.serverGroup, model.region, model.account)
+          log.warn("Unable to find pending cache refresh request (model: ${model})")
+
+          return false
+        }
+
+        if (forceCacheUpdate) {
+          if (!forceCacheUpdate.processedTime || !forceCacheUpdate.cacheTime) {
+            // there is a pending cache update that is still awaiting processing
+            log.warn("Awaiting processing on pending cache refresh request (model: ${model})")
```
cfieber: this might be too noisy for prod

cfieber:
LGTM

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
ajordens: Possibly ... I'd like to leave it in initially in case something doesn't go quite right, the added context would be beneficial.

PR #726, opened 2016-02-09
ServerGroupCacheForceRefreshTask should succeed rather than fail if a…

ajordens:
… timeout occurs.

I've seen this occur if a forceCache is queued up for a subsequently deleted ASG ...
in these cases the ttl should expire the record after 10 minutes and the task re-forceCache refresh.

If we see a significant # of forceCacheRefreshes getting hung up around deletes, I'd suggest doing
a re-forceCache call every few minutes.

Otherwise, this change preserves the original behavior of a forceCacheRefresh being fire and forget.


ajordens:
I suspect the last few hang-on-deletes I've seen were related to https://github.com/spinnaker/clouddriver/pull/294.

In those cases we would have attempted to onDemand an ASG in the process of being deleted. By the time the next loadData() cycle came around that ASG would have finished terminating and thus the onDemand record never processed (it would naturally decay out after it's 10 minute ttl expired).

Unless AWS is having an issue, I cannot think of any legitimate reason why caches would not have reconciled after 12 minutes hence the auto-succeed rather than failing the pipeline.

PR #727, opened 2016-02-09
Removed the @JsonIgnore on zone so that it would be properly deserial…

ajordens:
…ized in to


ajordens:
@duftler

duftler:
LGTM.
Will merge when (Metal) Travis completes and then test it out behind deck again.

duftler:
Thanks for the quick fix. That did the trick.

PR #729, opened 2016-02-10
Add an implicit resize to a rollback operation

ajordens:
- Will resize the currently disabled ASG such that it matches the active ASG

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/ResizeStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/ResizeStrategy.groovy
@@ -22,7 +22,7 @@ import groovy.transform.Canonical
 
 interface ResizeStrategy {
   static enum ResizeAction {
-    scale_exact, scale_up, scale_down, scale_to_cluster
+    scale_exact, scale_up, scale_down, scale_to_cluster, scale_to_server_group
   }
 
   static class OptionalConfiguration {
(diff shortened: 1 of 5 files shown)


ajordens:
Had this on my back burner for awhile ... this PR makes the explicit rollback stage a little safer by resizing the previous ASG prior to disabling the current ASG.

PTAL @cfieber @duftler @ttomsu

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/ResizeStrategy.groovy
```diff
@@ -52,6 +52,28 @@ interface ResizeStrategy {
     Integer min
   }
 
+  @Canonical
+  static class Source implements Serializable {
+    Collection<String> zones
+    Collection<String> regions
+    String region
+    String zone
```
ttomsu: Seems like a single (or collection of) `Location` object(s) could be used here to replace all of these vars. Is there a way to tell Jackson "anytime you see `region` give me a `Location.region(region)`?"

ttomsu:
Outside of comment, LGTM.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/ResizeStrategy.groovy
ajordens: Ya, this is unfortunate.

I don't think there's a way we can do it explicitly within Jackson (maybe @cfieber knows) ... I've not looked into custom deserializers or anything.

What could be done is a transformation when the pipeline/task first starts that say injects `_location` or `_locations` into the JSON if it ever sees zones/regions/region/zone. 

I say `_location` just to have a key that isn't going to collide with a normal user-provided value.

ajordens:
I would also like to make our ResizeAsg operation in clouddriver not require asgName.

PR #731, opened 2016-02-11
fail a pipeline if a jenkins trigger has been specified but no artifacts

tomaslin:


tomaslin:
@spinnaker/reviewers ping

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -72,9 +72,21 @@ class PackageInfo {
...
+    }
+
     List<Map> triggerArtifacts = trigger.buildInfo?.artifacts ?: trigger.parentExecution?.trigger?.buildInfo?.artifacts
-    List<Map> buildArtifacts = buildInfo.artifacts
-    if ((!triggerArtifacts && !buildArtifacts) || isUrl(request.package)) {
+    List<Map> buildArtifacts = buildInfo?.artifacts
+
+    if (!buildInfo || (buildInfo && !buildArtifacts)) {
+      if (!triggerArtifacts && (trigger.buildInfo != null || trigger.parentExecution?.trigger?.buildInfo != null)) {
+        throw new IllegalStateException("Could not find artifacts in Jenkins jobs")
```
ajordens: Have we verified if this error message is displayed in `deck`?

Given it's probably something we want to show users, might be better worded as: "Jenkins trigger detected but no artifacts found. Please ensure your job is archive your x,y,z."(?) or something that might be a little more self-explanatory for people.

Otherwise preventing this behavior seems safer. I've got no idea how many people this will affect in reality.

ajordens:
LGTM other than comment agove

PR #732, opened 2016-02-12
Updates the FindImage stage to allow equivalent image lookup in regions with no ServerGroup.

cfieber:
Adds a `resolveMissingLocations` attribute to the FindImage stage that indicates if there is
not a server group for one of the required locations (zones or regions), to attempt to find
an image in those regions that matches the existing image.

Adds a global config property for the default behaviour of that flag if unspecified in the
request (defaults to false to preserve existing behaviour).

An image is only looked up if all the other find image results provided only a single
imageName - if the results returned multiple names the resolution gives up.

This enables the use case of promotion from test to production without requiring that
an image is deployed in test for each region that it ends up in production.

This requires that image naming ends up consistent (e.g. all regions need to be baked
at the same time).

This only currently will work for AWS as the clouddriver endpoint `/{cloudProvider}/images/{region}/{imageId}`
is only implemented on that provider.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/OortService.groovy
```diff
@@ -86,8 +86,8 @@ interface OortService {
                                            @Path("type") String type)
 
   @GET("/{type}/images/{account}/{region}/{imageId}")
-  Response getByAmiId(@Path("type") String type,
-                      @Path("account") String account,
-                      @Path("region") String region,
-                      @Path("imageId") imageId)
+  List<Map> getByAmiId(@Path("type") String type,
```
cfieber: The only place using this was just immediately objectmappering it (hence some of the noise in this commit)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -64,58 +66,118 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
...
+  @Value('${findImage.defaultResolveMissingLocations:false}')
+  boolean defaultResolveMissingLocations = false
+
   @Autowired
   OortService oortService
   @Autowired
   ObjectMapper objectMapper
 
+  @Canonical
+  static class FindImageConfiguration {
```
cfieber: supports stage.mapTo instead of reading stuff out of the context directly

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/oort/pipeline/FindAmiFromClusterStage.groovy
```diff
@@ -1,38 +0,0 @@
...
-package com.netflix.spinnaker.orca.oort.pipeline
-
-import com.netflix.spinnaker.orca.oort.tasks.FindAmiFromClusterTask
-import com.netflix.spinnaker.orca.pipeline.LinearStage
-import com.netflix.spinnaker.orca.pipeline.model.Stage
-import org.springframework.stereotype.Component
-
-@Component
-@Deprecated
-class FindAmiFromClusterStage extends LinearStage {
```
cfieber: We didn't have anyone still calling this

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/oort/tasks/FindAmiFromClusterTask.groovy
```diff
@@ -1,160 +0,0 @@
...
-import com.netflix.spinnaker.orca.pipeline.model.Stage
-import groovy.util.logging.Slf4j
-import org.springframework.beans.factory.annotation.Autowired
-import org.springframework.stereotype.Component
-import retrofit.RetrofitError
-
-@Component
-@Slf4j
-@Deprecated
-class FindAmiFromClusterTask implements Task {
```
cfieber: This was directly referenced in Canary (updated in this PR)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -64,58 +66,118 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
...
     }
 
+    if (missingLocations) {
+      if (imageNames.size() != 1) {
+        throw new IllegalStateException("Request to resolve images for missing ${config.requiredLocations.first().pluralType()} requires exactly one image. (Found ${imageNames})")
+      }
+
+      def searchImage = imageIds.entrySet().first()
+
+      def deploymentDetailTemplate = imageSummaries.find { k, v -> v != null}.value
```
robfletcher: I don't get this, doesn't it amount to the same thing as `def deploymentDetailTemplate = null`?

orca-igor/src/test/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTaskSpec.groovy
```diff
@@ -465,8 +463,8 @@ class GetCommitsTaskSpec extends Specification {
     task.objectMapper = new ObjectMapper()
     def oortResponse = "{\"launchConfig\" : {\"imageId\" : \"${sourceImage}\"}}".stripIndent()
     Response response = new Response('http://oort', 200, 'OK', [], new TypedString(oortResponse))
-    Response sourceResponse = new Response('http://oort', 200, 'OK', [], new TypedString('[{ "tags" : { "appversion" : "myapp-1.143-h216.186605b/MYAPP-package-myapp/216" }}]'))
-    Response targetResponse = new Response('http://oort', 200, 'OK', [], new TypedString('[{ "tags" : { "appversion" : "myapp-1.144-h217.a86305d/MYAPP-package-myapp/217" }}]'))
+    List<Map> sourceResponse = [[ "tags" : [ "appversion" : "myapp-1.143-h216.186605b/MYAPP-package-myapp/216" ]]]
+    List<Map> targetResponse = [[ "tags" : [ "appversion" : "myapp-1.144-h217.a86305d/MYAPP-package-myapp/217" ]]]
```
robfletcher: Wow, that file reads a lot better now

robfletcher:
Looks good other than the fact it broke Metal Travis

PR #736, opened 2016-02-19
Add CF support for image details

gregturn:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryServerGroupCreator.groovy
```diff
@@ -46,6 +46,36 @@ class CloudFoundryServerGroupCreator implements ServerGroupCreator {
       operation.trigger = stage.execution.trigger
     }
 
+    // If this is a stage in a pipeline, look in the context for the baked image.
+    def deploymentDetails = (stage.context.deploymentDetails ?: []) as List<Map>
+
+    if (!operation.image && deploymentDetails) {
+      // Bakery ops are keyed off cloudProviderType
+      operation.image = deploymentDetails.find { it.cloudProviderType == 'cf' }?.ami
```
cfieber: there should also be an `imageId` attribute on the deploymentDetail, it would be better to use than `ami`

cfieber:
LGTM aside from ami->imageId, I can merge after its rebased

gregturn:
Done

PR #745, opened 2016-02-27
deduplicate stage ids for restarts

tomaslin:
dedups stage ids for restarts to resolve issue where restarting a pipeline will create multiple synthetic stages with the same exact id


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionStarter.groovy
```diff
@@ -62,6 +65,19 @@ abstract class ExecutionStarter<T extends Execution> {
 
   T startExecution(T subject) {
     def job = createJob(subject)
+    if (subject instanceof Pipeline) {
```
robfletcher: why only if it's a pipeline?

robfletcher:
I'm sure there's a good reason for only bothering if it's a pipeline. Other than that LGTM.

PR #749, opened 2016-02-29
allow the script stage to be restarted

tomaslin:


ajordens:
Something to possibly be said about ScriptStage extending JenkinsStage since they're effectively identical (would save the overridden and identical `prepareStageForRestart()` behavior).

Otherwise, LGTM.

PR #750, opened 2016-03-01
FindImageFromCluster should retry if a summary is missing image details

ajordens:


ajordens:
PTAL @cfieber 

Figured having FindImage wait for image details vs. having the deploy stages wait longer was perhaps less impactful. 

Open to alternatives.

cfieber:
LGTM, I agree findImage should retry until it gets what it expects to

PR #751, opened 2016-03-02
Pass ipIngress to clouddriver

claymccoy:
This seems to be why ip ingress disappears when editing a security group. Everything in cloud driver is tested and works as advertised. The ipIngress field has been on the cloud driver upsert sg description since 2014 (not a new field), but it seems that it has never been wired in at this level.


claymccoy:
@cfieber I think this is the reason that the ip ingress disappeared on you today.

ajordens:
`clouddriver` used to explicitly ignore cidr ingresses. I take it that's changed?

ajordens:
Did you run orca locally with this change and run a deck request through it?

If not, any chance you could setup the local environment (deck -> gate -> orca locally pointed at clouddriver-prestaging) and verify?

Bit of a one-time hassle to get the configs but once they're there it's nice to run end-to-end locally.

claymccoy:
No, I just hit orca locally with the json payload that deck is sending (from watching the network on an actual call) which should be sufficient. Gate is a pass through for this call. Deck sends a lot of redundant info.
I also tested hitting a local clouddriver (that is where I started the hunt) to make sure that the problem with was not there. No changes needed at that level.
Yes, clouddriver now does exactly what you tell it as far as ipIngress. We were simply dropping one of the parameters in orca. This fixes that.

ajordens:
LGTM then

PR #752, opened 2016-03-02
Ensure that each server group is inspected (at least once) when force…

ajordens:
…CacheRefresh'ing

- Avoid using .every() and instead inspect each 'deploy.server.groups'

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
@@ -73,6 +73,11 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
     }
 
     boolean allAreComplete = processPendingForceCacheUpdates(account, cloudProvider, stageData, stage.startTime)
+    if (allAreComplete) {
+      // ensure clean stage data such that a subsequent ServerGroupCacheForceRefresh (in this stage) starts fresh
+      stageData.reset()
+    }
+
     return new DefaultTaskResult(allAreComplete ? SUCCEEDED : RUNNING, convertAndStripNullValues(stageData))
   }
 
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -73,6 +73,11 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
     }
 
     boolean allAreComplete = processPendingForceCacheUpdates(account, cloudProvider, stageData, stage.startTime)
+    if (allAreComplete) {
+      // ensure clean stage data such that a subsequent ServerGroupCacheForceRefresh (in this stage) starts fresh
+      stageData.reset()
```
ajordens: We have a couple stages that have multiple `ServerGroupCacheForceRefreshTask`'s. 

This is an attempt at cleaning up the context such that subsequent tasks will start fresh.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -138,10 +143,17 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
 
     boolean finishedProcessing = true
     stageData.deployServerGroups.each { String region, Set<String> serverGroups ->
-      finishedProcessing = finishedProcessing && serverGroups.every { String serverGroup ->
```
ajordens: Uncovered a scenario involving smaller accounts that indexed quickly (< 30s cycles) and operations affecting multiple server groups.

It was not 100% reliable to expect that every single server group would be processed in the same `CloudDriver.loadData()` cycle. 

In reality we should check all server groups at once and avoid checking the same server group over and over again.

ajordens:
I've tested against a users pipeline that contained forceCacheRefresh tasks occasionally taking 10 minutes to complete.

The particular region/account in question contained 13 server groups and indexed quite quickly which I believe was a contributing factor. 

After applying this PR, the forceCacheRefresh tasks complete in a more respectable 1-2 minutes.

@cfieber - I believe this was John's problem this afternoon.

cfieber:
LGTM

PR #753, opened 2016-03-02
Allow CF to have repositories

gregturn:
Since CF doesn't bake, allow repository/artifact as well as images when creating new server groups.


gregturn:

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryServerGroupCreator.groovy
```diff
@@ -71,7 +71,7 @@ class CloudFoundryServerGroupCreator implements ServerGroupCreator {
       }
     }
 
-    if (!operation.image) {
+    if (!operation.image && !operation.repository && !operation.artifact) {
```
duftler: Seems like the error message should be updated to reflect the updated check?

gregturn:
I updated the error message and fixed the corresponding unit test.

duftler:
LGTM.

PR #756, opened 2016-03-04
provider/kubernetes: Find image by stage id

lwander:
Previously, it was possible to create a pipeline `FindImage -> ... -> Deploy`, go back and edit `FindImage` and have the pipeline fail, because the `Deploy` stage's pattern selector no longer matched that of the `FindImage` stage. Referring to the `RefId` of the `FindImage` stage in `Deploy` fixes that.

This only changes behavior for the Kubernetes provider, since it's the only one use a pattern to select images. Requires https://github.com/spinnaker/deck/pull/2056

@duftler

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesServerGroupCreator.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesServerGroupCreator.groovy
@@ -48,8 +48,10 @@ class KubernetesServerGroupCreator implements ServerGroupCreator {
     containers.forEach { container ->
       if (container.imageDescription.fromContext) {
         def image = deploymentDetails.find {
-          def cluster = Names.parseName(it.sourceServerGroup).cluster
-          cluster == container.imageDescription.cluster && it.imageName ==~ container.imageDescription.pattern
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesServerGroupCreator.groovy
```diff
@@ -48,8 +48,7 @@ class KubernetesServerGroupCreator implements ServerGroupCreator {
     containers.forEach { container ->
       if (container.imageDescription.fromContext) {
         def image = deploymentDetails.find {
-          def cluster = Names.parseName(it.sourceServerGroup).cluster
-          cluster == container.imageDescription.cluster && it.imageName ==~ container.imageDescription.pattern
+          it.refId == container.imageDescription.stageId
```
duftler: Might be worth dropping a comment here to explain why you're using the `refId (basically the text from the PR description).

duftler:
LGTM.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesServerGroupCreator.groovy
lwander: Good idea

PR #762, opened 2016-03-15
Akka actor that can run tasks

robfletcher:
This is a first step at moving the execution model from Spring Batch to Akka and solves some of the basic problems while giving us something that we can use to experimentally run the tasks of a particular stage or pipeline in a fairly unintrusive way.


robfletcher:
@ajordens @cfieber @claymccoy PTAL

robfletcher:
Closing as we're going to pull this onto a branch – see #771

PR #764, opened 2016-03-16
Making the jenkins stage cancellable

tonyrzhang:
Hello @robfletcher @ajordens @tomaslin !

This PR is to add the ability for the jenkins stage to be cancellable. The change involves changes on both Orca and Igor.

- Orca: Jenkins stage now implements the `CancellableStage` interface and there is now a new task that hits a `stop` endpoint on Igor
- Igor: There are two different apis for canceling a jenkins job. The first is `/job/{jobName}/{buildNumber}/stop` which will stop a job that is currently running. The other one is `/queue/cancelItem` which will remove a queued job. There is also some weird jenkins api issues I ran into that I commented on inline.
- Deck: I thought about adding some sort of flag on the jenkins stage that would allow the user to opt into the cancellable aspect of the stage, but the more I thought about it the more it seemed like a bug that it didn't cancel in the first place. Let me know what you guys think about this. 

[Igor PR](https://github.com/spinnaker/igor/pull/72)

Thanks!


robfletcher:
We've been discussing this a little and decided we'd like to merge it (we were just concerned that maybe having it non-optional would be a problem but I think we're good). Can you just rebase off the latest master and I'll merge it in.

tonyrzhang:
@robfletcher awesome. I've updated the branch. Should be good to go.

PR #765, opened 2016-03-16
provider/kubernetes: Use image tag resolved at runtime (from trigger) if needed

lwander:
When a user leaves their image tag blank for a docker trigger, the tag is now resolved in the deploy phase.

@duftler

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -126,6 +126,10 @@ class OperationsController {
           trigger.job as String
         )
       }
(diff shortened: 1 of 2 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -126,6 +126,10 @@ class OperationsController {
           trigger.job as String
         )
       }
+    } else if (trigger?.registry && trigger?.repository && trigger?.tag) {
+      trigger.buildInfo = [
+          taggedImages: [[registry: trigger.registry, repository: trigger.repository, tag: trigger.tag]]
```
duftler: One could argue this should be indented 2 spaces, not 4.

duftler:
LGTM.

PR #767, opened 2016-03-17
Customize queue behavior for pipeline waiting

jcusa:
Fixing the issue https://github.com/spinnaker/spinnaker/issues/805

Allowing queuing pipelines to stay in queue.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineStarter.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineStarter.groovy
@@ -54,6 +54,7 @@ class PipelineStarter extends ExecutionStarter<Pipeline> {
       .withAppConfig((Map<String, Serializable>) config.appConfig)
       .withParallel(config.parallel as Boolean)
       .withLimitConcurrent(config.limitConcurrent as Boolean)
+      .withKeepWaitingPipelines(config.keepWaitingPipelines as Boolean)
       .withExecutingInstance(currentInstance)
       .withNotifications((List<Map<String, Object>>) config.notifications)
       .build()
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineStarterListener.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineStarterListener.groovy
@@ -76,11 +76,13 @@ class PipelineStarterListener implements JobExecutionListener {
             def queuedExecution = executionRepository.retrievePipeline(id)
             log.info("starting pipeline ${nextPipelineId} due to ${execution.id} ending")
             pipelineStarter.startExecution(queuedExecution)
-          } else {
+            startTracker.removeFromQueue(execution.pipelineConfigId, id)
+          } else if (!execution.keepWaitingPipelines) {
             log.info("marking pipeline ${nextPipelineId} as canceled due to ${execution.id} ending")
             executionRepository.cancel(id)
-          }
-          startTracker.removeFromQueue(execution.pipelineConfigId, id)
+            startTracker.removeFromQueue(execution.pipelineConfigId, id)
+          } // else we want to keep the pipeline in the queue
+
         }
       }
     }
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
@@ -334,18 +334,19 @@ class JedisExecutionRepository implements ExecutionRepository {
     String key = "${prefix}:$execution.id"
 
     Map<String, String> map = [
-      application      : execution.application,
-      appConfig        : mapper.writeValueAsString(execution.appConfig),
-      canceled         : String.valueOf(execution.canceled),
-      parallel         : String.valueOf(execution.parallel),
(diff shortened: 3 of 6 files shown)


tomaslin:
tests failed

jcusa:
yeah. looking into it.

jcusa:
Fixed the test.

jcusa:
@tomaslin  I have fixed the test. Also, there's a related pull request here: https://github.com/spinnaker/spinnaker/pull/811

orca-core/src/test/groovy/com/netflix/spinnaker/orca/batch/pipeline/PipelineStarterListenerSpec.groovy
```diff
@@ -63,6 +63,7 @@ class PipelineStarterListenerSpec extends Specification {
 
   def "should mark waiting pipelines that are not started as canceled"() {
     given:
+    listener.optimizeWaitingQueue = true
```
tomaslin: There should be a test here for the two behaviors

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineStarterListener.groovy
```diff
@@ -76,11 +81,13 @@ class PipelineStarterListener implements JobExecutionListener {
             def queuedExecution = executionRepository.retrievePipeline(id)
             log.info("starting pipeline ${nextPipelineId} due to ${execution.id} ending")
             pipelineStarter.startExecution(queuedExecution)
-          } else {
+            startTracker.removeFromQueue(execution.pipelineConfigId, id)
```
tomaslin: not too keen on this being a global flag. Feels like you want to also be able to override this on a per pipeline config level.

jcusa: @tomaslin yeah, thanks, that makes sense.  you are suggesting to have a UI checkbox in the individual pipeline configuration level for this, correct?

jcusa:
@tomaslin yeah. thanks. that makes sense. you are saying, have a UI check box in the configuration of the pipeline

jcusa:
Will update current PR making the flag at the individual pipeline level. Also, will have a deck PR where a checkbox will be available if 'Disable concurrent pipeline' is checked. 

sample screen shot: 
this is when 'Disable concurrent pipeline is checked' the new flag will show up for you to check 
newflag

this is when 'Disable concurrent pipeline is unchecked' 
newflag2

tomaslin:
I would make that second checkbox only show up when the first one is selected. Visually, it feels like it should be indented so that the relationship to the parent checkbox is obvious ( might be optional if we're hiding and showing ).

jcusa:
thanks tomas, i added screen shots above.

tomaslin:
yeah, comments was related to screenshot. Also, make sure there is a help field explaining what it does

jcusa:
yes. Will hide the check box if the first box is not checked. and, thanks for mentioning. will add help text too.

jcusa:
related UI changes is here:
https://github.com/spinnaker/deck/pull/2119

jcusa:
I have merged in the new changes. No longer using global config. The flag is at the pipeline level now.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
```diff
@@ -345,7 +345,8 @@ class JedisExecutionRepository implements ExecutionRepository {
       executingInstance: execution.executingInstance,
       status           : execution.status?.name(),
       authentication   : mapper.writeValueAsString(execution.authentication),
-      paused           : mapper.writeValueAsString(execution.paused)
+      paused           : mapper.writeValueAsString(execution.paused),
+      limitWaiting     : String.valueOf(execution.limitWaiting)
```
tomaslin: the variable name here is not very clear. can we change it to something more descriptive like keepWaitingPipelines

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineStarter.groovy
```diff
@@ -54,6 +54,7 @@ class PipelineStarter extends ExecutionStarter<Pipeline> {
       .withAppConfig((Map<String, Serializable>) config.appConfig)
       .withParallel(config.parallel as Boolean)
       .withLimitConcurrent(config.limitConcurrent as Boolean)
+      .withLimitWaiting(config.limitWaiting == null ? true : config.limitWaiting as Boolean)
```
tomaslin: how about ```config.limitWaiting ?: true```

jcusa:
thank you. changed variable name per suggestion to "keepWaitingPipelines". with that, default is now false  (i.e.  by default pipelines will not wait in queue)

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineStarter.groovy
ajordens: That probably won't handle the case where `config.limitWaiting == false`.

jcusa: agree. 
in the latest code, i ended up not needing the null check due to changing of the variable name/meaning per tomas' suggestion. so i think we are good here.

anotherchrisberry:
@tomaslin good here?

tomaslin:
LGTM

PR #768, opened 2016-03-20
provider/kubernetes: Security group upserter

lwander:
@duftler 

Appears to be required by each cloud provider.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesSecurityGroupUpserter.groovy
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2016 Netflix, Inc.
```
duftler: s/Netflix/Google

duftler:
LGTM.

PR #769, opened 2016-03-21
add allow ingress from classic param

claymccoy:


claymccoy:
@cfieber

orca-tide/src/main/groovy/com/netflix/spinnaker/orca/tide/tasks/DeepCopyPipelineTask.groovy
```diff
@@ -36,11 +36,14 @@ class DeepCopyPipelineTask implements Task {
     Map<String, String> source = stage.context.source
     Map<String, String> target = stage.context.target
     def dryRun = stage.context.dryRun ?: false
+    def allowIngressFromClassic = stage.context.allowIngressFromClassic ?: true
```
cfieber: this is probably not right in either the case of someone trying to pass false, or someone passing the string "true" or "false"

I would make a pojo in there and use stage.mapTo

orca-tide/src/main/groovy/com/netflix/spinnaker/orca/tide/tasks/DeepCopyServerGroupTask.groovy
```diff
@@ -36,7 +36,9 @@ class DeepCopyServerGroupTask implements Task {
     Map<String, String> source = stage.context.source
     Map<String, String> target = stage.context.target
     def dryRun = stage.context.dryRun ?: false
-    def taskId = tideService.deepCopyServerGroup(source.account, source.region, source.asgName, dryRun, target)
+    def allowIngressFromClassic = stage.context.allowIngressFromClassic ?: true
```
cfieber: same as above

cfieber:
LGTM

PR #776, opened 2016-03-24
Making pipeline stage restartable.

jcusa:


jcusa:
accidentally included extra file. will remove

PR #779, opened 2016-03-30
Don't retry network errors for non-idempotent requests

robfletcher:
This fixes an issue where we'll retry things like create server group on CloudDriver if we get a network error. We should only ever retry things that are known to be idempotent.

### DIFF
--- a/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/RetrofitExceptionHandler.groovy
+++ b/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/RetrofitExceptionHandler.groovy
@@ -60,9 +60,13 @@ class RetrofitExceptionHandler implements ExceptionHandler<RetrofitError> {
       response.details.kind = e.kind
       response.details.status = properties.status ?: null
       response.details.url = properties.url ?: null
-      response.shouldRetry = (e.kind == NETWORK && findHttpMethodAnnotation(e) in ["GET", "HEAD"])
+      response.shouldRetry = (e.kind == NETWORK && isIdempotentRequest(e))
       return response
     }
+  }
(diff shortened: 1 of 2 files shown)


robfletcher:

robfletcher:
SPIN-1623

ajordens:
I dig it, LGTM

orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/RetrofitExceptionHandler.groovy
```diff
@@ -56,8 +60,31 @@ class RetrofitExceptionHandler implements ExceptionHandler<RetrofitError> {
       response.details.kind = e.kind
       response.details.status = properties.status ?: null
       response.details.url = properties.url ?: null
-      response.shouldRetry = (e.kind == RetrofitError.Kind.NETWORK)
+      response.shouldRetry = (e.kind == NETWORK && findHttpMethodAnnotation(e) in ["GET", "HEAD"])
```
cfieber: I think you could add PUT and DELETE here - they should be idempotent ( I'm sure there's somewhere we are doin it wrong but nothing comes to mind )

robfletcher: Yeah, I was wondering about DELETE specifically. I think the problem with PUT is it gets used wrong as often as it gets used right. I know I mix up PUT and POST all the time.

robfletcher: If it comes to that I guess PATCH could also be considered idempotent.

PR #780, opened 2016-03-31
core: Run job task & stage

lwander:
Closely modeled after the deploy server group stage; however, the force cache refresh doesn't wait for the refresh to take place since the job could easily have completed by the time the refresh finishes.

@duftler

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.groovy
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline.job
+
+import com.netflix.spinnaker.orca.clouddriver.tasks.MonitorKatoTask
+import com.netflix.spinnaker.orca.clouddriver.tasks.job.JobForceCacheRefreshTask
+import com.netflix.spinnaker.orca.clouddriver.tasks.job.RunJobTask
+import com.netflix.spinnaker.orca.clouddriver.tasks.job.WaitOnJobCompletion
+import com.netflix.spinnaker.orca.pipeline.LinearStage
+import com.netflix.spinnaker.orca.pipeline.model.Stage
+import org.springframework.batch.core.Step
+import org.springframework.stereotype.Component
+
+@Component
+class RunJobStage extends LinearStage {
+  public static final String PIPELINE_CONFIG_TYPE = "runJob"
+
+  RunJobStage() {
+    super(PIPELINE_CONFIG_TYPE)
+  }
+
+  @Override
+  List<Step> buildSteps(Stage stage) {
+    [
+        buildStep(stage, "runJob", RunJobTask),
+        buildStep(stage, "monitorDeploy", MonitorKatoTask),
+        buildStep(stage, "forceCacheRefresh", JobForceCacheRefreshTask),
+        buildStep(stage, "waitOnJobCompletion", WaitOnJobCompletion),
+    ]
+  }
+}
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
@@ -56,8 +56,14 @@ class MonitorKatoTask implements RetryableTask {
     }
 
     Map<String, ? extends Object> outputs = [:]
-    if (status == ExecutionStatus.SUCCEEDED && !stage.context.containsKey("deploy.server.groups")) {
-      outputs["deploy.server.groups"] = getServerGroupNames(katoTask)
+    if (status == ExecutionStatus.SUCCEEDED) {
+      def deployed = getDeployedNames(katoTask)
+      if (!stage.context.containsKey("deploy.server.groups")) {
+        outputs["deploy.server.groups"] = getServerGroupNames(katoTask)
+      }
+      if (!stage.context.containsKey("deploy.jobs") && deployed) {
+        outputs["deploy.jobs"] = deployed
+      }
     }
     if (status == ExecutionStatus.SUCCEEDED || status == ExecutionStatus.TERMINAL) {
       List<Map<String, Object>> katoTasks = []
@@ -96,8 +102,11 @@ class MonitorKatoTask implements RetryableTask {
     }
   }
 
-  // This is crappy logic
-  // TODO clean this up in Kato
+  /**
+   * @param The task being inspected for region/server group mappings.
+   * @return Server group names keyed by region.
+   * @deprecate In favor of getDeployedNames(Task task).
+   */
   @CompileStatic(TypeCheckingMode.SKIP)
   private static Map<String, List<String>> getServerGroupNames(Task task) {
     def result = [:]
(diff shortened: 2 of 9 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobRunner.groovy
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.tasks.job
+
+import com.netflix.spinnaker.orca.pipeline.model.Stage
+
+/**
+ * Deployments of server groups vary wildly across cloud providers. A ServerGroupCreator
```
ajordens: s/ServerGroupCreator/JobRunner

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.groovy
```diff
@@ -0,0 +1,43 @@
...
+@Component
+class RunJobStage extends LinearStage {
+  public static final String PIPELINE_CONFIG_TYPE = "runJob"
+
+  RunJobStage() {
+    super(PIPELINE_CONFIG_TYPE)
+  }
+
+  @Override
+  List<Step> buildSteps(Stage stage) {
```
ajordens: At first glance, I suspect we could generify the `CreateServerGroupStage` to support this?

Slightly different behavior (_as you called out_) around forceCacheRefresh ... but that doesn't seem unreasonably to allow cloud providers to parameterize?

lwander: Do you mean each cloud provider could provide its own set of stages to execute when running a job or creating a server group?

ajordens: Or perhaps the cloud provider could influence whether there's a need for a `waitForUpInstances` and perhaps what the `forceCacheRefresh` payload should be (serverGroupName vs. jobName).

Those seem like the most significant changes between RunJobStage and CreateServerGroupStage?

lwander:
I just realized, there should probably be a "Wait for Job completion" polling stage. That will definitely have to be provider specific, as there are many ways a job can succeed (or fail). For example, you might allow a job to retry until it succeeds, allowing up to N failures.

The thing is, then if we make this an ad-hoc stage under Clusters (or something), then the dialog won't close until the job is done. 

Also, there should probably be some sort of a "Cleanup Job" stage.

ajordens:
Where are we sitting with this @lwander?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.groovy
lwander: I'd rather wait to make those changes when they are necessary, since I can't guess now how other cloud providers will implement this stage. 

Having thought about it, I think this stage should block until the Job either succeeds or fails, since a Job is expected to terminate and its result is valuable (i.e. exit code, data on disk, etc...). That's another departure from `CreateServerGroupStage`, and makes combining the two tricky to reason about.

lwander:
All set on my end. PTAL @duftler @ajordens

duftler:
> The thing is, then if we make this an ad-hoc stage under Clusters (or something), then the dialog won't close until the job is done.

Isn't this consistent with the existing ad-hoc operation? The user can always close the dialog and monitor the in-flight operation under Tasks.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -56,8 +56,16 @@ class MonitorKatoTask implements RetryableTask {
...
 
     Map<String, ? extends Object> outputs = [:]
-    if (status == ExecutionStatus.SUCCEEDED && !stage.context.containsKey("deploy.server.groups")) {
-      outputs["deploy.server.groups"] = getServerGroupNames(katoTask)
+    if (status == ExecutionStatus.SUCCEEDED) {
+      def deployed = getDeployedNames(katoTask)
+      if (!stage.context.containsKey("deploy.server.groups")) {
+        outputs["deploy.server.groups"] = getServerGroupNames(katoTask)
+      }
+      if (!stage.context.containsKey("deploy.jobs") && deployed) {
```
duftler: Can `stage.context` ever contain both `deploy.server.groups` and `deploy.jobs`? If not, this can be `else if`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -56,8 +56,16 @@ class MonitorKatoTask implements RetryableTask {
...
+        if (deployed) {
```
duftler: This `if (deployed)` appears extraneous.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -97,7 +105,7 @@ class MonitorKatoTask implements RetryableTask {
   }
 
   // This is crappy logic
-  // TODO clean this up in Kato
+  // TODO clean this up in Kato (deprecate in favor of getDeployedNames)
```
duftler: Since this is in a comment, why not just make it a javadoc and use `@deprecated` (and indicate what to use instead, as you have already done here)?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobForceCacheRefreshTask.groovy
```diff
@@ -0,0 +1,51 @@
...
+  MortService mort
+
+  @Override
+  TaskResult execute(Stage stage) {
+    String cloudProvider = getCloudProvider(stage)
+    String account = getCredentials(stage)
+
+    stage.context."deploy.jobs"?.each { String location, List names ->
+      names.each { name ->
+        mort.forceCacheUpdate(
```
duftler: Shouldn't this use `oort`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
```diff
@@ -0,0 +1,81 @@
...
+  @Autowired
+  OortService oortService
+
+  @Autowired
+  ObjectMapper objectMapper
+
+  static final String REFRESH_TYPE = "Job"
+
+  @Autowired
+  MortService mort
```
duftler: This doesn't look to be used here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobForceCacheRefreshTask.groovy
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.tasks.job
+
+import com.netflix.spinnaker.orca.DefaultTaskResult
+import com.netflix.spinnaker.orca.ExecutionStatus
+import com.netflix.spinnaker.orca.Task
+import com.netflix.spinnaker.orca.TaskResult
+import com.netflix.spinnaker.orca.clouddriver.MortService
```
duftler: Extraneous import.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
```diff
@@ -0,0 +1,78 @@
...
+
+package com.netflix.spinnaker.orca.clouddriver.tasks.job
+
+import com.fasterxml.jackson.core.type.TypeReference
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.orca.DefaultTaskResult
+import com.netflix.spinnaker.orca.ExecutionStatus
+import com.netflix.spinnaker.orca.RetryableTask
+import com.netflix.spinnaker.orca.TaskResult
+import com.netflix.spinnaker.orca.clouddriver.MortService
```
duftler: Extraneous import.

PR #782, opened 2016-04-06
provider/kubernetes: Match against imageNamePattern

lwander:
Removed this in an earlier PR, but forgot it's needed to disambiguate between multiple images. 

@duftler


duftler:
LGTM.

PR #789, opened 2016-04-20
Fix UpsertServerGroupTagsTask to correctly determine the server group to issue forceCacheRefresh against

cfieber:
@ajordens PTAL


ajordens:
LGTM

PR #793, opened 2016-04-26
core: Send 'region=$region' to clouddriver operations instead of 'reg…

duftler:
…ions=[$region]'.


duftler:
Tested with all manner of AWS and GCE ad-hoc and pipeline operations.
Should be no breaking changes as it will tolerate deck sending both regions/region and zones/zone.

duftler:
@cfieber or @ajordens please review.

cfieber:
LGTM, nice cleanup

ajordens:
Tidy!

PR #799, opened 2016-04-28
Batch separation

ajordens:


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionLauncher.java
```diff
@@ -20,25 +20,33 @@
...
     this.objectMapper = objectMapper;
     this.currentInstance = currentInstance;
+    this.executionRepository = executionRepository;
     this.runner = runner;
   }
 
   public T start(String configJson) throws Exception {
     final T execution = parse(configJson);
+
+    persistExecution(execution);
```
ajordens: This jives with historical behavior where we persisted the skeleton (before stages got blown out) ... the SpringBatchExecutionRunner will re-persist once it has finished building the execution.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionRunnerSupport.java
```diff
@@ -68,6 +63,7 @@ public ExecutionRunnerSupport(Collection<StageDefinitionBuilder> stageDefinition
         task.setId(taskDef.getId());
         task.setName(taskDef.getName());
         task.setStatus(NOT_STARTED);
+        task.setImplementingClass(taskDef.getImplementingClass());
```
ajordens: The original behavior was looking up task implementations by `name` which doesn't really work in practice ... going by `implementingClass` seems a little more consistent.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineLauncher.java
```diff
@@ -81,8 +90,9 @@ private String getString(Map<String, ?> map, String key) {
   }
 
   @Override protected boolean shouldQueue(Pipeline execution) {
-    return startTracker == null ||
-      execution.getPipelineConfigId() == null ||
+    return execution.getPipelineConfigId() != null &&
```
ajordens: This is the correct logic for determining whether something should be queued ... 

queue iff:
- it's an execution of an existing pipeline (ie. has a pipelineConfigId)
- we're supposed to be limiting concurrent execution (this seems a little weird to trust what the user supplied vs. what is configured on the actual pipeline ...)
- we have a startTracker

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionLauncherSpec.groovy
```diff
@@ -102,28 +106,11 @@ class PipelineLauncherSpec extends ExecutionLauncherSpec<Pipeline, PipelineLaunc
...
     then:
+    1 * executionRepository.store(_)
     0 * runner.start(_)
 
     where:
-    config = [pipelineConfigId: "whatever", stages: []]
-    json = objectMapper.writeValueAsString(config)
-  }
-
-  def "does not start pipeline if it does not have a pipeline config id"() {
```
ajordens: I don't believe this is a correct test ... the pipeline *should* start if it doesn't have a pipelineConfigId.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionLauncherSpec.groovy
```diff
@@ -102,28 +106,11 @@ class PipelineLauncherSpec extends ExecutionLauncherSpec<Pipeline, PipelineLaunc
...
-
-    when:
-    launcher.start(json)
-
-    then:
-    0 * runner.start(_)
-
-    where:
-    config = [stages: []]
+    config = [id: "whatever", stages: [], limitConcurrent: true]
```
ajordens: s/pipelineConfigId/id (I guess we use 'id' instead of pipelineConfigId when kicking off pipelines)

orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/SpringBatchExecutionRunner.java
```diff
@@ -42,35 +44,48 @@
...
   public SpringBatchExecutionRunner(
     Collection<StageDefinitionBuilder> stageDefinitionBuilders,
+    ExecutionRepository executionRepository,
     JobLauncher jobLauncher,
     JobRegistry jobRegistry,
     JobBuilderFactory jobs,
     StepBuilderFactory steps,
     TaskTaskletAdapter taskTaskletAdapter,
-    Map<String, com.netflix.spinnaker.orca.Task> tasks) {
+    Collection<com.netflix.spinnaker.orca.Task> tasks
```
ajordens: A collection is fine as there's no unique name that a Task implementation can be referred by ... rather we'll just look it up by `Class`

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -187,7 +191,14 @@ class OperationsController {
   private Map<String, String> startPipeline(Map config) {
     def json = objectMapper.writeValueAsString(config)
     log.info('requested pipeline: {}', json)
-    def pipeline = pipelineStarter.start(json)
+
+    def pipeline
+    if (config.executionEngine == "v2") {
```
ajordens: Just needed some way of identifying a pipeline that should run against the "new" engine (engine isn't quite the right wording for where we're at but what the hell)

PR #800, opened 2016-04-29
Allow to pass multiple packages down to the Bakery in the …

kalymero:
…Bake Stage

When a package match one of the packages coming from the trigger or from the previous stage its name get replaced
with the actual package name. Otherwise its just passed down to the bakery, letting the bakery to resolve it

This is someway related to: https://github.com/spinnaker/rosco/pull/88 to correctly handle multiple packages

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
@@ -140,7 +140,7 @@ class CreateBakeTask implements RetryableTask {
     }
 
     if (!roscoApisEnabled) {
-      bakeRequest = bakeRequest.copyWith(templateFileName: null, extendedAttributes: null)
+      bakeRequest = bakeRequest.copyWith(templateFileName: null, extendedAttributes: null, allowMissingPackageInstallation: null)
     }
 
     return bakeRequest
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
@@ -139,8 +139,10 @@ class CreateBakeTask implements RetryableTask {
       bakeRequest = bakeRequest.copyWith(cloudProviderType: null)
     }
 
+    // The allowMissingPackageInstallation only affect Orca behavior. We don't want to propagate it to any bakery
+    bakeRequest = bakeRequest.copyWith(allowMissingPackageInstallation: null)
     if (!roscoApisEnabled) {
-      bakeRequest = bakeRequest.copyWith(templateFileName: null, extendedAttributes: null, allowMissingPackageInstallation: null)
+      bakeRequest = bakeRequest.copyWith(templateFileName: null, extendedAttributes: null)
     }
 
     return bakeRequest
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
@@ -139,8 +139,9 @@ class CreateBakeTask implements RetryableTask {
       bakeRequest = bakeRequest.copyWith(cloudProviderType: null)
     }
 
-    // The allowMissingPackageInstallation only affect Orca behavior. We don't want to propagate it to any bakery
+    // The allowMissingPackageInstallation only affect Orca behavior. We don't want to propagate it to any bakery.
     bakeRequest = bakeRequest.copyWith(allowMissingPackageInstallation: null)
+
     if (!roscoApisEnabled) {
       bakeRequest = bakeRequest.copyWith(templateFileName: null, extendedAttributes: null)
     }
--- a/orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTaskSpec.groovy
+++ b/orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTaskSpec.groovy
@@ -292,29 +292,6 @@ class CreateBakeTaskSpec extends Specification {
   }
 
   @Unroll
-  def "fails if pipeline trigger or context includes artifacts but no artifact for the bake package"() {
-    given:
-    Pipeline pipelineWithTrigger = new Pipeline.Builder().withTrigger([buildInfo: triggerInfo]).build()
-    Stage stage = new PipelineStage(pipelineWithTrigger, "bake", bakeConfig).asImmutable()
-    bakeConfig.buildInfo = contextInfo
-
-    when:
-    task.execute(stage)
-
-    then:
-    IllegalStateException ise = thrown(IllegalStateException)
-    ise.message.startsWith("Unable to find deployable artifact starting with hodor_ and ending with .deb in")
-
-    where:
-    contextInfo         | triggerInfo
-    null                | buildInfoNoMatch
-    buildInfoNoMatch    | null
-    buildInfoNoMatch    | buildInfoNoMatch
-    buildInfoNoMatch    | invalidArtifactList
-    invalidArtifactList | buildInfoNoMatch
-  }
-
-  @Unroll
   def "fails if pipeline trigger and context includes artifacts have a different match"() {
     given:
     Pipeline pipelineWithTrigger = new Pipeline.Builder().withTrigger([buildInfo: buildInfo]).build()
--- a/orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTaskSpec.groovy
+++ b/orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTaskSpec.groovy
@@ -292,6 +292,29 @@ class CreateBakeTaskSpec extends Specification {
   }
 
   @Unroll
+  def "fails if pipeline trigger or context includes artifacts but no artifact for the bake package"() {
+    given:
+    Pipeline pipelineWithTrigger = new Pipeline.Builder().withTrigger([buildInfo: triggerInfo]).build()
+    Stage stage = new PipelineStage(pipelineWithTrigger, "bake", bakeConfig).asImmutable()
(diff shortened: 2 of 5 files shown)


kalymero:
@SPINNAKER/reviewers PTAL

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -90,62 +90,72 @@ class PackageInfo {
       return request
     }
 
-    String prefix = "${request.package}${versionDelimiter}"
-    String fileExtension = ".${packageType}"
+    def requestPackages = request.package.split(" ")
```
duftler: Please use an explicit type here to fit in with the immediate neighbors.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -90,62 +90,72 @@ class PackageInfo {
...
+        }
+      }
 
-    if (packageName) {
-      request.put('package', packageName)
+      if (packageVersion) {
+        request.put('packageVersion', packageVersion)
+      }
+
+      // If the package cannot be found it's passed down as is
```
duftler: Please add a terminating period to the comment.

duftler: I think it might be helpful to add some more detail to the comment. Maybe something closer to the commit comment itself?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -90,62 +90,72 @@ class PackageInfo {
...
+            request.put('commitHash', commitHash)
+          }
         }
       }
-
-      return request
     }
 
-    throw new IllegalStateException("Unable to find deployable artifact starting with ${prefix} and ending with ${fileExtension} in ${buildArtifacts} and ${triggerArtifacts}. Make sure your deb package file name complies with the naming convention: name_version-release_arch.")
+    request.replace('package', requestPackages.join(" "))
```
duftler: Seems like there are 2 changes in this PR:
1) Allow the package that will be matched to a trigger/build artifact to occupy slots other than just sub 0.
2) Remove the requirement that a matching artifact in fact exists in the context.

Change (1) LGTM. It reads to me like we can make change (1) without change (2). Can you just track whether or not a match was made (I think this is what `if (packageName)` already does) and then throw that exception if no match was made over all of the specified packages? And if a match was made, then the augmented request is returned.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -0,0 +1,65 @@
...
+    when:
+      Map requestMap = packageInfo.createAugmentedRequest(trigger, buildInfo, request)
+
+    then:
+      requestMap.package == result
+
+    where:
+      filename                                      | requestPackage                                 | result
+      [["fileName": "test-package_1.0.0.deb"]]      | "test-package"                                 | "test-package_1.0.0"
+      [["fileName": "test-package_1.0.0.deb"]]      | "another-package"                              | "another-package"
```
duftler: If you make the change I described above, I think this test case needs to be removed.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -140,7 +140,7 @@ class CreateBakeTask implements RetryableTask {
     }
 
     if (!roscoApisEnabled) {
-      bakeRequest = bakeRequest.copyWith(templateFileName: null, extendedAttributes: null)
+      bakeRequest = bakeRequest.copyWith(templateFileName: null, extendedAttributes: null, allowMissingPackageInstallation: null)
```
duftler: `allowMissingPackageInstallation` should be elided whether `roscoApisEnabled` is enabled or not. We don't want to propagate the new bake request attribute `allowMissingPackageInstallation` from orca to the bakery whether it is Rosco or the internal Netflix bakery.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -139,10 +137,10 @@ class PackageInfo {
       }
 
       if (packageName) {
-        // At least one matched package must be present it the context.
-        if (!matched) {
-          notMachedPrefix = prefix
-          notMatchedFileExtension = fileExtension
+
+        // If it hasn't been possible to match a package and allowMissingPackageInstallation is false raise an exception.
+        if (!matched && !request.get("allowMissingPackageInstallation")) {
```
duftler: It's ok as is, but might be even better to build a full list of all package misses instead of throwing on the first miss. Otherwise, the user might be in a situation where they make a config change in response to a thrown exception, and then get a slightly different exception at the same spot on the next iteration.

duftler: If you decide to stick with throwing on the first miss, I think we can probably still drop `matched` entirely and move the `if (!request.get("allowMissingPackageInstallation"))` test within the `if (!triggerArtifact && !buildArtifact)` block.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -90,13 +90,14 @@ class PackageInfo {
       return request
     }
 
-    Boolean matched = false
+    List notMachedPrefixes = []
```
duftler: s/notMachedPrefixes/notMatchedPrefixes

Or maybe `missingPrefixes`?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -164,6 +159,11 @@ class PackageInfo {
       }
     }
 
+    // If it hasn't been possible to match a package and allowMissingPackageInstallation is false raise an exception.
+    if (!notMachedPrefixes.empty && !request.get("allowMissingPackageInstallation")) {
```
duftler: `if (missingPrefixes && !request.allowMissingPackageInstallation))`

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -139,8 +139,10 @@ class CreateBakeTask implements RetryableTask {
       bakeRequest = bakeRequest.copyWith(cloudProviderType: null)
     }
 
+    // The allowMissingPackageInstallation only affect Orca behavior. We don't want to propagate it to any bakery
```
duftler: Please add terminating period to comment.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -139,8 +139,10 @@ class CreateBakeTask implements RetryableTask {
...
+    bakeRequest = bakeRequest.copyWith(allowMissingPackageInstallation: null)
```
duftler: Please insert blank line before the following `if ()`.

orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTaskSpec.groovy
```diff
@@ -292,29 +292,6 @@ class CreateBakeTaskSpec extends Specification {
   }
 
   @Unroll
-  def "fails if pipeline trigger or context includes artifacts but no artifact for the bake package"() {
```
duftler: Can this test case be added back now that the `allowMissingPackageInstallation` property has been introduced, and defaults to `false`?

duftler:
LGTM.
@tomaslin Any concerns?

duftler:
Needs a rebase.

tomaslin:
This should pass a list of packages to the bakery instead of letting the bakery decide. Having a trigger means that you want the exact artifacts built by that trigger. If rosco has the ability to deal with multiple packages, then it should explicitly take these multiple packages. 

I would also like to see this only work when 'allowMultiplePackages' in the config allows this to work. There is an implicit assumption that one build -> one debian package. I am concerned that changing this behavior will impact existing pipelines.

kalymero:
Hi @tomaslin, thanks for commenting, please help me understand this better.

How could this affect existing pipelines? Original behaviour is preserved (the new behavior is guarded by the `allowMissingPackageInstallation`) and this change will only pass to the bakery packages name that are not matched by triggers, *if any*. At this point existing pipelines should have only one package configured and `allowMissingPackageInstallation` is **false** by default.  What am I missing?

To give you some context, the use case is: having multiple triggers for a pipeline that trigger a bake which install multiple packages.  
When a trigger happens only the actual trigger will report a full package name to the bake stage while the other packages will be passed down as unpinned packages to the bakery (which will fetch the latest available version if any or fail the installation due a 404).

tomaslin:
what's the use case for defining multiple debian packages vs. having the package itself define its own dependencies?

tomaslin:
Thank you for addressing the mechanism for a multiple package in your comment. 

There is a fundamental assumption in the Spinnaker UI at least that the build number is associated with the artifact being deployed ( at least at Netflix ). We recommend the use of nebula-ospackage mechanism for multiple package installations, and these packages can define their dependencies pinned to specific numbers. 

Any addition to the open source version of spinnaker around this should be optional. If `allowMissingPackageInstallation` is going to be what is guarding this, then the presence of this should be optional so we can turn it off.

tomaslin:
happy for this to go in if it is optionally enabled

duftler:
@tomaslin Since the `allowMissingPackageInstallation` attribute is optional, and defaults to `false`, are we good to go with this?

tomaslin:
LGTM

duftler:
Sorry, neglected to edit the 'squashed' commit message.

PR #804, opened 2016-05-04
Introduced StageListener and ExecutionListener sans Spring Batch

ajordens:
- Converted all existing listeners to the new interfaces
- Adapters make the simplified listener interfaces backwards compatible with Spring Batch
- Existing test cases have been refactored (or removed in situations where tests were rather worthless)


ajordens:
Tests pass and as good a time as any for some feedback ... PTAL @robfletcher.

My motivation with this change set was two-fold:
- Simplify the listener interfaces and remove all traces of Spring Batch (_largely successful_)
- Maintain backwards compatibility with all existing listeners (_there's some more work required to fully vet the compatibleness but I believe it's functional_)

Prior to getting tests passing, I did have the listeners firing as expected with pipelines launched through the new `Runner`.

I owe some unit tests for `ExecutionPropagationListener` and `StageStatusPropagationListener`. It wasn't worth refactoring the previous tests.

Need to do some additional manual testing against the echo notification and rolling push behaviors.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/listeners/ExecutionListener.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package com.netflix.spinnaker.orca.listeners;
+
+import com.netflix.spinnaker.orca.ExecutionStatus;
+import com.netflix.spinnaker.orca.pipeline.model.Execution;
+import org.springframework.core.Ordered;
+
+public interface ExecutionListener extends Ordered {
+  default void beforeExecution(Persister persister, Execution execution) {
```
robfletcher: Could the Persister be a DI injected dependency instead of a parameter?

ajordens: Ya, it could be now.

Originally I was delegating from the old listener interface to the new and passing as a parameter was trivial.

I'll make a note to refactor this.

PR #806, opened 2016-05-05
update spinnaker dependencies.

cfieber:
includes a fairly significant refactor of the orca-eureka bits due overlap of that functionality moving in kork-core.

updated to export the currentInstanceId as a bean in the application context, to get rid of the direct dependency on eureka's InstanceInfo object
added an InstanceStatusProvider interface to replace the one place we were using DiscoveryClient to look up instance status
added a TestConfiguration in orca-test for common test beans (currently just `currentInstanceId`) and removed EurekaComponents from spring based integration tests

@spinnaker/reviewers PTAL


orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/OrcaConfiguration.groovy
```diff
@@ -83,7 +81,8 @@ class OrcaConfiguration {
     Duration.of(3, MINUTES)
   }
 
-  @Bean Scheduler scheduler() {
+  @Bean(destroyMethod = "")
```
cfieber: newer RxJava scheduler adds a `close` method which Spring was autodiscovering and ruining multiple test case executions

orca-eureka/src/main/java/com/netflix/spinnaker/orca/eureka/DiscoveryStatusPoller.java
```diff
@@ -1,84 +0,0 @@
...
-import org.springframework.beans.factory.annotation.Qualifier;
-import org.springframework.context.ApplicationContext;
-import org.springframework.context.ApplicationListener;
-import org.springframework.context.event.ContextRefreshedEvent;
-import rx.Observable;
-import rx.Scheduler;
-import static com.netflix.appinfo.InstanceInfo.InstanceStatus.UNKNOWN;
-import static java.util.concurrent.TimeUnit.SECONDS;
-
-public class DiscoveryStatusPoller implements ApplicationListener<ContextRefreshedEvent> {
```
cfieber: We are getting reliable events from discovery client as to status changes (at least for application STARTING -> UP) but I need to test this a bit more with a proper eureka server and remote status toggle

orca-web/src/main/groovy/com/netflix/spinnaker/orca/Main.groovy
```diff
@@ -55,6 +56,7 @@ import org.springframework.scheduling.annotation.EnableAsync
 @EnableAutoConfiguration(exclude = [BatchAutoConfiguration, GroovyTemplateAutoConfiguration])
 @EnableBatchProcessing(modular = true)
 @Import([
+  PlatformComponents,
```
cfieber: needed this early in the import list or `@ConditionalOnBean(DiscoveryClient.class)` wasn't evaluating properly (previously we were getting this via autoconfiguration)

orca-core/src/main/groovy/com/netflix/spinnaker/orca/restart/PipelineRestartAgent.groovy
```diff
@@ -49,17 +47,19 @@ class PipelineRestartAgent extends AbstractPollingNotificationAgent {
...
-  private final LookupService discoveryClient
-  private final InstanceInfo currentInstance
+  private final InstanceStatusProvider instanceStatusProvider
+  private final String applicationName
+  private final String currentInstanceId
 
   @Autowired
-  PipelineRestartAgent(ObjectMapper mapper, Client jesqueClient, ExecutionRepository executionRepository, LookupService discoveryClient,
-                       @Qualifier("instanceInfo") InstanceInfo currentInstance) {
+  PipelineRestartAgent(ObjectMapper mapper, Client jesqueClient, ExecutionRepository executionRepository, InstanceStatusProvider instanceStatusProvider,
```
ajordens: FWIW this agent is never enabled (right now) ... the changes make sense.

ajordens:
LGTM.

Pretty sure we only run the `topApplicationExecutionCleanup` agent so the actual usage of instance discovery/eureka status is quite low in `orca`.

PR #810, opened 2016-05-06
updates retrofit okclient construction.

cfieber:
main goal of this change it just to expose some configuration around the okclient creation to experiment with addressing the EOFExceptions.  

* adds extra configuration for okclient connection pool and retry (defaults to the existing default values)
* updates a couple of classes that were not explicitly passing a retrofit client
* fixed a bug where the retrofit client user agent interceptor wasn't actually getting applied


ajordens:
LGTM

PR #814, opened 2016-05-09
Converted a handful of LinearStage's to StageDefinitionBuilders

ajordens:
- Only targeted simple LinearStages with no injectBefore/injectAfter behavior


ajordens:
PTAL @robfletcher

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/pipelines/CreateApplicationStage.groovy
```diff
@@ -17,23 +17,17 @@
...
 @Component
 @CompileStatic
-class CreateApplicationStage extends LinearStage {
-  public static final String PIPELINE_CONFIG_TYPE = "createApplication"
-
-  CreateApplicationStage() {
-    super(PIPELINE_CONFIG_TYPE)
-  }
-
+class CreateApplicationStage implements StageDefinitionBuilder {
```
ajordens: `LinearStageDefinitionBuilder` is an adapter over `LinearStage` that can handle a `StageDefinitionBuilder`.

Will need to revise this strategy a bit when migrating over other types of `StageBuilder`.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/StageBuilderProvider.groovy
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.orca.batch
+
+public interface StageBuilderProvider {
+  Collection<StageBuilder> all()
```
ajordens: The `StageBuilderProvider` is to be used in all situations where we were doing a `getBeansOfType(StageBuilder)`.

It's capable of returning legacy `StageBuilder` and adapted `StageDefinitionBuilder`.

PR #819, opened 2016-05-11
core: Run jobs in parallel

lwander:
@duftler


duftler:
LGTM.

ajordens:
Just expressing some caution ... we're really trying to get away from this style of parallel stages in favor of making everything refId/requisiteRefId-based.

Guessing jobs are created as orchestrations and not pipelines?

lwander:
The idea was to mirror the Deploy pipeline stage - how is that going to change?

ajordens:
Thoughts are that eventually we'll only have a single way of running things
in parallel.

PR #822, opened 2016-05-13
core: Propagate interestingHealthProviderNames from ShrinkClusterStag…

duftler:
…e to synthesized DisableClusterStage.


duftler:
@ajordens or @cfieber please review.

duftler:
This addresses a problem we were seeing where a shrink cluster stage would never get beyond the synthesized disable cluster task when at least one of the condemned server groups had no load balancers associated with it.

ajordens:
Makes sense. LGTM.

PR #826, opened 2016-05-18
change instance discovery calls to not append Description 

tomaslin:
at the end of actions.


tomaslin:
depends on https://github.com/spinnaker/clouddriver/pull/615

PR #830, opened 2016-05-20
SPIN-1767: Handle FAILD_CONTINUE status for stage join points.

zanthrash:
@cfieber PTAL


cfieber:
LGTM thanks for the quick turnaround

PR #831, opened 2016-05-20
Migrated TargetServerGroupLinearStage (and children)

ajordens:
- First real usages of synthetic stages

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupLinearStageSupport.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupLinearStageSupport.groovy
@@ -16,15 +16,17 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support
 
-import com.netflix.spinnaker.orca.batch.StageBuilder
 import com.netflix.spinnaker.orca.kato.pipeline.Nameable
-import com.netflix.spinnaker.orca.pipeline.LinearStage
+import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder
+import com.netflix.spinnaker.orca.pipeline.model.Execution
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
+import static com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder.StageDefinitionBuilderSupport.*
+import static com.netflix.spinnaker.orca.pipeline.model.Stage.SyntheticStageOwner.*
 
 @Slf4j
-abstract class TargetServerGroupLinearStageSupport extends LinearStage implements Nameable {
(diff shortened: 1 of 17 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupLinearStageSupport.groovy
```diff
@@ -34,50 +36,55 @@ abstract class TargetServerGroupLinearStageSupport extends LinearStage implement
 
   String name = this.type
 
-  TargetServerGroupLinearStageSupport(String name) {
-    super(name)
+  @Override
+  def <T extends Execution> List<Stage<T>> aroundStages(Stage<T> parentStage) {
```
ajordens: This logic necessitated a move away from separate `preStages`/`postStages` methods ... although it's probably possible to refactor the creation logic such that we could do a definitive split into pre/post.

Given that these stages include a `SyntheticStageOwner` indicating whether they're `STAGE_BEFORE` or `STAGE_AFTER`, I'm not sure whether a further split into methods is necessary.

That said, I'm not convinced we should be returning concrete `Stage` implementations here.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionRunnerSupport.java
```diff
@@ -32,30 +36,39 @@ public ExecutionRunnerSupport(Collection<StageDefinitionBuilder> stageDefinition
...
     reverse(
-      builder
-        .postStages(stage))
-      .forEach(postStage -> {
-        List<Stage> stages = stage.getExecution().getStages();
-        int index = stages.indexOf(stage);
-        stages.add(index + 1, postStage);
-        planStage(postStage);
-      });
+      allStages.getOrDefault(STAGE_BEFORE, Collections.emptyList())
```
ajordens: There is some historical preference that the order of STAGE_BEFORE's should be reversed.

Aside from digging through code, I'm not sure it's clear to anyone implementing a stage that this is the way it works. I've filed https://jira.netflix.com/browse/SPIN-1768 to circle around on this ... 

https://jira.netflix.com/browse/SPIN-1768

orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/stages/LinearStageDefinitionBuilder.java
```diff
@@ -16,25 +16,50 @@
...
 
   @Override
   public List<Step> buildSteps(Stage stage) {
+    Map<String, List<StageBuilder>> stageBuilders = stageBuilderProvider
+      .all()
+      .stream()
+      .collect(Collectors.groupingBy(StageBuilder::getType));
+
+    Collection<Stage> aroundStages = delegate.aroundStages(stage);
+    aroundStages.forEach(s -> {
```
ajordens: Support synthetic stages in a `StageDefinitionBuilder` that has been adapted to a `StageBuilder`.

ajordens:
PTAL @robfletcher

PR #836, opened 2016-05-23
Handle multiple rpm artifacts in random order

gardleopard:
Because of the `-` delimiter being nondeterministic, we have to have sophisticated parsing to extract the package name from rpm file names. We found this in our installation when a build produced two rpms, myapp and myapp-someotherpackage. When the bakery was told to install myapp, orca matched the myapp-someotherpackage and the bakery installed the wrong package.


duftler:
LGTM.
@tomaslin PTAL.

tomaslin:
LGTM

PR #837, opened 2016-05-23
provider/kubernetes: simplify resolve tag

lwander:
I realized that only a single trigger can start a pipeline at once, so there is no need to search through buildInfo to find a matching trigger.

@duftler


duftler:
LGTM.

PR #841, opened 2016-05-24
Favor namespace, then region, then zone when resolving location.

jtk54:
Favor namespace, then region, then zone when resolving location. @duftler PTAL. @ajordens FYI.


duftler:
LGTM.

ajordens:
For lack of confusion, what would it take to have parameter order match
logical order?

Right now zone is first parameter but namespace takes precedence, etc.

duftler:
Good point.
PR is here: https://github.com/spinnaker/orca/pull/842

PR #843, opened 2016-05-24
provider/kubernetes: Cluster wide operations fix

lwander:
At some point I switched back from region -> namespace, breaking cluster wide ops


lwander:
@duftler

duftler:
LGTM.

ajordens:
This seems rather difficult to read? (and probably not tested?)

lwander:
Yeah you're right on both accounts. @duftler, can we drop `zone` here? That will at least make it a little easier to read. I'll work on tests tomorrow.

duftler:
Would like to leave the `zones` support for a while still. If we remove it, pipelines that are still using the pre-region style will break. Could leave a todo to remove it in a month if you like.

Maybe dropping the use of the chained ternary operators will make it more readable.

PR #844, opened 2016-05-25
Optional stage support

ajordens:
- Parameter or Expression based

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
@@ -87,7 +87,7 @@ class ExecutionPropagationListener extends JobExecutionListenerSupport implement
       jobExecution.exitStatus = ExitStatus.STOPPED
     }
 
-    if (orcaTaskStatus == STOPPED) {
+    if (orcaTaskStatus?.isSuccessful()) {
       orcaTaskStatus = SUCCEEDED
     }
 
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/OptionalStageSupport.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/OptionalStageSupport.groovy
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.orca.pipeline.model
+
+import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor
+
+class OptionalStageSupport {
+  public static Map<String, Class<? extends OptionalStageEvaluator>> OPTIONAL_STAGE_TYPES = [
+    "expression": ExpressionOptionalStageEvaluator
+  ]
+
+  /**
+   * A Stage is optional if it has an {@link OptionalStageEvaluator} in it's context that evaluates {@code true}.
+   */
+  static boolean isOptional(Stage stage) {
+    def optionalType = (stage.context.stageEnabled?.type as String)?.toLowerCase()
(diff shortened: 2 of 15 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/OptionalStageSupport.groovy
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * An {@link OptionalStageEvaluator} that will evaluate an expression against the current execution.
+   */
+  private static class ExpressionOptionalStageEvaluator implements OptionalStageEvaluator {
+    String expression
+
+    @Override
+    boolean isOptional(Stage stage) {
+      def augmentedContext = [:] + stage.context
```
ajordens: Similar to what `ExpressionPreconditionTask` is doing.

ajordens:
Motivation behind this is to allow people to configure pipelines with particular stages being optional (_by parameter or expression_).

Optional stages should be skipped but any subsequent stages will be executed normally.

PTAL @spinnaker/netflix-reviewers / @spinnaker/google-reviewers

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
```diff
@@ -87,7 +87,7 @@ class ExecutionPropagationListener extends JobExecutionListenerSupport implement
       jobExecution.exitStatus = ExitStatus.STOPPED
     }
 
-    if (orcaTaskStatus == STOPPED) {
+    if (orcaTaskStatus?.isSuccessful()) {
```
tomaslin: suspect we would need to modify echo to bypass skipped stages as well. https://github.com/spinnaker/echo/blob/78500ab1bff77bf1dc1072e52bd00c2cada64043/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/AbstractEventNotificationAgent.groovy#L70-L72

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/OptionalStageSupport.groovy
```diff
@@ -0,0 +1,91 @@
...
+      if (stage.execution instanceof Pipeline) {
+        augmentedContext.put('trigger', ((Pipeline) stage.execution).trigger)
```
tomaslin: I think we're doing this in 3 places, should move to a common utils class perhaps

tomaslin:
I think calling the field 'enabled' instead of 'optional' would be clearer, otherwise LGTM

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionPropagationListener.groovy
ajordens: Good call ... don't really want to send notifications for `SKIPPED` stages.

ajordens:
@tomaslin Thoughts some some sort of a prefix to denote context keys as being Spinnaker-only? I worry about something like `optional` or `enabled` in that they're pretty fairly and could just as easily be a result of a `jenkins` stages which would (albeit edge-casey) cause confusion.

tomaslin:
stageEnabled

ajordens:
Was thinking the same thing ... aligns with `stageTimeoutMs`.

cfieber:
think this is missing a change in `WaitForRequisiteStage` to handle the skipped state (@zanthrash just added the `FAILED_CONTINUE` state in there in #830)

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/OptionalStageSupport.groovy
```diff
@@ -0,0 +1,91 @@
...
+   * Determines whether a stage is optional and should be skipped
+   */
+  private static interface OptionalStageEvaluator {
+    boolean isOptional(Stage stage)
+  }
+
+  /**
+   * An {@link OptionalStageEvaluator} that will perform a case-insensitive comparison against a pipeline parameter.
+   */
+  private static class ParameterOptionalStageEvaluator implements OptionalStageEvaluator {
```
cfieber: thoughts on special casing this instead of just using the equivalent expression? `parameters.key == value` as the expression doesn't seem much worse and saves having an extra case both here and in deck

ajordens: That's fair ... chalk it up to starting with the parameter based implementation and tests.

About the only thing it gives you is perhaps a little more flexibility around being case-insensitive on matching. 

I'll strip it out for now.

ajordens:
- removed the explicit parameter support (for now)
- added `SKIPPED` to `WaitForRequisiteStage`
- moved the `augmentedContext` creation to `ContextParameterProcessor#buildExecutionContext` and retrofitted existing usages
- suppressed events for SKIPPED and CANCELED statuses

PR #845, opened 2016-05-25
SPIN-1387 parallel stage execution

robfletcher:


robfletcher:
Updated to make synthetic stages work properly. PTAL @ajordens

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionRunnerSupport.java
```diff
@@ -36,13 +34,27 @@ public ExecutionRunnerSupport(Collection<StageDefinitionBuilder> stageDefinition
...
-  public <T extends Execution> void start(T execution) throws Exception {
+  /**
+   * The default implementation plans all stages.
+   * Implementations may override this method completely or use it to plan the entire execution before running it.
+   *
+   * @param execution the execution instance with skeleton stages.
+   * @param <T>       the execution type.
+   * @throws Exception
+   */
+  @Override public <T extends Execution<T>> void start(T execution) throws Exception {
```
ajordens: Was thinking (the other day) if `start` is the appropriate name for this method vs. `plan`?

It's overridden in subclasses (right?) with a super.start(execution) and then the behavior for actually running it in spring batch.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AbstractStage.groovy
```diff
@@ -27,42 +30,34 @@ import com.netflix.spinnaker.orca.ExecutionStatus
...
   String name
   Long startTime
   Long endTime
   ExecutionStatus status = NOT_STARTED
-  void setStatus(ExecutionStatus status) {
-    this.status = status
-  }
-  @JsonBackReference
-  T execution
+  @JsonBackReference T execution
```
ajordens: Never been a big fan of the annotations on same line as declaration ...

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AbstractStage.groovy
```diff
@@ -27,42 +30,34 @@ import com.netflix.spinnaker.orca.ExecutionStatus
...
+  Collection<String> requisiteStageRefIds = []
+  SyntheticStageOwner syntheticStageOwner
   List<InjectedStageConfiguration> beforeStages = []
   List<InjectedStageConfiguration> afterStages = []
   long scheduledTime
 
   @JsonIgnore
   AtomicInteger stageCounter = new AtomicInteger(0)
 
+  @Deprecated
```
ajordens: Do we not need this anymore?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AbstractStage.groovy
```diff
@@ -98,15 +93,7 @@ abstract class AbstractStage<T extends Execution> implements Stage<T>, Serializa
   }
 
   AbstractStage(Execution execution, String type) {
-    this(execution, type, EMPTY_MAP)
-  }
-
-  @Override
-  Stage preceding(String type) {
```
ajordens: This `preceding` is a horrible method ... only works consistently if the pipeline isn't parallel (which all should).

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -0,0 +1,232 @@
...
+  /**
+   * sets the execution status for this stage
+   */
+  void setStatus(ExecutionStatus status);
+
+  /**
+   * Gets the last stage preceding this stage that has the specified type.
+   */
+  default Stage preceding(String type) {
+    int i = getExecution()
```
ajordens: This is still a poor method.

If anything delegating to `ancestors()` seems more appropriate (provided ancestors handles the situation where there are no refIds on stages).

orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/SpringBatchExecutionRunner.java
```diff
@@ -127,23 +165,124 @@ private Step initStep() {
       .build();
   }
 
-  private <E extends Execution> FlowBuilder<FlowJobBuilder> buildStepsForStage(FlowBuilder<FlowJobBuilder> flow, Stage<E> stage) {
+  private <E extends Execution<E>, Q> FlowBuilder<Q> buildStepsForStageAndDownstream(FlowBuilder<Q> flow, Stage<E> stage, Set<Serializable> alreadyBuilt) {
+    if (alreadyBuilt.contains(stage.getRefId())) {
+      log.info("Already built {}", stage.getRefId());
```
ajordens: DEBUG maybe ... going to get verbose.

ajordens:
LGTM but definitely need to run through a significant variety of pipelines both `linear` and `parallel` to ensure behavior hasn't changed.

Hopefully the generic/stage work won't increase the difficulty in rebasing off master until this is in a position to get merged down.

ajordens:
The direction that this is headed in (_and the general cleanup we've been making along the way_) is great and much needed in `orca`.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AbstractStage.groovy
robfletcher: Probably just my auto-format in IntelliJ

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionRunnerSupport.java
robfletcher: Correct.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AbstractStage.groovy
robfletcher: It's only used in `StageBuilder` so we can remove it once that's redundant.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
robfletcher: Yeah, I don't disagree but it's not a change in behavior here.

orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/SpringBatchExecutionRunner.java
robfletcher: Meant to take that out, my bad

PR #846, opened 2016-05-25
Minor adjustment to how an <optional> expression is interpreted

ajordens:
- Aligns with a UI that now prompts for an expression that determines whether a stage should run (vs determining that a stage should not run)
- Any issue evaluating an expression will result in the stage being executed


ajordens:
image

PR #851, opened 2016-05-26
Don't inject an allowLaunch operation if there is no amiName

cfieber:
This is a legitimate case if issuing a cloneServerGroup wanting to inherit the AMI from the
source serverGroup


ajordens:
LGTM

PR #853, opened 2016-05-27
Invert from and to

gbougeard:
The source commit should be the `from` and the target commit the `to`.
If it's not in the right order, the commits list will be empty (at least when using github)

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
@@ -122,7 +122,7 @@ class GetCommitsTask implements DiffTask {
 
   List getCommitsList(String repoType, String projectKey, String repositorySlug, String sourceCommit, String targetCommit) {
     List commitsList = []
-    List commits = buildService.compareCommits(repoType, projectKey, repositorySlug, [to: sourceCommit, from: targetCommit, limit: 100])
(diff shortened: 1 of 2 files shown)


tomaslin:
need to verify that this behavior is the same in stash

gbougeard:
@tomaslin do you have any news about this?

tomaslin:
@dzapata PTAL

dzapata:
This is what's working for Stash currently.  I'll dig some more and see if it's flipped somewhere in the code.

dzapata:
The stash rest api documentation (https://developer.atlassian.com/static/rest/stash/3.0.1/stash-rest.html) is not clear on this, but I just verified that 'to' = older/source commit and 'from' = newer/target commit.  If it's the reverse in GitHub (haven't verified) the proper fix may be to modify the igor service (https://github.com/spinnaker/igor) to account for this (which is what orca calls for the commit diff).

gbougeard:
https://developer.github.com/v3/repos/commits/#compare-two-commit

gbougeard:
I think the most confusing is the naming of `sourceCommit` and `targetCommit`. In my mind it's make more sense that `sourceCommit` is the `from` and the `targetCommit` is the `to`.

gbougeard:
@dzapata any news about this?

gbougeard:
With the curren implementation in igor, to retrieve the commit history you need to set as from the older commit and as to the newest. So the reverse of stash.

dzapata:
Maybe abstract both of them out and make the igor api use 'oldest' 'newest' and each repo type implements it their way?

gbougeard:
https://github.com/spinnaker/igor/issues/135

PR #854, opened 2016-05-29
Disable before scaling down old server groups in red/black

cfieber:
The existing implementation was an attempt to speed things up by not waiting on the disable step for instances we were about to terminate, but not doing the disable is disruptive to active clients and causes unnecessary errors / retries


ajordens:
LGTM

PR #857, opened 2016-06-02
Fix for build stage with empty package name.

zanthrash:
This is to take care of a use case for Google to allow the build stage to have an empty package name, and not extract a package name from the trigger context or stage generated artifacts.

@duftler PTAL and let me know if this doesn't fix your issue.


duftler:
LGTM.
Thanks for such a quick turnaround.

PR #858, opened 2016-06-03
core/bakery: Also support passing null for package on bake request, i…

duftler:
…n addition to empty string.
Follow-on to #857.


duftler:
@zanthrash please review.

duftler:
The issue with requiring an empty string (or whitespace-only string) to indicate 'no package' is that the bake stage configuration does not include the `package` attribute when the field is left blank.
Would this change break your use-case?

zanthrash:
LGTM

PR #860, opened 2016-06-07
Add bake data to deploy operation information

JargoonPard:
@duftler @spinnaker/netflix-reviewers PTAL


JargoonPard:
@duftler @spinnaker/netflix-reviewers PTAL, this is the last piece that we need to have our initial crack at a working pipeline for bake and deploy to Azure.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/azure/AzureServerGroupCreator.groovy
```diff
@@ -38,6 +39,14 @@ class AzureServerGroupCreator implements ServerGroupCreator {
       operation.credentials = operation.account
     }
 
+    def bakeStage = getPreviousStageWithImage(stage, operation.region)
+
+    if (bakeStage) {
+      operation.image.isCustom = true
+      operation.image.uri = bakeStage?.context?.imageId
```
ajordens: null dereferencing with ? shouldn't be necessary as you've already asserted bakeStage is not null

JargoonPard:
Can I bribe anyone to merge this?  :-)

PR #861, opened 2016-06-10
stop instance service using fully configured Retrofit client

robfletcher:
The tasks that use these services do their own timeout management.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ allprojects {
       exceptionFormat = 'full'
     }
   }
-  
+
   configurations.all {
     exclude group: "javax.servlet", module: "servlet-api"
     exclude group: "org.slf4j", module: "slf4j-log4j12"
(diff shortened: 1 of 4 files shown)


robfletcher:
PTAL @ajordens

ajordens:
Took a look ... I believe the default `OkClient` has retryOnConnectionFailure = true.

In a pinch you could inject `OkHttpClientConfiguration` and just new OkClient(okHttpClientConfiguration.create()).

robfletcher:
ah, right. I should have checked that. I'll update the PR

build.gradle
```diff
@@ -38,7 +38,11 @@ allprojects {
       exceptionFormat = 'full'
     }
   }
-  
+
+  repositories {
+    maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
```
ajordens: Don't need this eh?

ajordens:
LGTM

PR #866, opened 2016-06-21
SPIN-1840: Fix Instance Rebooting

zanthrash:
The WaitForUpInstanceHealthTask & WaitForDownInstanceHealthTasks would never get into a success state when the Aamzon health state (which is always "Unknown") was used in determining health.

This would cause the "Reboot Instance Task" to hang on these tasks.  The reboot of the instance would complete but the task would appear hung, causing end user confusion.

This fix simply allows success of the tasks if the only health state is Unknown.

In the case of multiple health checks:
For "Up" health check it will accept combinations of "Up" and "Unknown" state for success
For "Down" a combination of "Down", "OutOfService", & "Unknown" state for success

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForDownInstanceHealthTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForDownInstanceHealthTask.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.instance
 
+import com.netflix.spinnaker.orca.clouddriver.utils.HealthHelper
 import groovy.transform.CompileStatic
 import org.springframework.stereotype.Component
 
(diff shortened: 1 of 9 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForDownInstanceHealthTask.groovy
```diff
@@ -28,8 +28,10 @@ class WaitForDownInstanceHealthTask extends AbstractWaitForInstanceHealthChangeT
       return true
     }
 
+    boolean allAreUnknown = healthProviders.every {it.state == 'Unknown'}
```
cfieber: I'm not sure we should be using this flag here unless the 'only consider platform health' option is selected

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForUpInstanceHealthTask.groovy
```diff
@@ -28,10 +28,18 @@ class WaitForUpInstanceHealthTask extends AbstractWaitForInstanceHealthChangeTas
       return false
     }
 
-    if (healthProviders.any { it.state != "Up"}) {
-      return false
+    if (allUp(healthProviders) || mixOfUpAndUnknown(healthProviders)) {
```
cfieber: the original looks wrong, but here I think we want at least one UP, and the rest up or unknown. and in the case of only platform health then allow that to be considered up

cfieber:
there is a `HealthHelper` class that has some of the platform health logic rolled up in it, and the WaitForUpInstancesTask (plural) and WaitForAllInstancesDownTask are using those

I wonder if we couldn't just kill off these single instance tasks and use the corresponding multi-instances task to reduce the proliferation of this health checking logic?

zanthrash:
@cfieber PTAL when you have a moment

cfieber:
LGTM, thanks!

PR #867, opened 2016-06-21
Support for paging an application owner (see spinnaker/echo#92)

ajordens:


ajordens:
PTAL @tomaslin

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
```diff
@@ -0,0 +1,59 @@
...
+class PageApplicationOwnerTask implements RetryableTask {
+  long backoffPeriod = 15000
+  long timeout = TimeUnit.MINUTES.toMillis(5)
+
+  @Autowired
+  EchoService echoService
+
+  @Override
+  TaskResult execute(Stage stage) {
+    echoService.create(
```
tomaslin: probably makes sense to move the logic for pdApi key lookup here and keep echo's integration generic

ajordens: I initially had it here ... but opt'd for the generic bit being that we want to page "by application" and orca not having to actually know whether it's pager duty, victor ops, etc.

From echo's perspective, I suspect there could be multiple notification services that can handle an `INCIDENT` (but expect only one would be enabled).

PR #874, opened 2016-07-03
Allow deployment details to be resolved from a parent pipeline.

duftler:


duftler:
@cfieber @ajordens or @tomaslin please review.

tomaslin:
is this behavior also present in https://github.com/spinnaker/orca/blob/2f1fe18906cbbbf31f883fa6d32b1c215f961973/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/AmazonServerGroupCreator.groovy ? Would be weird if the clone and deploy had different behaviors?

ajordens:
I'm not sure there's a good reason to not have `withImageFromPrecedingStage` look to upstream pipelines automatically if there's no image details in the current pipeline?

tomaslin:
how about grandparent pipelines and grandgrandparent pipelines?

tomaslin:
Also feels like if we're looking for deployment details, we should also look at Triggers, I think both Titus and Kubernetes default to trigger for looking for imageIds

ajordens:
@tomaslin I'd be fine with looking all the way up the genealogy for details

duftler:
Updated this to search all the way up the ancestor tree as discussed. The logic is smart about which branch to take if a child pipeline was triggered via a pipeline stage. Searches both individual stage contexts as well as for deployment details in global context.

I didn't touch Titus or Kubernetes. I _think_ Titus will benefit from this somewhat since it delegates to `AmazonServerGroupCreator`. But the logic that explicitly looks in the trigger hasn't been touched.

duftler:
Note that the bulk of the changes are tests.

duftler:
@ajordens, @tomaslin please review.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
```diff
@@ -61,30 +63,80 @@ trait DeploymentDetailsAware {
...
+            return getAncestors(null, parentPipelineExecution)
+          }
+        }
+      }
+
       return []
     }
   }
 
+  private getParentPipelineExecution(def execution) {
```
ajordens: Normally don't use `def` as parameter types ... in this case `Execution` would probably suffice.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
```diff
@@ -61,30 +63,80 @@ trait DeploymentDetailsAware {
...
+    // Want to avoid groovy.lang.MissingPropertyException when execution is an Orchestration.
+    return !(execution in Orchestration) ? execution?.trigger?.parentExecution : null
```
ajordens: Haven't seen `in` before ...  is it equivalent to `instanceof`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
```diff
@@ -61,30 +63,80 @@ trait DeploymentDetailsAware {
...
     Stage stage,
     String targetRegion,
     String targetCloudProvider,
     Closure callback) {
     def result = [:]
     def deploymentDetails = (stage.context.deploymentDetails ?: []) as List<Map>
+
+    if (!deploymentDetails) {
+      // If no deployment details were found in the stage context, check the global context of each pipeline up the tree.
+      def pipelineExecutions = getPipelineExecutions(stage?.execution)
```
ajordens: What does `getPipelineExecutions(null)` give back?

No sure the situation where we'd have a `null` stage?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/CreateServerGroupTaskSpec.groovy
```diff
@@ -76,4 +79,577 @@ class CreateServerGroupTaskSpec extends Specification {
...
+                             parentPipeline:
+                                             | trigger
+                                             v
+                                               childPipeline: manualJudgment -> createServerGroup
+     */
+    given:
+    KatoService katoService = Mock(KatoService)
+    MortService mortService = Mock(MortService)
+    def deployRegion = "us-east-1"
+    def deployConfig = [
```
ajordens: Would be nice to eliminate some of the repeated config in these test cases ... 

Probably beneficial to have some helpers that build these graphs and stitch stages together by refId/requisiteIds.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
ajordens: needs a return type as well

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
duftler: Yup.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
duftler: `execution` can be a `Pipeline` or a `Map`, so my only option is `Object`.

duftler: Seems the most idiomatic Groovy thing here is to drop `def` from the declarations if there isn't a proper type to use: http://groovy-lang.org/style-guide.html#_def_and_type

ajordens: Sorry, I missed the bit about dealing with a `Map` as well. 

That's coming from `execution.trigger.parentExecution`? 

In my mind I was thinking we would fetch the parent pipeline out of the `ExecutionRepository` on-demand rather than always expecting it to be passed around as `parentExecution` ... might just be me tho. 

If we can't do ^^, no change to `ObjectMapper.convert` the parentExecution to a proper Pipeline. I can almost guarantee we're going to get hit with an issue related to the extremely loose typing in here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
ajordens: I think `instanceof` is more consistently used throughout `orca`, particularly for single type comparisons.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
duftler: Makes sense. I'll rework this.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/CreateServerGroupTaskSpec.groovy
duftler: Created helpers to wire up all the test config.

duftler:
Reworked this to drop all the dynamic typing.
PTAL @ajordens .

duftler:
Rebased.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.groovy
```diff
@@ -106,4 +109,10 @@ class CloudDriverConfiguration {
       .build()
       .create(KatoRestService)
   }
+
+  @Bean
+  ObjectMapper pipelineObjectMapper() {
+    new ObjectMapper()
+      .registerModule(new SimpleModule("MyConverterModule").addAbstractTypeMapping(Stage, PipelineStage))
```
ajordens: No need for `OrchestrationStage`? I suppose if we're doing an `instanceof Orchestration; return null` than it's unnecessary?

ajordens: You might have been able to get away with just doing a

`private static final ObjectMapper objectMapper = new ObjectMapper()....` in `DeploymentDetailsAware` and not creating a new bean and passing it all around.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/CreateServerGroupTaskSpec.groovy
```diff
@@ -76,4 +82,484 @@ class CreateServerGroupTaskSpec extends Specification {
...
+
+    def childTrigger = [
+      parentExecution: parentPipeline
+    ]
+    def childPipeline = new Pipeline.Builder().withName("child").withTrigger(childTrigger).build()
+    def manualJudgmentStage = buildStageForPipeline(childPipeline, "manualJudgment")
+
+    def deployConfig = buildDeployConfig(deployRegion, operationCloudProvider)
+    def deployStage = buildStageForPipeline(childPipeline, "createServerGroup", deployConfig)
+    makeDependentOn(deployStage, manualJudgmentStage)
```
ajordens: FWIW ... @robfletcher and I chatted and he's going to come up with a little builder that makes it easier to stitch together pipelines and hierarchical children, etc. 

We do this all to much.

duftler: Sounds good.

duftler:
Dropped DI of `pipelineObjectMapper` in favor of a field declared directly on the trait.

PR #875, opened 2016-07-04
Allow missing build info when resolving missing locations.

varikin:
Build info is not required without `defaultResolveMissingLocations` enabled, so it seems reasonable to not require it with `defaultResolveMissingLocations` enabled as well.


cfieber:
LGTM

PR #876, opened 2016-07-05
(titus) Remove special cases when destroying server group

tomaslin:
It now disables before destroy


tomaslin:
@cfieber PTAL

cfieber:
L
G
T
M

👍

PR #879, opened 2016-07-06
Update logrotate file permissions

robertsigler:
This change will fix problems where logrotate ignores
/etc/logrotate.d/orca due to 'bad file mode.'

For Spinnaker issue [#904](https://github.com/spinnaker/spinnaker/issues/904)


cfieber:
LGTM needs rebase

PR #881, opened 2016-07-07
Support propagation the authentication context of a Manual Judgment s…

ajordens:
…tage's approver

 Handles the scenario where a pipeline was triggered by credentials that did not allow
 access to a particular account.

 In such a situation it would be nice if the pipeline could continue to run under the context
 of the individual that approved a manual judgment step.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
@@ -28,6 +28,7 @@ import com.netflix.spinnaker.orca.pipeline.model.OptionalStageSupport
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import com.netflix.spinnaker.orca.pipeline.model.Task as TaskModel
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
+import com.netflix.spinnaker.orca.pipeline.util.StageNavigator
 import com.netflix.spinnaker.security.AuthenticatedRequest
 import com.netflix.spinnaker.security.User
 import groovy.transform.CompileStatic
@@ -48,15 +49,18 @@ class TaskTasklet implements Tasklet {
   private final ExecutionRepository executionRepository
   private final List<ExceptionHandler> exceptionHandlers
   private final Registry registry
+  private final StageNavigator stageNavigator
 
   TaskTasklet(Task task,
               ExecutionRepository executionRepository,
               List<ExceptionHandler> exceptionHandlers,
-              Registry registry) {
+              Registry registry,
+              StageNavigator stageNavigator) {
     this.task = task
     this.executionRepository = executionRepository
     this.exceptionHandlers = exceptionHandlers
     this.registry = registry
+    this.stageNavigator = stageNavigator
   }
 
   Class<? extends Task> getTaskType() {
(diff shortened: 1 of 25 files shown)


ajordens:
This turned into a bit more than a "quick" hack.

Essentially this allows a pipeline to continue under the authentication context of the individual that approved a manual judgment step. 

Is this a terrible idea?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
```diff
@@ -467,51 +471,6 @@ class JedisExecutionRepository implements ExecutionRepository {
     }
   }
 
-  @Deprecated
-  @CompileDynamic
-  private <T extends Execution> T sortStages(JedisCommands jedis, T execution, Class<T> type) {
```
ajordens: @robfletcher These methods weren't being called so took the liberty of removing

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -54,6 +56,20 @@ class ManualJudgmentStage extends LinearStage implements RestartableStage {
...
+    def stageData = stage.mapTo(StageData)
+    if (stageData.state != StageData.State.CONTINUE || !stage.lastModified?.user || !stageData.propagateAuthenticationContext) {
+      return Optional.empty()
+    }
+
+    def user = new User()
+    user.setAllowedAccounts(stage.lastModified.allowedAccounts)
+    user.setUsername(stage.lastModified.user)
+    user.setEmail(stage.lastModified.user)
+    return Optional.of(user)
```
ttomsu: `return Optional.of(user.asImmutable())`

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -54,6 +56,20 @@ class ManualJudgmentStage extends LinearStage implements RestartableStage {
...
 
+  @Override
+  Optional<User> authenticatedUser(Stage stage) {
+    def stageData = stage.mapTo(StageData)
+    if (stageData.state != StageData.State.CONTINUE || !stage.lastModified?.user || !stageData.propagateAuthenticationContext) {
+      return Optional.empty()
+    }
+
+    def user = new User()
+    user.setAllowedAccounts(stage.lastModified.allowedAccounts)
```
ttomsu: We'll likely need to yank this out once Fiat comes online.

ttomsu:
Interesting idea. Though, if someone without access presses the "Go" button, the whole pipeline fails, right? And without controls on who else can see your applications pipelines, lots of unauthorized people could press the buttons of paused pipelines.

ajordens:
They could click the pipeline and it would fail ... that's the behavior we currently have (which is kind of annoying with locked accounts).

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
ajordens: Certainly ... with fiat checks happening in clouddriver we would only need to inject the user and not the accounts.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
ajordens: Good catch!

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
```diff
@@ -173,9 +177,17 @@ class TaskTasklet implements Tasklet {
     }
 
     try {
-      def currentUser = new User(
+      // An AuthenticatedStage can override the default pipeline authentication credentials
+      def authenticatedUser = stageNavigator.findAll(stage, { Stage ancestorStage, StageBuilder stageBuilder ->
```
cfieber: its not clear what the ordering would be here (its it finding the stages by traversing backwards up the graph to the nearest ancestor?) If that is the case maybe adding a findNearestAncestor on stageNavigator would be useful

cfieber:
I think this makes sense, definite improvement over the current state of things

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
ajordens: Ya, that is what it's doing (traversing up the graph ... equivalent to a nearest ancestor).

PR #887, opened 2016-07-14
Wrap the StageStatus and StageTask propagation listeners in a Composi…

ajordens:
…teStepExecutionListener to ensure consistent execution ordering

Observationally, I'm seeing what looks like the StageTaskPropagationListener running after the StageStatusPropagationListener which results in the stage status not always getting updated correctly.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/OrcaConfiguration.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/OrcaConfiguration.groovy
@@ -20,7 +20,11 @@ import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.orca.batch.ExecutionPropagationListener
 import com.netflix.spinnaker.orca.libdiffs.ComparableLooseVersion
 import com.netflix.spinnaker.orca.libdiffs.DefaultComparableLooseVersion
+import com.netflix.spinnaker.orca.pipeline.model.Stage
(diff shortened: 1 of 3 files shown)


ajordens:

orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/OrcaConfiguration.groovy
```diff
@@ -145,13 +149,51 @@ class OrcaConfiguration {
...
+     (1).beforeTask
+     (2).beforeTask
+     (1).afterTask
+     (2).afterTask
+
+     To achieve this we can overload listeners while providing no-op implementations to
+     ensure a consistent execution ordering.
+     */
+    def compositeStepExecutionListener = new CompositeStepExecutionListener()
+    compositeStepExecutionListener.setListeners([
```
ajordens: On a whim I was able to reproduce the behavior consistently by 

```
def compositeStepExecutionListener = new CompositeStepExecutionListener()
    compositeStepExecutionListener.setListeners([
      new StageTaskPropagationListener(executionRepository),
      new StageStatusPropagationListener(executionRepository)
    ] as StepExecutionListener[])
```

which led me to believe that it has something to do with the `StageStatusPropagationListener.afterTask()` (occasionally?) running before `StageTaskPropagationListener.afterTask()`.

This is not an identification of the root cause but rather an attempt to make consistent to order in which step (_aka task_) listeners execute.

PR #889, opened 2016-07-14
core: Serialize application task & stage

nwwebb:
Created a stage and task for serializing an application. Modeled it closely after the stages and tasks in orca. I tested it by sending a curl command to gate and it properly forwards the request to clouddriver. @lwander @duftler PTAL.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/serialize/SerializeApplicationTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/serialize/SerializeApplicationTask.groovy
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/serialize/SerializeApplicationTask.groovy
```diff
@@ -0,0 +1,52 @@
...
+      .toBlocking()
+      .first()
+    Map outputs = [
+      "notification.type"  : "destroyjob",
+      "kato.last.task.id"  : taskId,
+      "delete.name"        : stage.context.jobName,
+      "delete.region"      : stage.context.region,
+      "delete.account.name": account
+    ]
+    new DefaultTaskResult(ExecutionStatus.SUCCEEDED, outputs)
```
lwander: Put a `return` here

lwander:
LGTM

lwander:
Needs a rebase

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/serialize/SerializeApplicationTask.groovy
```diff
@@ -0,0 +1,52 @@
...
+
+  @Override
+  TaskResult execute(Stage stage) {
+    String cloudProvider = getCloudProvider(stage)
+    String account = getCredentials(stage)
+    def taskId = kato.requestOperations(cloudProvider, [[serializeApplication: stage.context]])
+      .toBlocking()
+      .first()
+    Map outputs = [
+      "notification.type"  : "destroyjob",
```
duftler: This looks like a copy/paste error to me.

duftler: Actually, the three `delete.*` values below do as well.
The idea with the stage outputs is to populate it with contextual information that may be useful to downstream stages (or for script callers, or for troubleshooting, ...). So maybe the name of the application being serialized and related bits?

PR #893, opened 2016-07-15
Handle any exceptions that may arise when canceling a jenkins stage

ajordens:


ajordens:
@robfletcher / @tomaslin --> This fixes the `TERMINAL` pipeline ends up being `CANCELED` problem talked about in the channel.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/JenkinsStage.groovy
```diff
@@ -77,7 +77,11 @@ class JenkinsStage extends LinearStage implements RestartableStage, CancellableS
   CancellableStage.Result cancel(Stage stage) {
     log.info("Cancelling stage (stageId: ${stage.id}, executionId: ${stage.execution.id}, context: ${stage.context as Map})")
 
-    stopJenkinsJobTask.execute(stage)
+    try {
+      stopJenkinsJobTask.execute(stage)
```
cfieber: should we retry at all on this?

cfieber:
LGTM

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/JenkinsStage.groovy
ajordens: In this particular case, a retry would have had 0 effect on the result so I don't think it's actually worthwhile.

Upon digging, I'm going to actually fix `igor` because the cancel/stop behavior was very blatantly broken ~3 months ago.

PR #901, opened 2016-07-22
move quip version determination into a task so it runs at stage execution time not stage evaluation time

cfieber:
the quick patch stage used to do work at stage evaluation time in order to determine how many synthetic stages to add for the rolling patch case

it still does some (it looks up all the instances in the cluster) which is not great, but with this change the resolution of which version to deploy happens in a task

this means that the `bulkQuickPatch` stage now has to handle the "skip if up to date" behaviour, and the retrieval of the version from the context has to traverse back up to the parent stage's context

this change enables a pipeline that has `[some random stages]-->[jenkins stage]-->[quick patch]`

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/quip/TriggerQuipTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/quip/TriggerQuipTask.groovy
@@ -16,13 +16,15 @@
 
 package com.netflix.spinnaker.orca.kato.tasks.quip
 
-import groovy.util.logging.Slf4j
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.orca.DefaultTaskResult
 import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.RetryableTask
 import com.netflix.spinnaker.orca.TaskResult
+import com.netflix.spinnaker.orca.batch.StageBuilder
+import com.netflix.spinnaker.orca.clouddriver.InstanceService
 import com.netflix.spinnaker.orca.pipeline.model.Stage
+import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 import retrofit.RetrofitError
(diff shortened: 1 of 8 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/quip/TriggerQuipTask.groovy
```diff
@@ -31,42 +33,88 @@ import retrofit.client.Client
...
+      try {
+        instanceResponse = instanceService.getCurrentVersion(packageName)
+        version = objectMapper.readValue(instanceResponse.body.in().text, Map)?.version
+        if (version && !version.isEmpty()) {
+          return version
+        }
+      } catch (RetrofitError e) {
+        //retry
+      }
+      sleep(INSTANCE_VERSION_SLEEP)
```
cfieber: not super happy about this but it mirrors what was happening at stage evaluation time
ideally this would be its own retryable task

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/quip/VerifyQuipTask.groovy
```diff
@@ -45,19 +45,18 @@ class VerifyQuipTask extends AbstractQuipTask implements Task {
     String cluster = stage.context?.clusterName
     String region = stage.context?.region
     String account = stage.context?.account
-    String app = stage.context?.application
     Map instances = stage.context?.instances
     ArrayList healthProviders = stage.context?.healthProviders
     Map stageOutputs = [:]
     ExecutionStatus executionStatus = ExecutionStatus.SUCCEEDED
-    if (cluster && region && account && healthProviders != null && app && instances) {
+    if (cluster && region && account && healthProviders != null && instances) {
```
cfieber: elsewhere we are deriving app from cluster name, so I've taken it out where it was being enforced as a required context parameter

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/kato/pipeline/QuickPatchStageSpec.groovy
```diff
@@ -16,302 +16,125 @@
...
-    region = "us-east-1"
-    account = "account"
-    asgNames = ["deck-prestaging-v300"]
-    expectedInstances = ["i-1234": [hostName: "foo.com", healthCheckUrl: "http://foo.com:7001/healthCheck"]]
+  /**
+   * This noops out the spring batch aspects of stage building which aren't really a concern for the purposes
+   * of this test
+   */
+  @InheritConstructors
+  static class NoBatchQuickPatchStage extends QuickPatchStage {
```
cfieber: I could do this with a Spy - I tore this test apart because it was too much Mock magic but this last bit makes sense to keep the test simple

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.groovy
```diff
@@ -90,7 +92,7 @@ interface Stage<T extends Execution> {
    * Gets all ancestor stages that satisfy {@code matcher}, including the current stage.
    * @return
    */
-  List<StageNavigator.Result> ancestors(Closure<Boolean> matcher)
+  List<StageNavigator.Result> ancestors(@ClosureParams(value=SimpleType, options="com.netflix.spinnaker.orca.pipeline.model.Stage, com.netflix.spinnaker.orca.batch.StageBuilder") Closure<Boolean> matcher)
```
cfieber: We should be doing this anywhere we have Closure as a method signature - especially on an abstract class or interface

(alternatively we should be using java8 functional interfaces )

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/kato/tasks/quip/TriggerQuipTaskSpec.groovy
```diff
@@ -170,10 +229,15 @@ class TriggerQuipTaskSpec extends Specification {
...
     account = 'test'
     region = "us-east-1"
-    app = "foo"
     packageName | patchVersion | instances
     null | "1.2" | ["i-1234": ["hostName" : "foo.com"]]
     "bar" | null | ["i-1234": ["hostName" : "foo.com"]]
     "bar" | "1.2" | null
   }
+
+  def "some instances are skipped due to skipUpToDate"() {
```
cfieber: oops - will fill these in

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/kato/pipeline/QuickPatchStageSpec.groovy
cfieber: or not - Spy doesn't lend itself really well to stubbing something out when you want to assert the other mock calls. leaving it as is

cfieber:

PR #904, opened 2016-07-27
Fix for Canaries Timing out

zanthrash:
There is an edge case when a canary is with both a duration and warm-up periods are set the canary times out before completion.
Basically there was no padding built into the stage timeout time.

This add a 2 hour padding to the time out to allow for the canary to complete and report back before timeout.


robfletcher:
LGTM. Sorry I really should have seen that when talking to you about this issue a couple of weeks back.

PR #905, opened 2016-07-27
Fix flaky test

robfletcher:
It could break due to a millisecond boundary happening between invocations of `new Date`. Also mocking a `Clock` is kind of pointless.


robfletcher:
PTAL @tomaslin — this is the test we discussed separated into its own PR

tomaslin:
LGTM

PR #906, opened 2016-07-28
Support for tagging of machine images (AWS-only implementation right …

ajordens:
…now)

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/image/UpsertImageTagsStage.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/image/UpsertImageTagsStage.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline.image;
+
+import com.netflix.spinnaker.orca.clouddriver.tasks.MonitorKatoTask;
+import com.netflix.spinnaker.orca.clouddriver.tasks.image.ImageForceCacheRefreshTask;
+import com.netflix.spinnaker.orca.clouddriver.tasks.image.UpsertImageTagsTask;
+import com.netflix.spinnaker.orca.clouddriver.tasks.image.WaitForUpsertedImageTagsTask;
+import com.netflix.spinnaker.orca.pipeline.LinearStage;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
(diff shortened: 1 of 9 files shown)


ajordens:
Premise is that we'll update tags in the `defaultBakeAccount` and re-share the image to any other accounts that it exists in (to force an update of the tags).

`clouddriver` doesn't currently have force cache refresh support for `Image` so that task is essential a no-op.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/image/MachineImageForceCacheRefreshTask.java
```diff
@@ -0,0 +1,65 @@
...
+import com.netflix.spinnaker.orca.clouddriver.OortService;
+import com.netflix.spinnaker.orca.clouddriver.tasks.AbstractCloudProviderAwareTask;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class MachineImageForceCacheRefreshTask extends AbstractCloudProviderAwareTask implements RetryableTask {
```
tomaslin: Not a fan of calling this MachineImage tagger when everywhere else we call it an 'image'. Would be more consistent if this was just Image.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/AmazonMachineImageTagger.java
```diff
@@ -0,0 +1,150 @@
...
+    MatchedImage matchedImage = findImage(targetMachineImage.imageName);
+
+    AtomicBoolean isUpserted = new AtomicBoolean(true);
+    targetMachineImage.regions.forEach(region -> {
+        List<String> imagesForRegion = matchedImage.amis.get(region);
+        imagesForRegion.forEach(image -> {
+          Map<String, String> allImageTags = matchedImage.tagsByImageId.get(image);
+          targetMachineImage.tags.entrySet().forEach(entry -> {
+            // assert tag equality
+            isUpserted.set(isUpserted.get() && allImageTags.get(entry.getKey()).equals(entry.getValue()));
```
tomaslin: would it make sense to have a set of tags that we won't update such as 'appversion' and 'buildhost'?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/image/UpsertMachineImageTagsTask.java
```diff
@@ -0,0 +1,61 @@
...
+  KatoService kato;
+
+  @Autowired
+  List<MachineImageTagger> machineImageTaggers;
+
+  @Override
+  public TaskResult execute(Stage stage) {
+    String cloudProvider = getCloudProvider(stage);
+
+    MachineImageTagger tagger = machineImageTaggers.stream()
```
tomaslin: feels like it might be useful to have a list of the previous tag values written to the context

ajordens:
Agreed. I'll address the comments in a followup commit to this PR.

ajordens:
This can be run in one of two modes:
- as an API call where the `imageName` is explicitly provided
- within a pipeline (needs a preceding Find Image or Bake Stage)

I'm only looking upstream for the first image (and then applying tags to all regions/accounts that image is available in) ... I was trying to avoid needing users to explicitly specify regions.

Inclined to wait for actual usage before tweaking further (_initial caller will be API anyways_).

PR #911, opened 2016-07-31
provider/google: Implement DeploymentDetailsAware instead of relying …

duftler:
…on one-off logic to resolve images.
core: In DeploymentDetailsAware, don't filter on region if region is unspecified.


duftler:
@lwander or @jtk54 please review.
@cfieber @ajordens @tomaslin fyi: The tweak to region-filtering in `DeploymentDetailsAware`. It allows providers that don't rely on regional images to still employ the common logic.

lwander:
LGTM

jtk54:
LGTMA

PR #914, opened 2016-08-03
Apply artifact filtering to the archived artifacts of a Jenkins stage

ajordens:
- This filtering was previously only applied to incoming trigger artifacts

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/BuildArtifactFilter.groovy
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/BuildArtifactFilter.groovy
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/BuildArtifactFilter.groovy
```diff
@@ -0,0 +1,78 @@
...
+
+package com.netflix.spinnaker.orca.igor
+
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.core.env.Environment
+import org.springframework.stereotype.Component
+
+@Component
+class BuildArtifactFilter {
+  static final String MAX_ARTIFACTS_PROP = BuildArtifactFilter.simpleName + ".maxArtifacts"
```
ajordens: Any concern with the property name changing (other than needing to update our config)?

Not sure anyone in the community was actively tweaking these defaults.

ajordens:
A bit of self-preservation as we encountered a pipeline with two Jenkins jobs that archived a total of 20k files resulting in some weird execution results (_timeouts most noticeably_).

Any reason why we should apply the same filtering to jenkins stage outputs that we do to artifacts on a trigger? cc: @tomaslin @cfieber

FWIW the filtering logic was extracted unchanged from the `OperationsController`.

ajordens:
Verified that the artifact filtering works as I would expect ... also verified that a properties file is always available for retrieval even if it's filtered out of the artifact list.

Sound reasonable to merge this down @tomaslin?

PR #915, opened 2016-08-03
No health indicators are an indicator of an instance being down

ajordens:
Fixes an issue in the `DeregisterInstancesFromLoadBalancerStage` where `interestingHealthProviderNames = ['LoadBalancer']'`.

In this scenario, the instance will have *no* health indicators when it has been removed from the load balancer.

Historically, the `WaitForDownInstancesTask` would short-circuit `true` when there were no health indicators. Of course `WaitForAllDownInstancesTask` didn't do this and was inconsistent.


ajordens:
PTAL @zanthrash -- you most recently changed the health checkers and there was a slight behavior change in `WaitForDownInstancesTask` that makes the `DeregisterInstancesFromLoadBalancerStage` never complete.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForDownInstanceHealthTaskSpec.groovy
```diff
@@ -35,13 +35,13 @@ class WaitForDownInstanceHealthTaskSpec extends Specification {
     ["Discovery"]                 || [health: [[type: "Discovery", state: "Down"]]]                                                                      || true
     ["Discovery"]                 || [health: [[type: "Discovery", state: "OutOfService"]]]                                                              || true
     ["Discovery", "Other"]        || [health: [[type: "Other", state: "Down"]]]                                                                          || true
-    ["Amazon"]                    || [health: []]                                                                                                        || false
```
ajordens: Talked about it offline ... `Amazon` is no different than `LoadBalancer` for the purpose of this test.

PR #916, opened 2016-08-03
Consider `stage.context.region` when determining whether a server gro…

ajordens:
…up has been destroyed or not

Removed the second `ServerGroupCacheForceRefreshTask` to pushing more complication into
clouddriver's ClusterCachingAgent ... specifically around folding in OnDemand changes vs. Edda.

This potentially adds ~20-30s to the overall execution time but is arguably simpler.


ajordens:
This should fix the inconsistencies we've seen when destroying server groups.

image

zanthrash:
LGTM

PR #918, opened 2016-08-05
Make ACATask Stage Restartable

zanthrash:


zanthrash:
@spinnaker/netflix-reviewers ping.  Can someone take a look at this for me.

ajordens:
Needs some rebase love.

Although it's questionable whether restarting a pipeline with a canary would work in practice ... `orca` is dependent on building the "identical" pipeline model and I'm not sure that's likely given the complexity of most canary pipelines.

PR #917, opened 2016-08-05
core/bake: Don't strip extraneous attributes from bake requests sent to Netflix's internal bakery.

duftler:
core/bake: Temporarily comment-out logic that strips extraneous attributes from bake requests sent to Netflix's internal bakery.

PR #922, opened 2016-08-08
core/bake: Don't strip extraneous attributes from bake requests sent …

duftler:
…to Netflix's internal bakery.


duftler:
This re-opens the accidentally-merged #917.

PR #933, opened 2016-08-10
Add xenial OperatingSystem

cfieber:
I don't always edit `@Deprecated` classes, but when I do it's to hard-code more things that should be configuration driven.


tomaslin:
LGTM

PR #935, opened 2016-08-11
Continues ACATask Stage if ACA health is UNKNOWN.

zanthrash:
If continueOnUnhelthy is checked and the ACA has a health status of UNKNOWN the stage
will be marked as FAILED_CONTINUE. Presently this will just throw an IlleagialStateException and cancel the pipeline.

@robfletcher PTAL


robfletcher:
LGTM apart from the grammar in that test name

PR #936, opened 2016-08-11
`FindImageFromTags` should emit `deploymentDetails` to the global con…

ajordens:
…text

- `deck` expects jenkins build info to exist in `deploymentDetails`

image

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/image/FindImageFromTagsTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/image/FindImageFromTagsTask.java
@@ -46,9 +46,11 @@ public class FindImageFromTagsTask extends AbstractCloudProviderAwareTask implem
       .orElseThrow(() -> new IllegalStateException("ImageFinder not found for cloudProvider " + cloudProvider));
 
     StageData stageData = (StageData) stage.mapTo(StageData.class);
+    Collection<ImageFinder.ImageDetails> imageDetails = imageFinder.byTags(stage, stageData.packageName, stageData.tags);
     return new DefaultTaskResult(
       ExecutionStatus.SUCCEEDED,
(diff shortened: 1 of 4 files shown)


ajordens:
Thanks for pointing this out @anotherchrisberry.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/image/FindImageFromTagsTask.java
```diff
@@ -46,9 +46,11 @@ public TaskResult execute(Stage stage) {
       .orElseThrow(() -> new IllegalStateException("ImageFinder not found for cloudProvider " + cloudProvider));
 
     StageData stageData = (StageData) stage.mapTo(StageData.class);
+    Collection<ImageFinder.ImageDetails> imageDetails = imageFinder.byTags(stage, stageData.packageName, stageData.tags);
     return new DefaultTaskResult(
       ExecutionStatus.SUCCEEDED,
-      Collections.singletonMap("amiDetails", imageFinder.byTags(stage, stageData.packageName, stageData.tags))
+      Collections.singletonMap("amiDetails", imageDetails),
+      Collections.singletonMap("deploymentDetails", imageDetails)
```
ajordens: Not a fan of putting more stuff in the global context but this will require some work in `deck` to move away from (_it's an already established pattern so this isn't necessarily making anything worse_).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/image/ImageFinder.java
```diff
@@ -28,5 +29,14 @@
   interface ImageDetails {
     String getImageId();
     String getImageName();
+    JenkinsDetails getJenkins();
+  }
+
+  class JenkinsDetails extends HashMap<String, String> {
```
ajordens: The global context has different serialization requirements but making it a Map skirts all of that.

PR #940, opened 2016-08-15
core: Added a pipeline stage for restoring infrastructure

nwwebb:
Added a pipeline stage for restoring infrastructure. Named this deserialization for now to match the serialization pipeline stage, I will soon change them to restoreSnapshot and saveSnapshot to match front50. PTAL @lwander @duftler

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/serialize/RestoreSnapshotStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/serialize/RestoreSnapshotStage.groovy
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline.serialize
+
+import com.netflix.spinnaker.orca.clouddriver.tasks.MonitorKatoTask
+import com.netflix.spinnaker.orca.clouddriver.tasks.serialize.RestoreSnapshotTask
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/serialize/DeserializeApplicationTask.groovy
```diff
@@ -0,0 +1,52 @@
...
+    String cloudProvider = getCloudProvider(stage)
+    String account = getCredentials(stage)
+    def taskId = kato.requestOperations(cloudProvider, [[deserializeApplication: stage.context]])
+      .toBlocking()
+      .first()
+    Map outputs = [
+      "notification.type"       : "deserializeapplication",
+      "kato.last.task.id"       : taskId,
+      "deserialize.application" : stage.context.applicationName,
+      "deserialize.snapshot"    : stage.context.snapshotId,
```
lwander: Will this effectively be the version/timestamp of the bucket?

lwander:
LGTM

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/serialize/DeserializeApplicationTask.groovy
nwwebb: yeah, I'm not positive what it will look like now (might be a generation or a timestamp). I can change it to a more descriptive name when its fully implemented

ajordens:
Curious ... what are we expecting `deserializeApplication` to do in `clouddriver`? 

By foisting all that logic into `clouddriver`, we miss out on any events that would normally be published from `orca` (_events are never published directly from `clouddriver`_).

nwwebb:
In clouddriver I was planning on having `deserializeApplication` use terraform to restore the infrastructure to the state specified by some snapshot. Maybe I could make it more fine grained (by restoring each resource separately) to publish more of what's happening. Are there any events you were thinking of specifically?

ajordens:
Operationally ... I think you want an event for every resource that's being restored. I'm not quite sure how that would work with `terraform` though.

Right now that happens in `orca` with each stage emitting events to `echo` (_which we then ship off to a couple internal systems for change tracking_).

lwander:
Hey @ajordens, that's possible to do with terraform, but will require a bit of work parsing `terraform plan` output. Do you mind if we drop this in as a TODO? Currently this work is feature flagged, and only supports GCP so I don't think it should get in the way in it's current state. I figure when completed, the stage outputs will be a list/map of resource ids, regions and their types similar to the `deployServerGroup` output.

ajordens:
I think a `TODO` is legitimate ... as it stands I'm not sure we would use such a feature (_even if it had AWS support_) but I do believe a consistent story around eventing is important for any production installation.

PR #941, opened 2016-08-17
Fix account name in security group force cache update.

varikin:
The targets object in the security group stage has a securityGroupName,
region, and accountName which came from the MortService.SecurityGroup
object. When constructing the object being sent to clouddrive,
SecurityGroupForceCacheRefreshTask was trying to access
target.credentials which is always null.

Effectively, on demand caching for all providers was not working since
clouddriver got a null account name. Since security groups are
infrequently upserted, it was not noticed.


varikin:
@spinnaker/reviewers PTAL

lwander:
@ajordens could you double check this?

varikin:
@ajordens PTAL

ajordens:
Do you mind rebasing this @varikin and I'll merge it down

varikin:
Done

PR #944, opened 2016-08-18
core/bake: Add varFileName attribute to bake request.

duftler:


duftler:
@tomaslin @ajordens or @cfieber please review.
We verified a week or two back that new bake request attributes will not break the internal bakery.
This pr relates to https://github.com/spinnaker/rosco/pull/131, but they can be merged in any order.

ajordens:
Should `BakeRequest` start using @JsonAnyGetter/@JsonAnySetter and avoid the need to keep adding attributes that are just passed through to `rosco`?

ajordens:
Otherwise this LGTM

duftler:
I experimented with @JsonAnyGetter/@JsonAnySetter and the problem was that the property naming strategy is ignored for these 'dynamic' properties. So I'd either have to use the snake case naming for just the new dynamic properties (in deck/api callers) or mangle the names myself in the dynamic getter. Seems like that'd be less clean than just including the exact properties in `BakeRequest`. WDYT?

ajordens:
Tough one.

I'd probably go for applying the `LowerCaseWithUnderscoresStrategy.translate()` on the `@JsonAnySetter` and avoid the need to release `orca` whenever a new bake property goes in.

I don't feel overly strong about this ... (_if we could go back in time, I'd suggest making the snake case an internal concern of our bakery and feed regular unprocessed camel-case to rosco_).

duftler:
That's not bad. I'll make that change to the setter.

duftler:
Done.

ajordens:
LGTM!

robfletcher:
This is breaking baking in AWS because we're dumping a spurious `status` property from the context into the bake request.

PR #946, opened 2016-08-19
Handle canary failure case

zanthrash:
We were not handling the FAILURE case for canaries. This was causing an IllegialStateException to be thrown instead fo setting the stage to TERMINAL or FAILED_CONTINUE if 'continueOnUnhealthy'  is true

### DIFF
--- a/orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/CompleteCanaryTask.groovy
+++ b/orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/CompleteCanaryTask.groovy
@@ -32,10 +32,14 @@ class CompleteCanaryTask implements Task {
       return new DefaultTaskResult(ExecutionStatus.CANCELED)
     } else if (canary.canaryResult?.overallResult == 'SUCCESS') {
       return new DefaultTaskResult(ExecutionStatus.SUCCEEDED)
-    } else if (stage.context.continueOnUnhealthy && canary.health.health in ["UNHEALTHY", "UNKNOWN"]) {
-      return new DefaultTaskResult(ExecutionStatus.FAILED_CONTINUE)
+    } else if (canary?.health?.health in ["UNHEALTHY", "UNKNOWN"] || canary.canaryResult?.overallResult == 'FAILURE') {
+      return new DefaultTaskResult(
+        stage.context.continueOnUnhealthy == true
+          ? ExecutionStatus.FAILED_CONTINUE
+          : ExecutionStatus.TERMINAL
+      )
     } else {
-      throw new IllegalStateException("Canary failed")
+      throw new IllegalStateException("Canary in unhandled state")
     }
   }
 }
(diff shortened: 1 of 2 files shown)


orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/CompleteCanaryTask.groovy
```diff
@@ -32,10 +32,12 @@ class CompleteCanaryTask implements Task {
...
     } else if (canary.canaryResult?.overallResult == 'SUCCESS') {
       return new DefaultTaskResult(ExecutionStatus.SUCCEEDED)
-    } else if (stage.context.continueOnUnhealthy && canary.health.health in ["UNHEALTHY", "UNKNOWN"]) {
+    } else if (stage.context.continueOnUnhealthy == true && (canary?.health?.health in ["UNHEALTHY", "UNKNOWN"] || canary.canaryResult?.overallResult == 'FAILURE') ) {
       return new DefaultTaskResult(ExecutionStatus.FAILED_CONTINUE)
+    } else if (canary.canaryResult?.overallResult == 'FAILURE') {
+      return new DefaultTaskResult(ExecutionStatus.TERMINAL)
     } else {
-      throw new IllegalStateException("Canary failed")
+      throw new IllegalStateException("Canary failed: [overallResult: ${canary?.canaryResult?.overallResult}, health: ${canary?.health?.health}, status: ${canary?.status?.status}, continueOnUnhealthy: ${canary?.context?.continueOnUnhealthy}]")
```
ajordens: Does this make it's way to the UI as an error message or anything?

If we're looking at this in logs, I'd suggest including the `executionId` (perhaps more valuable than the other bits as they should be included in the context).

orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/CompleteCanaryTask.groovy
```diff
@@ -32,10 +32,12 @@ class CompleteCanaryTask implements Task {
       return new DefaultTaskResult(ExecutionStatus.CANCELED)
     } else if (canary.canaryResult?.overallResult == 'SUCCESS') {
       return new DefaultTaskResult(ExecutionStatus.SUCCEEDED)
-    } else if (stage.context.continueOnUnhealthy && canary.health.health in ["UNHEALTHY", "UNKNOWN"]) {
+    } else if (stage.context.continueOnUnhealthy == true && (canary?.health?.health in ["UNHEALTHY", "UNKNOWN"] || canary.canaryResult?.overallResult == 'FAILURE') ) {
```
ajordens:
```

orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/CompleteCanaryTask.groovy
zanthrash: Good point

orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/CompleteCanaryTask.groovy
zanthrash: Agreed

PR #947, opened 2016-08-22
Orchestration and Pipeline migration (multi-redis)

ajordens:
These migrators support a migration from <old> redis to <new> redis.


ajordens:
Nothing terribly sophisticated here, a couple simple migrations for `Orchestration` and `Pipeline` that support migrating to a new redis node.

PR #954, opened 2016-08-26
provider/kubernetes: Job stage implemented & tested

lwander:
@duftler 

After talking with @tomaslin we decided not to worry about explicitly caching jobs, and worry only about reporting status back the user. This gets rid of the instance/job confusion.


duftler:
Nice, that's cleaner.
LGTM.

PR #955, opened 2016-08-29
Avoid sending fields to the bakery that we shouldn't

robfletcher:
Because we now use `@JsonAnySetter` we can send all kinds of garbage to the Bakery. Recently we've had problems where `status` gets sent and breaks the bake request.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
@@ -139,6 +139,19 @@ class CreateBakeTask implements RetryableTask {
 
     Map requestMap = packageInfo.findTargetPackage(allowMissingPackageInstallation)
 
+    if (!roscoApisEnabled) {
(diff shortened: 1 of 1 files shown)


robfletcher:
PTAL @cfieber

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -139,6 +139,19 @@ class CreateBakeTask implements RetryableTask {
...
+      requestMap.remove("history")
+      requestMap.remove("id")
+      requestMap.remove("log")
+      requestMap.remove("status")
+      requestMap.remove("statusId")
+      requestMap.remove("statusUri")
+      requestMap.remove("uri")
+    }
+
     return mapper.convertValue(requestMap, BakeRequest)
```
cfieber: what about:
````groovy
def request = mapper.convertValue(requestMap, BakeRequest)
if (!roscoApisEnabled) {
  request.other.clear()
}
return request
````

cfieber:
LGTM

PR #956, opened 2016-08-30
(core) Nest completion details for easy retrieval

lwander:
The idea is to make it easier to render the completed job information in the executionDetails view in deck.

@danielpeach FYI @duftler PTAL


duftler:
LGTM.

PR #957, opened 2016-08-30
Added `DetermineHealthProvidersTask` to stages that otherwise rely on…

ajordens:
… instance health

`deck` will no longer automatically include `interestingHealthProviderNames` when
`platformHealthOnly` is true.

The expectation is that `orca` will determine whether or not `interestingHealthProviderNames`
are required dynamically at the point in time that an individual stage is executed.

Relates to spinnaker/deck#2615

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.tasks;
+
+import com.google.common.collect.ImmutableMap;
+import com.netflix.frigga.Names;
+import com.netflix.spinnaker.orca.DefaultTaskResult;
+import com.netflix.spinnaker.orca.ExecutionStatus;
+import com.netflix.spinnaker.orca.RetryableTask;
+import com.netflix.spinnaker.orca.TaskResult;
+import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.ServerGroupCreator;
+import com.netflix.spinnaker.orca.clouddriver.utils.CloudProviderAware;
+import com.netflix.spinnaker.orca.front50.Front50Service;
+import com.netflix.spinnaker.orca.front50.model.Application;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+@Component
+public class DetermineHealthProvidersTask implements RetryableTask, CloudProviderAware {
+  private static final Logger log = LoggerFactory.getLogger(DetermineHealthProvidersTask.class);
+
+  private final Front50Service front50Service;
+  private final Map<String, String> healthProviderNamesByPlatform;
+
+  @Autowired
+  public DetermineHealthProvidersTask(Front50Service front50Service,
+                                      Collection<ServerGroupCreator> serverGroupCreators) {
+    this.front50Service = front50Service;
+    this.healthProviderNamesByPlatform = serverGroupCreators
+      .stream()
+      .filter(serverGroupCreator ->  serverGroupCreator.getHealthProviderName().isPresent())
+      .collect(
+        Collectors.toMap(
+          ServerGroupCreator::getCloudProvider,
+          serverGroupCreator -> serverGroupCreator.getHealthProviderName().orElse(null)
+        )
+      );
+  }
+
+  @Override
+  public TaskResult execute(Stage stage) {
+    if (stage.getContext().containsKey("interestingHealthProviderNames")) {
+      // should not override any stage-specified health providers
+      return new DefaultTaskResult(ExecutionStatus.SUCCEEDED);
+    }
+
+    String platformSpecificHealthProviderName = healthProviderNamesByPlatform.get(getCloudProvider(stage));
+    if (platformSpecificHealthProviderName == null) {
+      log.warn("Unable to determine platform health provider for unknown cloud provider '{}'", getCloudProvider(stage));
+      return new DefaultTaskResult(ExecutionStatus.SUCCEEDED);
+    }
+
+    try {
+      String applicationName = (String) stage.getContext().get("application");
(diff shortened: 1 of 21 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
```diff
@@ -0,0 +1,99 @@
...
+
+    String platformSpecificHealthProviderName = HEALTH_PROVIDER_NAMES_BY_PLATFORM.get(getCloudProvider(stage));
+    if (platformSpecificHealthProviderName == null) {
+      log.warn("Unable to determine platform health provider for unknown cloud provider '{}'", getCloudProvider(stage));
+      return new DefaultTaskResult(ExecutionStatus.SUCCEEDED);
+    }
+
+    try {
+      String applicationName = (String) stage.getContext().get("application");
+      if (applicationName == null && stage.getContext().containsKey("serverGroupName")) {
```
ajordens: This was sufficient to cover the various scenarios that I tested ... did I miss anything obvious?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
```diff
@@ -0,0 +1,99 @@
...
+        applicationName = Names.parseName((String) stage.getContext().get("serverGroupName")).getApp();
+      } else if (applicationName == null && stage.getContext().containsKey("asgName")) {
+        applicationName = Names.parseName((String) stage.getContext().get("asgName")).getApp();
+      } else if (applicationName == null && stage.getContext().containsKey("cluster")) {
+        applicationName = Names.parseName((String) stage.getContext().get("cluster")).getApp();
+      }
+
+      Application application = front50Service.get("default", applicationName);
+      if (application.platformHealthOnly == Boolean.TRUE && application.platformHealthOnlyShowOverride != Boolean.TRUE) {
```
ajordens: The UI lets you configured this ... 

`platformHealthOnly: true` and `platformHelathOnlyShowOverride: true` which seems counter-intuitive to me.

Right now, if both are set to `true`, we won't automatically inject the `interestingHealthProviderNames`.

ajordens:
PTAL @duftler / @spinnaker/netflix-reviewers 

Took a spin through a variety of `aws` and `titus` use-cases ... both from within pipelines and ad-hoc orchestrations.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.stereotype.Component;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class DetermineHealthProvidersTask implements RetryableTask, CloudProviderAware {
+  private static final Logger log = LoggerFactory.getLogger(DetermineHealthProvidersTask.class);
+  private static final Map<String, String> HEALTH_PROVIDER_NAMES_BY_PLATFORM = ImmutableMap
```
duftler: Wonder if it would be more helpful to build this list by adding a `getHealthProviderName()` function to each `ServerGroupCreator` and then autowiring the list of server group creators in here. Asking for the self-serving reason that I can then reuse that to solve a problem here: https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/WaitForClusterDisableTask.groovy#L60-L67
I want to be able to determine the platform health type even when a server group has zero instances.

Can always build the list from the set of server group creators and just delete kubernetes from that list, since it is not presently desired in the `HEALTH_PROVIDER_NAMES_BY_PLATFORM` list.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
duftler: Maybe it's the way the name `platformHealthOnlyShowOverride` can be read. I intended this to read:
platformHealthOnly -> showOverride
not
platformHealth -> onlyShowOverride
The goal was to allow the application-level `platformHealthOnly` setting to be overridden on any stage/operation.

duftler: I can't think of any situation that would require anything outside of deck itself to consider `platformHealthOnlyShowOverride`. It was intended only to control the rendering of the checkbox.

ajordens: Sure ... I think it gets tricky with:

`platformHelathOnly: true` and `platformHealthOnlyShowOverride: true` and then the user actually deselects the "Use Platform Health" checkbox when the go to run the operation.

In such a scenario, `interestingHealthProviderNames` would not come across in the request.

This is quite an edge-case ... one that I'm not too sure is all that relevant. Let me do some quick analysis on all our applications to see how many have `platformHealth: true` and `platformHealthOnlyShowOverride: true`. I still think it's confusing that we allow both to be selected.

Are you thinking we should update deck and prevent both options from being selected? I could get behind that.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
ajordens: I could get behind this ... although maybe it seems like something that should be sourced from `clouddriver`? 

Kubernetes was excluded because it appeared to have a few possible options for platform health that were largely task-specific.

ajordens:
@duftler Pushed an update that adds 

```

File: orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCreator.groovy [hunk 4]
```diff
@@ -40,4 +40,9 @@ interface ServerGroupCreator {
    * @return The cloud provider type that this object supports.
    */
   String getCloudProvider()
-}
\ No newline at end of file
+
+  /**
+   * @return The platform health provider name for this cloud provider
+   */
+  Optional<String> getHealthProviderName()
```
duftler: Nice.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/model/Application.groovy
```diff
@@ -29,6 +29,8 @@ class Application {
   String accounts
   String updateTs
   String createTs
+  public Boolean platformHealthOnly;
```
duftler: All that java 8 has foisted scope modifiers and semi-colons on you.

ajordens: Needs the scope modifier because it's accessed from java ... but doesn't need the semi-colons.

These jackson classes are somewhat bogus, I'll make everything public for consistency.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
duftler: This looks right to me.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
duftler: Ok, I get your point now. I think the way you've done it here makes sense.

duftler:
LGTM.

PR #959, opened 2016-09-01
(core/WaitForClusterDisable) If platform health name cannot be determ…

duftler:
…ined from instances, try to resolve it from cloud provider.


duftler:
@ajordens please review.

ajordens:
LGTM

PR #960, opened 2016-09-02
provider/core: changed name from serialize application to save snapshot

nwwebb:
Changed the name of the stage and task from serializeApplication to saveSnapshot. PTAL @lwander


lwander:
LGTM

PR #961, opened 2016-09-02
Convert linear pipelines to their parallel equivalents

ajordens:


ajordens:
PTAL @robfletcher 

Initially I did this in the `PipelineStarter` ... but ran into a handful of failing test cases that I'm sure you also addressed in the `batch-separation` branch when you took out linear support.

 Anyhow, the `OperationsController` should catch anyone that is programmatically submitting linear pipelines.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -66,6 +66,11 @@ class OperationsController {
       throw new DisabledPipelineException("Pipeline is disabled and cannot be started.")
     }
 
+    def parallel = pipeline.parallel as Boolean
+    if (!parallel) {
+      pipeline = convertLinearToParallel(pipeline)
```
robfletcher: it looks like it mutates in place, do you need to reassign?

robfletcher:
LGTM

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ajordens: fair comment, removed the return type to prevent confusion.

PR #962, opened 2016-09-07
Adds permission calls during create/update/delete of application config

ttomsu:
@jtk54 @duftler PTAL

@ajordens, @cfieber - this PR relies on https://github.com/spinnaker/front50/pull/129, so Front50 will need to go out before Orca in your next release.

Closes https://github.com/spinnaker/fiat/issues/43


jtk54:
LGTM

ajordens:
Should front50 itself just delete permissions when an app is removed?

We've done that for pipelines, notifications, etc.

ajordens:
Just looking at the PR on my phone but I don't grok the Application.user?
Is that supposed to be the logged in user?

At what point at the required groups distinguished by read vs write?

Also, I think we've been referring to an unknown user as "anonymous" vs
"unknown".

PR #963, opened 2016-09-07
failPipeline/continuePipeline should be propagated to monitorCanaryStage

ajordens:
relates to spinnaker/deck#2668


ajordens:
Fixes an issue where canary failures weren't overridable.

Specifically, there's a desire to continue even in the event of a failure.

PR #965, opened 2016-09-08
listeners promote stage and task status correctly in v1 and v2

robfletcher:
SPIN-1990

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/listeners/StageStatusPropagationListener.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/listeners/StageStatusPropagationListener.groovy
@@ -41,7 +41,9 @@ class StageStatusPropagationListener implements StageListener {
   @Override
   void afterTask(Persister persister, Stage stage, Task task, ExecutionStatus executionStatus, boolean wasSuccessful) {
     if (executionStatus) {
-      def nonBookendTasks = stage.tasks.findAll { !it.bookend }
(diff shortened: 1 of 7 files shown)


robfletcher:
PTAL @ajordens

orca-core/src/main/groovy/com/netflix/spinnaker/orca/listeners/StageStatusPropagationListener.groovy
```diff
@@ -41,7 +41,9 @@ class StageStatusPropagationListener implements StageListener {
   @Override
   void afterTask(Persister persister, Stage stage, Task task, ExecutionStatus executionStatus, boolean wasSuccessful) {
     if (executionStatus) {
-      def nonBookendTasks = stage.tasks.findAll { !it.bookend }
+      def nonBookendTasks = stage.tasks.findAll {
+        !it.bookend && !it.stageStart && !it.stageEnd
```
ajordens: Do we actually need `bookend` any more?

ajordens:
LGTM

orca-core/src/main/groovy/com/netflix/spinnaker/orca/listeners/StageStatusPropagationListener.groovy
robfletcher: `stageStart` and `stageEnd` will _not_ be set in old mode. In that case it will rely on `bookend`. Ultimately that should go away.

PR #968, opened 2016-09-12
Various fixes for deploy stages

ajordens:


ajordens:
PTAL @robfletcher

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingPushStrategy.groovy
```diff
@@ -83,6 +74,15 @@ class RollingPushStrategy implements Strategy {
       )
     }
 
+    stages << newStage(
```
ajordens: This reversal of stage order matches what is happening on `master`.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionRunnerSupport.java
```diff
@@ -95,6 +86,7 @@ public ExecutionRunnerSupport(Collection<StageDefinitionBuilder> stageDefinition
 
       // main stage name can be overridden
       stage.setName(branchBuilder.parallelStageName(stage, parallelStages.size() > 1));
+      stage.setInitializationStage(true);
```
ajordens: This governs whether or not this stage is displayed in `deck`

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionRunnerSupport.java
```diff
@@ -113,13 +105,29 @@ public ExecutionRunnerSupport(Collection<StageDefinitionBuilder> stageDefinition
         branchBuilder.postBranchGraph(stage, postBranch);
       });
       callback.accept(singleton(stage), afterGraph);
+
+      // ensure parallel stages have the correct stage type (ie. createServerGroup -> deploy to satisfy deck)
+      parallelStages.forEach(it -> it.setType(branchBuilder.getType()));
```
ajordens: For BranchingStageDefinitionBuilder, the generated parallel stages may (in the case of deploy) need a different stage type.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionRunnerSupport.java
```diff
@@ -113,13 +105,29 @@ public ExecutionRunnerSupport(Collection<StageDefinitionBuilder> stageDefinition
...
     } else {
+      Map<SyntheticStageOwner, List<Stage<T>>> aroundStages = builder
+        .aroundStages(stage)
+        .stream()
+        .collect(groupingBy(Stage::getSyntheticStageOwner));
+
+      if (stage.getExecution().getBuiltPipelineObjects().contains(stage)) {
+        aroundStages = new HashMap<>();
```
ajordens: Avoid duplicate processing of before / after stages ... previously you would get:

`Deploy Stage -> Disable -> Disable -> Shrink`

when it should have been

`Deploy Stage -> Disable -> Shrink`

Not sure this is the ideal fix, but it does generate the expected graph shape.

robfletcher: do we need to attach the cache to the execution rather than just keeping it in a field in this class?

ajordens: The execution bit was something we were doing on the old model ... the `ExecutionRunnerSupport` is a component so it's life cycle doesn't really jive with per-execution/

robfletcher: Oh right,  I forgot it's not prototyped.

robfletcher:
LGTM, let's merge it

PR #969, opened 2016-09-12
Propagate status correctly on parallel tasks

robfletcher:
- inject a no-op task if there are no tasks before the parallel part of the stage
- fix handling of status propagation in v2

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/TaskNode.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/TaskNode.java
@@ -24,6 +24,7 @@ import java.util.function.Consumer;
 import com.google.common.annotations.VisibleForTesting;
 import com.netflix.spinnaker.orca.ExecutionStatus;
 import lombok.Value;
+import static com.netflix.spinnaker.orca.pipeline.TaskNode.GraphType.LOOP;
 
 /**
  * A node in a {@link TaskGraph} which can be either an individual task or a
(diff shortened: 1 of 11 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/TaskNode.java
```diff
@@ -24,26 +24,40 @@
...
 /**
  * A node in a {@link TaskGraph} which can be either an individual task or a
  * sub-graph.
  */
 public interface TaskNode {
 
-  static TaskGraph build(Consumer<Builder> closure) {
-    Builder builder = new Builder();
+  static enum GraphType {
+    full, loop, head, tail
```
ajordens: lowercase enum threw me for a loop when reading this ... I'd suggest capitalized as they're constants.

orca-test/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionRunnerSpec.groovy
```diff
@@ -445,6 +446,41 @@ abstract class ExecutionRunnerSpec<R extends ExecutionRunner> extends Specificat
     contexts = [[region: "a"], [region: "b"]]
   }
 
+  def "does not consider before tasks of an internal parallel stage to complete the stage"() {
```
ajordens: similarly the stage isn't complete until the post tasks of a parallel stage complete?

ajordens:
Beyond changing the enums, this seems reasonable tom e.

orca-test/src/main/groovy/com/netflix/spinnaker/orca/pipeline/ExecutionRunnerSpec.groovy
robfletcher: yeah, needs a test, I'll add it

PR #973, opened 2016-09-15
(core) wait for server group during up instance check

lwander:
It's possible to have the "wait for up instances to fail" task if a server group isn't caught in the "force cache refresh" preceding it. This is easy to verify by modifying deploy handler to deploy nothing (but report deployment details), causing the "force cache refresh" task to simply report that the server doesn't exist (and attempt to evict it from the cache), which succeeds happily. Finally "wait for up instances" tries to verify that the server group exists, which throws an exception.

This only happens when a no blocking action happens during a deploy stage, and takes around a minute after the deploy for the server group to even show up (ignoring instances).

@ajordens @duftler

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
@@ -32,10 +32,14 @@ import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import retrofit.RetrofitError
 
+import java.util.concurrent.TimeUnit
+
 @Slf4j
 abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask implements RetryableTask {
-  long backoffPeriod = 10000
-  long timeout = 7200000
+  long backoffPeriod = TimeUnit.SECONDS.toMillis(10)
+  long timeout = TimeUnit.HOURS.toMillis(2)
+  long serverGroupWaitTime = TimeUnit.MINUTES.toMillis(10)
+  Long serverGroupReportedMissing = null
 
   @Autowired
   OortService oortService
@@ -58,6 +62,13 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
   abstract protected Map<String, List<String>> getServerGroups(Stage stage)
 
   abstract protected boolean hasSucceeded(Stage stage, Map serverGroup, List<Map> instances, Collection<String> interestingHealthProviderNames)
+
+  // When waiting for up instances during a "Deploy" stage, it is OK for the server group to not exist coming into this
+  // task. Instead of failing on a missing server group, we retry the stage until it either succeeds, fails, or times out.
+  // For the other uses of this task, we will fail if the server group doesn't exist.
+  boolean waitForUpServerGroup() {
+    return false
+  }
 
   @Override
   TaskResult execute(Stage stage) {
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
```diff
@@ -59,6 +59,11 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
 
   abstract protected boolean hasSucceeded(Stage stage, Map serverGroup, List<Map> instances, Collection<String> interestingHealthProviderNames)
 
+  // When waiting for up instances during a "Deploy" stage, it is OK for the server group to not exist coming into this
+  // task. Instead of failing on a missing server group, we retry the stage until it either succeeds, fails, or times out.
+  // For the other uses of this task, we will fail if the server group doesn't exist.
+  abstract boolean waitForUpServerGroup()
```
duftler: Why not actually implement this method here and return `false`? Then you only need to override it in the one subclass that returns `true`.

lwander: TIL OOP

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
```diff
@@ -119,7 +126,16 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
         }
       }
 
-      verifyServerGroupsExist(stage)
+      try {
+        verifyServerGroupsExist(stage)
+      } catch (MissingServerGroupException e) {
+        if (waitForUpServerGroup()) {
```
ajordens: Seems reasonable ... it's probably overkill to wait for up to 2hrs for a server group to exist.

I'd suggest short circuiting to failure after 5 or 10 minutes of waiting.

lwander: Yeah makes sense, should the timeout for the whole abstract task be changed? Or just wait for up instances? Or do are theses classes instantiated once per stage, and keep their state through the execution allowing me to timeout only when the server group is missing?

lwander: I figured out they maintain state during the execution of the task, and added a 10 minute timeout.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
```diff
@@ -59,6 +62,16 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
...
   abstract protected boolean hasSucceeded(Stage stage, Map serverGroup, List<Map> instances, Collection<String> interestingHealthProviderNames)
 
+  // When waiting for up instances during a "Deploy" stage, it is OK for the server group to not exist coming into this
+  // task. Instead of failing on a missing server group, we retry the stage until it either succeeds, fails, or times out.
+  // For the other uses of this task, we will fail if the server group doesn't exist.
+  boolean waitForUpServerGroup() {
+    return false
+  }
+
+  long serverGroupWaitTime = TimeUnit.MINUTES.toMillis(10)
```
ajordens: I'd maybe move these up to the top below the other backoff/timeout related properties.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
ajordens: 10 minutes sounds good to me, I take it you gave this change a whirl locally?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
lwander: Sure thing

lwander:
Yup I've tested locally with and without the waiting happening.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
lwander: Yup, works with and without waiting

ajordens:
:shipit:

PR #977, opened 2016-09-16
Rolling push adapter for v1

robfletcher:


robfletcher:
PTAL @ajordens

orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/stages/RollingPushStageDefinitionBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * A listener that resets the task status of everything in a loop so that it can be re-run without interfering
+   * with restart semantics.
+   */
+  static class RedirectResetListener implements StageListener {
+    @Override
+    public <T extends Execution<T>> void afterTask(Persister persister, Stage<T> stage, Task task, ExecutionStatus executionStatus, boolean wasSuccessful) {
+      if (executionStatus == REDIRECT) {
+        // TODO: same hideous index reference was in v1
+        stage.getTasks().subList(1, 11).forEach(it -> {
```
robfletcher: This sucks but it's the same on master. The new way of doing loops doesn't hardcode indexes.

ajordens: Sorry, I must be missing something ... what's up with the `1, 11`? Can you link to where we were doing it in master?

robfletcher: https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/RollingPushStage.groovy#L103

Indexes are different because the old one is an inclusive index.

orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/stages/RollingPushStageDefinitionBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+   * with restart semantics.
+   */
+  static class RedirectResetListener implements StageListener {
+    @Override
+    public <T extends Execution<T>> void afterTask(Persister persister, Stage<T> stage, Task task, ExecutionStatus executionStatus, boolean wasSuccessful) {
+      if (executionStatus == REDIRECT) {
+        List<Task> tasks = stage.getTasks();
+        int startIndex = Iterables.indexOf(tasks, it -> it.getName().equals("determineCurrentPhaseTerminations"));
+        int endIndex = Iterables.indexOf(tasks, it -> it.getName().equals("checkForRemainingTerminations"));
+        tasks.subList(startIndex, endIndex + 1).forEach(it -> {
```
robfletcher: That's a little better than hardcoding the indexes.

PR #978, opened 2016-09-16
Persist `executionEngine` as part of the Execution

ajordens:
- Dependent pipelines should be launched with the executionEngine of their parent pipeline


ajordens:
PTAL @robfletcher

PR #980, opened 2016-09-19
Adjust timeout on GetCommits to handle situations where an HTTP call …

ajordens:
…might time out after 60s


ajordens:
PTAL @zanthrash, I think this might help out with SPIN-2006.

If the underlying issue was that of an HTTP timeout (I believe that's 60s) than the previous task timeout of 60s would not have been long enough.

It looks like the `GetCommitsTask` has logic to only make 3 attempts before auto-succeeding, happy scenario would see this succeeding or failing within seconds. In the event of an http timeout, I think we need the task timeout to be at least be 3 x http timeout.

We might also consider adding `TaskTasklet` support for overriding the result of certain tasks such that they **always** complete (_really only GetCommits and JarDiff_).

PR #981, opened 2016-09-19
Made cancel callback work with v2 engine (and v1)

robfletcher:
Fixes SPIN-2018

### DIFF
--- a/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTaskSpec.groovy
+++ b/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTaskSpec.groovy
@@ -55,7 +55,7 @@ class AbstractClusterWideClouddriverTaskSpec extends Specification {
       ]
     ])
 
-    (pipeline.stages[0] as AbstractStage).stageNavigator = new StageNavigator(Mock(ApplicationContext))
+    (pipeline.stages[0] as AbstractStage).stageNavigator = new StageNavigator(Stub(ApplicationContext))
     pipeline.stages[0].parentStageId = pipeline.stages[1].id
     pipeline.stages[1].requisiteStageRefIds = ["2"]
     pipeline.stages[2].refId = "2"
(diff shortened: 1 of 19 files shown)


orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTaskSpec.groovy
```diff
@@ -55,7 +55,7 @@ class AbstractClusterWideClouddriverTaskSpec extends Specification {
       ]
     ])
 
-    (pipeline.stages[0] as AbstractStage).stageNavigator = new StageNavigator(Mock(ApplicationContext))
+    (pipeline.stages[0] as AbstractStage).stageNavigator = new StageNavigator(Stub(ApplicationContext))
```
robfletcher: So it turns out that when using `Mock` a call to a method that returns `Map` such as `ApplicationContext.getBeansOfType` will return `null` whereas when using a `Stub` it will return an empty map. Without this change (and the same change in a bunch of other files in this PR) `StageNavigator.stageBuilders` was NPEing all over the place.

robfletcher:
@ajordens PTAL

I didn't eliminate the `CancellableStage` interface as we discussed as I didn't need to. Still think it's a good idea long term, though.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/WaitStage.java
```diff
@@ -16,15 +16,23 @@
...
 import com.netflix.spinnaker.orca.pipeline.model.Execution;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
 import com.netflix.spinnaker.orca.pipeline.tasks.WaitTask;
 import org.springframework.stereotype.Component;
+import static java.lang.System.currentTimeMillis;
+import static java.util.Collections.singletonMap;
 
 @Component
-public class WaitStage implements StageDefinitionBuilder {
+public class WaitStage implements StageDefinitionBuilder, CancellableStage {
```
ajordens: Bingo Bango.

robfletcher: fixed

ajordens:
Ship it!

PR #982, opened 2016-09-20
Handle an NPE that chex just identified

ajordens:


ajordens:
FYI @jeyrschabu ... this was throwing an NPE when we attempted to delete a non-existent application (which `chex` can do).

It didn't happen before as the empty `return` simply existed the `.each {}` whereas it actually returns `null` now.

PR #983, opened 2016-09-20
Adds Fiat integration annotations.

ttomsu:
I haven't tested all of these endpoints yet, but Orca/Fiat integration went much smoother than I thought it would.

@jtk54 @duftler @ajordens @cfieber PTAL

Closes https://github.com/spinnaker/fiat/issues/37


duftler:
LGTM.

PR #984, opened 2016-09-21
handles the case of a join point waiting on a STOPPED requisite stage

cfieber:


ajordens:
LGTM

PR #985, opened 2016-09-21
Child pipelines should be canceled whenever a parent pipeline is canc…

ajordens:
…eled


ajordens:

PR #987, opened 2016-09-23
Specify services.fiat.baseUrl.

duftler:
@ttomsu Going to push this through temporarily so Orca can start up.
Not sure how you intend to configure this, and guessing Clouddriver might need the same treatment.


duftler:
Closes https://github.com/spinnaker/spinnaker/issues/1166.

PR #992, opened 2016-09-26
Only no-op a previously completed task if the stage is NOT optional

ajordens:
Handles the following situation:

`Jenkins` -> `Manual Judgment` -> `Deploy`

- `Manual Judgment` is conditional on `Jenkins` being FAILED_CONTINUE.
- `Manual Judgment` is set to not fail the pipeline on failure

A problem arises if `Jenkins` is FAILED_CONTINUE and `Manual Judgment` is STOPPED.

If `Jenkins` is restarted and passes, `Manual Judgment` should be skipped ... even if it
was STOPPED previously.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
@@ -77,8 +77,8 @@ class TaskTasklet implements Tasklet {
         setStopStatus(chunkContext, ExitStatus.STOPPED, ExecutionStatus.CANCELED)
         contribution.exitStatus = ExitStatus.STOPPED
         return cancel(stage)
-      } else if (task.status.complete || task.status.halt) {
-        // no-op
+      } else if (!OptionalStageSupport.isOptional(stage) && (task.status.complete || task.status.halt)) {
+        // no-op if task is already complete AND stage is NOT optional
         log.warn "Skipping task $task.name because its status is $task.status"
         chunkContext.stepContext.stepExecution.executionContext.put("orcaTaskStatus", task.status)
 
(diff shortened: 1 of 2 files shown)


ajordens:

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
```diff
@@ -77,8 +77,8 @@ class TaskTasklet implements Tasklet {
         setStopStatus(chunkContext, ExitStatus.STOPPED, ExecutionStatus.CANCELED)
         contribution.exitStatus = ExitStatus.STOPPED
         return cancel(stage)
-      } else if (task.status.complete || task.status.halt) {
-        // no-op
+      } else if (!OptionalStageSupport.isOptional(stage) && (task.status.complete || task.status.halt)) {
```
ajordens: The issue is that a subsequent restart would see the tasks as complete even if the overall stage became optional.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTaskletSpec.groovy
```diff
@@ -118,6 +118,24 @@ class TaskTaskletSpec extends Specification {
...
+    ]
+    executionRepository.store(pipeline)
+
+    when:
+    tasklet.execute(stepContribution, chunkContext)
+
+    then:
+    0 * task.execute(_)
+
+    stepContext.stepExecution.executionContext.get("orcaTaskStatus") == SKIPPED
```
robfletcher: Does it then update the STOPPED tasks to a different status? Should it?

ajordens: Ya ... the listeners would kick in at this point and propagate the `orcaTaskStatus` ... expectation is that the stage would be re-evaluated on restart and STOPPED -> SKIPPED.

robfletcher: Sure the stage, but what about the individual tasks? Do they get updated? Does it matter?

PR #993, opened 2016-09-26
change check preconditions stage to v2

robfletcher:


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/BranchingStageDefinitionBuilder.java
```diff
@@ -40,7 +40,9 @@
    * TODO-AJ Only necessary for v1 execution support
    */
   @Deprecated
-  Class<Task> completeParallelTask();
+  default Class<? extends Task> completeParallelTask() {
+    return NoOpTask.class;
+  }
```
robfletcher: Thought this seemed like a sane default

orca-spring-batch/src/main/java/com/netflix/spinnaker/config/SpringBatchConfiguration.java
```diff
@@ -77,10 +78,10 @@ ExecutionListenerProvider springBatchStepExecutionListenerProvider(ExecutionRepo
 
   @Bean
   StageBuilderProvider springBatchStageBuilderProvider(ApplicationContext applicationContext,
-                                                       Collection<StageBuilder> stageBuilders,
+                                                       Optional<Collection<StageBuilder>> stageBuilders,
```
robfletcher: There are no StageBuilders left in orca-core now so tests started failing because they couldn't autowire this

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/CheckPreconditionsStage.groovy
```diff
@@ -16,42 +16,38 @@
...
-  CheckPreconditionsStage() {
-    super(PIPELINE_CONFIG_TYPE)
+  @Autowired
+  CheckPreconditionsStage(List<? extends PreconditionTask> preconditionTasks) {
+    this.preconditionTasks = preconditionTasks
   }
 
   @Override
-  List<Step> buildSteps(Stage stage) {
+  def <T extends Execution<T>> void taskGraph(Stage<T> stage, TaskNode.Builder builder) {
```
ajordens: that `def` shouldn't be necessary?

robfletcher: Groovy struggles with a generic signature like that when it's not preceded by a keyword

PR #996, opened 2016-09-28
(core) move serverGroupReported missing logic out of global scope

lwander:
Instead of making global state changes, I now just check when this task was started to decide when to stop waiting for a server group to show up. @ajordens PTAL


ajordens:
LGTM but needs a rebase ... I think @anotherchrisberry is going to push this through quick because we may have just run into it here.

ajordens:
@anotherchrisberry copied this into another PR and merged it down ... closing this one.

lwander:
Thanks guys

PR #1000, opened 2016-09-30
Fix canary issues

robfletcher:


robfletcher:
@ajordens PTAL

ajordens:
LGTM

orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/stages/ParallelStageDefinitionBuilder.java
```diff
@@ -47,7 +46,7 @@ public String parallelStageName(Stage stage, boolean hasParallelFlows) {
   @Override
   public Task completeParallel() {
     try {
-      return delegate.completeParallelTask().newInstance();
+      return delegate.completeParallelTask();
```
robfletcher: In v1 tasks are manually created so need to give stages a chance to pass stuff to constructors. In v2 they can be auto-wired.

PR #1003, opened 2016-10-05
Orca: Bumps spinnaker-dependencies version

ttomsu:
@duftler @jtk54 @cfieber @ajordens PTAL


ttomsu:
@ajordens - I'm going to split out the exception logging + fiat version bump to another PR. I'm sending out instructions for fiat alpha testers to have a go and I need them on the latest API module.

PR #1004, opened 2016-10-05
migrate a bunch of new stages to v2

robfletcher:


robfletcher:
PTAL @anotherchrisberry

PR #1006, opened 2016-10-07
changes to support rolling push in titus

tomaslin:


tomaslin:
@cfieber PTAL

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/DetermineTerminationCandidatesTask.groovy
```diff
@@ -38,11 +38,13 @@ class DetermineTerminationCandidatesTask implements Task {
   @Override
   TaskResult execute(Stage stage) {
     def stageData = stage.mapTo(StageData)
-    def response = oortService.getServerGroup(stageData.application, stageData.account, stageData.cluster, stage.context.asgName, stage.context.region, stage.context.providerType ?: 'aws')
+    def response = oortService.getServerGroup(stageData.application, stageData.account, stageData.cluster, stage.context.asgName, stage.context.region, stage.context.cloudProvider ?: 'aws')
```
tomaslin: providerType doesn't seem to be set anymore, so using cloudProvider field

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/WaitForNewInstanceLaunchTask.groovy
```diff
@@ -41,11 +41,12 @@ class WaitForNewInstanceLaunchTask implements RetryableTask {
   @Override
   TaskResult execute(Stage stage) {
     StageData stageData = stage.mapTo(StageData)
-    def response = oortService.getServerGroup(stageData.application, stageData.account, stageData.cluster, stage.context.asgName as String, stage.context.region as String, stageData.providerType ?: 'aws' )
+    def response = oortService.getServerGroup(stageData.application, stageData.account, stageData.cluster, stage.context.asgName as String, stage.context.region as String, stageData.cloudProvider ?: 'aws' )
```
tomaslin: Same comment on cloudprovidertype

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/WaitForNewInstanceLaunchTask.groovy
```diff
@@ -41,11 +41,12 @@ class WaitForNewInstanceLaunchTask implements RetryableTask {
...
 
     Map serverGroup = objectMapper.readValue(response.body.in(), Map)
 
-    Set<String> currentIds = serverGroup.instances*.instanceId
+    def serverGroupInstances = serverGroup.instances
```
tomaslin: titus uses task ID so making the same change here

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DisableInstancesTask.groovy
```diff
@@ -29,18 +29,24 @@ import org.springframework.stereotype.Component
...
+  @Autowired
+  KatoService katoService
 
   @Override
   TaskResult execute(Stage stage) {
-    def actions = [[disableInstancesInDiscoveryDescription: stage.context], [deregisterInstancesFromLoadBalancerDescription: stage.context]]
-    def taskId = katoService.requestOperations(actions)
+    def actions = [[disableInstancesInDiscovery: stage.context]]
+    def healthProviders = ["Discovery"]
+    if (stage.context.cloudProvider != 'titus') {
```
ajordens: I know we've got a one or two of these `!= 'titus'` blocks ... but should we use a constant for slightly improved traceability?

or

If `CloudProvider.supportsLoadBalancer(stage.context.cloudProvider)`? (with a default of true). Then at least we have these decisions centralized somewhere.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/WaitForNewInstanceLaunchTask.groovy
ajordens: Same comment as above wrt handling `id` vs `instanceId` elsewhere such that it's not a concern for orca.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/DetermineTerminationCandidatesTask.groovy
```diff
@@ -38,11 +38,13 @@ class DetermineTerminationCandidatesTask implements Task {
...
     def serverGroup = objectMapper.readValue(response.body.in(), Map)
     boolean ascending = stage.context.termination?.order != 'newest'
     int totalRelaunches = getNumberOfRelaunches(stage.context.termination, serverGroup.instances.size())
-    def knownInstanceIds = serverGroup.instances.sort { ascending ? it.launchTime : -it.launchTime }*.instanceId
+    def serverGroupInstances = serverGroup.instances.sort { ascending ? it.launchTime : -it.launchTime }
+    // need to use id instead of instanceIds for titus as the titus API doesn't yet support this
+    def knownInstanceIds = stage.context.cloudProvider == 'titus' ? serverGroupInstances*.id : serverGroupInstances*.instanceId
```
tomaslin: There is an instance id, the problem is that titus doesn't support it in operations in v2, only for discovery registration. This will go away once we switch to instanceIds everywhere but for now this feels like the right place for the workaround

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DisableInstancesTask.groovy
tomaslin: it goes away if I add a noop disable in discovery as well. I'll do that.

ajordens: That's reasonable!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/DetermineTerminationCandidatesTask.groovy
tomaslin: yes, still works for aws

PR #1012, opened 2016-10-11
Adding expression parsing to Manual Judgement stage

robzienert:
cc @ajordens

### DIFF
--- a/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
+++ b/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
@@ -27,6 +27,7 @@ import com.netflix.spinnaker.orca.echo.EchoService
 import com.netflix.spinnaker.orca.pipeline.LinearStage
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
+import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor
 import com.netflix.spinnaker.security.User
 import groovy.util.logging.Slf4j
 import org.springframework.batch.core.Step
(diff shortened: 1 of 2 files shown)


orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -162,7 +163,9 @@ class ManualJudgmentStage extends LinearStage implements RestartableStage, Authe
           application: stage.execution.application
         ),
         additionalContext: [
-          instructions: stage.context.instructions ?: ""
+          instructions: ContextParameterProcessor.process([
```
ajordens: @tomaslin - Shouldn't the processing happen automatically without the need to explicitly process it?

orca-echo/src/test/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStageSpec.groovy
```diff
@@ -133,4 +133,25 @@ class ManualJudgmentStageSpec extends Specification {
     ""             | true                           || false
     null           | true                           || false
   }
+
+  @Unroll
```
ajordens: Don't need the `@Unroll` in this case

orca-echo/src/test/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStageSpec.groovy
```diff
@@ -133,4 +133,25 @@ class ManualJudgmentStageSpec extends Specification {
...
+    def result = task.execute(new PipelineStage(new Pipeline(), "", [
+      notifications: [new Notification(type: "email", address: "test@netflix.com")],
+      instructions: '${replaceMe}',
+      replaceMe: 'newValue'
+    ]))
+
+    then:
+    1 * echoService.create({ EchoService.Notification n ->
+      assert n.additionalContext.instructions == 'newValue'
+      true
```
ajordens: minor but also don't think you need the `true`

ajordens:
LGTM ... I'd check with Tomas as I would have naively thought that `instructions` would have been processed automatically.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
tomaslin: I see that it does. I suspect it might be that error where the expression cannot be evaluated.

PR #1013, opened 2016-10-11
Resume a previously paused pipeline when restarting

ajordens:
Previously a pipeline that was paused for 12hrs would automatically go TERMINAL and
subsequent attempts to restart would fail.

- Adjusted max pause time upwards from 12hrs -> 72hrs to handle pausing over a weekend


orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/RetryableTaskTasklet.groovy
```diff
@@ -42,7 +42,7 @@ import static com.netflix.spinnaker.orca.pipeline.model.Stage.STAGE_TIMEOUT_OVER
 
 @CompileStatic
 class RetryableTaskTasklet extends TaskTasklet {
-  static final long MAX_PAUSE_TIME_MS = TimeUnit.HOURS.toMillis(12)
+  static final long MAX_PAUSE_TIME_MS = TimeUnit.DAYS.toMillis(3)
```
cfieber: what if its a 4 day weekend?!

PR #1014, opened 2016-10-12
Add guard around canceling a canary for restart

zanthrash:
Can't guarantee that there will be a canary id on the context when a call to cancel from the restart.

@robfletcher PTAL


robfletcher:
LGTM, ship it

PR #1015, opened 2016-10-12
Separate Spring Batch from core of Orca execution model

robfletcher:
This change fundamentally alters the way that stages are constructed and the execution plan for an Orca pipeline is turned into a Spring Batch job.
The goal is to have `orca-core` not depend on Spring Batch at all.
To that end there is a new `orca-spring-batch` module containing all the code required to construct a Job from a pipeline model.

In order to have a controlled migration to the new way of running jobs we're currently supporting the existing model as well.
That means that in the near future there will be a second part of this work that removes all the existing Spring Batch stuff from `orca-core`.
Pipelines submitted with `executionEngine: 'v2'` are run with the new code.
That way we can flip some of our own pipelines over to the new model with (hopefully) minimal disruption to existing pipelines.
Once we're confident in the reliability of the changes we can remove the `executionEngine` flag and just flip everything over to run in "v2" mode.

Some advantages:

- Tests for `ExecutionRunner` are generic and independent of Spring Batch so when we want to consider an alternate runtime platform we can re-use the tests.
- We now have a generic model for defining stages like _rolling push_.
- No Spring Batch types leak into the new `StageDefinitionBuilder` interface.
- Some of the tasks and stages that existed only to get Spring Batch to hang together are no longer necessary (e.g. stage boundary tasks and pipeline initializer).
- Task definitions are not affected, only stages.

Still pending

- Orchestrations are not run using v2 yet. They should be simple enough to convert.
- Remove all the existing v1 execution engine code. Much of it is marked with `@Deprecated`.
- The only module directly dependent on Spring Batch should be `orca-spring-batch`. The v1 stuff needs to be removed first, though.


cfieber:
LGTM

cfieber:
just kidding, going on blind faith on this one

PR #1020, opened 2016-10-15
Remove `targetReferences` from any successful completed execution

ajordens:
- `targetReferences` contains complete asg/instance details and is no longer
necessary after an execution has completed

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/OrcaConfiguration.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/OrcaConfiguration.groovy
@@ -16,6 +16,9 @@
 
 package com.netflix.spinnaker.orca.config
 
+import com.netflix.spinnaker.orca.listeners.CompositeExecutionListener
+import com.netflix.spinnaker.orca.listeners.ExecutionCleanupListener
+
 import java.time.Clock
 import java.time.Duration
 import java.util.concurrent.ThreadPoolExecutor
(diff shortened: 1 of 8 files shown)


anotherchrisberry:
One thing to consider: we currently show a very limited set of fields in Deck (name, region, zone) in the `determineTargetReferences` stage details, so we should either blacklist that stage in Deck, or retain just those fields. My preference is the former, as I don't think there's much value in having it in the list of steps on a stage.

ajordens:
Blacklisting seems reasonable to me

ajordens:
@anotherchrisberry removed the `determineTargetReferences` stage as part of https://github.com/spinnaker/deck/pull/2877

orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/OrcaConfiguration.groovy
```diff
@@ -162,15 +165,11 @@ class OrcaConfiguration {
...
-  ExecutionPropagationListener executionPropagationListenerBefore() {
-    // need a dedicated beforeJob listener due to how spring boot ordered listeners
-    new ExecutionPropagationListener(true, false)
-  }
-
-  @Bean
-  ExecutionPropagationListener executionPropagationListenerAfter() {
-    // need a dedicated afterJob listener due to how spring boot ordered listeners
-    new ExecutionPropagationListener(false, true)
+  CompositeExecutionListener executionListeners() {
```
cfieber: thoughts on just accepting an autowired List<ExecutionListener> 

moving these into explicit construction means that you can't DI a Listener

maybe even down below the places where a CompositeExecutionListener is expected, accept the list of listeners there and internally construct one of these

ajordens:
I consciously did it like it like this because these are the only two
listeners that have an execution order dependency (there are others that
don't).

Rob had done something similar for StageListeners previously.

Personally I think being explicit for now is easier to understand than
@Order.

ajordens:
Don't recall the history behind only having a subset of ExecutionListeners
attached to an Orchestration vs Pipeline.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/listeners/ExecutionPropagationListener.groovy
```diff
@@ -26,20 +26,8 @@ import static com.netflix.spinnaker.orca.ExecutionStatus.*
 @Slf4j
 @CompileStatic
 class ExecutionPropagationListener implements ExecutionListener {
-  private boolean isBeforeJobEnabled = false
-  private boolean isAfterJobEnabled = false
```
robfletcher: I like that we're getting rid of this stuff but does it have anything directly to do with this change?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/OrcaConfiguration.groovy
robfletcher: You mean you can't add a listener at runtime?

PR #1021, opened 2016-10-15
Fix a bug where calling a task with custom strategies would fail beca…

jeyrschabu:
…use of casting error

- Added a guard to ensure the parent task is a pipeline before attempting to cast to a Pipeline


jeyrschabu:
going to do it with a different branch for sanity

PR #1022, opened 2016-10-15
Fix a bug where calling a task with custom strategies would fail beca…

jeyrschabu:
…use of casting error

- Added a guard to ensure the parent task is a pipeline before attempting to cast to a Pipeline

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
@@ -133,7 +133,7 @@ trait DeploymentDetailsAware {
     if (execution instanceof Pipeline) {
       if (execution.trigger.parentExecution instanceof Pipeline) {
         return execution.trigger.parentExecution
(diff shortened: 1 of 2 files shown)


jeyrschabu:
@ajordens PTAL.

This is the PR for master

jeyrschabu:
@robfletcher PTAL

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
```diff
@@ -133,7 +133,7 @@ trait DeploymentDetailsAware {
     if (execution instanceof Pipeline) {
       if (execution.trigger.parentExecution instanceof Pipeline) {
         return execution.trigger.parentExecution
-      } else if (execution.trigger.parentExecution) {
+      } else if (execution.trigger?.isPipeline) {
```
robfletcher: What happens if it _isn't_ a pipeline?

jeyrschabu: > In this situation, the task execution can't have a parent pipeline.

PR #1023, opened 2016-10-17
(core) get health providers from stage context

lwander:
@robfletcher @ajordens @duftler PTAL

Kubernetes uses different health providers for different tasks, so the health providers need to be pulled from the stage context. Is there a neater/more consistent way for Kubernetes to get its health providers?

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
@@ -101,10 +101,7 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
         }
 
         seenServerGroup[name] = true
-        Collection<String> interestingHealthProviderNames = stage.execution.appConfig.interestingHealthProviderNames as Collection
-        if (interestingHealthProviderNames == null) {
-          interestingHealthProviderNames = stage.context?.appConfig?.interestingHealthProviderNames as Collection
-        }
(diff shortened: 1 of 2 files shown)


lwander:
Closes https://github.com/spinnaker/spinnaker/issues/1240

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
```diff
@@ -101,10 +101,7 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
         }
 
         seenServerGroup[name] = true
-        Collection<String> interestingHealthProviderNames = stage.execution.appConfig.interestingHealthProviderNames as Collection
-        if (interestingHealthProviderNames == null) {
```
lwander: After talking to @duftler, it looks like this line isn't necessary due to this change here: https://github.com/spinnaker/orca/pull/957/files#diff-f68af5115f175c0b21337307aac71d46R92

ajordens: Agreed ... stage.context will have appConfig folded in.

ajordens:
LGTM

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
duftler: We ran some tests with additional instrumentation and agree that `stage.context.interestingHealthProviderNames` is the right place to look, and the `stage.context.appConfig` references are extraneous.

PR #1026, opened 2016-10-18
A couple of queueing fixes

robfletcher:
- ensure pipeline queue is only checked after the status is propagated otherwise queued pipelines aren't started as the original pipeline status is incomplete
- run v2 pipelines in v2 mode from the queue


robfletcher:
PTAL @ajordens

ajordens:
Could we have fixed the ordering issue if the Order of the composite
listener was such that it always ran first?

ajordens:
Otherwise LGTM ... I'm satisfied with being very explicit about the
ordering of listeners.

robfletcher:
We possibly could manage the ordering with the annotations but I have to say I prefer doing it in code when it's that important.

PR #1036, opened 2016-10-25
Adding latest pipeline executions endpoint

robzienert:
Given a list of pipeline config ids, the `/pipelines` endpoint will return the latest executions that have a start time or have not yet started. This new endpoint will be used in echo to offer a viable deployment & restart story using an in-memory scheduler.

Also open to a different endpoint; perhaps `/latestPipelines` or something, so `/pipelines` can stay open for something a little more generic?

@cfieber @ajordens PTAL


ajordens:
I'm alright w/ `/pipelines` (we used to have a horrible `/pipelines` that would attempt to give everything back).

FYI @robfletcher

PR #1040, opened 2016-10-25
STOPPED stage in a branch should not fail the pipeline if all requisite stages succeeded

jeyrschabu:
- Replaced explicitly exception throw with a STOPPED return result


jeyrschabu:
ref: https://jira.netflix.com/browse/SPIN-2091

ajordens:
LGTM

PR #1041, opened 2016-10-26
Fixing stage state when restarting a ACA Task.

zanthrash:
Restarting an ACA Task stage will cancel the Canary and, in doing so will put the stage into a STOPED state, preventing the
restarted pipeline from continuing.

This fix just put the stage state back to the RUNNING state that the 'prepareStageForRestart' puts it in.

@robfletcher PTAL


orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/pipeline/AcaTaskStage.groovy
```diff
@@ -65,6 +65,9 @@ class AcaTaskStage implements StageDefinitionBuilder, CancellableStage, Restarta
       stage.context.canary.remove("canaryResult")
       stage.context.canary.remove("status")
       stage.context.canary.remove("health")
+
+      //Canceling the canary marks the stage as CANCELED preventing it from restarting.
+      stage.setStatus(ExecutionStatus.RUNNING)
```
robfletcher: Should it be `NOT_STARTED`?

PR #1043, opened 2016-10-28
GetCommitsTask : Use amiId instead of ami name

gbougeard:
`List<Map> amiDetails = oortService.getByAmiId("aws", account, region, ami)` needs an `amiId` so if you use an ami name it can not resolve the amiDetails.

PTAL @tomaslin @ttomsu @dzapata


gbougeard:
I need to modify the tests too :(

gbougeard:
cc @cfieber

gbougeard:
@robfletcher PTAL

ajordens:
Needs a quick rebase

gbougeard:
rebased

icfantv:
@gbougeard Hey, just a heads up that this broke us.  I'm going to leave your change in there, but I am first going to look for the `amiId` and if it's not there, which it's not in our case, I will defer to the original property of `ami`.

PR #1045, opened 2016-10-29
Clarify listener operation

robfletcher:
This fixes SPIN-2119 where Echo is receiving invalid status for certain conditions such as STOPPED or FAILED_CONTINUE.
Listener order is now programmatic instead of relying on Spring ordering annotations, etc.


robfletcher:
PTAL @ajordens

orca-core/src/test/groovy/com/netflix/spinnaker/orca/batch/lifecycle/LinearStageSpec.groovy
```diff
@@ -46,9 +45,9 @@ class LinearStageSpec extends AbstractBatchLifecycleSpec {
   @Shared
   def stageNavigator = new StageNavigator(Stub(ApplicationContext))
 
-  List<StepExecutionListener> listeners = [
-    new SpringBatchStageListener(executionRepository, new StageStatusPropagationListener())
-  ]
+//  List<StepExecutionListener> listeners = [
```
ajordens: no longer necessary?

robfletcher: Yeah, it comes from the SpringBatchExecutionListenerProvider now — see below. I'll remove the commented out code.

PR #1052, opened 2016-11-03
(core) add waitForCompletion field to fire and forget

tomaslin:
sometimes you just want to send a request and not care if the downstream pipeline / stage / job / etc completes

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.groovy
@@ -34,6 +34,9 @@ class RunJobStage implements StageDefinitionBuilder {
     builder
       .withTask("runJob", RunJobTask)
       .withTask("monitorDeploy", MonitorKatoTask)
-      .withTask("waitOnJobCompletion", WaitOnJobCompletion)
+
+    if (!stage.getContext().getOrDefault("waitForCompletion", "true").toString().equalsIgnoreCase("false")) {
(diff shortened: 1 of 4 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.groovy
```diff
@@ -34,6 +34,9 @@ class RunJobStage implements StageDefinitionBuilder {
     builder
       .withTask("runJob", RunJobTask)
       .withTask("monitorDeploy", MonitorKatoTask)
-      .withTask("waitOnJobCompletion", WaitOnJobCompletion)
+
+    if (!stage.getContext().containsKey("ignoreResults") || (Boolean.parseBoolean(stage.getContext().get("ignoreResults").toString()) == true)) {
```
ajordens: I'd maybe call it `waitForCompletion` rather than `ignoreResults`.

```
def waitForCompletion = !stage.getContext().getOrDefault("waitForCompletion", "true").equalsIgnoreCase("false")
if (Boolean.valueOf(waitForCompletion)) {
   builder.withTask(...)
}
```

^^ might be a tad tidier.

Feels like it needs decent messaging in the UI but otherwise I think this is alright ... what does @robfletcher think?

ajordens: Tweaked the `code` above to be strict about only skipping if `waitForCompletion == "false"` rather than `waitForCompletion != "true"` ... seems slightly safer

ajordens:
ship it ... wcgw

PR #1055, opened 2016-11-04
supply stage details for v2 executions

robfletcher:
In v1 there is a special task to add these details to the context but it seems way more reliable to do it in a listener. 

This is what was missing for SPIN-2129.


ajordens:
lgtm

PR #1058, opened 2016-11-07
Fail Jenkins Stage when properties file expected, but not provided

robzienert:
@tomaslin PTAL


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/MonitorJenkinsJobTask.groovy
```diff
@@ -67,11 +67,14 @@ class MonitorJenkinsJobTask implements RetryableTask {
       }
 
       if (statusMap.containsKey(result)) {
+        ExecutionStatus status = statusMap[result]
         Map<String, Object> properties = [:]
         if (stage.context.propertyFile) {
           properties = buildService.getPropertyFile(buildNumber, stage.context.propertyFile, master, job)
+          if (properties.size() == 0) {
+            status = ExecutionStatus.TERMINAL
```
tomaslin: is it better to throw an exception to say `expected a properties file called blah.properties but this didn't exist` ?

PR #1059, opened 2016-11-08
Ensure stage with FAILED_CONTINUE tasks is marked as FAILED_CONTINUE …

robfletcher:
…even if it wasn't the last task

Fixes SPIN-2122


robfletcher:
PTAL @ajordens

ajordens:
LGTM

PR #1061, opened 2016-11-09
Quip `InstanceHealthCheckTask` should favor `instance.privateIpAddress`

ajordens:


ajordens:
FYI @tomaslin

tomaslin:
works for me

PR #1062, opened 2016-11-10
Extract buildInfo from parent pipelines

gardleopard:
Before this orca can extract artifacts from triggering pipelines of depth 1.
This code enables orca to extract artifacts from pipelines in infinite depth
and also extract the build details from the build that produced the artifact.
The objective of this PR is to populate buildInfo in amis created with artifacts
from parent pipelines.


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -166,6 +166,16 @@ class PackageInfo {
     return request
   }
 
+  @CompileDynamic
+  Map getArtifactSourceBuildInfo(Map trigger){
+    if (trigger.buildInfo?.artifacts) {
```
lwander: Might want to check if trigger isn't null  (`trigger?.`) since you do that in the below `if` block

gardleopard: Fixing and squashing. Thanks!

PR #1064, opened 2016-11-10
Fix v2 stage notifications

robfletcher:
Echo expects `stageDetails` key to be present in context. Because it wasn't there at the time the echo listener fires before the stage the "stage starting" notification was never triggered.


ajordens:
LGTM

PR #1068, opened 2016-11-14
Propagates security context when a pipeline triggers another pipeline

ttomsu:
PTAL @ajordens @jtk54 @cfieber @tomaslin 

Fixes https://github.com/spinnaker/fiat/issues/114

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
@@ -22,6 +22,8 @@ import com.netflix.spinnaker.orca.pipeline.PipelineStarter
 import com.netflix.spinnaker.orca.pipeline.model.Execution
 import com.netflix.spinnaker.orca.pipeline.model.Pipeline
 import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor
+import com.netflix.spinnaker.security.AuthenticatedRequest
+import com.netflix.spinnaker.security.User
 import groovy.util.logging.Slf4j
 import org.springframework.beans.BeansException
 import org.springframework.beans.factory.annotation.Autowired
@@ -44,9 +46,11 @@ class DependentPipelineStarter implements ApplicationContextAware {
     def json = objectMapper.writeValueAsString(pipelineConfig)
     log.info('triggering dependent pipeline {}:{}', pipelineConfig.id, json)
 
+    User principal = getUser(parentPipeline)
+
     pipelineConfig.trigger = [
       type                     : "pipeline",
-      user                     : user ?: '[anonymous]',
+      user                     : principal?.username ?: user ?: '[anonymous]',
       parentPipelineId         : parentPipeline.id,
       parentPipelineApplication: parentPipeline.application,
       parentStatus             : parentPipeline.status,
@@ -85,17 +89,18 @@ class DependentPipelineStarter implements ApplicationContextAware {
 
     def pipeline
 
-    def t1 = new Thread(new Runnable() {
-      @Override
-      public void run() {
-        if (parentPipeline.executionEngine == Execution.V2_EXECUTION_ENGINE) {
-          pipeline = pipelineLauncher().start(json)
-        } else {
-          pipeline = pipelineStarter.start(json)
-        }
+    log.debug("Source thread: MDC user: " + AuthenticatedRequest.getAuthenticationHeaders() +
+                  ", principal: " + principal?.toString())
+    def runnable = AuthenticatedRequest.propagate({
+      log.debug("Destination thread user: " + AuthenticatedRequest.getAuthenticationHeaders())
+      if (parentPipeline.executionEngine == Execution.V2_EXECUTION_ENGINE) {
+        pipeline = pipelineLauncher().start(json)
+      } else {
+        pipeline = pipelineStarter.start(json)
       }
-    })
+    }, true, principal) as Runnable
 
+    def t1 = new Thread(runnable)
     t1.start()
 
     try {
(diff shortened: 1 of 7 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -88,6 +91,14 @@ class DependentPipelineStarter implements ApplicationContextAware {
     def t1 = new Thread(new Runnable() {
       @Override
       public void run() {
+        if (user) {
+
+          def securityCtx = SecurityContextHolder.createEmptyContext()
+          // Using RunAsUserToken is an arbitrary Authentication objection - it's merely named the best for this use case.
```
jtk54: Why not [PreAuthenticatedAuthenticationToken](http://docs.spring.io/autorepo/docs/spring-security/4.1.2.RELEASE/apidocs/org/springframework/security/web/authentication/preauth/PreAuthenticatedAuthenticationToken.html) to get rid of the random key?

ajordens: This also probably isn't the right "user", looks like it's coming out of a trigger and/or context variable.

https://github.com/spinnaker/orca/blob/master/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy#L57

https://github.com/spinnaker/orca/blob/master/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/StartPipelineTask.groovy#L76

Feels like doing something along the lines of what `AuthenticatedRequest.propagate()` is doing might be appropriate?

https://github.com/spinnaker/orca/blob/master/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy#L194

^^ is what we're doing in the TaskTasklet to propagate the auth context from a running execution.

ttomsu: The problem is that `DependentPipelineExecutionListener` is running on a separate thread from independent pipeline, so there is nothing in the SecurityContext/MDC mechanism for `Authenticatedreuest.propagate()` to read values from.

To put it a bit more clearly, we're dealing with 3 separate threads:

1. The original/independent pipeline thread
2. The `DependendPipelineExecutionListener` thread
3. The Thread created within 2.) to run the dependent pipeline.

The security context needs to propagate from 1 -> 3, but is being dropped between 1 and 2, and I don't control the invocation, so I can't wrap it with `propagate`

ttomsu:
Updated this to rely on the `stage.execution.authentication` much more. 

Checked with implicit pipeline trigger (dependent pipeline has "Automatic trigger" on the completion of the independent pipeline) and explicit trigger (dependent pipeline is a stage in the independent pipeline).

I also checked that the authentication propagation during a manual judgement continues as expected, which I think is behavior that doesn't currently work today (can you confirm @ajordens?)

xkvrhpv82ft 1

@skim1420 @duftler also PTAL

ajordens:
I'm OOO (and largely mobile) until Tuesday, but you should still be able to
use propagate() but pass in an explicit user based on the execution's
authentication. That's what the TaskTasklet does and it spans threads much
the same way.

Either way, using the auth context from the execution (as you've done) is
certainly preferable to trusting the user-supplied values from the trigger.

ajordens:
Also not sure it's a reasonable side effect of the authenticatedUser()
method to modify the executions authentication?

(Sorry for the review by email ... only have phone)

ttomsu:
You're right about the `authenticatedUser` side effect - I'll try to find a better place to modify that.

ttomsu:
Ok, I reworked this PR to not have those side effects.

The current behavior now will prefer the MDC user and then the `parentExecution.authentication.user` when setting the MDC context for the triggered pipeline.

This means that both implicitly and explicitly run pipelines will pick up the user/service account that originally invoked the parent pipeline.

In the case of a Manual Judgement within the parent pipeline that propagates authentication, only explicitly invoked pipelines within that MJ branch will use the user's auth. _Implicitly_ invoked pipelines just inherit the same auth that the parent was invoked with.

I think that makes the most sense. It also means we avoid the multiple MJs on different branches problem.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -105,9 +122,24 @@ class DependentPipelineStarter implements ApplicationContextAware {
     }
 
     log.info('executing dependent pipeline {}', pipeline.id)
+    return pipeline
+  }
 
-    pipeline
+  // There are currently two sources-of-truth for the user:
+  // 1. The MDC context, which are the values that get propagated to downstream services like Front50.
+  // 2. The Execution.AuthenticationDetails object.
```
ajordens: FWIW the MDC context is set based on the Execution.AuthenticationDetails each time a task run (via `TaskTasklet`).

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -85,9 +90,16 @@ class DependentPipelineStarter implements ApplicationContextAware {
 
     def pipeline
 
+    Map mdc = null
     def t1 = new Thread(new Runnable() {
       @Override
       public void run() {
+        log.debug("MDC destination thread 1: " + MDC.getCopyOfContextMap())
```
ajordens: Manually copying around the MDC context seems a little suspect, would the following work?

```
    def runnable = AuthenticatedRequest.propagate({
      if (parentPipeline.executionEngine == Execution.V2_EXECUTION_ENGINE) {
        pipeline = pipelineLauncher().start(json)
      } else {
        pipeline = pipelineStarter.start(json)
      }
    }, true, principal) as Runnable

    def t1 = new Thread(runnable)
    t1.start()
```

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
ttomsu: Right, but in terms of "source of truth" to downstream services, it doesn't matter what's in the Execution.AuthenticationDetails if MDC holds different values. 

In the case of the implicit pipeline invocation, the MDC is empty, which is why we fall back to Execution.AuthenticationDetails.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
ttomsu: That worked nicely. Great suggestion!

PR #1072, opened 2016-11-15
handle multiple root stages in v2 engine

robfletcher:
Fixes SPIN-2165


robfletcher:
PTAL @ajordens

ajordens:
I trust that  this works, LGTM

PR #1073, opened 2016-11-17
(core) Send additional details to echo for manual judgment notification.

duftler:


duftler:
@tomaslin please review.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -155,9 +156,25 @@ class ManualJudgmentStage implements StageDefinitionBuilder, RestartableStage, A
...
         additionalContext: [
+          stageName: stage.name,
+          stageId: stage.refId,
+          restrictExecutionDuringTimeWindow: stage.context.restrictExecutionDuringTimeWindow,
           instructions: stage.context.instructions ?: ""
         ]
-      ))
+      )
+
+      if (stage.execution instanceof Pipeline) {
```
tomaslin: in other places we send the entire execution. Might make sense to do that as well here so that if we decided to improve templating in the future manual judgment is not a special snowflake

duftler: Good idea. I've updated this to send the entire execution instead of cherry-picking those two attributes.

duftler:
Manual judgment subject template has also been updated to reflect these changes: https://github.com/spinnaker/echo/pull/115

PR #1079, opened 2016-11-23
Add tests for RPM in artifact resolving

gardleopard:
This just adds some missing tests on RPM version extraction.


robzienert:
Still valid?

PR #1089, opened 2016-12-13
(manualJudgment) Add support for manualJudgmentContinue and manualJud…

duftler:
…gmentStop notifications.
(manualJudgment) Employ same notification configuration for manualJudgment as used by the other stages (even though the actual 'when' sets are different).

### DIFF
--- a/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
+++ b/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
@@ -76,29 +76,58 @@ class ManualJudgmentStage implements StageDefinitionBuilder, RestartableStage, A
 
     @Override
     TaskResult execute(Stage stage) {
-      def stageData = stage.mapTo(StageData)
+      StageData stageData = stage.mapTo(StageData)
+      NotificationState notificationState
+      ExecutionStatus executionStatus
+
       switch (stageData.state) {
         case StageData.State.CONTINUE:
-          return new DefaultTaskResult(ExecutionStatus.SUCCEEDED)
+          notificationState = NotificationState.manualJudgmentContinue
+          executionStatus = ExecutionStatus.SUCCEEDED
+          break
         case StageData.State.STOP:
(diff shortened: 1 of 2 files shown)


duftler:
Depends on https://github.com/spinnaker/echo/pull/118 for the new notifications to actually be sent.

duftler:
@tomaslin please review.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -77,30 +77,58 @@ class ManualJudgmentStage implements StageDefinitionBuilder, RestartableStage, A
...
+          break
       }
 
-      def outputs = [:]
+      processNotifications(stage, stageData, outputs, notificationState)
+
+      return new DefaultTaskResult(executionStatus, outputs)
+    }
+
+    void processNotifications(Stage stage, def stageData, def outputs, NotificationState notificationState) {
```
ajordens: Would suggest typing the arguments and returning the outputs rather than modifying the `outputs` parameter as a side-effect.

orca-echo/src/test/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStageSpec.groovy
```diff
@@ -63,21 +63,58 @@ class ManualJudgmentStageSpec extends Specification {
...
+      ],
+      judgmentStatus: judgmentStatus
+    ]))
+
+    then:
+    result.status == executionStatus
+    if (sent) result.stageOutputs.notifications?.getAt(0)?.lastNotifiedByNotificationState?.containsKey(notificationState)
+
+    where:
+    sendNotifications | notificationState                        | judgmentStatus | executionStatus || sent
```
ajordens: can you reformat this table

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
duftler: Done.

PR #1091, opened 2016-12-14
Fix problem where stages with empty task graphs never complete

robfletcher:


ajordens:
Just needs a quick rebase

PR #1093, opened 2016-12-15
(manualJudgment) Include judgedBy in additional context sent to echo.

duftler:


duftler:
@lwander or @ajordens please review.

PR #1094, opened 2016-12-16
Any failure in a canary stage should auto cancel (ie. cleanup) the cr…

ajordens:
…eated clusters

- If the executing stage fails (TERMINAL, STOPPED, FAILED_CONTINUE) and is `Cancellable`, it should be canceled
- Improved resilience when canceling canaries that failed prior to registration

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/ExecutionStatus.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/ExecutionStatus.groovy
@@ -95,6 +95,7 @@ enum ExecutionStatus {
   final ExitStatus exitStatus
 
   private static final Collection<ExecutionStatus> SUCCESSFUL = [SUCCEEDED, STOPPED, SKIPPED]
+  private static final Collection<ExecutionStatus> FAILURE = [TERMINAL, STOPPED, FAILED_CONTINUE]
 
   ExecutionStatus(boolean complete, boolean halt, ExitStatus exitStatus) {
     this.complete = complete
@@ -104,5 +105,9 @@ enum ExecutionStatus {
 
   boolean isSuccessful() {
     return SUCCESSFUL.contains(this)
+  }
+
+  boolean isFailure() {
+    return FAILURE.contains(this)
   }
 }
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
@@ -103,6 +103,15 @@ class TaskTasklet implements Tasklet {
           executionRepository.cancel(stage.execution.id)
         }
 
+        if (result.status.isFailure()) {
+          try {
+            // ensure that any cleanup behavior is invoked when a task fails (if stage is `Cancellable`)
+            cancel(stage, false)
+          } catch (Exception e) {
+            log.error("Error occurred canceling stage '${stage.id}' in execution '${stage.execution.id}'", e)
+          }
+        }
+
         def stageOutputs = new HashMap(result.stageOutputs)
         if (result.status.complete) {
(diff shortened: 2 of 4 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/ExecutionStatus.groovy
```diff
@@ -95,6 +95,7 @@ enum ExecutionStatus {
   final ExitStatus exitStatus
 
   private static final Collection<ExecutionStatus> SUCCESSFUL = [SUCCEEDED, STOPPED, SKIPPED]
+  private static final Collection<ExecutionStatus> FAILURE = [TERMINAL, STOPPED, FAILED_CONTINUE]
```
ajordens: Thoughts on if `STOPPED` should result in the auto cancel behavior?

I was inclined to say yes (_because that's how the pipeline in question that exposed the issue is setup_) ... also STOPPED is an indication of a failure that was configured to not halt the overall pipeline.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
```diff
@@ -143,19 +151,26 @@ class TaskTasklet implements Tasklet {
     return result
   }
 
-  private RepeatStatus cancel(Stage stage) {
+  private RepeatStatus cancel(Stage stage, boolean adjustStageStatusAndTasks = true) {
+    doCancel(stage, adjustStageStatusAndTasks)
+    return RepeatStatus.FINISHED
+  }
+
+  protected void doCancel(Stage stage, boolean adjustStageStatusAndTasks) {
```
ajordens: I'll write a test case around this cancel behavior as we don't currently have any coverage

orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/pipeline/MonitorCanaryStage.groovy
```diff
@@ -49,8 +49,18 @@ class MonitorCanaryStage implements StageDefinitionBuilder, CancellableStage {
...
+    def canaryId = stage.context.canary?.id as String
+
+    try {
+      if (canaryId) {
+        // will not have a `canaryId` if the failure occured prior to registration
+        cancelCanaryResults = mineService.cancelCanary(canaryId, "Pipeline execution (${stage.execution?.id}) canceled")
+        log.info("Canceled canary in mine (canaryId: ${canaryId}, stageId: ${stage.id}, executionId: ${stage.execution.id})")
+      }
+    } catch (Exception e) {
+      log.error("Unable to cancel canary '${canaryId}' in mine", e)
```
ajordens: The real point here is to ensure that launched canary clusters are cleaned up upon failure ... it would seem preferable that a failure to cancel in `mine` did NOT prevent cleanup.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
```diff
@@ -143,19 +152,26 @@ class TaskTasklet implements Tasklet {
     return result
   }
 
-  private RepeatStatus cancel(Stage stage) {
+  private RepeatStatus cancel(Stage stage, boolean adjustStageStatusAndTasks = true) {
+    doCancel(stage, adjustStageStatusAndTasks)
```
robfletcher: why the indirection on this method?

robfletcher:
Beat you to the rebase!

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/adapters/TaskTasklet.groovy
ajordens: Really just so I could override it with an assertion in a test and not worry about the batch-specific `RepeatStatus` response

PR #1098, opened 2016-12-21
expressions in stage context should be parsed before first task runs

robfletcher:
SPIN-2239


orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/ExecutionContextManager.groovy
```diff
@@ -41,6 +42,17 @@ class ExecutionContextManager {
     }
   }
 
+  private
+  static Map<String, Object> processEntries(Map<String, Object> map, Stage<? extends Execution> stage, ChunkContext context) {
```
ajordens: probably no line break needed here but the rest LGTM

PR #1102, opened 2016-12-27
Add support for cc: field to email notifications.

duftler:


duftler:
@ajordens or @tomaslin please review.
Relates to: https://github.com/spinnaker/echo/pull/123

ajordens:
LGTM

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/EchoService.groovy
```diff
@@ -35,9 +36,11 @@ interface EchoService {
   @POST("/notifications")
   Response create(@Body Notification notification)
 
+  @JsonInclude(JsonInclude.Include.NON_NULL)
```
duftler: @ajordens Added this following your LGTM as per our offline discussion.

PR #1103, opened 2016-12-29
provider/appengine start / stop tasks and stages

danielpeach:
@duftler or @lwander please review

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AbstractWaitForAppEngineServerGroupStopStartTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AbstractWaitForAppEngineServerGroupStopStartTask.groovy
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AbstractWaitForAppEngineServerGroupStopStartTask.groovy
```diff
@@ -0,0 +1,87 @@
...
+  @Override
+  TaskResult execute(Stage stage) {
+    String cloudProvider = getCloudProvider(stage)
+    String account = getCredentials(stage)
+    String serverGroupName = (stage.context.serverGroupName ?: stage.context.asgName) as String
+    Names names = Names.parseName(serverGroupName)
+    try {
+      def response = oortService.getCluster(names.app, account, names.cluster, cloudProvider)
+
+      if (response.status != 200) {
```
lwander: Maybe a `!response.isSuccessful()` check instead?

lwander: Then again, if the response fails it'll throw an exception that you'll catch below - what's this check for?

danielpeach: I see this in a few places (e.g., `WaitForDestroyedServerGroupTask` or `AbstractInstancesCheckTask`). I'm assuming that Clouddriver can return a `2xx` response such that we'll want to keep the task status as `RUNNING` but not want to check the response body for pass/fail conditions.

lwander: (resolved offline) those checks exist because those tasks can return a `202` to indicate the response is underway in the background.

PR #1105, opened 2017-01-02
Use parentPipelineId and not isPipeline to identify parent executions

gardleopard:
Use parentPipelineId to identify that there is a parent pipeline. isPipeline does not get set when triggering pipelines manually with a specified parent execution. This change should make orca treat manual and automatic triggers the same way.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
@@ -133,7 +133,7 @@ trait DeploymentDetailsAware {
     if (execution instanceof Pipeline) {
       if (execution.trigger.parentExecution instanceof Pipeline) {
         return execution.trigger.parentExecution
-      } else if (execution.trigger?.isPipeline) {
+      } else if (execution.trigger?.parentPipelineId) {
         return pipelineObjectMapper.convertValue(execution.trigger.parentExecution, Pipeline)
       }
     }
--- a/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/CreateServerGroupTaskSpec.groovy
+++ b/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/CreateServerGroupTaskSpec.groovy
@@ -202,7 +202,7 @@ class CreateServerGroupTaskSpec extends Specification {
     ]
     // Building these as maps instead of using the pipeline model objects since this configuration was observed in testing.
     def parentTrigger = [
-      isPipeline: true,
+      parentPipelineId: "abc",
       parentExecution: [
         name: "grandparent",
         context: grandparentGlobalContext,
@@ -214,7 +214,7 @@ class CreateServerGroupTaskSpec extends Specification {
     ]
 
     def childTrigger = [
-      isPipeline: true,
+      parentPipelineId: "abc",
       parentExecution: [
         name: "parent",
         trigger: parentTrigger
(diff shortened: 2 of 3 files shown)


gardleopard:
@spinnaker/reviewers PTAL

tomaslin:
No. `isPipeline` is used to distinguish the type of the parent execution between Orchestration and Pipeline.

This PR will break deployment strategies that are triggered by hand, i.e, clusters screen -> clone -> select a custom strategy. For this scenario, the parentPipelineId will be set, but it will be a task, hence we need to check the isPipeline scenario. Tasks cannot be cast into Pipeline. 

If isPipeline is not set when triggering manually, the solution should be to set isPipeline in the trigger  and not change this behavior.

gardleopard:
@tomaslin if the parent is a task, why has it got a parentPipelineId? Shouldn't that be a parentTaskId instead then?

tomaslin:
it should really be called parentExecutionId because that is what is being captured.

tomaslin:
Historically I think we did dependent pipelines before strategies, so pipeline parents were always Pipelines and the name made sense then. Once we introduced strategies, parentPipelineId could now refer to a Task or a Pipeline.

tomaslin:
In any case this PR should not be merged since it will break the scenario I have described.

gardleopard:
So we should:
1. rename references to parentPipelineId to parentExecutionId
2. introduce a new field called parentExecutionType with values like pipeline, task and strategy

Does that sound good or am I missing something?

tomaslin:
I think there are 2 issues here and I don't want to confuse them:

1. The functionality of DeploymentDetailsAware and ancestor resolution, which depends on looking up the chain of Pipeline objects. 

2. Better metadata on triggers for downstream consumers like deck or echo. 

From a metadata perspective having a field called parentExecutionType makes sense. But if we're doing that, we should just a field in the actual Pipeline or orchestration object that says "type":"pipeline" | "orchestration" | "strategy". 

From a ancestor resolution perspective, strategies are Pipeline / Pipeline Objects. You don't really want to treat them any differently. Tasks don't have ancestors so any resolution should stop there. 

I'm ok with renaming things to pipelineExecutionId as long as this is fixed in all the downstream services ( echo | deck ).

tomaslin:
While I agree that parentExecutionId is a better name, I think it is a terrible idea to do it now. It will introduce a breaking change for anyone relying on it in expressions and downstream jobs. While changing echo / deck would be easy, it will inevitably break many workflows today.

tomaslin:
Is the solution to this issue not just adding trigger.isPipeline = true to this? https://github.com/spinnaker/orca/blob/master/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy#L97-L99

gardleopard:
If we go for parentPipelineId and parentTaskId and drop the parentExecutionType we should be safe? The data of the triggers lives in parentExecution so we should be safe for expressions.

Is tasks like cloning only triggered from the gui?

tomaslin:
What is the actual reason to change to parentTaskId and parentPipelineId. I would close this PR and open a new one that sets trigger.isPipeline in OperationController code I pointed out earlier. 

I suspect there are other parts of the code that use the isPipeline field ( IIRC in the UI to generate the link back ). Your changes would remove this.

gardleopard:
@tomaslin the real reason is that I spent about six hours to figure out why the amiName does not get resolved on manually triggered pipelines, while it gets resolved if its triggered by a parent. I want to make the code cleaner, so that the next person that looks at this part of the code will understand it. Im fine with keeping the isPipeline there for a while, so that it can be fixed in the ui as well.

The reason that I'm continuing to discuss this, is to figure out what the next PR should be like so that I have a good start on that one and its aligned with how you think it should be.

So what I have now is:
1. Keep isPipeline as it is now and add it to https://github.com/spinnaker/orca/blob/master/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy#L97-L99.
2. Rename parentPipelineId to parentTaskId for tasks.
3. Remove the isPipeline usage in deck.
4. Remove isPipeline from orca.

Am I missing something here?

PR #1106, opened 2017-01-05
(google) Implement upsertImageTags for GCE.

duftler:


duftler:
@lwander or @danielpeach please review.
@ajordens fyi

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleImageTagger.java
```diff
@@ -0,0 +1,192 @@
...
+
+    Collection<MatchedImage> matchedImages = findImages(stageData.imageNames, stage);
+
+    stageData.imageNames = matchedImages.stream()
+      .map(matchedImage -> matchedImage.imageName)
+      .collect(Collectors.toSet());
+
+    // Built-in tags are not updatable
+    Map<String, String> tags = stageData.tags.entrySet().stream()
+      .filter(entry -> !BUILT_IN_TAGS.contains(entry.getKey()))
```
lwander: Why is this ignored here? Shouldn't this be caught by the `Validator` in clouddriver?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleImageTagger.java
```diff
@@ -0,0 +1,192 @@
...
+   * Return true iff the tags on the current machine image match the desired.
+   */
+  @Override
+  public boolean areImagesTagged(Collection<Image> targetImages, Stage stage) {
+    Collection<MatchedImage> matchedImages = findImages(
+      targetImages.stream().map(targetImage -> targetImage.imageName).collect(Collectors.toSet()),
+      stage
+    );
+
+    AtomicBoolean isUpserted = new AtomicBoolean(true);
```
lwander: Why does this need to be `Atomic`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleImageTagger.java
```diff
@@ -0,0 +1,192 @@
...
+
+    return isUpserted.get();
+  }
+
+  @Override
+  public String getCloudProvider() {
+    return "gce";
+  }
+
+  private Collection<MatchedImage> findImages(Collection<String> imageNames, Stage stage) {
```
lwander: If I'm reading this correctly, it'll throw whenever it doesn't find a desired image in `imageNames`. Why bother with [this check](https://github.com/spinnaker/orca/pull/1106/files#diff-b832c3399f681baf48877ccde1287963R118) if `targetImages` contains all images in `matchedImages` given that it hasn't thrown at that point?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleImageTagger.java
```diff
@@ -0,0 +1,192 @@
...
+
+        log.info(format("Found upstream image '%s' (executionId: %s)", upstreamImageName, stage.getExecution().getId()));
+      }
+    }
+
+    return imageNames.stream()
+      .map(targetImageName -> {
+        List<Map> allMatchedImages = oortService.findImage(getCloudProvider(), targetImageName, null, null, null);
+        Map matchedImage = allMatchedImages.stream()
+          .filter(image -> image.get("imageName").equals(targetImageName))
```
lwander: It feels like this is adding a lot of logic that will need to be redone if we ever decide to respect the GCP "project" namespacing for images.

duftler:
Relates to https://github.com/spinnaker/clouddriver/pull/1354.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleImageTagger.java
ajordens: `Atomic` is a bit of a hack but likely because it's doing a new school `forEach` that doesn't want to let you re-assign a variable.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleImageTagger.java
duftler: Mainly going for consistency with the aws implementation of image tagger here. We could handle it in either spot, but if done in the validator it would throw an error instead of just ignoring it. That list of built-in tags is 'known' by orca/clouddriver/rosco/deck.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleImageTagger.java
duftler: If/when we decide to do that, we'll have to find all the name<->name matches throughout deck/orca/clouddriver/rosco, and make them (roughly) resource link <-> resource link or project/name <-> project/name. I don't think this new bit adds much to that workload.

ajordens:
Now that there's Google + Amazon ImageTagger's, is there an opportunity to pull anything up?

Specifically thinking about what's in `areImagesTagged()` as it looks quite similar.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleImageTagger.java
duftler: That's the search so we can then try to match all the key/value pairs. I suppose we could drop the `if (matchedImage == null)`, but seems more future-proof this way.

duftler:
Was kind of hoping there'd be more opportunity for code-sharing, but the regional and `tagsByImageId` content in the aws find images response is not present in the gce response.

duftler:
Looks like a `findImages()` can probably be made common. Would probably need to make *ImageTaggerSupport common at the same time. I can work on this tomorrow.

duftler:
@ajordens PTAL. Have added a 2nd commit to this PR where I factor out some common code. I can squash the 2 commits after you have a look.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/image/ImageTagger.java
```diff
@@ -18,32 +18,88 @@
...
+        }
+
+        String upstreamImageName = (String) allMatchedImages.get(0).get("imageName");
+        imageNames.add(upstreamImageName);
+
+        log.info(format("Found upstream image '%s' (executionId: %s)", upstreamImageName, stage.getExecution().getId()));
+      }
+    }
+
+    Collection foundImages = new ArrayList();
```
duftler: Lost a fight here involving the type arg and java 8 type inferencing so had to drop the stream/map in favor of the older-style for-each.

duftler:
I tested this after the refactoring with both providers.

ajordens:
LGTM

PR #1107, opened 2017-01-05
(amazon) Lowercase image tag keys before using them to determine tagg…

duftler:
…ing operation completeness.


duftler:
Without this change, specifying a tag with uppercase characters causes the stage to never complete.
The tag keys are lowercased in clouddriver before being returned to orca: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/controllers/AmazonNamedImageLookupController.groovy#L113

duftler:
@ajordens or @anotherchrisberry please review.

ajordens:
LGTM

PR #1108, opened 2017-01-05
Add isPipeline = true if a parentPipeline is found.

gardleopard:
@tomaslin as you suggested in https://github.com/spinnaker/orca/pull/1105


tomaslin:
LGTM

PR #1109, opened 2017-01-05
Add the ability to delete a Fast Property via Pipeline Stage

zanthrash:
Allows for a boolean flag be set on the stage context to make a call to
delete one or more fast properties via a pipeline stage.

This will be primarily used for the new Global Fast Property work that is in flight.

### DIFF
--- a/orca-mahe/src/test/groovy/com/netflix/spinnaker/orca/mahe/tasks/CreatePropertiesTaskSpec.groovy
+++ b/orca-mahe/src/test/groovy/com/netflix/spinnaker/orca/mahe/tasks/CreatePropertiesTaskSpec.groovy
@@ -51,24 +51,11 @@ class CreatePropertiesTaskSpec extends Specification {
 
   def "assemble the property list from the context"() {
     given:
-    def scope = [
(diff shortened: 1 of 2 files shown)


orca-mahe/src/test/groovy/com/netflix/spinnaker/orca/mahe/tasks/CreatePropertiesTaskSpec.groovy
```diff
@@ -138,6 +138,56 @@ class CreatePropertiesTaskSpec extends Specification {
     }
   }
 
+  def "delete single persisted properties"() {
```
ajordens: Might be nice if things were refactored slightly such that we didn't need a 50 line test case, is this possible?

I don't quite know the context behind  deleting properties in the `CreatePropertiesTask`, but at a minimum you could extract the bit that's iterating through properties and upserting/deleting into it's own method that could be tested directly.

Or maybe you don't need such a verbose `given` to test the task? It doesn't seem like you need two pipeline stages, the email address, etc?

Otherwise this LGTM.

PR #1111, opened 2017-01-10
Adds orca-extension module to define extension interfaces.

cfieber:
Adds an orca-extensionpoint module as a no-dependency API for adding extensions to orca core functionality.
Defines PipelinePreprocessor extension point and wires it into OperationsController /orchestrate endpoint to transform supplied JSON object.
Adds orca-pipelinetemplate module as placeholder to start work on declarative continuous delivery efforts, with initial (no-op) PipelinePreprocessor implementation (also not enabled by default without `pipelineTemplate.enabled=true`

Some build updates to get rid of all subprojects having groovy as a compile time dependency.

@spinnaker/netflix-reviewers PTAL - does adding a separate module for extension interfaces seem like a dumb idea? (orca-core is a pretty heavy dependency to bring in currently, was trying to avoid that...)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -94,14 +94,10 @@ allprojects {
     }
   }
 }
-
 subprojects {
   dependencies {
     spinnaker.group('spockBase')
-    compile spinnaker.dependency('groovy')
-    compile spinnaker.dependency('guava')
-    compile spinnaker.dependency('slf4jApi')
-    testRuntime spinnaker.dependency('slf4jSimple')
+    testCompile spinnaker.dependency('groovy')
   }
 }
 
--- a/orca-core/orca-core.gradle
+++ b/orca-core/orca-core.gradle
@@ -17,6 +17,12 @@
 apply plugin: "io.franzbecker.gradle-lombok"
 
 dependencies {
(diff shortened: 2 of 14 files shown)


robzienert:
I like it!

robfletcher:
Orca core will be losing Spring Batch as a dependency in the near future (working on it now) but this might be a good idea anyway, I guess.

build.gradle
```diff
@@ -94,14 +94,10 @@ allprojects {
...
 }
-
 subprojects {
   dependencies {
     spinnaker.group('spockBase')
-    compile spinnaker.dependency('groovy')
-    compile spinnaker.dependency('guava')
-    compile spinnaker.dependency('slf4jApi')
-    testRuntime spinnaker.dependency('slf4jSimple')
+    testCompile spinnaker.dependency('groovy')
```
robfletcher: shouldn't this be implied by including the "spockBase" group?

cfieber: we would get some version of groovy as a transitive dependency of spock but not explicitly the version we use elsewhere.

most of the other modules have groovy in compile scope explicitly, but this doesn't (intending it to be java-only)

PR #1112, opened 2017-01-10
provider/appengine dynamic branch determined by webhook

danielpeach:
@lwander or @duftler please review.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineBranchFinder.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineBranchFinder.groovy
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineBranchFinder.groovy
```diff
@@ -0,0 +1,43 @@
...
+
+import com.netflix.spinnaker.orca.pipeline.model.Pipeline
+import com.netflix.spinnaker.orca.pipeline.model.Stage
+
+import java.util.regex.Pattern
+
+class AppEngineBranchFinder {
+  static void populateFromStage(Map operation, Stage stage) {
+    if (operation.fromTrigger && stage.execution instanceof Pipeline) {
+      def (source, project, slug, branch) = operation.branch.tokenize(":")
```
lwander: If there's not enough tokens, some of these could be `null`, right? Where does `operation.branch` come from?

danielpeach: This felt like a good idea when I wrote it. I'll just send the same information in a map under `operation.trigger` (the properties on `operation` come from the deploy description).

PR #1113, opened 2017-01-11
fix QUIP stage when build artifacts are missing

icfantv:
fixes the case were the build info/artifacts in the QUIP stage has been overwritten by a parallel or parent Jenkins stage.  the fix navigates the ancestors looking for the first match that contains the correct build info/artifacts based on the QUIP stage's properties.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
@@ -23,6 +23,8 @@ import com.netflix.spinnaker.orca.pipeline.model.Stage
 import groovy.transform.CompileDynamic
 import groovy.util.logging.Slf4j
 
+import java.util.regex.Pattern
+
 @Slf4j
 class PackageInfo {
   ObjectMapper mapper
@@ -32,6 +34,7 @@ class PackageInfo {
   boolean extractBuildDetails
   boolean extractVersion
   BuildDetailExtractor buildDetailExtractor
+  Pattern packageFilePattern
 
   PackageInfo(Stage stage, String packageType, String versionDelimiter, boolean extractBuildDetails, boolean extractVersion, ObjectMapper mapper) {
     this.stage = stage
@@ -41,6 +44,8 @@ class PackageInfo {
     this.extractVersion = extractVersion
     this.mapper = mapper
     this.buildDetailExtractor = new BuildDetailExtractor()
+
+    packageFilePattern = Pattern.compile("${stage.context.package}.*\\.${packageType}")
   }
 
   @VisibleForTesting
@@ -60,6 +65,11 @@ class PackageInfo {
       if (requestMap.buildInfo) { // package was built as part of the pipeline
         buildInfo = mapper.convertValue(requestMap.buildInfo, HashMap)
       }
+
+      if (!buildInfo?.artifacts) {
+        buildInfo = findBuildInfoInUpstreamStage(stage, packageFilePattern) ?: buildInfo
+      }
+
       return createAugmentedRequest(trigger, buildInfo, requestMap, allowMissingPackageInstallation)
     }
     return requestMap
@@ -215,4 +225,16 @@ class PackageInfo {
     }
   }
 
+  private static Map findBuildInfoInUpstreamStage(Stage currentStage, Pattern packageFilePattern) {
+    def upstreamStage = currentStage.ancestors().find {
+      artifactMatch(it.stage.context.buildInfo?.artifacts as List<Map<String, String>>, packageFilePattern)
+    }?.stage
+    return upstreamStage ? upstreamStage.context.buildInfo as Map : null
+  }
+
+  private static boolean artifactMatch(List<Map<String, String>> artifacts, Pattern pattern) {
+    artifacts?.find {
+      Map artifact -> pattern.matcher(artifact.get('fileName') as String ?: "").matches()
+    }
+  }
 }
--- a/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
+++ b/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
@@ -16,18 +16,108 @@
 package com.netflix.spinnaker.orca.pipeline.util
 
 import com.fasterxml.jackson.databind.ObjectMapper
-import com.netflix.spinnaker.orca.pipeline.model.Execution
 import com.netflix.spinnaker.orca.pipeline.model.Pipeline
 import com.netflix.spinnaker.orca.pipeline.model.PipelineStage
 import com.netflix.spinnaker.orca.pipeline.model.Stage
+
 import org.springframework.beans.factory.annotation.Autowired
 import spock.lang.Specification
 import spock.lang.Unroll
+
+import java.util.regex.Pattern
 
 class PackageInfoSpec extends Specification {
 
   @Autowired
   ObjectMapper mapper
+
+  @Unroll
+  def "Test that the artifact matching method returns whether or not there is a match"() {
+
+    given:
+    def pattern = Pattern.compile(artifactPattern)
+    def artifacts = artifactFilenames.collect { [fileName: it] }
+
+    expect:
+    PackageInfo.artifactMatch(artifacts, pattern) == expectedMatch
+
+    where:
+    artifactPattern  || artifactFilenames                          || expectedMatch
+    "foo.*"          || []                                         || false
+    "bat.*"          || [[], []]                                   || false
+    "testFileName.*" || ["testFileName_1"]                         || true
+    "baz.*"          || [[fileName: "badTestFileName_1"]]          || false
+    "testFileName.*" || ["testFileName_1", "testFileName_2"]       || true
+    "blah.*"         || ["badTestFileName_1", "badTestFileName_1"] || false
+    "testFileName.*" || ["badTestFileName", "testFileName_1"]      || true
+    "testFileName.*" || ["testFileName_1", "badTestFileName"]      || true
+  }
+
+  def "If no artifacts in current build info or ancestor stages, code should execute as normal"() {
+
+    given:
+    def quipStage = new PipelineStage(
+      execution: Pipeline.builder().withTrigger(["buildInfo": ["artifacts": [["fileName": "testFileName"]]]]).build(),
+      context: [buildInfo: [name: "someName"], package: "testPackageName"]
+    )
+
+    def packageType = PackageType.DEB
+    def packageInfo =
+      new PackageInfo(quipStage, packageType.packageType, packageType.versionDelimiter, true, false, new ObjectMapper())
+
+    when:
+    packageInfo.findTargetPackage(false)
+
+    then:
+    def ex = thrown(IllegalStateException)
+    ex.message.startsWith("Unable to find deployable artifact starting with")
+  }
+
+  @Unroll
+  def "If no artifacts in current build info, find from first ancestor that does"() {
+    given:
+    Stage quipStage = new PipelineStage(new Pipeline(), "type", "name", [buildInfo: [name: "someName"], package: "testPackageName"])
+    quipStage.stageNavigator = Mock(StageNavigator) {
+      findAll(_, _) >> {
+        return inputFileNames.collect() {
+          new StageNavigator.Result(new PipelineStage(context: [
+            buildInfo: [
+              artifacts: [
+                [fileName: "${it}.deb"],
+                [fileName: "${it}.war"],
+                [fileName: "build.properties"]
+              ],
+              url: "http://localhost",
+              name: "testBuildName",
+              number: "1"
+            ]
+          ]))
+        }
+      }
+    }
+
+    PackageType packageType = PackageType.DEB
+    PackageInfo packageInfo =
+      new PackageInfo(quipStage, packageType.packageType, packageType.versionDelimiter, true, false, new ObjectMapper())
+
+    when:
+    def requestMap = packageInfo.findTargetPackage(true)
+
+    then:
+    requestMap.package == expected
+
+    where:
+    inputFileNames                                                                         || expected
+    ["testPackageName_1.0abc298"]                                                          || "testPackageName_1.0abc298"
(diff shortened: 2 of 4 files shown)


icfantv:

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -215,4 +228,20 @@ class PackageInfo {
     }
   }
 
+  @CompileDynamic
+  private Map findBuildInfoInUpstreamStage() {
+
+    stage?.ancestors()?.findAll {
```
ajordens: Maybe something like ...

```
def upstreamStage = stage.ancestors().find { 
  artifactMatch(it.stage.context.buildInfo?.artifacts as List<Map<String, String>)
}?.stage
return upstreamStage ? upstreamStage.context.buildInfo as Map : null
```

If you made this method static and passed in the `Stage`, you could unit it test it independently of anything else.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -215,4 +228,20 @@ class PackageInfo {
...
+      it?.stage?.context?.buildInfo?.artifacts
+    }?.find {
+      artifactMatch(it.stage.context?.buildInfo?.artifacts as List)
+    }?.stage?.context?.buildInfo as Map
+  }
+
+  @CompileDynamic
+  private boolean artifactMatch(final List artifacts) {
```
ajordens: We don't generally apply `final` to parameters unless necessary ... although I'd probably suggest a stronger type `List<Map<String, String>>`.

That might actually let you get rid of the `@CompileDynamic`.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/model/TestPipelineStage.groovy
```diff
@@ -0,0 +1,34 @@
+package com.netflix.spinnaker.orca.pipeline.model
+
+import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder
+import com.netflix.spinnaker.orca.pipeline.TaskNode
+import groovy.transform.CompileStatic
+import groovy.transform.ToString
+
+@CompileStatic
+@ToString(includeNames = true, includeSuper = true)
+class TestPipelineStage extends PipelineStage implements StageDefinitionBuilder {
```
ajordens: This looks nearly identical to `TestStage`?

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -16,19 +16,135 @@
...
 
+import java.util.function.BiFunction
+
 class PackageInfoSpec extends Specification {
 
   @Autowired
   ObjectMapper mapper
 
+  def "If no artifacts in current build info or ancestor stages, code should execute as normal"() {
+    final Pipeline pipeline = Pipeline.builder().withTrigger(["buildInfo": ["artifacts": [["fileName": "testFileName"]]]]).build()
```
ajordens: should wrap in a `given:` block

Also, the `final` is really necessasry

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -16,19 +16,135 @@
...
+    final Stage quipStage = new PipelineStage()
```
ajordens: Believe there's a constructor on `PipelineStage` that would let you pass in the pipeline and context.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
ajordens: Also ... it's groovy so we can `def pipeline`.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -16,19 +16,135 @@
...
+    quipStage.execution = pipeline
+    quipStage.context = [buildInfo: [name: "someName"], package: "testPackageName"]
+
+    final PackageType packageType = PackageType.DEB
```
ajordens: we can `def` all these variables as well

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -16,19 +16,135 @@
...
+        url: "http://localhost",
+        name: "testBuildName",
+        number: "1"
+      ]
+    ]
+
+    final Pipeline pipeline = new Pipeline()
+    final Stage quipStage = new PipelineStage() {
+      @Override
+      List<StageNavigator.Result> ancestors(BiFunction<Stage<Pipeline>, StageDefinitionBuilder, Boolean> matcher) {
```
ajordens: I don't think you should need to override `ancestors` ... vs just giving it a proper StageNavigator (even if said `StageNavigator` is a Mock that returns whatever you'd like from

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -16,19 +16,135 @@
...
+    when:
+    final Map requestMap = packageInfo.findTargetPackage(false)
+
+    then:
+    final fileName = context.buildInfo.artifacts[0].fileName
+    requestMap.package == fileName.substring(0, fileName.indexOf('.deb'))
+  }
+
+  def "If no artifacts in current build info, find from first ancestor that does"() {
+    final Map goodContext = [
```
ajordens: Any opportunity to reduce the duplication and cut these tests down from ~50 lines each?

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/model/TestPipelineStage.groovy
icfantv: Yes, it is identical in content, but it only implements `StageDefinitionBuilder` and I need something that also extends `AbstractStage` or I don't have `context`, amongst other things.  Open to better alternatives here.

icfantv: To be sure, we have classes that do both, but not in this artifact.

icfantv:
@ajordens updated per your recommendations/requests.  PTAL && LMK.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -41,6 +44,8 @@ class PackageInfo {
     this.extractVersion = extractVersion
     this.mapper = mapper
     this.buildDetailExtractor = new BuildDetailExtractor()
+
+    filePattern = Pattern.compile("${stage?.context?.package}.*\\.${packageType}")
```
ajordens: I don't think you need to `stage?.context?.package` ... should be able to get away with `stage.context.package`.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -215,4 +225,17 @@ class PackageInfo {
     }
   }
 
+  private static Map findBuildInfoInUpstreamStage(Stage currentStage, Pattern pattern) {
+
```
ajordens: minor but new line isn't necessary ... would probably give `Pattern pattern` a more descriptive name (maybe `artifactFilePattern`)

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -16,19 +16,119 @@
...
+
+    when:
+    packageInfo.findTargetPackage(false)
+
+    then:
+    def ex = thrown(IllegalStateException)
+    ex.message.startsWith("Unable to find deployable artifact starting with")
+  }
+
+  private static Map gc(String name) {
```
ajordens: Sorry ... I think there was some confusion earlier.

In this particular usage, I don't believe you need to extract this logic into a method like `gc`.

I would normally use the short alias methods when I want to tidy up the colums in `where:` block.

Since `sr()` is only called in a single location, I'd probably just inline this there.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -16,19 +16,119 @@
...
+    [
+      buildInfo: [
+        artifacts: [
+          [fileName: "${name}.0abc298.deb"],
```
ajordens: Generally speaking I'd suggest avoiding tying constants across methods ...

ie) `0abc298` is used here and in `requestMap.package == "testPackageName_1.0abc298"`.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -16,19 +16,119 @@
...
+
+    PackageType packageType = PackageType.DEB
+    PackageInfo packageInfo =
+      new PackageInfo(quipStage, packageType.packageType, packageType.versionDelimiter, true, false, new ObjectMapper())
+
+    when:
+    Map requestMap = packageInfo.findTargetPackage(true)
+
+    then:
+    requestMap.package == "testPackageName_1.0abc298"
```
ajordens: Normally I'd suggest something like ...

```
where:
package   | inputFileNames      || expectedPackage
"appName" | ["appName_1.0.deb"] || "appName_1.0.deb"
...
```

Does that jive with what you're trying to cover in this test?

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -16,19 +16,119 @@
...
+    def pattern = Pattern.compile("${input.fileName}.*")
+
+    when:
+    def result = PackageInfo.artifactMatch(input.artifacts, pattern)
+
+    then:
+    result == matchFound
+
+    where:
+    input                                                                                                     || matchFound
```
ajordens: Maybe a bit of confusion again ...

```
where:
artifactPattern || artifactFilenames        || expectedMatch
"foobar.*"      || []                       || false
"foobar.*"      || ["foobar_1", "foobar_2"] || true

File: orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy [hunk 15]
```diff
@@ -16,19 +16,119 @@
...
   ObjectMapper mapper
 
+  @Unroll
+  def "Test that the artifact matching method returns whether or not there is a match"() {
+
+    given:
+    def pattern = Pattern.compile("${input.fileName}.*")
+
+    when:
+    def result = PackageInfo.artifactMatch(input.artifacts, pattern)
```
ajordens: Could also be written as:

```
given:
def pattern = Pattern.compile(artifactPattern)
def artifacts = artifactFilenames.collect { [fileName: it] }

expect:
PackageInfo.artifactMatch(artifacts, pattern) == expectedMatch
```

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -16,19 +16,119 @@
...
+        url: "http://localhost",
+        name: "testBuildName",
+        number: "1"
+      ]
+    ]
+  }
+
+  private static List<StageNavigator.Result> sr(List<String> names) {
+    names.collect() {
+      new StageNavigator.Result(new TestPipelineStage(context: gc(it)))
```
ajordens: Again don't actually need `TestPipelineStage` ... should be able to use `new PipelineStage(context: [:])`

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
icfantv: While it's probably the case in practice when the app is running, some of the other, preexisting tests pass `null` values which will cause exceptions to be thrown if the null-safe operators aren't present.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
icfantv: Not exactly, but I get your drift.  I'll run with it.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
ajordens: I'd suggest fixing the tests if there aren't too many to align with what's expected when running in production.

icfantv:
@ajordens ok.  issues/concerns addressed.

icfantv:
Test fixed.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
ajordens: I'd still suggest inlining this method down below.

ajordens:
Nice work following through on this @icfantv!

PR #1115, opened 2017-01-11
orca-clouddriver: Make backoff and timeout configurable for WaitForUpsertedImageTagsTask

gogreen53:
Since some providers often take >10min to tag an image.


robzienert:
Seems fine to me, although I'd tend to prefer a `tasks.waitForUpsertedImageTags` path so it directly matches the name of the task.

ajordens:
Curious what provider is taking > 10 minutes?

We haven't seen that with AWS.

gogreen53:
@robzienert I can change it out, not a big deal

@ajordens Yup, AWS. Happens pretty often to us during the mid-afternoon to mid-evening PST

ajordens:
How many images are you indexing per account/region?

If you manually specified a longer timeout on the pipeline, how long does it actually take to complete?

Lastly, if you searched for "ImageCachingAgent" and "completed in" in your clouddriver logs, what sort of timings do you see?

Not against making the defaults customizable (more so than needing to change them per pipeline), but it does feel like a bit of a bandaid unless we're talking very significant numbers of images (_in which case it might be worth investigating force cache refresh on an individual image_).

robzienert:
Closing - if you'd still like to make this change, feel free to re-open and address the comments.

PR #1116, opened 2017-01-12
Avoid building downstream flows for stages that have been previously …

ajordens:
…built

Failure to do this results in the following exception:
`No start state was found. You must specify at least one step in a job.`


ajordens:
PTAL @robfletcher

PR #1122, opened 2017-01-14
Try orchestrations on v2 again

robfletcher:


orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/SpringBatchExecutionRunner.java
```diff
@@ -139,7 +138,11 @@ public SpringBatchExecutionRunner(
 
   private <E extends Execution<E>> JobParameters createJobParameters(E subject) {
     JobParametersBuilder params = new JobParametersBuilder();
-    params.addString("pipeline", subject.getId());
+    if (subject instanceof Pipeline) {
+      params.addString("pipeline", subject.getId());
+    } else if (subject instanceof Orchestration) {
+      params.addString("orchestration", subject.getId());
+    }
```
robfletcher: This was missing from previous PR

PR #1123, opened 2017-01-15
pipeline templates config wip

robzienert:
Still a way, way early start into the implementation of the [proposed dcd-spec](https://github.com/robzienert/dcd-spec), but this is the general direction I'm headed down. Looking for initial feedback before I start dumping a bunch of time into it next week. Stole a bit of code from Cam's spike.

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/config/PipelineTemplateConfiguration.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/config/PipelineTemplateConfiguration.java
@@ -16,11 +16,28 @@
 
 package com.netflix.spinnaker.orca.config;
 
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.github.jknack.handlebars.Handlebars;
 import com.netflix.spinnaker.orca.pipelinetemplate.PipelineTemplateModule;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 
 @ConditionalOnProperty("pipelineTemplate.enabled")
 @ComponentScan(basePackageClasses = PipelineTemplateModule.class)
 public class PipelineTemplateConfiguration {
+
+  @Bean
+  Handlebars handlebars() {
+    return new Handlebars();
+  }
+
+  @Bean
+  ObjectMapper pipelineTemplateObjectMapper() {
+    return new ObjectMapper()
+      .enable(SerializationFeature.INDENT_OUTPUT)
+      .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+  }
 }
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplateObjectMapper.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplateObjectMapper.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.pipelinetemplate;
+
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
+public class PipelineTemplateObjectMapper extends ObjectMapper {
+
+  public PipelineTemplateObjectMapper() {
+      enable(SerializationFeature.INDENT_OUTPUT);
+      disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+  }
+}
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/MapMerge.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/MapMerge.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.pipelinetemplate.v1schema;
+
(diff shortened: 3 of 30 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/TemplateConfiguration.java
```diff
@@ -0,0 +1,195 @@
...
+
+    public List<String> getInherit() {
+      return inherit;
+    }
+
+    public void setInherit(List<String> inherit) {
+      this.inherit = inherit;
+    }
+
+    public Map<String, Object> getConcurrentExecutions() {
```
jeyrschabu: Minor comment here. IMHO, it would be beneficial from a user's standpoint, as well as to avoid errors, to have this come out as a ```[String:Bool]```.  So for example ```onStageFailure``` we would have ```haltEntirePiple: true```.
This allows the user to just enable the option by passing true.

jeyrschabu:
Full disclosure, I miss some of the talks on the game plan last week. 
Wanted to share some of the high level concepts and ideas I had in mind so I'm just gonna write them here. I think the following items, at least from a high level view, are concretely addressed in the proposed spec, so I don't think I am too far off.  (for example, the idea of a config being the end result of one or more templates). 

- Hierarchical Configurations
- We can ingest a somehow invariant representation of a pipeline or a set of configs
- At each level, a config will interpret portions it's tied to, either directly or indirectly
- The output of these levels is an atomic part of the execution model. (a fully executable chuck)
- The top level helps guide how each individual chucks are layered
- The final output is an inherently deterministic way of generating flavors of a pipeline

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/StageDefinition.java
```diff
@@ -0,0 +1,131 @@
...
+  private Map<String, Object> config;
+  private List<Map<String, Object>> notifications;
+  private String comments;
+
+  @Override
+  public String getName() {
+    return id;
+  }
+
+  public static class InjectionRule {
```
jeyrschabu: I'm not 100% sure I understand the concept of "Injection". I would imagine, in this declarative model, that a user has more control over the definition of a stage/configuration/module. Do you have a concrete example of how this would work ?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/PipelineTemplate.java
```diff
@@ -0,0 +1,171 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+public class PipelineTemplate {
+
+  private String schema;
+  private String id;
+  private String source;
+  private List<Variable> variables;
```
jeyrschabu: Are variables meant to be like parameters ?

robzienert: In that they're runtime-configuration, yes, however variables are pipeline template-level only. They won't cascade values to pipeline (ui) parameters.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/StageDefinition.java
robzienert: Yep. In my mind, inject is a "break glass in case of fire" or "mostly the same" feature. Let's say the company paved road policy fits your use case almost exactly and you want to use it, but your team is responsible for a new canary feature that should be run just before the deploy stage. You want to stay in-sync with the paved road, but also want to offer teams the "canary-enabled paved road".

```yaml
# Paved road pipeline template
# spinnaker://netflix/pavedRoad
id: pavedRoad
stages:
- id: findImage
  type: findImageFromTags
  config: '{ /* */ }'
- id: deploy
  type: deploy
  config: '{ /* */ }
```
  
```yaml
# canary team + paved road pipeline template
# spinnaker://canaryteam/pavedRoad
id: canaryPavedRoad
source: spinnaker://pavedRoad
variables:
- name: canaryScoreThreshold
  description: The score (0-100) which a canary should meet to be considered a success.
  type: int
stages:
- id: canary
  inject:
    before: findImage
  type: canary
  config:
    scoreThreshold: "{{ canaryScoreThreshold }}"
```
  
Then, the canary team starts distributing their version of the pipeline for teams, so a team switches their pipeline to use the new canary-enhanced pipeline. Once canary becomes best-practice, it'd get rolled into upstream and then eventually the canary team would deprecate their pipeline (somehow - not part of this scope).

```diff
# spinnaker://myTeam/myApp
id: myApp
pipeline:
  template:
-    source: spinnaker://netflix/pavedRoad
+    source: spinnaker://canaryteam/pavedRoad
  variables:
    canaryScoreThreshold: 90
    # ...
```

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/TemplateConfiguration.java
robzienert: Agree. A lot of my early code is pretty dynamic-y, easier for me to concept stuff out.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplateObjectMapper.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.pipelinetemplate;
+
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
+public class PipelineTemplateObjectMapper extends ObjectMapper {
```
cfieber: don't extend object mapper, just configure a standard one

extending sets us up for failure on jackson dependency update

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/config/PipelineTemplateConfiguration.java
```diff
@@ -16,11 +16,24 @@
...
 @ComponentScan(basePackageClasses = PipelineTemplateModule.class)
 public class PipelineTemplateConfiguration {
+
+  @Bean
+  Handlebars handlebars() {
+    return new Handlebars();
+  }
+
+  @Bean
+  PipelineTemplateObjectMapper objectMapper() {
```
cfieber: name this method something other than 'objectMapper' (`pipelineTemplateObjectMapper` is probably a good bet - want to avoid conflicting with a standard object mapper in the context.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/Graph.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.pipelinetemplate.v1schema.graph;
+
+
+import java.util.Comparator;
+import java.util.PriorityQueue;
+
+public class Graph {
```
cfieber: [jgrapht](http://jgrapht.org/) is a pretty solid graph library, probably worth a look

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/MapMerge.java
```diff
@@ -0,0 +1,91 @@
...
+   *   - the value is taken from overrides
+   *
+   * Each remaining key in overrides is then added to the resulting map.
+   *
+   * @param original the original Map
+   * @param overrides the Map to override/merge into original
+   * @return a new Map containing the merge of original and overrides (never null)
+   * @throws IllegalStateException if incompatible types exist between original and overrides
+   */
+  public static Map<String, Object> merge(Map<String, Object> original, Map<String, Object> overrides) {
```
cfieber: pretty sure there was a test case for this in the spike code - worth CAP theoreming that over here too

robzienert:
I'll address some of these comments, then get a merge going so people can take what I've got so far and rip / tear things up, extend and so-forth. As I mentioned IRL, I'm not particularly married to any of the code.

PR #1128, opened 2017-01-18
Support sourcing of pipeline templates from `file`, `http` and `https…

ajordens:
…` URIs

Currently only the first parent source is fetched (_limitation_)

```
curl -X "POST" "https://localhost:8083/orchestrate" \
     -H "Content-Type: application/context+json" \
     -d $'{
  "type": "templatedPipeline",
  "config": {
    "pipeline": {
		"template": {
			"source": "https://gist.githubusercontent.com/ajordens/8a227811ad994434927c4f410f8252d1/raw/3e2f67c106bcf4bb636946bce8e4a172ea5762b7/simple-v001.yml"
	    }
	}
  }
}'
```


robzienert:
👍 💯

PR #1129, opened 2017-01-18
Base orca halconfig

lwander:
@duftler PTAL (also, what's the status of that TODO in there)?


duftler:
Wish I could remember what flex is. I would just try starting without that attribute and drop it if it starts ok.

PR #1130, opened 2017-01-18
Handling interesting  health provider for titus

jeyrschabu:
- Added an interface that allows processing and augmentation of the context. For titus we are adding a label and use it in checking whether or not to set the interestingProviderNames for discovery purposes

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.orca.DefaultTaskResult;
 import com.netflix.spinnaker.orca.ExecutionStatus;
 import com.netflix.spinnaker.orca.RetryableTask;
 import com.netflix.spinnaker.orca.TaskResult;
+import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.InterestingHealthProviderNamesSupplier;
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.ServerGroupCreator;
 import com.netflix.spinnaker.orca.clouddriver.utils.CloudProviderAware;
 import com.netflix.spinnaker.orca.front50.Front50Service;
@@ -33,7 +34,10 @@ import org.springframework.stereotype.Component;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
+import java.util.HashMap;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
@@ -43,11 +47,14 @@ public class DetermineHealthProvidersTask implements RetryableTask, CloudProvide
 
   private final Front50Service front50Service;
   private final Map<String, String> healthProviderNamesByPlatform;
+  private final Collection<InterestingHealthProviderNamesSupplier> interestingHealthProviderNamesSuppliers;
 
   @Autowired
   public DetermineHealthProvidersTask(Front50Service front50Service,
+                                      Collection<InterestingHealthProviderNamesSupplier> interestingHealthProviderNamesSuppliers,
                                       Collection<ServerGroupCreator> serverGroupCreators) {
     this.front50Service = front50Service;
+    this.interestingHealthProviderNamesSuppliers = interestingHealthProviderNamesSuppliers;
     this.healthProviderNamesByPlatform = serverGroupCreators
       .stream()
       .filter(serverGroupCreator ->  serverGroupCreator.getHealthProviderName().isPresent())
@@ -83,6 +90,23 @@ public class DetermineHealthProvidersTask implements RetryableTask, CloudProvide
       }
 
       Application application = front50Service.get(applicationName);
+      Optional<InterestingHealthProviderNamesSupplier> healthProviderNamesSupplierOptional = interestingHealthProviderNamesSuppliers
+        .stream()
+        .filter(supplier -> supplier.supports(getCloudProvider(stage), stage))
+        .findFirst();
+
+      if (healthProviderNamesSupplierOptional.isPresent()) {
+        List<String> interestingHealthProviderNames = healthProviderNamesSupplierOptional.get().process(getCloudProvider(stage), stage);
+        Map<String, List<String>> results = new HashMap<>();
+
+        if (interestingHealthProviderNames != null) {
+          // avoid a `null` value that may cause problems with ImmutableMap usage downstream
+          results.put("interestingHealthProviderNames", interestingHealthProviderNames);
+        }
+
+        return new DefaultTaskResult(ExecutionStatus.SUCCEEDED, results);
+      }
+
       if (application.platformHealthOnly == Boolean.TRUE && application.platformHealthOnlyShowOverride != Boolean.TRUE) {
         // if `platformHealthOnlyShowOverride` is true, the expectation is that `interestingHealthProviderNames` will
         // be included in the request if it's desired ... and that it should NOT be automatically added.
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesSupplier.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesSupplier.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.tasks.servergroup;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.orca.clouddriver.OortService;
+import com.netflix.spinnaker.orca.kato.pipeline.support.SourceResolver;
+import com.netflix.spinnaker.orca.kato.pipeline.support.StageData;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import retrofit.client.Response;
+import java.util.List;
+import java.util.Map;
+import java.util.Arrays;
+
+@Slf4j
+@Component
+public class TitusInterestingHealthProviderNamesSupplier implements InterestingHealthProviderNamesSupplier {
+  private static final String TITUS = "titus";
+  private static final String INTERESTING_HEALTH_PROVIDER_NAMES = "interestingHealthProviderNames";
+  private static final String ROLLING_PUSH = "rollingpush";
+  private static final List<String> SUPPORTED_STAGES = Arrays.asList("enableServerGroup", "createServerGroup", "cloneServerGroup");
+
+  private final OortService oortService;
+  private final ObjectMapper objectMapper;
+  private final SourceResolver sourceResolver;
+
+  @Autowired
+  public TitusInterestingHealthProviderNamesSupplier(OortService oortService,
+                                                     SourceResolver sourceResolver,
+                                                     ObjectMapper objectMapper) {
+    this.oortService = oortService;
+    this.sourceResolver = sourceResolver;
+    this.objectMapper = objectMapper;
+  }
+
+  @Override
+  public boolean supports(String cloudProvider, Stage stage) {
+    if (!cloudProvider.equals(TITUS)) {
+      return false;
+    }
+
+    String strategy = (String) stage.getContext().get("strategy");
+    return SUPPORTED_STAGES.contains(stage.getType()) && ROLLING_PUSH.equals(strategy);
+  }
+
+  @Override
+  public List<String> process(String cloudProvider, Stage stage) {
+    try {
+      StageData stageData = (StageData) stage.mapTo(StageData.class);
+      StageData.Source source = sourceResolver.getSource(stage);
+      String serverGroupName =  source.getServerGroupName() != null ? source.getServerGroupName() : source.getAsgName();
+
+      Response response = oortService.getServerGroup(
+        stageData.getApplication(),
+        source.getAccount(),
+        stageData.getCluster(),
+        serverGroupName,
+        source.getRegion(),
+        cloudProvider
+      );
+
+      Map serverGroup = objectMapper.readValue(response.getBody().in(), Map.class);
+      Map titusServerGroupLabels = (Map) serverGroup.get("labels");
+
+      if (titusServerGroupLabels != null && titusServerGroupLabels.containsKey(INTERESTING_HEALTH_PROVIDER_NAMES)) {
+        String healthProviderNames = (String) titusServerGroupLabels.get(INTERESTING_HEALTH_PROVIDER_NAMES);
+        return Arrays.asList(healthProviderNames.split(","));
+      }
+
+    } catch (Exception e) {
+      log.error("Failed to process interesting health provider names for cloud provider {} on stage {} ", cloudProvider, stage, e);
+    }
+
+    return null;
+  }
+}
--- a/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTaskSpec.groovy
+++ b/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTaskSpec.groovy
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.clouddriver.tasks.providers.aws.AmazonServerGroupCreator
 import com.netflix.spinnaker.orca.clouddriver.tasks.providers.gce.GoogleServerGroupCreator
 import com.netflix.spinnaker.orca.clouddriver.tasks.providers.kubernetes.KubernetesServerGroupCreator
+import com.netflix.spinnaker.orca.clouddriver.tasks.providers.titus.TitusServerGroupCreator
 import com.netflix.spinnaker.orca.front50.Front50Service
 import com.netflix.spinnaker.orca.front50.model.Application
 import com.netflix.spinnaker.orca.pipeline.model.Pipeline
@@ -35,7 +36,8 @@ class DetermineHealthProvidersTaskSpec extends Specification {
   @Subject
   def task = new DetermineHealthProvidersTask(
     front50Service,
-    [new KubernetesServerGroupCreator(), new AmazonServerGroupCreator(), new GoogleServerGroupCreator()]
+    [],
+    [new KubernetesServerGroupCreator(), new AmazonServerGroupCreator(), new GoogleServerGroupCreator(), new TitusServerGroupCreator()]
   )
 
   @Unroll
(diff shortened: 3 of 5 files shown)


jeyrschabu:

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+  @Override
+  public boolean supports(String cloudProvider, Application application, Stage stage) {
+    if (cloudProvider.equalsIgnoreCase(TITUS)) {
+      if (application.platformHealthOnly == Boolean.TRUE && application.platformHealthOnlyShowOverride == Boolean.TRUE) {
+        return true;
+      }
+
+      // checks titus job label
+      Map<String, Object> labels = (Map<String, Object>)stage.getContext().get("labels");
+      return (labels != null && labels.containsKey(INTERESTING_HEALTH_PROVIDER_NAMES) && String.valueOf(labels.get(INTERESTING_HEALTH_PROVIDER_NAMES)).equalsIgnoreCase(TITUS));
```
tomaslin: are we mapping a list to a string in the label? 

It feels like we want to set a more explicit discoveryEnabled label and convert this into interestingHealthProviderNames, but can be convinced otherwise. This doesn't feel very future proof once Titus starts adding ALB support and we can then have [ "Titus", "Discovery" ] or [ "Titus", "Loadbalancer" ]

jeyrschabu: The choice on the label name is quite flexible. I'm not particularly attached to ```interestingHealthProviderNames```.  A label from Titus perspective is an arbitrary key-value pair, so we can adjust. I'm planning on doing more testing before it becomes a merge candidate. Wanted to get some opinion on the direction first.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/InterestingHealthProviderNamesProcessor.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.tasks.servergroup;
+
+import com.netflix.spinnaker.orca.front50.model.Application;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+
+import java.util.List;
+
+public interface InterestingHealthProviderNamesProcessor {
```
ajordens: Maybe `InterestingHealthProviderNameSupplier`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
```diff
@@ -83,6 +88,18 @@ public TaskResult execute(Stage stage) {
       }
 
       Application application = front50Service.get(applicationName);
+      List<String> processedInterestingHealthProviderNames = interestingHealthProviderNamesProcessors
+        .stream()
+        .filter(healthProcessor -> healthProcessor.supports(getCloudProvider(stage), application, stage))
+        .flatMap(healthProcessor -> healthProcessor.process(stage).stream())
```
ajordens: What about if we filtered it down and took the first match ... it's probably a little confusing to process multiple times. I'm not sure there would ever be a legitimate reason to do so.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
```diff
@@ -83,6 +88,18 @@ public TaskResult execute(Stage stage) {
...
+        .collect(Collectors.toList());
+
+      if (!processedInterestingHealthProviderNames.isEmpty()) {
```
ajordens: IMO we should only do this if we didn't find a specific processor ... as it's "possible" that a process might happen to return an empty list.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesProcessor.java
ajordens: Not sure the `String.valueOf().equalsIgnoreCase()` is needed or legitimate.

I was thinking that it would be more along the lines of ... if there's a label, we use it without any expectation of it's value.

Maybe prefix the label with something that namespaces it to Spinnaker ... `spinnaker:interestingHealthProviderNames` or something maybe?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+  @Override
+  public boolean supports(String cloudProvider, Application application, Stage stage) {
+    if (cloudProvider.equalsIgnoreCase(TITUS)) {
+      if (application.platformHealthOnly == Boolean.TRUE && application.platformHealthOnlyShowOverride == Boolean.TRUE) {
+        return true;
+      }
+
+      // checks titus job label
+      Map<String, Object> labels = (Map<String, Object>)stage.getContext().get("labels");
```
ajordens: By convention I think there's a space between the cast and variable ... `(Map<String, Object>) stage`

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTaskSpec.groovy
```diff
@@ -55,14 +58,16 @@ class DetermineHealthProvidersTaskSpec extends Specification {
...
+    [interestingHealthProviderNames: null]                                                          | null                      || [:]
+    [application: "app"]                                                                            | bA("app", true, false)    || [interestingHealthProviderNames: ["Amazon"]]
+    [application: "app", cloudProvider: "gce"]                                                      | bA("app", true, false)    || [interestingHealthProviderNames: ["Google"]]
+    [application: "app"]                                                                            | bA("app", true, true)     || [:]                                           // no health provider names when platformHealthOnlyShowOverride is true
+    [application: "app", cloudProvider: "titus"]                                                    | bA("app", true, true)     || [interestingHealthProviderNames: ["Titus"]]
+    [application: "app", cloudProvider: "titus", labels: [interestingHealthProviderNames: "Titus"]] | bA("app", false, true)    || [interestingHealthProviderNames: ["Titus"]]
```
ajordens: How about a test where the label is `Titus` and `Discovery`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/InterestingHealthProviderNamesProcessor.java
jeyrschabu: Fair

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
jeyrschabu: I think, the current pattern works as well. but sure thats an easy change

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTaskSpec.groovy
jeyrschabu: this ``` String.valueOf().equalsIgnoreCase()``` not legitimate  as in not necessary or made up? the label field is a [String:Object] dictionary. The string.valueOf() was meant to do the job of the toString() on the object to the the comparison against the string. I am happy to rewrite that, no big deal

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+@Component
+public class TitusInterestingHealthProviderNamesProcessor implements InterestingHealthProviderNamesProcessor {
+  private static final String TITUS = "Titus";
+  private static final String INTERESTING_HEALTH_PROVIDER_NAMES = "interestingHealthProviderNames";
+
+  @Override
+  public boolean supports(String cloudProvider, Application application, Stage stage) {
+    if (cloudProvider.equalsIgnoreCase(TITUS)) {
+      if (application.platformHealthOnly == Boolean.TRUE && application.platformHealthOnlyShowOverride == Boolean.TRUE) {
```
ajordens: I don't think you actually need this ... this `Processor` kicks in only if the current server group being acted upon has a label right?

Two things:
- Should it be restricted to a DeployStage w/ RollingPush
- Should it be looking at labels on the current server group (ie. fetching it from clouddriver) vs. what's in the `stage.getContext()`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Map;
+import java.util.Arrays;
+
+@Slf4j
+@Component
+public class TitusInterestingHealthProviderNamesSupplier implements InterestingHealthProviderNamesSupplier {
+  private static final String TITUS = "Titus";
+  private static final String INTERESTING_HEALTH_PROVIDER_NAMES = "interestingHealthProviderNames";
+  private static final String ROLLING_PUSH = "rollingpush";
+  private static final List<String> SUPPORTED_STAGES = Arrays.asList("deploy", "enableServerGroup", "createServerGroup", "cloneServerGroup", "disableServerGroup");
```
jeyrschabu: @tomaslin @ajordens are these all the stages we care about ?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @Override
+  public boolean supports(String cloudProvider, Stage stage) {
+    return cloudProvider.equalsIgnoreCase(TITUS) && SUPPORTED_STAGES.contains(stage.getType()) && stage.getContext().containsKey("strategy") && stage.getContext().get("strategy").equals(ROLLING_PUSH);
+  }
+
+  @Override
+  public List<String> process(String cloudProvider, OortService oortService, ObjectMapper objectMapper, Stage stage) {
+    String region = (String) stage.getContext().get("region");
+    String asgName = (String) stage.getContext().get("asgName");
```
ajordens: Typically we would check both `serverGroupName` and `asgName` (former is _more_ correct but later is legacy and still commonly used).

Also, I believe that `asgName` / `serverGroupName` is contained within `stage.context.source`. See `SourceResolver.getSource(stage)` that `RollingPushStrategy` is using.

`deck` is sending the following payload for a rolling push:

```
{
    "job": [{
        "application": "spintest",
        "strategy": "rollingpush",
        "stack": "",
        "freeFormDetails": "",
        "credentials": "test",
        "cooldown": 10,
        "healthCheckGracePeriod": 600,
        "healthCheckType": "EC2",
        "terminationPolicies": ["Default"],
        "loadBalancers": [],
        "useSourceCapacity": true,
        "capacity": {
            "min": 1,
            "max": 1,
            "desired": 1
        },
        "targetHealthyDeployPercentage": 100,
        "availabilityZones": {
            "us-east-1": ["us-east-1c", "us-east-1d", "us-east-1e"]
        },
        "source": {
            "account": "test",
            "region": "us-east-1",
            "asgName": "spintest-v002"
        },
        "suspendedProcesses": [],
        "tags": {},
        "useAmiBlockDeviceMappings": false,
        "subnetType": "internal (vpc0)",
        "instanceType": "t2.micro",
        "iamRole": "........",
        "keyPair": "........",
        "instanceMonitoring": false,
        "ebsOptimized": false,
        "base64UserData": null,
        "securityGroups": [........],
        "amiName": "........I",
        "virtualizationType": "hvm",
        "termination": {
            "order": "oldest",
            "relaunchAllInstances": true,
            "concurrentRelaunches": 1,
            "totalRelaunches": 1
        },
        "type": "cloneServerGroup",
        "cloudProvider": "aws",
        "account": "test",
        "user": "ajordens@netflix.com"
    }],
    "application": "spintest",
    "description": "Create Cloned Server Group from spintest-v002"
}
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+
+    try {
+      StageData stageData = (StageData) stage.mapTo(StageData.class);
+      Response response = oortService.getServerGroup(stageData.getApplication(), stageData.getAccount(), stageData.getCluster(), asgName, region, cloudProvider);
+      Map serverGroup = objectMapper.readValue(response.getBody().in(), Map.class);
+      Map titusServerGroupLabels = (Map) serverGroup.get("labels");
+
+      if (titusServerGroupLabels != null && titusServerGroupLabels.containsKey(INTERESTING_HEALTH_PROVIDER_NAMES)) {
+        String healthProviderNames = (String) titusServerGroupLabels.get(INTERESTING_HEALTH_PROVIDER_NAMES);
+        stage.getContext().put(INTERESTING_HEALTH_PROVIDER_NAMES, healthProviderNames);
```
ajordens: Should this method be modifying the stage context as a side effect?

Seems like the health provider names could be returned and dealt with by the caller?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTaskSpec.groovy
```diff
@@ -65,6 +76,38 @@ class DetermineHealthProvidersTaskSpec extends Specification {
...
+    taskResult.status == ExecutionStatus.SUCCEEDED
+    (taskResult.getStageOutputs() as Map) == expectedStageOutputs
+
+    where:
+    stageType             | stageContext                                                           | labels                                                  | invocationCount || expectedStageOutputs
+    "createServerGroup"   | [application: "app", strategy: "rollingpush", cloudProvider: "titus"]  | [interestingHealthProviderNames: "Titus"]               | 1               || [interestingHealthProviderNames: ["Titus"]]
+    "createServerGroup"   | [application: "app", strategy: "rollingpush", cloudProvider: "aws"]    | [interestingHealthProviderNames: "Titus"]               | 0               || [:]
+    "createServerGroup"   | [application: "app", strategy: "rollingpush", cloudProvider: "titus"]  | [interestingHealthProviderNames: "Titus,Discovery"]     | 1               || [interestingHealthProviderNames: ["Titus", "Discovery"]]
+    "createServerGroup"   | [application: "app", strategy: "rollingpush", cloudProvider: "titus"]  | null                                                    | 1               || [interestingHealthProviderNames: []]
```
ajordens: Hmm.

If it doesn't have a label, I reckon we don't want `interestingHealthProviderNames: []` as I suspect that will short-circuit all health provider checks.

See `HealthHelper.filterHealths()`.

I believe `interestingHealthProviderNames: null` would be more reasonable.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+  private static final String ROLLING_PUSH = "rollingpush";
+  private static final List<String> SUPPORTED_STAGES = Arrays.asList("deploy", "enableServerGroup", "createServerGroup", "cloneServerGroup", "disableServerGroup");
+
+  @Override
+  public boolean supports(String cloudProvider, Stage stage) {
+    return cloudProvider.equalsIgnoreCase(TITUS) && SUPPORTED_STAGES.contains(stage.getType()) && stage.getContext().containsKey("strategy") && stage.getContext().get("strategy").equals(ROLLING_PUSH);
+  }
+
+  @Override
+  public List<String> process(String cloudProvider, OortService oortService, ObjectMapper objectMapper, Stage stage) {
```
ajordens: IMO `oortService` and `objectMapper` should be `@Autowire` into the constructor of this class rather than pollute a public constructor.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTaskSpec.groovy
```diff
@@ -65,6 +76,38 @@ class DetermineHealthProvidersTaskSpec extends Specification {
     [application: "app"]                          | null                   || [:]                                           // no health provider names when an exception is raised
   }
 
+  @Unroll
+  def "should check interestingHealthProviderNames based server group 'interestingHealthProviderNames' label for titus "() {
```
ajordens: Might be easier to follow if this test were extracted into a dedicated test spec for `TitusInterestingHealthProviderNamesSupplier`?

As for `DetermineHealthProvidersTask` ... we really just need a test that asserts that we look for a `Supplier` and if found use whatever it returns. 

We don't care if it's Kubernetes/Amazon/Google, etc.

Thoughts?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTaskSpec.groovy
jeyrschabu: In this case, it seems more readable to only set it if that collection isn't empty.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesSupplierSpec.groovy
```diff
@@ -0,0 +1,82 @@
...
+    cloudProvider | stageType           | stageContext                || supports
+    "titus"       | "createServerGroup" | [strategy: "rollingpush"]   || true
+    "titus"       | "enableServerGroup" | [strategy: "rollingpush"]   || true
+    "titus"       | "cloneServerGroup"  | [strategy: "rollingpush"]   || true
+    "aws"         | "createServerGroup" | [strategy: "rollingpush"]   || false
+    "titus"       | "createServerGroup" | [strategy: "custom"]        || false
+  }
+
+  @Unroll
+  def "should process interestingHealthNames by inspecting lables on titus serverGroup"() {
```
ajordens: labels

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
```diff
@@ -83,6 +89,21 @@ public TaskResult execute(Stage stage) {
...
       Application application = front50Service.get(applicationName);
+      Optional<InterestingHealthProviderNamesSupplier> healthProviderNamesSupplierOptional = interestingHealthProviderNamesSuppliers
+        .stream()
+        .filter(supplier -> supplier.supports(getCloudProvider(stage), stage))
+        .findFirst();
+
+      if (healthProviderNamesSupplierOptional.isPresent()) {
+        List<String> interestingHealthProviderNames = healthProviderNamesSupplierOptional.get().process(getCloudProvider(stage), stage);
+
+        if (!interestingHealthProviderNames.isEmpty()) {
```
ajordens: As discussed ...

```
Map<String, List<String>> results = new HashMap<>();
if (interestingHealthProviderNames != null) {
  // avoid a `null` value that may cause problems with ImmutableMap usage downstream
  results.put("interestingHealthProviderNames", interestingHealthProviderNames);
}
return new DefaultTaskResult(ExecutionStatus.SUCCEEDED, results);
```

PR #1132, opened 2017-01-18
Support for loading more than one level of template source

ajordens:
- raises an exception if a cycle is detected (A -> B -> A)


ajordens:
Nothing too tricky here.

PTAL @robzienert or @cfieber

robzienert:
Looks good.

PR #1133, opened 2017-01-18
poller to clean up pipelines that died if an orca instance goes out

robfletcher:
See #1353

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineLauncher.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineLauncher.java
@@ -61,7 +61,6 @@ public class PipelineLauncher extends ExecutionLauncher<Pipeline> {
       .withExecutingInstance(currentInstanceId)
       .withExecutionEngine(Execution.V2_EXECUTION_ENGINE)
       .withNotifications((List<Map<String, Object>>) config.get("notifications"))
-      .withId()
       .build();
   }
 
--- a/orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
+++ b/orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
@@ -1,0 +1,208 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.zombie;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.ZoneId;
+import com.netflix.spinnaker.orca.RetryableTask;
+import com.netflix.spinnaker.orca.pipeline.model.Pipeline;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import com.netflix.spinnaker.orca.pipeline.model.Task;
+import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.context.ApplicationContext;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+import rx.Observable;
+import static com.netflix.spinnaker.orca.ExecutionStatus.CANCELED;
+import static com.netflix.spinnaker.orca.ExecutionStatus.RUNNING;
+import static java.lang.String.format;
+import static java.time.Clock.systemDefaultZone;
+import static java.time.Duration.ZERO;
+import static java.time.Duration.between;
+import static java.time.format.DateTimeFormatter.ISO_LOCAL_DATE_TIME;
+import static rx.Observable.just;
+
+/**
+ * This poller cancels any pipelines that appear to be still running but are
+ * more than {@link #TOLERANCE} overdue.
+ */
+@Component
+@ConditionalOnExpression("${pollers.zombiePipeline.enabled:true}")
+@Slf4j
+public class ZombiePipelineCleanupAgent {
+
+  /**
+   * How often the agent runs.
+   */
+  private static final long FREQUENCY_MS = 21600000; // 6 hours
+
+  /**
+   * Delay before running the agent the first time.
+   */
+  private static final long INITIAL_DELAY_MS = 300000; // 5 minutes
+
+  private final ExecutionRepository repository;
+  private final Clock clock;
+  private final ApplicationContext applicationContext;
+  private final Lock lock;
+  private final String currentInstanceId;
+
+  /**
+   * Duration a task can be overdue before we decide to kill it.
+   */
+  static final Duration TOLERANCE = Duration.ofMinutes(30);
+
+  /**
+   * Duration a non-retryable task can run before being considered overdue.
+   * {@link #TOLERANCE} is added on.
+   */
+  static final Duration DEFAULT_TASK_TIMEOUT = Duration.ofMinutes(2);
+
+  @Autowired
+  public ZombiePipelineCleanupAgent(ExecutionRepository repository,
+                                    ApplicationContext applicationContext,
+                                    Lock lock,
+                                    String currentInstanceId) {
+    this(
+      repository,
+      applicationContext,
+      lock,
+      currentInstanceId,
+      systemDefaultZone()
+    );
+  }
+
+  ZombiePipelineCleanupAgent(ExecutionRepository repository,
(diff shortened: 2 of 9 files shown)


orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
```diff
@@ -0,0 +1,137 @@
...
+    this.clock = clock;
+    this.applicationContext = applicationContext;
+  }
+
+  @Override public void run() {
+    timer(10, MINUTES)
+      .subscribe((tick) -> slayZombies());
+  }
+
+  void slayZombies() {
```
robzienert: 💀

orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
```diff
@@ -0,0 +1,137 @@
...
+
+  private Duration pausedTime(Pipeline pipeline) {
+    if (pipeline.getPaused() == null) {
+      return ZERO;
+    }
+    return Duration.ofMillis(pipeline.getPaused().getPausedMs());
+  }
+
+  private RetryableTask taskBean(Task task) {
+    return (RetryableTask) applicationContext.getBean(task.getImplementingClass());
```
ajordens: What if it's not a `RetryableTask`?

orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
```diff
@@ -0,0 +1,137 @@
...
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static rx.Observable.just;
+import static rx.Observable.timer;
+
+/**
+ * This poller cancels any pipelines that appear to be still running but are
+ * more than {@link #TOLERANCE} overdue.
+ */
+@Component
+@ConditionalOnExpression("${pollers.zombiePipeline.enabled:false}")
```
ajordens: Would probably just default this to `true`

robfletcher:
@ajordens Updated to handle non-retryable tasks and enable poller by default

ajordens:
LGTM

orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
```diff
@@ -0,0 +1,145 @@
...
+  public ZombiePipelineCleanupAgent(ExecutionRepository repository,
+                                    Clock clock,
+                                    ApplicationContext applicationContext) {
+    this.repository = repository;
+    this.clock = clock;
+    this.applicationContext = applicationContext;
+  }
+
+  @Override public void run() {
+    timer(10, MINUTES)
```
duftler: Is it by design to only have this run once, or am I reading it wrong?

orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
```diff
@@ -0,0 +1,145 @@
...
+import static java.time.Duration.between;
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static rx.Observable.just;
+import static rx.Observable.timer;
+
+/**
+ * This poller cancels any pipelines that appear to be still running but are
+ * more than {@link #TOLERANCE} overdue.
+ */
+@Component
```
duftler: Where does this actually get scanned by spring? I don't see this package or class referenced via `Main` or somewhere similar.

orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
```diff
@@ -0,0 +1,145 @@
...
+  @Autowired
+  public ZombiePipelineCleanupAgent(ExecutionRepository repository,
+                                    Clock clock,
+                                    ApplicationContext applicationContext) {
+    this.repository = repository;
+    this.clock = clock;
+    this.applicationContext = applicationContext;
+  }
+
+  @Override public void run() {
```
duftler: Does this `runnable` somehow get automatically kicked off by the spring lifecycle stuff, or does this need to be scheduled to call `run()`?

ajordens: Fair question, this isn't normally how we set up these agents (`OrchestrationCleanupPollingNotificationAgent` or `PipelineCleanupPollingNotificationAgent` are alternative examples).

Any idea how this is actually going to run on a large database of pipelines? It wouldn't surprise me if it took much longer than 10 minutes.

duftler: Since the main thing prompting the public spinnaker/spinnaker issues related to this is the inability to cancel zombie pipelines via the ui, maybe the new logic in this pr can be exploited to allow for explicit cancelation of zombie pipelines via the api/ui? If that were supported, this interval could then be much longer I think.

Might be nice to also indicate in the ui when a pipeline is a zombie.

robfletcher:
@duftler @ajordens Poller now uses Spring's `@Scheduled` annotation to run repeatedly and a simple Redis based lock to ensure only one node is killing zombies at a time.

orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
```diff
@@ -0,0 +1,201 @@
...
+                             String currentInstanceId,
+                             Clock clock) {
+    this.repository = repository;
+    this.clock = clock;
+    this.applicationContext = applicationContext;
+    this.currentInstanceId = currentInstanceId;
+    this.lock = lock;
+  }
+
+  @Scheduled(fixedDelay = FREQUENCY_MS, initialDelay = INITIAL_DELAY_MS)
```
duftler: Don't we need something like `@EnableScheduling` on `ZombiePipelineCleanupConfig`, or is this scheduled some other way?

orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupConfig.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import redis.clients.jedis.JedisCommands;
+
+@Configuration
+@ComponentScan(basePackageClasses = ZombiePipelineCleanupConfig.class)
+public class ZombiePipelineCleanupConfig {
+
+  @Bean
+  Lock zombiePollerLock(JedisCommands redis) {
```
duftler: When running this with no additional config, I get this spring exception:
```
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type [redis.clients.jedis.JedisCommands] found for dependency: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {}
```

orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
```diff
@@ -0,0 +1,201 @@
...
+      format("%s.%d", currentInstanceId, clock.millis()),
+      this::findZombies,
+      this::slayZombie
+    );
+  }
+
+  private Observable<Pipeline> findZombies() {
+    log.info("Starting sweep for zombie pipelines...");
+    return repository
+      .retrievePipelines()
```
duftler: Best as I can tell, this `retrievePipelines()` call always returns (for me at least) _just_ the v1 pipeline executions, so no executions survive the v2 filter.

I _think_ the reason that only v1 executions are returned is that they are the only ones I see added to `allJobs:pipeline`. The reason for that seems to be that v2 executions already have `execution.id` set when this line is reached: https://github.com/spinnaker/orca/blob/master/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy#L453

robfletcher: oy, that's bad.

orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupConfig.java
robfletcher: How are you running it? If I execute the `Main` class from orca-web or run `gradle bootRun` it's working fine.

orca-spring-batch/src/main/java/com/netflix/spinnaker/zombie/ZombiePipelineCleanupAgent.java
robfletcher: Damn, yeah. I added that at the last minute — was using rx.Observable.timer while I tested it all out.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/PipelineLauncher.java
```diff
@@ -61,7 +61,6 @@ public PipelineLauncher(ObjectMapper objectMapper,
       .withExecutingInstance(currentInstanceId)
       .withExecutionEngine(Execution.V2_EXECUTION_ENGINE)
       .withNotifications((List<Map<String, Object>>) config.get("notifications"))
-      .withId()
```
robfletcher: @duftler this was the line that was breaking the indexing in redis. Didn't even appear to be necessary for anything.

PR #1136, opened 2017-01-19
super rough first pass at layering configs for deployment model

cfieber:
just looking for some feedback on the approach, the code is an inlined pile of garbage currently.

generally:
* a DeploymentModel represents a single yaml document, containing clusters, loadBalancers, accounts, environments (potentially more in the future) all keyed on name/id pointing to an object
* object can extend another object with `extends: name`
* a resolver takes an ordered list of documents (priority order). when looking up an object by name (for retrieval or extension) the first object found by name wins.
* resolving an object by name follows the `extends` and keeps layering on the configuration via MapMerge
* resolving to a typed object (`Cluster` from `Map`) resolves the object to its fully merged format, as well resolving its relationships to other configurable types. (e.g Cluster has an Environment has an Account, so cluster Map is formed, and cluster.get("environment") is used to resolve Environment which causes environment Map to merge and environment.get("account") to be resolved.

end state it is intended that a pipeline template should be able to take a variable of type `cluster` (or `environment` or `account`) and the resolver will supply a resolved strongly typed object for that variable. I also intend to build a transformer that knows how to take the current model (Cluster to start) which (while incomplete) is intended not to be cloud provider specific, and produce the proper payload for a cloud provider deploy.

feels like there is some overlap with the existing pipeline template composition bit (and I'm I think a bit out of date with that). I was going to circle back to that at some point.

I think this does a good job of addressing the inheritance selector issues with the one exception of lists not being treated well (they are just replaced as opposed to merged - both of those choices have bad tradeoffs).

Probably the most interesting thing to look at would be `gate.yml` which produces the following once loaded, and the gateMain cluster is mapped to the strongly typed cluster object, then serialized back out (DeploymentModelSpec does that):
````json
{
  "name" : "gateMain",
  "application" : "gate",
  "environment" : {
    "account" : {
      "name" : "mgmt",
      "cloudProvider" : "aws",
      "regions" : {
        "us-east-1" : {
          "name" : "us-east-1",
          "zones" : [ "us-east-1c", "us-east-1d", "us-east-1e" ]
        },
        "us-west-2" : {
          "name" : "us-west-2",
          "zones" : [ "us-west-2a", "us-west-2b", "us-west-2c" ]
        }
      }
    },
    "stack" : "main",
    "detail" : null,
    "regions" : {
      "us-west-2" : {
        "name" : null,
        "zones" : null
      }
    }
  },
  "requires" : {
    "resources" : {
      "cpu" : "2",
      "ram" : "16gb",
      "disk" : "20gb",
      "network" : "10mbit"
    },
    "membership" : [ "spinnaker-internal-service" ],
    "ingress" : null
  },
  "loadBalancers" : [ {
    "name" : "gate-main-internal",
    "type" : "internal",
    "ingress" : null
  }, {
    "name" : "gate-main-external",
    "type" : "external",
    "ingress" : null
  } ]
}
````

### DIFF
--- a/orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/DeploymentModelSpec.groovy
+++ b/orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/DeploymentModelSpec.groovy
@@ -1,0 +1,227 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model
+
+import com.fasterxml.jackson.annotation.JsonBackReference
+import com.fasterxml.jackson.annotation.JsonManagedReference
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.SerializationFeature
+import groovy.transform.Canonical
+import org.yaml.snakeyaml.Yaml
+import spock.lang.Specification
+
+/**
+ * DeploymentModelSpec.
+ */
+class DeploymentModelSpec extends Specification {
+/* jackson foo:
+  @Canonical
+  static class Child {
+    @JsonBackReference
+    Parent p
+    int age
+
+    String getName() {
+      return p.children.entrySet().findResult { e -> if (e.value.is(this)) return e.key else return null }
+    }
+  }
+
+  @Canonical
+  static class Parent {
+    @JsonManagedReference
+    Map<String, Child> children = [:]
+  }
+
+  def "it should serialize"() {
+    given:
+    def p = new Parent()
+    p.children.jimmy = new Child(p: p, age: 3)
+    p.children.billy = new Child(p: p, age: 5)
+
+    when:
+    def result = new ObjectMapper().writeValueAsString(p)
+
+    then:
+    result != null
+    println result
+  }
+
+  def "it should deserialize"() {
+    given:
+    def json = '{"children":{"jimmy":{"age":3},"billy":{"age":5}}}'
+
+    when:
+    Parent p = new ObjectMapper().readValue(json, Parent)
+
+    then:
+    p.children.size() == 2
+    p.children.jimmy.name == 'jimmy'
+  }
+*/
+
+  def "should layer content"() {
+    given:
+    def models = [
+        new DeploymentModel(clusters: [leaf: enhm([extends: 'extendo', someOtherAttribute: 'mustard', someExtendAttribute: 'bologna'])]), //highest precidence
+        new DeploymentModel(clusters: [extendo: enhm([extends: 'base', someExtendAttribute: 'ham'])]),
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/DeploymentModelSpec.groovy [hunk 1]
```diff
@@ -0,0 +1,227 @@
...
+import com.fasterxml.jackson.databind.SerializationFeature
+import groovy.transform.Canonical
+import org.yaml.snakeyaml.Yaml
+import spock.lang.Specification
+
+/**
+ * DeploymentModelSpec.
+ */
+class DeploymentModelSpec extends Specification {
+/* jackson foo:
```
cfieber: ah, ignore this jackon bit, was looking at options for jackson mapping to address the case of modelling things as map<`id`, `the thing`> and `the thing` being able to know its id

orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/DeploymentModelSpec.groovy
```diff
@@ -0,0 +1,227 @@
...
+
+    then:
+    thrown(NoSuchElementException)
+  }
+
+  def "should read yaml filearooni"() {
+
+    given:
+    def mapper = new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT)
+    def models = new Yaml().loadAll(getClass().getResourceAsStream("/gate.yml")).collect {
```
cfieber: there is an open issue in that jackson yaml doesn't support a multi document file

orca-pipelinetemplate/src/test/resources/gate.yml
```diff
@@ -0,0 +1,150 @@
+#document netflix-conventions
+clusters:
+  netflix-conventions:
+    requires:
+      membership:
```
cfieber: this isn't really effective because list gets clobbered by an override instead of merged.  could restructure to Map<String, Object>, but there isn't really anything I want to put in the actual object payload

orca-pipelinetemplate/src/test/resources/gate.yml
```diff
@@ -0,0 +1,150 @@
...
+    cloudProvider: aws
+    regions:
+      us-east-1:
+        name: us-east-1 #baerf
+        zones:
+          - us-east-1c
+          - us-east-1d
+          - us-east-1e
+      us-west-2:
+        name: us-west-2  #backrefs?
```
cfieber: name vs the containing key thing kind of sucks - the json backref or the stuff @robzienert was looking at (using a Map type that passes id down to the child) or something else could work here. alternatively just sorting it out in the translation from Map -> typed object could work too

orca-pipelinetemplate/src/test/resources/gate.yml
```diff
@@ -0,0 +1,150 @@
...
+
+
+---
+
+#document spinnaker-conventions
+
+environments:
+  spinBase:
+    regions:
+      us-west-2: {}
```
cfieber: an example of maps vs lists (in this case region actually potentially has useful values which warrant it being an object)

orca-pipelinetemplate/src/test/resources/gate.yml
```diff
@@ -0,0 +1,150 @@
...
+
+  main:
+    extends: spinProd
+    stack: main
+
+  failover:
+    extends: main
+    regions:
+      us-west-2: null
+      us-east-1: {}
```
cfieber: this addresses one aspect of controlling the inheritence bit without selectors

orca-pipelinetemplate/src/test/resources/gate.yml
robzienert: One concern I have with the jackson route is that we're making changes to the graph (adding, removing) based on ID in stages & modules after deserialization. If we go the jackson route, won't we need to jump through some hoops to get ids assigned correctly to child objects?

orca-pipelinetemplate/src/test/resources/gate.yml
robzienert: I kind of prefer the list being clobbered, personally. Otherwise, how would you remove an element from the list?

PR #1138, opened 2017-01-23
Adding inheritance control support for template stages

robzienert:
Allows for finer-grained control over what gets inherited
or overridden by child templates or configurations in nested
data structures.

This is being run after the render stage, so inheritance controls
cannot use handlebar templates. We can change this later on, but
I wanted to limit the amount of side-effects as part of introducing
this level of modification.

Implements this guy: https://github.com/robzienert/dcd-spec#inheritance-control


cfieber:
LGTM

PR #1141, opened 2017-01-24
Preliminary support for validating stages against json schemas

ajordens:
Ideally we have a single json schema document per stage type describing
relevant properties with some additional metadata linking the property
to 0..* cloud providers.

`orca` will process these schema documents and generate a valid json
schema containing only the properties supported by the current cloud
provider.


ajordens:
PTAL @cfieber / @robzienert 

It's early days but FYI @spinnaker/google-reviewers

robzienert:
Cool, this is nice.

PR #1142, opened 2017-01-25
First pass at pipeline executions

robzienert:
Pretty naive implementation at this point, but it does work.

PTAL @cfieber @ajordens

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessor.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessor.java
@@ -19,8 +19,10 @@ package com.netflix.spinnaker.orca.pipelinetemplate;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.orca.extensionpoint.pipeline.PipelinePreprocessor;
+import com.netflix.spinnaker.orca.pipelinetemplate.generator.ExecutionGenerator;
 import com.netflix.spinnaker.orca.pipelinetemplate.loader.TemplateLoader;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.TemplateMerge;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.V1SchemaExecutionGenerator;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.graph.GraphMutator;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.PipelineTemplate;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.TemplateConfiguration;
@@ -66,8 +68,6 @@ public class PipelineTemplatePipelinePreprocessor implements PipelinePreprocesso
 
     TemplateConfiguration templateConfiguration = request.getConfig();
 
-    // TODO find all templates via request.config.pipeline.template.source
-    // list needs to be FIFO, where the first template is the root
     List<PipelineTemplate> templates = templateLoader.load(templateConfiguration.getPipeline().getTemplate());
 
     PipelineTemplate template = TemplateMerge.merge(templates);
@@ -75,9 +75,11 @@ public class PipelineTemplatePipelinePreprocessor implements PipelinePreprocesso
     GraphMutator graphMutator = new GraphMutator(templateConfiguration, renderer, registry);
     graphMutator.mutate(template);
 
-    // TODO final validation & marshal to pipeline json
+    // TODO validation
 
-    return pipeline;
+    ExecutionGenerator executionGenerator = new V1SchemaExecutionGenerator();
+
+    return executionGenerator.generate(template, templateConfiguration);
   }
 
   private static class TemplatedPipelineRequest {
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGenerator.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGenerator.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.pipelinetemplate.v1schema;
+
+import com.netflix.spinnaker.orca.pipelinetemplate.generator.ExecutionGenerator;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.PipelineTemplate;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.PipelineTemplate.Configuration;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.StageDefinition;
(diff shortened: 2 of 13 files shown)


orca-pipelinetemplate/src/test/resources/templates/simple-001.yml
```diff
@@ -1,30 +1,36 @@
...
   name: Bake
   config:
-    regions: "{{ var.regions }}"
-    package: "{{ var.application }}-package"
+    regions: |
+      [
+      {{#each regions}}
+      "{{ this }}"{{#unless @last}},{{/unless}}
+      {{/each}}
+      ]
```
robzienert: One thing I hadn't really thought about until writing the integration spec is that handlebar templates need to emit  JSON, _not_ YAML. I'm not sure if we want to support YAML directly, or if telling people to emit JSON is fine. Might cause some room for confusion, though.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGenerator.java
```diff
@@ -0,0 +1,85 @@
...
+        stage.put("context", s.getConfig());
+        stage.put("requisiteStageRefIds", getStageRequisiteIds(s, template.getStages()));
+        return stage;
+      })
+      .collect(Collectors.toList()));
+
+    return pipeline;
+  }
+
+  private static List<Integer> getStageRequisiteIds(StageDefinition stage, List<StageDefinition> allStages) {
```
robzienert: I've moved this into the graph transforms in a PR I'll be opening after this is merged.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
```diff
@@ -74,6 +77,24 @@ private void replaceStages(PipelineTemplate pipelineTemplate) {
...
+
+    if (!configStages.isEmpty() || !templateStages.isEmpty()) {
+      throw new UnsupportedOperationException("stage injection is not yet supported");
+    }
+  }
+
+  private void assignRefIds(PipelineTemplate pipelineTemplate) {
+    AtomicInteger refId = new AtomicInteger(1);
+    pipelineTemplate.getStages().forEach(s -> {
+      s.setRefId(refId.getAndIncrement());
```
cfieber: refId is actually a string in the JSON and just happens to be done as an incrementing integer in the UI. If there is already some other unique identifier that we have for each stage could probably just use that (and the resulting JSON maybe has some more useful meaning to it when looking at it). If we don't have an already unique identifier at this point I'm totally okay with just using ints

cfieber:
LGTM

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessor.java
```diff
@@ -66,18 +68,18 @@ public PipelineTemplatePipelinePreprocessor(ObjectMapper pipelineTemplateObjectM
...
     PipelineTemplate template = TemplateMerge.merge(templates);
 
     GraphMutator graphMutator = new GraphMutator(templateConfiguration, renderer, registry);
     graphMutator.mutate(template);
 
-    // TODO final validation & marshal to pipeline json
+    // TODO validation
 
-    return pipeline;
+    ExecutionGenerator executionGenerator = new V1SchemaExecutionGenerator();
```
robfletcher: Does the v1 here refer to orca's execution engine? Why do we want to run this stuff in v1?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGenerator.java
```diff
@@ -0,0 +1,85 @@
...
+public class V1SchemaExecutionGenerator implements ExecutionGenerator {
+
+  @Override
+  public Map<String, Object> generate(PipelineTemplate template, TemplateConfiguration configuration) {
+    Map<String, Object> pipeline = new HashMap<>();
+    pipeline.put("id", configuration.getRuntimeId());
+    pipeline.put("application", configuration.getPipeline().getApplication());
+    pipeline.put("name", Optional.ofNullable(configuration.getPipeline().getName()).orElse("Unnamed Execution"));
+    // TODO Doesn't look like we even use this anywhere in the codebase.
+//    pipeline.put("appConfig", "unsupported");
```
robfletcher: I _think_ this is for preferences set at an application level — notifications and the like

orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessorSpec.groovy
```diff
@@ -0,0 +1,134 @@
...
+class PipelineTemplatePipelinePreprocessorSpec extends Specification {
+
+  ObjectMapper objectMapper = new ObjectMapper()
+
+  TemplateLoader templateLoader = new TemplateLoader([new FileTemplateSchemeLoader(objectMapper)])
+
+  Renderer renderer = new HandlebarsRenderer(objectMapper)
+
+  Registry registry = Mock() {
+    clock() >> Mock(Clock) {
```
robfletcher: What's this used for? I don't see anything referencing it elsewhere.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
robzienert: Oh fancy. If I change this from an integer value, do you think we'd wind up having render issues on the UI?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessor.java
robzienert: It references the pipeline template schema version, not the execution engine.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGenerator.java
robzienert: I was looking around in the codebase and didn't see it actually used anywhere of significance.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
robfletcher: Would be easy to try with an API call

orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessorSpec.groovy
robzienert: There's a metric in the `PipelineTemplatePipelinePreprocessor` that records the elapsed time for processing templates into executions. This spec came sometime after that class was written.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
robzienert: Looks to work in the UI just fine! I'll update stuff. Thanks for the suggestion.

robzienert:
@robfletcher Not the purpose of this PR. `PipelinePreprocessor` objects are [run before executions occur](https://github.com/spinnaker/orca/blob/master/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy#L67), and just are expected to result in pipeline execution object maps. This PR is just transforming the pipeline template into an execution.

If you like, I can give you a quick pass through the workflow.

PR #1143, opened 2017-01-25
A set of exemplar json schema for each stage type

ajordens:
- No properties are marked required
- No nested definitions for properties of type 'array' or 'object'
- Cloud provider mappings are incomplete or missing


robzienert:
What's `builtIn` mean?

ajordens:
I was treating `builtIn` as a bit of metadata indicating that it's a field that **all** stages have.

robzienert:
I see, thanks. I didn't scope the individual schemas with too much detail; rest looks good.

:+1:

PR #1146, opened 2017-01-27
Add NUPKG as a supported PackageType (Windows support)

robumberger:
PR's will be opened against Rosco and Spinnaker repos that require this change

### DIFF
--- a/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
+++ b/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
@@ -161,14 +161,7 @@ class PackageInfoSpec extends Specification {
   def "Find the right package, don't be dependant on artifact order"() {
     given:
     Stage bakeStage = new PipelineStage()
-    PackageType packageType = PackageType.RPM
     boolean extractBuildDetails = false
-    PackageInfo packageInfo = new PackageInfo(bakeStage,
-      packageType.packageType,
-      packageType.versionDelimiter,
-      extractBuildDetails,
-      false,
-      mapper)
     def allowMissingPackageInstallation = true
 
     Map trigger = ["buildInfo": ["artifacts": filename]]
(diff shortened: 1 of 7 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -225,6 +227,13 @@ class PackageInfo {
     }
   }
 
+  @CompileDynamic
+  private Map filterNUPKGArtifacts(List<Map> artifacts, String prefix) {
```
duftler: Looks like this method does the same thing as the `else` branch at the call-site. Is it necessary to add this method?

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -192,6 +192,41 @@ class PackageInfoSpec extends Specification {
 
   }
 
+  @Unroll
+  def "Find the right package (NUPKG), don't be dependant on artifact order"() {
```
duftler: Since this new test case appears to be the same as the one just above, with the exception of `packageType` and the fileName extensions, we could just add a `packageType` column to the `where:` table and add the additional cases to the table. Please add some `.deb` cases as well since your hands are already in the code.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfoSpec.groovy
```diff
@@ -161,34 +161,47 @@ class PackageInfoSpec extends Specification {
...
+     ["fileName": "package-something-4.11.4h-1.x86_64.rpm"]]         || "package"            || PackageType.RPM   || "package-4.11.4h-1.x86_64"
     [["fileName": "package-something-4.11.4h-1.x86_64.rpm"],
-     ["fileName": "package-4.11.4h-1.x86_64.rpm"]]           || "package"        || "package-4.11.4h-1.x86_64"
+     ["fileName": "package-4.11.4h-1.x86_64.rpm"]]                   || "package"            || PackageType.RPM   || "package-4.11.4h-1.x86_64"
+    [["fileName": "package_4.11.4-h02.sha123_amd64.deb"]]            || "package"            || PackageType.DEB   || "package_4.11.4-h02.sha123_amd64"
+    [["fileName": "package-something_4.11.4-h02.sha123_amd64.deb"]]  || "package"            || PackageType.DEB   || "package"
+    [["fileName": "package_4.11.4-h02.deb"],
+     ["fileName": "package-something_4.11.4-h02.deb"]]               || "package"            || PackageType.DEB   || "package_4.11.4-h02"
+    [["fileName": "package_4.11.4-h02.sha123.deb"],
+      ["fileName": "package-something_4.11.4-h02.deb"]]               || "package-something" || PackageType.DEB   || "package-something_4.11.4-h02"
```
duftler: Indentation is off on this line.

rguthriemsft:
@scotmoor @EricJizbaMSFT LGTM.

PR #1148, opened 2017-01-30
make sure determine target server group stage is injected in the right place for v2

robfletcher:


robzienert:
So this is basically making the behavior non-silly? 👍

robfletcher:
Yeah, the original behavior was to run "before" stages in reverse order, because… reasons. Nü Örca doesn't do that.

PR #1150, opened 2017-01-31
Stage DAG injection

robzienert:
Allows templates and configuration to inject stages at the beginning / end of pipelines, as well as before and after other stages.

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
@@ -18,10 +18,12 @@ package com.netflix.spinnaker.orca.pipelinetemplate.v1schema.graph.transform;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.PipelineTemplateVisitor;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.PipelineTemplate;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.StageDefinition;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.StageDefinition.InjectionRule;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.TemplateConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -32,11 +34,6 @@ import java.util.stream.Collectors;
  * will be added. This can be useful in situations where users want to publish
  * pipeline traits, then wire them together through configuration-level stages.
  * This offers more flexibility in addition to inheritance relatively for free.
- *
- * When template stages do exist, a configuration-defined stage must either
- * be replacing a stage by the same name, or have an inject stanza. Stages
- * without these semantics will hard-fail, to safe-guard against underlying
- * templates changing in unexpected ways.
  */
 public class ConfigStageInjectionTransform implements PipelineTemplateVisitor {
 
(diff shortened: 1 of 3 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
```diff
@@ -75,17 +72,107 @@ private void replaceStages(PipelineTemplate pipelineTemplate) {
...
+    // Create initial graph via dependsOn.
+    createDag(pipelineTemplate.getStages());
+
+    // Handle stage injections.
+    injectStages(pipelineTemplate.getStages(), pipelineTemplate.getStages());
+    injectStages(templateConfiguration.getStages(), pipelineTemplate.getStages());
+  }
+
+  private static void createDag(List<StageDefinition> stages) {
+    stages
```
jeyrschabu: Do we have a sensible way to detect cycles ? Or is the assumption that the user will never introduce a cycle.
I would be inclined to do a first pass through the iteration to make sure we aren't going to get hit with a stack overflow. Thoughts ?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
```diff
@@ -75,17 +72,107 @@ private void replaceStages(PipelineTemplate pipelineTemplate) {
...
+      .forEach(s -> s.setRequisiteStageRefIds(
+        stages
+          .stream()
+          .filter(parentCandidate -> s.getDependsOn().contains(parentCandidate.getId()))
+          .map(StageDefinition::getId)
+          .collect(Collectors.toList())
+      ));
+  }
+
+  private static void injectStages(List<StageDefinition> stages, List<StageDefinition> templateStages) {
```
jeyrschabu: Would it make sense to move these static methods to a helper class ?
Is there a reason for them to be static ?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
```diff
@@ -75,17 +72,107 @@ private void replaceStages(PipelineTemplate pipelineTemplate) {
...
-    }
+  private static void injectLast(StageDefinition stage, List<StageDefinition> allStages) {
+    int numStages = allStages.size();
+    allStages
+      .stream()
+      .filter(s -> !s.getId().equals(stage.getId()))
+      .filter(candidate -> allStages
+        .stream()
+        .filter(s -> !s.getRequisiteStageRefIds().contains(candidate.getId()))
+        .count() == numStages)
```
jeyrschabu: Not sure why we need the count here. Is this a way to ensure we have the right node ?

orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransformSpec.groovy
```diff
@@ -65,26 +66,105 @@ class ConfigStageInjectionTransformSpec extends Specification {
...
+    where:
+    subject || requisiteStages
+    's1'    || []
+    's2'    || ['s1']
+    's3'    || ['s1']
+    's4'    || ['s2']
+    's5'    || ['s3']
+  }
+
+  def 'should inject stage into dag'() {
```
jeyrschabu: nice!

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
robzienert: The transforms are only intended to get the graph to the configured state, not really validate that it's acceptable. There's a validation step that we haven't added in yet that will perform validation that this would be included in: https://github.com/spinnaker/orca/blob/master/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessor.java#L78

Things like the `orca-validation` module (which does JSON schema validation of required fields, etc) would probably happen here, as well as doing this cyclic detection. I'm quite bad at graph stuff, so if you wanna take on the cycle detection, 💯 💯 💯 . :)

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
robzienert: They're just in-class helper methods. I tend to like to break methods up where I can so it's easier to follow each piece. Since they're only useful as far as the internals of this class, I don't think it makes a lot of sense to have a helper class.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
robzienert: Because I didn't build up an vertex list, so there's no real elegant way to discover relations. As discussed in person, it works - so we could merge it now and make it elegant in a second pass, or close this. Let me know how you'd like to proceed there.

PR #1153, opened 2017-02-01
Fast Property Rollback on TERMINAL, CANCEL

zanthrash:
Actually adding code to make the FastPropertyCleanupListener rollback a Fast Properties to a last know good state when a
 Execution TERMINALs or CANCELs.

 - Creating a new FP will delete it on failure/cancel
 - Updating a FP will upsert the original property on failure/cancel
 - Deleting a FP will also upsert the original property on failure/cancel

@robfletcher PTAL


robfletcher:
LGTM.

I wonder if we really care about the difference between creating and updating a property. But I guess it makes sense from a UI standpoint.

zanthrash:
We care about Create vs. Update only for the rollback.  Create we will delete on fail and Update we will Upsert the original on fail.

robfletcher:
Right but we can detect that just by the presence or absence of a previous value

PR #1160, opened 2017-02-06
(core) allows canary analysis to also work with other cloud providers…

tomaslin:
…. Resolves region based on cluster's region tag as well as AWS's availabilityZones.


cfieber:
feels like theres only one way to find out if this works or not

:shipit:

PR #1165, opened 2017-02-07
Adding support for interestingHealthProviderNames resolution when stage is of type rollingPush

jeyrschabu:
- interestingHealthProviderNames resolved from job label will have priority
- Added rollingPush to the list of supported stages


jeyrschabu:
@tomaslin PTAL

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesSupplier.java
```diff
@@ -35,7 +35,7 @@
   private static final String TITUS = "titus";
   private static final String INTERESTING_HEALTH_PROVIDER_NAMES = "interestingHealthProviderNames";
   private static final String ROLLING_PUSH = "rollingpush";
-  private static final List<String> SUPPORTED_STAGES = Arrays.asList("enableServerGroup", "createServerGroup", "cloneServerGroup");
+  private static final List<String> SUPPORTED_STAGES = Arrays.asList("enableServerGroup", "createServerGroup", "cloneServerGroup", "rollingPush");
```
tomaslin: why do we need create and clone if rollingpush and enable seem to be the only ones involved in migration?

jeyrschabu: Probably don't need the create but definitely a clone

PR #1168, opened 2017-02-08
Remove '-' before level in logback format to be like rest of Spinnaker

ewiseblatt:
@ttomsu 

You changed this here https://github.com/spinnaker/orca/commit/7514e2ad115798aa9b44c97c742ea8876f398928#diff-8af87d1b446779657d406343bfb9c928L20

I assume it was an accident.


ewiseblatt:
ping

PR #1172, opened 2017-02-09
pretty urgent fix for any orchestrations that use determine target se…

robfletcher:
…rver groups — wrong execution engine flag meant stages inserted in wrong order


orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/SpringBatchExecutionRunner.java
```diff
@@ -110,9 +111,9 @@ public SpringBatchExecutionRunner(
 
   @Override
   public <T extends Execution<T>> void start(T execution) throws JobExecutionException {
-    Job job = createJob(execution);
+    execution.setExecutionEngine(V2_EXECUTION_ENGINE); // force this here because we ARE running v2
 
-    execution.setExecutionEngine("v2"); // force this here because we ARE running v2
+    Job job = createJob(execution);
```
robfletcher: executionEngine flag is checked by `DetermineTargetServerGroupLinearStageSupport` so it needs to be set *before* the job is created.

PR #1176, opened 2017-02-11
Injecting trigger infor into pipeline prior to preprocessors

jeyrschabu:
- Parsing pipeline trigger info prior to preprocessors
- Ensuring trigger info is set in pipeline after preprocessors


jeyrschabu:
@cfieber @robzienert PTAL

robzienert:
We just need to get this wired up into [RenderContext](https://github.com/spinnaker/orca/blob/master/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/RenderContext.java) so that the trigger information is available from within the handlebars template rendering process.

jeyrschabu:
Now passing the trigger down to the render context

PR #1177, opened 2017-02-11
(rrb) Adjust success conditions for percent enable/disable

lwander:
@duftler @ajordens @jtk54 PTAL

This uses the bounds from [my doc](https://docs.google.com/a/google.com/document/d/1rHe6JUkKGt58NaVO_3fHJt456Pw5kiX_sdVn1gwTZxk/edit?usp=sharing) to determine if an enable disable with a specified desired percentage succeeds or fails while accounting for autoscaling behavior.

I renamed `WaitForAllInstancesDownTask` to `WaitForRequiredInstancesDownTask` because it can now succeed if the bounds above are met.


lwander:
ping

PR #1183, opened 2017-02-14
(halconfig) Fix orca endpoint references

lwander:


lwander:
This affects the halconfig only

@duftler FYI

PR #1184, opened 2017-02-15
Rebumps Fiat API up to 0.21

ttomsu:
@cfieber @zanthrash Please test in your environment and lets get down to why it was returning 401s a few weeks ago.


ttomsu:
Ping

cfieber:
testing this out

ttomsu:
Thanks for looking into this again. I'll revisit Gate and see if I can refactor to remove that property, as I agree that it isn't super straightforward.

PR #1187, opened 2017-02-19
Ensuring interesting health provider names are preserved for rolling …

jeyrschabu:
…push

The Titus migration uses a clone/create server group with a Rolling Push strategy. 
Additionally, the strategy creates two stages, a modify launch config stage and an actual rolling push stage.

Previously, ```interestingHealthProviderNames``` was set at the clone/create server group stage. However this was isolated to the said stage and not available in the rolling push stage context. 
As a consequence, the titus migration would get stuck waiting for up instances because ```interestingHealthProviderNames``` would default to ```["Load Balancer, Discovery"]```

This was never caught because of a false positive, the image used during testing was "discoverable", thus concealing the ugly behavior.

Now, this change revisits the implementation & ensures ```interestingHealthProviderNames``` are present (resolved through a titus job label) :

- Moved logic to determine interesting health provider names in rolling push to a task
- Updated RollingPush to ensure interesting health names before executing tasks

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
@@ -21,7 +21,6 @@ import com.netflix.spinnaker.orca.DefaultTaskResult;
 import com.netflix.spinnaker.orca.ExecutionStatus;
 import com.netflix.spinnaker.orca.RetryableTask;
 import com.netflix.spinnaker.orca.TaskResult;
-import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.InterestingHealthProviderNamesSupplier;
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.ServerGroupCreator;
 import com.netflix.spinnaker.orca.clouddriver.utils.CloudProviderAware;
 import com.netflix.spinnaker.orca.front50.Front50Service;
@@ -34,10 +33,7 @@ import org.springframework.stereotype.Component;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.List;
 import java.util.Map;
-import java.util.HashMap;
-import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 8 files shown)


ajordens:
I suppose alternatively there could be a task on `RollingPushStage` that looked to the parent for any `interestingHealthProviderNames` and copied  them down?

RollingPushStage is synthetic so it'll have a parent.

jeyrschabu:
Yup, the only issue in that case would be ```interestingHealthProviderNames``` get resolved after the rolling push stage is already constructed. 

I tested the current solution and it works like a charm :)

ajordens:
`interestingHealthProviderNames` gets resolved at runtime but that should be fine provided that the parent's `DeterminateHealthProviderNames` task is complete prior to the `RollingPushStage` starting ... which I think is the case?

Do we not also need the "lookup from label" behavior for non-RollingPush's?

jeyrschabu:
Meant to say resolving ```interestingHealthProviderNames``` from label.  ```DeterminateHealthProviderNames```.  I'm guessing the task you are talking about to get ```interestingHealthProviderNames``` from the parent stage would reside exactly where I have the ```ensureInterestingHealthProviderNamesTask``` ?

As far as I know, this is only for Rolling Pushes. @tomaslin thoughts ?

ajordens:
Yup, it would live in the same spot.

Definitely needed for RollingPush as that's what SPaaS is doing, I was thinking it might also be appropriate behavior for a regular non-strategy Clone/Deploy, no? (or maybe in those situations it's explicitly specified in the UI?)

robfletcher:
Other strategies inherit the determine health providers task from `AbstractDeployStrategyStage` which rolling push doesn't extend

jeyrschabu:
Probably something we could extend with minimal friction in the future as the need arise.

We could also make RollingPushStage extends ```AbstractDeployStrategyStage``` to make it consistent with the other strategies per @robfletcher 's note but the loop in RollingPush scares me.

I will move forward with this PR, and maybe follow up with another later to make RollingPush consistent if needed

ajordens:
Pretty sure that `AbstractDeployStrategyStage` doesn't factor in here ... the only subclasses of that are `CreateServerGroupStage` and `CloneServerGroupStage`, which should be the parent of the RollingPushStage right?

The rolling push stage is a little different than the others in that the actual work happens in a synthetic stage.

I still believe the correct long-term solution is to have RollingPush pull the health providers from the DetermineHealthProviderNames task ... regardless of whether it's titus or not. 

This is alright in the short-term but it's worth taking a stab at consolidating it sooner than later IMO.

PR #1188, opened 2017-02-20
Add optional force flag to the cancel pipeline endpoint

jervi:
This option will make it possible to force cancel a pipeline (kill a Zombie process) before the default timeout has passed.

The intent is to make a feature in Deck that lets a user force cancel a zombie process before the default timeout interval of the pipeline has elapsed.


jervi:
This is related to the closed Spinnaker issue [#1353](https://github.com/spinnaker/spinnaker/issues/1353).
cc/ @robfletcher

jervi:
I'll start on the Deck integration tomorrow.

robfletcher:
Seems reasonable, I guess.

PR #1191, opened 2017-02-23
Refactoring & adding cycle detection in DAG

jeyrschabu:
- Updated injectFirst to add directly to the beginning of the list since graph is topologically sorted
   - Updated injectLast. Now finding leaf nodes through dfs as well
   - Ensuring no cycle is introduced during injection


jeyrschabu:
WIP: NO MERGE

robzienert:
👍

PR #1193, opened 2017-02-24
feat(pipelinetemplates): Adding basic plan capability

robzienert:
A very rudimentary plan capability for pipeline templates. I have some ideas on how to improve (and increase) debug output if people get to using this and need more output. I'd prefer to see how people start to use this feature before diving deep into enhanced output.

Should we want more details, I think I'll do some refactoring to allow us to collect consumer-friendly errors and debug output as we compile the execution. Before tackling this, I'd like to see what the UI might want to display so we're providing the right context.

@spinnaker/reviewers PTAL

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
@@ -15,6 +15,7 @@
  */
 package com.netflix.spinnaker.orca.pipelinetemplate.v1schema.graph.transform;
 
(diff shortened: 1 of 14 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
```diff
@@ -97,7 +103,7 @@ private static void dfs(String stageId,
 
     for (String n : stage.getRequisiteStageRefIds()) {
       Status status = state.get(n);
-      if (status == Status.VISITING) throw new IllegalStateException("Cycle detected in graph");
+      if (status == Status.VISITING) throw new IllegalTemplateConfigurationException("Cycle detected in graph (discovered on stage: " + stageId + ")");
```
jrsquared: Everything else in here uses `String.format()` instead of string concatenation, would be nice to be consistent.

jrsquared:
One minor comment, but otherwise LGTM

PR #1196, opened 2017-03-01
prevent disabled pipelines from launching when queued / sub-pipelines

robfletcher:


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -201,6 +202,13 @@ class OperationsController {
...
   }
 
+  @ResponseStatus(HttpStatus.BAD_REQUEST)
+  @ExceptionHandler(PipelineValidator.PipelineValidationFailed)
+  Map disabledPipelineHandler(PipelineValidator.PipelineValidationFailed e) {
+    return [message: e.message, status: HttpStatus.BAD_REQUEST]
+  }
+
+  @Deprecated
   static class DisabledPipelineException extends RuntimeException {
```
robfletcher: We can get rid of this previous exception when we remove v1. For now we might as well keep it. It only traps manually triggered executions.

PR #1197, opened 2017-03-01
feat(rrb) Bare bones RRB support

lwander:
This only supports an incremental rollout without any attention paid to autoscaling or calling out to custom strategies.
rrb
6nwsz4cute7
wrfvo0uo6kf

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStageData.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStageData.groovy
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
+
+import com.netflix.spinnaker.orca.kato.pipeline.support.StageData
+
+class RollingRedBlackStageData extends StageData {
+  List<Integer> targetPercentages = []
+  Map capacity
+  Map originalCapacity
+  Integer targetSize
+}
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
@@ -1,4 +1,136 @@
 package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
 
-class RollingRedBlackStrategy {
+import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.DisableClusterStage
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerGroupStage
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.DetermineTargetServerGroupStage
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup
+import com.netflix.spinnaker.orca.kato.pipeline.support.SourceResolver
+import com.netflix.spinnaker.orca.pipeline.model.Execution
+import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner
+import groovy.util.logging.Slf4j
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.ApplicationContext
+import org.springframework.context.ApplicationContextAware
+import org.springframework.stereotype.Component
+
+import static com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder.StageDefinitionBuilderSupport.newStage
+
+@Slf4j
+@Component
+class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
+  final String name = "rollingredblack"
+
+  @Autowired
+  DisableClusterStage disableClusterStage
+
+  @Autowired
+  ResizeServerGroupStage resizeServerGroupStage
+
+  @Autowired
+  DetermineTargetServerGroupStage determineTargetServerGroupStage
+
+  @Autowired
+  SourceResolver sourceResolver
+
+  @Override
+  <T extends Execution<T>> List<Stage<T>> composeFlow(Stage<T> stage) {
+    def stages = []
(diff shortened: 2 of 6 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -1,4 +1,136 @@
...
+
+    stage.context.savedCapacity = savedCapacity
+    stage.context.capacity = [
+        min: 0,
+        max: 0,
+        desired: 0
+    ]
+
+    def targetPercentages = stageData.getTargetPercentages()
+    if (targetPercentages.size() == 0 || targetPercentages[-1] != 100) {
```
duftler: Would `.last()` work here?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -1,4 +1,136 @@
...
+
+    Map baseContext = [
+        (cleanupConfig.location.singularType()): cleanupConfig.location.value,
+        cluster                                : cleanupConfig.cluster,
+        credentials                            : cleanupConfig.account,
+        cloudProvider                          : cleanupConfig.cloudProvider,
+    ]
+
+    Map savedCapacity = stageData.savedCapacity ?: stageData.capacity
+    if (!savedCapacity) {
```
lwander: Ah right, @robfletcher - I added this because this code path seems be called twice when building the stages, do you know why that is?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: Yup, that reads nicer too - I'll switch it.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DisableClusterTask.groovy
```diff
@@ -36,6 +36,16 @@ class DisableClusterTask extends AbstractClusterWideClouddriverTask {
     boolean preferLargerOverNewer = false
   }
 
+  public static class DisableOperation {
+    Integer desiredPercentage
```
cfieber: is there an equivalent clouddriver update that makes disableServerGroup understand this option?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -1,4 +1,136 @@
...
+        "Determine Deployed Server Group",
+        findContext,
+        stage,
+        SyntheticStageOwner.STAGE_AFTER
+    )
+
+    // java .forEach rather than groovy .each, since the nested .each closure sometimes omits parent context
+    targetPercentages.forEach({ p ->
+      def resizeContext = baseContext + [
+          target: TargetServerGroup.Params.Target.current_asg_dynamic,
```
cfieber: what would happen ~~if~~when this is running on a cluster and someone manually deploys a new server group in there?

Wondering if we can get the target server group once instead of per iteration?

cfieber: alternatively maybe inserting a checkPreconditions to assert the cluster is exactly 2 server groups each iteration

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DisableClusterTask.groovy
lwander: Yup I extended all the enable/disable descriptions a little while ago: https://github.com/spinnaker/clouddriver/commit/cc10d6f8cb8191ed72294fc751b4cf520476b124

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: Good question - my quick read of the target resolver stuff was that it would always resolve to whatever that first run of the `determineTargetServerGroup` stage found before all the resizes & disables start running, so even if something new was deployed it would keep resizing the same server group. But now that you mention it, the disable cluster stage will start disabling the wrong stuff... CheckPreconditions sounds like the best answer

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: pinging @robfletcher

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStageData.groovy
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
+
+import com.netflix.spinnaker.orca.kato.pipeline.support.StageData
+
+class RollingRedBlackStageData extends StageData {
+  List<Integer> targetPercentages = []
+  Map capacity
+  Map savedCapacity
```
jtk54: +1

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
jtk54: Bueller... Bueller... Bueller...

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: I've thought about this some more and the checkPreconditions feels cumbersome, since none of the other strategies so far are limited to a cluster size of 2, and potentially suffer from the same problem where an inappropriate deploy during a strategy could cause the wrong server group to be destroyed/disabled. I've verified that the `current_asg_dynamic` always resolves to the same server group that was first resolved before the operation started. 

So, if someone does manually deploy a new cluster halfway through, they're just left with an extra disabled server group, and their manually deployed server group taking all the traffic. If this becomes a problem, a more generic TrafficGuard-like solution is probably more appropriate.

lwander: Just to demonstrate, I created a server group with size 1 in the cluster being rrb'd after the strategy started: 
rrb1
rrb2
rrb3

lwander:
Merging due to lack of complaints

PR #1204, opened 2017-03-03
add details of running executions to execution tracker

robfletcher:
Also don't flush them when an instance goes stale so we can see zombie executions.

### DIFF
--- a/orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/SpringBatchActiveExecutionTracker.java
+++ b/orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/SpringBatchActiveExecutionTracker.java
@@ -16,13 +16,15 @@
 
 package com.netflix.spinnaker.orca.batch;
 
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
+import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


robfletcher:

robzienert:
👍

orca-spring-batch/src/main/java/com/netflix/spinnaker/orca/batch/SpringBatchActiveExecutionTracker.java
```diff
@@ -95,15 +104,16 @@ public SpringBatchActiveExecutionTracker(JobOperator jobOperator,
   }
 
   /**
-   * Records the count of active executions running on _this_ instance.
+   * Records the details of active executions running on _this_ instance.
    */
   @Scheduled(fixedDelay = FREQUENCY_MS)
   public void countRunningExecutions() {
```
duftler: Would probably help future readers if we rename the various count* methods (and maybe tests too).

robfletcher: Good point

robfletcher:
@duftler should be resolved now

PR #1206, opened 2017-03-06
fix(core): Avoid capturing security group ingress rules with empty ci…

duftler:
…dr's.


duftler:
@danielpeach please review.

PR #1208, opened 2017-03-06
fix(netflix): ensure older instances report counts

robfletcher:
also, expire details of zombie executions after a week


robfletcher:

PR #1210, opened 2017-03-08
fix(executions): avoid intermittent StackOverflowError

robfletcher:
The use of `Sets.union` was causing such a deeply nested data structure with large amounts of running executions that we get StackOverflowError trying to iterate over it. Changed the implementation to much simpler and RxJava-y form.


robfletcher:

jrsquared:
👍

jeyrschabu:
LGTM

PR #1211, opened 2017-03-08
feature(netflix): simplify zombie pipeline cleanup

robfletcher:
Instead of using a heuristic to determine if a pipeline is in a zombie state we now just look at the reported state of the instance running the pipeline.


robfletcher:
PTAL @spinnaker/netflix-reviewers @duftler

robzienert:
👍

duftler:
Nice. A belated LGTM.

PR #1213, opened 2017-03-09
feat(travis stage): Add explicit Travis stage

jervi:
Adds support for a new Travis stage. It reuses the Jenkins api towards Igor.

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/JenkinsStage.groovy
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/JenkinsStage.groovy
@@ -43,11 +43,11 @@ class JenkinsStage implements StageDefinitionBuilder, RestartableStage, Cancella
   @Override
   <T extends Execution<T>> void taskGraph(Stage<T> stage, TaskNode.Builder builder) {
     builder
-      .withTask("startJenkinsJob", StartJenkinsJobTask.class)
-      .withTask("waitForJenkinsJobStart", MonitorQueuedJenkinsJobTask.class)
+      .withTask("start${getType().capitalize()}Job", StartJenkinsJobTask.class)
+      .withTask("waitFor${getType().capitalize()}JobStart", MonitorQueuedJenkinsJobTask.class)
 
     if (!stage.getContext().getOrDefault("waitForCompletion", "true").toString().equalsIgnoreCase("false")) {
-      builder.withTask("monitorJenkinsJob", MonitorJenkinsJobTask.class);
+      builder.withTask("monitor${getType().capitalize()}Job", MonitorJenkinsJobTask.class)
     }
   }
 
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/TravisStage.groovy
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/TravisStage.groovy
@@ -1,0 +1,7 @@
+package com.netflix.spinnaker.orca.igor.pipeline
+
+import org.springframework.stereotype.Component
+
+@Component
+class TravisStage extends JenkinsStage {
+}


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/BuildStage.groovy
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright 2017 Schibsted ASA.
```
tomaslin: would keep this a netflix header since you just moved code here

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/BuildStage.groovy
```diff
@@ -0,0 +1,87 @@
...
+      .withTask("waitFor${getType().capitalize()}JobStart", MonitorQueuedBuildJobTask.class)
+
+    if (!stage.getContext().getOrDefault("waitForCompletion", "true").toString().equalsIgnoreCase("false")) {
+      builder.withTask("monitor${getType().capitalize()}Job", MonitorBuildJobTask.class);
+    }
+  }
+
+  @Override
+  Stage prepareStageForRestart(ExecutionRepository executionRepository, Stage stage, Collection<StageDefinitionBuilder> allStageBuilders) {
+    stage = StageDefinitionBuilder.StageDefinitionBuilderSupport
```
tomaslin: is the travis stage the same as the jenkins stage in terms of buildInfo and buildNumber?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/JenkinsStage.groovy
```diff
@@ -17,72 +17,10 @@
...
-import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
-import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
-@Slf4j
 @Component
 @CompileStatic
-class JenkinsStage implements StageDefinitionBuilder, RestartableStage, CancellableStage {
-  @Autowired StopJenkinsJobTask stopJenkinsJobTask
+class JenkinsStage extends BuildStage {
```
tomaslin: If the travis stage is just the same thing as the jenkins stage, why do this major refactoring and renaming in orca instead of just making TravisStage extend JenkinsStage? Seems like a much less disruptive change

tomaslin: Also not sure we have a good story renaming tasks. IIRC changing these task names makes trying to load a previous pipeline / execution incompatible with the previous release. Not really comfortable doing this massive renaming of tasks for the sake of supporting the travis stage. I would roll back these changes and just have a TravisStage extends JenkinsStage

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/MonitorBuildJobTask.groovy
```diff
@@ -32,7 +32,7 @@ import java.util.concurrent.TimeUnit
 
 @Slf4j
 @Component
-class MonitorJenkinsJobTask implements RetryableTask {
+class MonitorBuildJobTask implements RetryableTask {
```
tomaslin: Yeah, don't rename tasks. This causes serialization problems trying to read pipelines using the old task name. Here is an example of the type of work that needs to be done afterwards.  https://github.com/spinnaker/orca/pull/1198

jervi:
Thanks @tomaslin! I actually had just the simpler implementation first. I'll revert and update the PR.

jervi:
@tomaslin Please have a look at the updated PR when you have time

tomaslin:
LGTM. Needs to be squashed.

jervi:
I've re-squashed

PR #1214, opened 2017-03-09
Get rid of v1 execution engine and isolate Spring Batch dependency

robfletcher:
This removes all the legacy code involved in the "v1" execution engine. Now all Spring Batch dependent code is isolated in the `orca-spring-batch` module.


robfletcher:
PTAL @spinnaker/netflix-reviewers @spinnaker/google-reviewers

orca-spring-batch/src/main/groovy/com/netflix/spinnaker/orca/batch/RetryableTaskTasklet.groovy
```diff
@@ -99,7 +97,8 @@ class RetryableTaskTasklet extends TaskTasklet {
   /**
    * In order to accurately determine whether this task has timed out, any time spent in a paused state must be discounted.
    */
-  private static long determineCurrentExecutionTime(Clock clock, long startTime, PausedDetails pausedDetails) {
+  private
+  static long determineCurrentExecutionTime(Clock clock, long startTime, PausedDetails pausedDetails) {
```
robzienert: Unnecessary new line.

robzienert: And in a few other places below.

robzienert:
Did a look-through, but I'm not really familiar enough to technically vet it.

orca-spring-batch/src/main/groovy/com/netflix/spinnaker/orca/batch/RetryableTaskTasklet.groovy
robfletcher: That's IntelliJ auto-wrapping lines, sorry.

ttomsu:
Onlookers FYI: This caused some startup failures in our environment. @robfletcher and I are looking into it.

PR #1217, opened 2017-03-10
fix(Jenkins): Ensuring there is a buildNumber present in the context before proceeding

jeyrschabu:
- Added a guard to return with a TERMINAL status when we dont have a buildNumber


jeyrschabu:
@robfletcher PTAL

robfletcher:
👍 
:shipit:

PR #1218, opened 2017-03-10
any execution that does not specify a version will run v2

robfletcher:
This will any "anonymous" pipelines submitted via the REST API will run on v2. That should mean everything is running v2.

We should ensure this is happy before merging #1214


ajordens:
LGTM

PR #1221, opened 2017-03-11
feat(pipelinetemplate): Basic template & config validation

robzienert:
Working in a better interface for performing validation of templates and configurations. This doesn't include working in any of the `orca-validation` module stuff. My goal here is to emit more user-friendly errors than stacktraces that are surfaced in the UI or Halyard.


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessor.java
```diff
@@ -82,33 +86,33 @@ public PipelineTemplatePipelinePreprocessor(ObjectMapper pipelineTemplateObjectM
...
     ExecutionGenerator executionGenerator = new V1SchemaExecutionGenerator();
 
     Map<String, Object> generatedPipeline = executionGenerator.generate(template, templateConfiguration);
 
-    validateGeneratedPipeline(generatedPipeline);
-
     return generatedPipeline;
   }
 
-  private static void validateGeneratedPipeline(Map<String, Object> generatedPipeline) {
```
robzienert: I rolled this check into the configuration validator.

PR #1231, opened 2017-03-17
fix(config): Make front50, bakery & igor optional.

lwander:


lwander:
Don't merge, there is a better fix incoming...

lwander:
@robfletcher orca used to ignore errors connecting to non-critical dependencies like front50, which is something halyard depends on. I think the last few orca changes have changed that behavior, so it feels cleaner to toggle those services on and off. PTAL.

orca-web/config/orca.yml
```diff
@@ -6,6 +6,7 @@ default:
     account: default
 
 front50:
+  enabled: true
```
lwander: As far as I was able to verify, adding these flags here ensures default behavior is unaffected.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -58,6 +59,9 @@ class CreateBakeTask implements RetryableTask {
   @Override
   TaskResult execute(Stage stage) {
     String region = stage.context.region
+    if (!bakery) {
+      throw new RuntimeException("You have not enabled baking for this orca instance. Set bakery.enabled = true")
```
duftler: In most of the other messages you use this style: `bakery.enabled: true`. Here and one other spot you use `=`.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/EnabledPipelineValidator.java
```diff
@@ -31,12 +31,13 @@
...
-  private final Front50Service front50Service;
-
-  @Autowired
-  public EnabledPipelineValidator(Front50Service front50Service) {this.front50Service = front50Service;}
+  @Autowired(required = false)
+  private Front50Service front50Service;
 
   @Override public void checkRunnable(Pipeline pipeline) {
+    if (front50Service == null) {
+      throw new RuntimeException("Front50 not enabled, no way validate pipeline. Fix this by setting front50.enabled: true");
```
duftler: s/no way/no way to

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -147,7 +142,12 @@ class OperationsController {
     if (trigger.master && trigger.job && trigger.buildNumber) {
       def buildInfo = buildService.getBuild(trigger.buildNumber, trigger.master, trigger.job)
       if (buildInfo?.artifacts) {
-        buildInfo.artifacts = buildArtifactFilter.filterArtifacts(buildInfo.artifacts)
+        if (!buildArtifactFilter) {
+          log.warn("Igor is not enabled, unable to lookup build artifacts. Fix this by settings igor.enabled: true")
```
duftler: s/settings/setting

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/pipelines/UpsertProjectStage.groovy
```diff
@@ -45,13 +45,17 @@ class UpsertProjectStage implements StageDefinitionBuilder {
...
-    @Autowired
+    @Autowired(required = false)
     Front50Service front50Service
 
     UpsertProjectTask() {}
 
     @Override
     TaskResult execute(Stage stage) {
+      if (!front50Service) {
+        throw new RuntimeException("Unable to modify projects, front50 has not been enabled. Fix this by setting front50.enabled: true")
```
robfletcher: It feels like a more specific exception class would be good here (and in the similar places).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -33,15 +33,11 @@
...
-
   @Autowired
-  public TrafficGuard(OortHelper oortHelper, Front50Service front50Service) {
-    this.oortHelper = oortHelper;
-    this.front50Service = front50Service;
-  }
+  private OortHelper oortHelper;
+
+  @Autowired(required = false)
+  private Front50Service front50Service;
```
robfletcher: I have to say I vastly prefer constructor injection. I know Spinnaker generally isn't consistent on that point, though. It's possible to make constructor injected dependencies optional by declaring them as e.g. `Optional<Front50Service>`.

lwander: Ahh that `Optional` is what I was looking for. Thank, I'll switch it over.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -271,6 +271,10 @@ class TaskController {
   List<Pipeline> getPipelinesForApplication(@PathVariable String application,
                                             @RequestParam(value = "limit", defaultValue = "5") int limit,
                                             @RequestParam(value = "statuses", required = false) String statuses) {
+    if (!front50Service) {
+      throw new RuntimeException("Cannot lookup pipelines, front50 has not been enabled. Fix this by settings front50.enabled: true")
```
duftler: s/settings/setting

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
```diff
@@ -42,11 +42,14 @@ class PageApplicationOwnerTask implements RetryableTask {
...
   EchoService echoService
 
-  @Autowired
+  @Autowired(required = false)
   Front50Service front50Service
 
   @Override
   TaskResult execute(Stage stage) {
+    if (!front50Service) {
+      throw new RuntimeException("Front50 is not enabled, no way to fetch pager duty. Fix this by settings front50.enabled: true");
```
duftler: s/settings/setting

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/pipelines/UpsertProjectStage.groovy
lwander: Sure thing. How about `ServiceNotEnabledException`?

duftler: What about https://docs.oracle.com/javase/7/docs/api/java/lang/UnsupportedOperationException.html#UnsupportedOperationException()?

lwander: STGM

lwander:
Comments addressed, PTAL

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/pipelines/UpsertProjectStage.groovy
robfletcher: it's more like an illegal state, no?

lwander: I'd argue state is internal to the system, and is someplace orca can arrive when something unexpected happens, while on the other hand orca with a specific config doesn't support certain operations.

lwander: Up to you what to call it though, I don't mind changing it

lwander:
Thanks for the quick review @robfletcher

PR #1233, opened 2017-03-19
fix(triggers): pipeline triggers may not deserialize parent pipeline …

robfletcher:
…correctly in some circumstances

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
@@ -18,11 +18,8 @@ package com.netflix.spinnaker.orca.kato.tasks
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.module.SimpleModule
-import com.netflix.spinnaker.orca.pipeline.model.Execution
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
```diff
@@ -34,7 +31,7 @@ import com.netflix.spinnaker.orca.pipeline.model.Stage
 trait DeploymentDetailsAware {
 
   private ObjectMapper pipelineObjectMapper =
-    new ObjectMapper().registerModule(new SimpleModule("MyConverterModule").addAbstractTypeMapping(Stage, PipelineStage))
+    new OrcaObjectMapper().registerModule(new SimpleModule("MyConverterModule").addAbstractTypeMapping(Stage, PipelineStage))
```
tomaslin: is that second part needed? instead of just  new OrcaObjectMapper()

robfletcher: Yes, because there's no type specifier in the stages as encoded in the trigger so it has to know to interpret them as PipelineStages.

tomaslin: makes sense. LGTM

PR #1234, opened 2017-03-22
feat(pipelinetemplate): Adding a Jinja template renderer alternative

robzienert:
Refactoring pipeline templates codebase to use a Jinja rendering engine over Handlebars.

While converting our internal pipelines to use templates, I found that I was spending a lot of time fighting against Handlebars. Jinja is a lot more expressive, so it's easier to work with (especially around maps and mutating objects). As a bonus, the standard Jinja tags, filters and expression are fairly well-featured, which should allow us to maintain a smaller codebase (whereas Handlebars, I was already having to build a number of helpers for simple quality of life). I opted for doing this work now, rather than later so we don't need to support a migration effort.

I'm leaving the handlebars code in place (and also the default) for the time being while I convert our already existing pipelines over to Jinja, as the formats are incompatible. A separate PR will rip it out.

Lastly, there's another (lower priority quality of life) task of switching the rendered output from JSON to YAML, which will make templates less verbose - so I created an interface to support that in `RenderedValueConverter`.

### DIFF
--- a/orca-pipelinetemplate/orca-pipelinetemplate.gradle
+++ b/orca-pipelinetemplate/orca-pipelinetemplate.gradle
@@ -5,6 +5,9 @@ dependencies {
 
   compile('com.github.jknack:handlebars:4.0.6')
   compile('com.github.jknack:handlebars-jackson2:1.0.0')
+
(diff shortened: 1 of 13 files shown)


jrsquared:
👍

PR #1235, opened 2017-03-22
Adds respawn to orca upstart conf to deal with it being more sensitive

skim1420:
to redis being down.

@spinnaker/google-reviewers

### DIFF
--- a/orca-web/etc/init/orca.conf
+++ b/orca-web/etc/init/orca.conf
@@ -11,4 +11,4 @@ stop on stopping spinnaker
 
 exec /opt/orca/bin/orca 2>&1 > /var/log/spinnaker/orca/orca.log &
 
-respawn 20 15
+respawn limit 20 15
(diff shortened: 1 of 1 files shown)


orca-web/etc/init/orca.conf
```diff
@@ -10,3 +10,5 @@ expect fork
 stop on stopping spinnaker
 
 exec /opt/orca/bin/orca 2>&1 > /var/log/spinnaker/orca/orca.log &
+
+respawn 20 15
```
lwander: I think you need a `limit` between `respawn` and `20`

lwander:
LGTM, let's have @robfletcher take a look as well since this might affect how autohealing of orca nodes works.

skim1420:
I'm going to close this for now. It doesn't solve the problem we're dealing with immediately, so want to avoid any possible adverse effects on stuff that ain't broke.

PR #1238, opened 2017-03-22
feat(pipelinetemplate): Metadata for UI; template protection

robzienert:
* Adds `protect` flag to templates, allowing template creators to prevent configuration-level modification of the stage graph. Going for the simplest level to start with, we can get more granular down the road if necessary.
* Adds new (optional) metadata block to templates containing `name`, `description` and `owner` so we can surface this in the UI for additional context.

@spinnaker/netflix-reviewers @spinnaker/google-reviewers PTAL

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/TemplateMerge.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/TemplateMerge.java
@@ -31,16 +31,22 @@ public class TemplateMerge {
     for (PipelineTemplate template : templates) {
       appendSource(result, template);
 
-      // TODO rz - should validate var types are the same here or elsewhere?
+      // If any of the templates are set as protected, configurations won't be allowed to alter stages.
+      if (template.getProtect()) {
+        result.setProtect(true);
+      }
+
       result.setVariables(mergeNamedContent(result.getVariables(), template.getVariables()));
 
       mergeConfiguration(result, template);
 
-      // TODO rz - might make sense that any child template must use inject syntax / modules.
       result.setStages(mergeIdentifiable(result.getStages(), template.getStages()));
-
       result.setModules(mergeIdentifiable(result.getModules(), template.getModules()));
     }
+
+    // Apply the last template's metadata to the final result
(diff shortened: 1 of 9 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/TemplateMerge.java
```diff
@@ -31,16 +31,21 @@ public static PipelineTemplate merge(List<PipelineTemplate> templates) {
...
 
       mergeConfiguration(result, template);
 
-      // TODO rz - might make sense that any child template must use inject syntax / modules.
       result.setStages(mergeIdentifiable(result.getStages(), template.getStages()));
-
       result.setModules(mergeIdentifiable(result.getModules(), template.getModules()));
     }
+
+    result.setMetadata(templates.get(templates.size() - 1).getMetadata());
```
jeyrschabu: What is the story around this ?

jeyrschabu:
One question. Other than that LGTM

danielpeach:
It's still possible for config level parameters to mutate the graph, right? Mostly checking if I understand everything properly. e.g., if a parent template had:

```
id: manualJudgment
type: manualJudgment
inject:
  before: stage.{{variable}}
```

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/TemplateMerge.java
robzienert: Will add a comment to the code, but we're only interested in applying the template metadata from the last template. Maybe this won't be necessary for the sake of building up pipeline executions, though.

robzienert:
@danielpeach It isn't, no. I'm checking to see if any stages are defined as part of the configuration (what you pass in at exec/plan time). If any are defined, you'll get an error. Also, your example calls out that I need to update dcd-spec:

```
stages:
- id: manualJudgement
  type: manualJudgement
  inject:
    before: myStage <-- No "stage." or "module." namespacing anymore
```

Furthermore, `ConfigStageInjectionTransform` does not evaluate the templating language, so the injection rules are static.

PR #1242, opened 2017-03-23
refactor(core): stop extending ObjectMapper

robfletcher:
PTAL @cfieber


orca-core/src/main/groovy/com/netflix/spinnaker/orca/jackson/OrcaObjectMapper.java
```diff
@@ -0,0 +1,20 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.guava.GuavaModule;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+import static com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES;
+
+public class OrcaObjectMapper {
+  private OrcaObjectMapper() {}
+
+  public static ObjectMapper newInstance() {
```
robfletcher: Static factory method for an instance of `ObjectMapper` with our config.

PR #1246, opened 2017-03-26
fix(core): Remove unsafe bool cast

lwander:
Undefined "strategy" can't be cast to `boolean`.


lwander:
Or if any of the @spinnaker/netflix-reviewers are online and can give this a quick look. Seems that Matt is out.

jrsquared:
LGTM, curious why there was a boolean cast there to begin with? 🤔

lwander:
Thanks guys

lwander:
I thought `"strategy"` was a string? e.g. `"redblack"` or `"highlander"`

robfletcher:
Not in this case. It's a Boolean in the trigger (IIRC) that indicates if the child pipeline is a custom deploy strategy or not. You should probably do `null as Boolean ?: false`

tomaslin:
No, this is to indicate that that pipeline is a custom strategy execution. The string you're thinking of stays in the deploy stage

lwander:
Ah ok let me fix this

lwander:
https://github.com/spinnaker/orca/pull/1247

PR #1248, opened 2017-03-27
fix(authentication): subtly different handling of varargs between Java & Groovy

robfletcher:


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.java
```diff
@@ -80,7 +80,7 @@ public AuthenticationDetails(String user, String... allowedAccounts) {
       if (spinnakerUserOptional.isPresent() || spinnakerAccountsOptional.isPresent()) {
         return Optional.of(new AuthenticationDetails(
           spinnakerUserOptional.orElse(null),
-          spinnakerAccountsOptional.map(s -> s.split(",")).orElse(null)
+          spinnakerAccountsOptional.map(s -> s.split(",")).orElse(new String[0])
```
robfletcher: Passing `null` here from Groovy would effectively call `new AuthenticationDetails(user)` in Java it calls `new AuthenticationDetails(user, null)` -> NPE.

PR #1250, opened 2017-03-28
feat(pipelinetemplate): Support YAML output from templates

robzienert:
Big quality of life improvement for pipeline templating. Makes templates like these:

```yaml
- id: bake
  type: bake
  config:
    regions: |-
      [
        {% for region in bakeRegions %}
          "{{ region }}"{% if not loop.last %},{% endif %}
        {% endfor %}
      ]
```

Easier to read and write, like so:

```yaml
- id: bake
  type: bake
  config:
    regions: |
      {% for region in bakeRegions %}
      - {{ region }}
      {% endfor %}
```

Still supports JSON, of course, so there's no incompatibility with existing templates.

@spinnaker/reviewers PTAL


jeyrschabu:
LGTM

PR #1251, opened 2017-03-28
(netflix) Enhance rollbacks for Fast Prop Pipelines with multiple stages.

zanthrash:
Fast Property pipelines can now have multiple persisted property stages in them.

The current FastPropertyCleanUpListener will only rollback the last stage in the case of a
pipeline failure leaving all previous stages in tact.

This change looks at each stage context, instead of the global context to determine the rollback strategy.


zanthrash:
@robfletcher wow!  Thanks for getting around to this.

PR #1253, opened 2017-03-29
refactor(core): removed DefaulTaskResult and made TaskResult a class

robfletcher:
There's just no reason for having a `TaskResult` interface and a single default implementation, it's unnecessary complexity.


orca-core/src/main/groovy/com/netflix/spinnaker/orca/DefaultTaskResult.groovy
```diff
@@ -1,56 +0,0 @@
...
- * limitations under the License.
- */
-
-package com.netflix.spinnaker.orca
-
-import groovy.transform.CompileStatic
-import com.google.common.collect.ImmutableMap
-
-@CompileStatic
-final class DefaultTaskResult implements TaskResult {
```
robfletcher: Gone

orca-core/src/main/groovy/com/netflix/spinnaker/orca/TaskResult.java
```diff
@@ -1,14 +1,37 @@
...
-import java.io.Serializable;
+import java.util.Map;
 import com.google.common.collect.ImmutableMap;
+import lombok.Value;
+import static java.util.Collections.emptyMap;
 
-public interface TaskResult {
-  ExecutionStatus getStatus();
+@Value
+public final class TaskResult {
```
robfletcher: This becomes a simple immutable value type.

PR #1270, opened 2017-04-07
feat(loadtest): Skeleton loadtest

robzienert:
Sets up the bare minimum structure for load tests in orca. This will be progressively enhanced to help enable vetting of nu-orca's reliability characteristics.

The plan here is to get this skeleton in place to run local load to vet what metrics will and will not be valuable for nu-orca. From there, my goal is to start injecting more diversified orchestations, polling and what have you to uncover any issues, as well as vet more advanced features like rate limiting, execution logs and so-on.

It's laid out kinda strange for such a simple implementation, but the structure scales well as we introduce more complexity. Thanks @jonsie for the pointers & refreshers.

PTAL @robfletcher

### DIFF
--- a/orca-loadtest/README.md
+++ b/orca-loadtest/README.md
@@ -1,0 +1,5 @@
+# orca-loadtest
+
+This module is a work in progress for testing the reliability and 
+scale characteristics of Orca. It is not run as part of CI: It is 
+currently intended for ad-hoc execution.
--- a/orca-loadtest/src/main/resources/orca-simulation.conf
+++ b/orca-loadtest/src/main/resources/orca-simulation.conf
@@ -1,0 +1,11 @@
+service.orca {
+  serviceUrl = "http://localhost:8083"
+
+  rampUpPeriod = 10
+  duration = 30
+
+  submitOrchestration {
+    rampUsersPerSec = 1
+    rampUsersTo = 3
+  }
+}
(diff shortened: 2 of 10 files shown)


orca-loadtest/README.md
```diff
@@ -0,0 +1,5 @@
+# orca-loadtest
+
+This module is a work in progress for testing the reliability and 
+scale characteristics of Orca. It is not run as part of CI: It is 
+currently intended for ad-hoc executinon.
```
jonsie: Typo -- `executinon`.

orca-loadtest/src/main/resources/orca-simulation.conf
```diff
@@ -0,0 +1,15 @@
...
+
+  rampUpPeriod = 10
+  duration = 30
+
+  submitOrchestration {
+    rampUsersPerSec = 1
+    rampUsersTo = 3
+  }
+
+  setup {
```
jonsie: If you don't think you need it, I'd remove this - helps declutter things.  You can always add it back later.

orca-loadtest/src/main/scala/com/netflix/spinnaker/orca/loadtest/OrcaSimulation.scala
```diff
@@ -0,0 +1,61 @@
...
+  val config = new OrcaSimulationConfig(OrcaSimulationConfig.loadConfig())
+
+  // TODO rz - Make a circular feeder of RawFileBody instead.
+  lazy val CircularTaskFeeder: Feeder[String] = {
+    Iterator.continually(Map(
+      "body" -> ""
+    ))
+  }
+
+  def dataSetup(): Unit = {
```
jonsie: If you don't think you need it, I'd remove this - helps declutter things.  You can always add it back later.

robzienert:
Thanks.

PR #1272, opened 2017-04-11
feat(pipelinetemplate): Removing handlebars rendering option

robzienert:
Removing handlebars rendering support: All templates will now only support jinja.


jrsquared:
RIP IT OUT

PR #1274, opened 2017-04-13
chore(aws): remove no longer used amiId

icfantv:
* remove no longer used `amiId` property from `deploymentDetails` map on
the stage context for AWS.  this is cleanup as a result of the
discussion in #1264.


icfantv:
@tomaslin @ajordens PTAL.  This is what we discussed removing in #1264.

ajordens:
LGTM

PR #1278, opened 2017-04-14
fix(url restriction) - includes url restriction on webhook stages

cfieber:
breaking: this changes the configuration namespace for the recently introduced URL restriction configuration from `expressions` to `userConfiguredUrlRestrictions`


cfieber:
fyi @jervi

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
```diff
@@ -281,32 +281,8 @@ abstract class ContextUtilities {
...
-          throw new IllegalArgumentException('invalid address for ' + u.getHost())
-        }
-      }
-      if (cfg.rejectLinkLocal && addr.isLinkLocalAddress()) {
-        throw new IllegalArgumentException('invalid address for ' + u.getHost())
-      }
-    }
-
-    return u.toURL().getText()
+    URL u = contextFunctionConfiguration.get().urlRestrictions.validateURI(url).toURL()
```
cfieber: moved this validation into the UserConfiguredUrlRestrictions class

PR #1282, opened 2017-04-18
make sure synthetic stages don't get re-generated on restart

robfletcher:
This was partly masked by the fact that restarting on the same orca instance would use the same Spring Batch job so nothing got rebuilt at all. However, restarting on a new instance would duplicate all the synthetic and parallel stages.

 SPIN-2505


jeyrschabu:
LGTM

PR #1284, opened 2017-04-18
feat(web): Adding resolve pipeline template endpoint

robzienert:
Endpoint resolves & merges a template inheritance tree by source. This will be used to help Deck render variable bindings while configuring a pipeline template.

@spinnaker/reviewers PTAL

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/PipelineTemplateController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/PipelineTemplateController.groovy
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.controllers
+
+import com.netflix.spinnaker.orca.pipelinetemplate.PipelineTemplateService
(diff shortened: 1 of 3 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/PipelineTemplateController.groovy
```diff
@@ -0,0 +1,68 @@
...
+    return [message: e.message, status: HttpStatus.BAD_REQUEST]
+  }
+
+  @ResponseStatus(HttpStatus.BAD_REQUEST)
+  @ExceptionHandler(TemplateLoaderException)
+  Map templateLoadingErrorHandler(TemplateLoaderException e) {
+    return [message: e.message, status: HttpStatus.BAD_REQUEST]
+  }
+
+  static class InvalidTemplateSourceException extends RuntimeException {
```
danielpeach: Could use `@InheritConstructors` here.

jeyrschabu:
LGTM other than the comment above

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/PipelineTemplateController.groovy
robzienert: Oh Groovy, you so magical. Thanks.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/PipelineTemplateController.groovy
```diff
@@ -0,0 +1,66 @@
...
+@ConditionalOnProperty("pipelineTemplate.enabled")
+@RestController
+@Slf4j
+class PipelineTemplateController {
+
+  @Autowired
+  PipelineTemplateService pipelineTemplateService
+
+  @RequestMapping(value = "/pipelineTemplate", method = RequestMethod.GET)
+  PipelineTemplate getPipelineTemplate(@QueryParam("source") String source) {
```
danielpeach: I see `RequestParam` in a lot of places - any reason to use `QueryParam` instead?

robzienert: Updated. Turns out `QueryParam` is JAX-RS. It works, but probably preferential to use the Spring `RequestParam` instead.

PR #1285, opened 2017-04-18
fix(pipelinetemplate): Child stages of conditional stages should be p…

jeyrschabu:
…reserved

- Fixed an issue where children of a rendered conditional stage are dropped off the stages graph
- If s2 is a conditional stage then:
Input:
 s1 <- s2 <- s4
       \- s3 

Output:
   s1 <- s4
	  \- s3

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConditionalStanzaTransform.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConditionalStanzaTransform.java
@@ -25,6 +25,7 @@ import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.render.Renderer;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 public class ConditionalStanzaTransform implements PipelineTemplateVisitor {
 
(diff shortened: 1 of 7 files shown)


jeyrschabu:

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConditionalStanzaTransform.java
```diff
@@ -46,25 +47,18 @@ public void visitPipelineTemplate(PipelineTemplate pipelineTemplate) {
...
+      .stream()
+      .filter(stage -> stage.getWhen() != null && !stage.getWhen().isEmpty())
+      .forEach(stage -> {
+        RenderContext context = new DefaultRenderContext(templateConfiguration.getPipeline().getApplication(), template, trigger);
+        context.getVariables().putAll(templateConfiguration.getPipeline().getVariables());
+        for (String conditional : stage.getWhen()) {
+          String rendered = renderer.render(conditional, context);
+          if (!Boolean.parseBoolean(rendered)) {
+            stage.setConditional(true);
+          }
```
robzienert: This logic reads weird. I see why you wrote it the way you did, but looking at this conditional on its own (without seeing `ConfigStageInjectionTransform`), you'd think it was logically a bug. Definitely warrants a comment like you have below.

robzienert:
🥇

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConditionalStanzaTransform.java
jeyrschabu: Will rename the method.

PR #1289, opened 2017-04-20
Task.implementing class becomes a String to avoid CNFE when deserializing

robfletcher:
Previously deserializing old executions could blow up if someone renamed a Task implementation class. Now at least it will only prevent running a new execution. Still not ideal but having a bean name reference or whatever is just as prone to breakage and this solution doesn't require any data migration.


robzienert:
👍

PR #1290, opened 2017-04-20
Pin `min` capacity to `desired` capacity when rolling back

ajordens:
This prevents the target server group from unnecessarily scaling down
and causing the rollback to timeout / take far longer than expected.

`min` capacity is restored when the rollback is complete.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ApplySourceServerGroupCapacityTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ApplySourceServerGroupCapacityTask.groovy
@@ -49,38 +49,78 @@ class ApplySourceServerGroupCapacityTask extends AbstractServerGroupTask {
   @Override
   Map convert(Stage stage) {
     try {
-      def ancestorDeployStage = getAncestorDeployStage(executionRepository, stage)
-      def ancestorDeployStageData = ancestorDeployStage.mapTo(DeployStageData)
+      TargetServerGroupContext operationContext = getTargetServerGroupContext(stage)
 
-      def deployServerGroup = oortHelper.getTargetServerGroup(
-        ancestorDeployStageData.account,
-        ancestorDeployStageData.deployedServerGroupName,
-        ancestorDeployStageData.region,
-        getCloudProvider(ancestorDeployStage)
+      def context = operationContext.context
+      def sourceServerGroupCapacitySnapshot = operationContext.sourceServerGroupCapacitySnapshot
+
+      def targetServerGroup = oortHelper.getTargetServerGroup(
+        context.credentials as String,
+        context.serverGroupName as String,
+        context.region as String,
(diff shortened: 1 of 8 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -197,7 +197,7 @@ public Stage preceding(String type) {
    * Maps the stage's context to a typed object at a provided pointer. Uses
    * <a href="https://tools.ietf.org/html/rfc6901">JSON Pointer</a> notation for determining the pointer's position
    */
-  <O> O mapTo(String pointer, Class<O> type) {
+  public <O> O mapTo(String pointer, Class<O> type) {
```
ajordens: @robfletcher We have a number of callers to this ... they were still working "because groovy"

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ApplySourceServerGroupCapacityTask.groovy
```diff
@@ -49,50 +49,87 @@ class ApplySourceServerGroupCapacityTask extends AbstractServerGroupTask {
...
     try {
-      def ancestorDeployStage = getAncestorDeployStage(executionRepository, stage)
-      def ancestorDeployStageData = ancestorDeployStage.mapTo(DeployStageData)
-
-      def deployServerGroup = oortHelper.getTargetServerGroup(
-        ancestorDeployStageData.account,
-        ancestorDeployStageData.deployedServerGroupName,
-        ancestorDeployStageData.region,
-        getCloudProvider(ancestorDeployStage)
+      TargetServerGroupContext operationContext = getTargetServerGroupContext(stage)
```
ajordens: Split this out to support both standard deploy and rollback scenarios.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/RollbackServerGroupStage.groovy
```diff
@@ -91,14 +101,17 @@ class RollbackServerGroupStage implements StageDefinitionBuilder {
...
-        source : {
+        action            : ResizeStrategy.ResizeAction.scale_to_server_group.toString(),
+        source            : {
           def source = parentStage.mapTo(ResizeStrategy.Source)
           source.serverGroupName = rollbackServerGroupName
           return source
         }.call(),
-        asgName: restoreServerGroupName
+        asgName           : restoreServerGroupName,
+        pinMinimumCapacity: true
```
ajordens: `pinMinimumCapacity` is new but this particular ResizeAction (`scale_to_server_group`) is only used during a rollback.

ajordens:
FYI @spinnaker/google-reviewers 

@duftler I don't think this breaks anything but just a heads up

PR #1292, opened 2017-04-21
fix(core): reset synthetic stages when restarting a pipeline

robfletcher:


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/StageDefinitionBuilder.java
```diff
@@ -111,24 +112,35 @@ public static String getType(Class<? extends StageDefinitionBuilder> clazz) {
...
-        List<Task> childStageTasks = childStage.getTasks();
-        childStageTasks.forEach(it -> {
-          it.setStartTime(null);
-          it.setEndTime(null);
-          it.setStatus(ExecutionStatus.NOT_STARTED);
+      List<String> restartingStageAndChildren = Stream.concat(Stream.of(stage), childStages.stream()).map(Stage::getId).collect(toList());
+      List<Stage<T>> syntheticStages = stages
+        .stream()
+        .filter(it -> {
+          System.out.printf("Stage %s complete %b in list %b%n", it.getName(), it.getStatus().isComplete() , restartingStageAndChildren.contains(it.getParentStageId()));
```
jeyrschabu: Print statement intentional ?

jeyrschabu:
LGTM other than the comment above

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/StageDefinitionBuilder.java
robfletcher: nope, good spot

PR #1295, opened 2017-04-25
feat(oraclebmcs): Add Oracle BMCS server group creator

simonlord:
This adds an Oracle BMCS server group creator pipeline task. This has been tested against latest clouddriver where we've recently finished integration of the Oracle BMCS provider.


duftler:
Needs a rebase.

PR #1296, opened 2017-04-25
fix(azure): Set imageName on server groups

EricJizbaMSFT:
This was displaying as an empty string in Deck


EricJizbaMSFT:
@rguthriemsft @dcaro @clguimanmsft PTAL

EricJizbaMSFT:
@spinnaker/reviewers PTAL

PR #1298, opened 2017-04-25
Nü Orca

robfletcher:
This a new queue-based workflow engine for Orca.

The key components are:

* a `Queue` interface that allows message receipt with acknowledgement, pushing messages for immediate delivery or delivery after a delay. Messages that are not acknowledged are re-delivered after a timeout.
* a `QueueProcessor` that uses a single thread to read messages from the queue and pass them to the appropriate handler based on the type of message. Handlers do their work using a thread pool so the queue processor itself is not blocked. If the handler does not throw an exception the queue processor acknowledges the message.
* `MessageHandler` implementations responsible for dealing with small atomic units of work such as state changes at a task, stage or execution level and running tasks. The idea is these should execute very rapidly, in the order of milliseconds or seconds at most. For example when running tasks that may wait for something, a single attempt is made and if the task is not yet complete the `RunTask` message is re-queued with a delay.
* Pub/sub events raised by message handlers that listeners can subscribe to in order to respond to updates on pipeline progress. For example, notifications sent by [Echo](https://github.com/spinnaker/echo) and…
* An append-only log for each individual execution. This is at least in part due to the fact that operations on a single execution are no longer pinned to a single Orca instance so tracking activity via `catalina.out` is not practical.

The main advantages of this new implementation are:

* Work on a particular execution is no longer pinned to a single Orca instance so there is no need to drain work from old instances when updating Orca and no danger of losing in-flight work in the case of instance failure.
* The acknowledgement / re-delivery mechanism means that in the case of an Orca instance failing while handling a message the message will simply be re-delivered and handled elsewhere.
* Work for individual stages is determined "just in time" (on handling of the `StartStage` message) which theoretically allows greater flexibility in determining pipeline topology based on preceding stages.
* It's far easier to handle user-requested restart of failed pipeline stages.
* We can throttle processing of the queue in interesting ways (prioritizing running executions over new work, rate-limiting applications, etc.)
* Work is distributed across the Orca cluster and there is less risk of a single instance getting choked due to multiple long-running executions.

Things are configured so that pipelines can be individually set to use the new engine by setting `executionEngine` to `v3`. Orchestrations will run using the new engine if the config key `orchestration.executionEngine` is set to `v3`.


robfletcher:
PTAL @spinnaker/reviewers

robzienert:
Some additional context into this rather lengthy PR:

*Implementation / motivation*

This implementation, amongst many things, is primarily focused around making orca more easy to operate. Most notably is that with nü-orca enabled, orca becomes a stateless service via a message queue. This means executions will not be lost or become stuck on service restart or instance terminations. nü-orca is discovery aware and will stop processing the queue (and safely self-drain) when disabled / terminated.

Along for the ride in this PR are a few extra operational improvements which will continue be refined and added to:

* An append-only log is aimed to help tracking and debugging execution timelines across all instances. 
* Optionally, traffic shaping logic can be written and applied to queue poll operations. We expect to use this to help manage bursty/peak traffic use cases, as well as help provide capacity safeties via queue- and/or message-scoped interceptors. Early interceptor implementations will be application-level message rate limiting and an emergency stop-the-world switch.

This PR contains an in-memory queue, redis-backed queue and an SQS-backed queue. Support for other queuing mechanisms should be fairly straight forward to implement using the existing examples.

robfletcher:
It should be noted that the ultimate aim here is to remove `orca-spring-batch` and make a number of other simplifications and have all executions run using this new engine. We're doing this by incrementally cutting existing pipelines over using the `executionEngine` flag. Once we're confident we've addressed any issues there will be a further PR deleting a significant amount of code including the entire `orca-spring-batch` module.

ttomsu:
Can you elaborate on how to signal an individual instance to safely self-drain?

robfletcher:
@ttomsu When an Orca server group is disabled there is an application event we respond to and simply stop processing messages from the queue. Individual messages are things like "mark this stage complete and queue the next one" or "run this task once" so draining should be very rapid as even under the worst load they represent seconds of work. Because pipelines are broken down like this work is naturally distributed across the cluster and immediately begins to "flow" across to a newly launched server group. 

I spoke to @duftler last week about potential alternatives signals for scenarios where discovery is not used and I think he has an idea of how to proceed. Take a look at the `DiscoveryEnabled` interface and the `ExecutionWorker` class that implements it and is the thing that pops messages from the queue.

duftler:
@ttomsu One idea we discussed was having each orca node query clouddriver to determine if it is in a disabled state (as an alternative to assuming Eureka is present).

Keep in mind that the new execution engine has to be explicitly enabled for a given pipeline. So we have to make sure we solve this before enabling it for any pipelines; but nothing should break in the meantime.

PR #1299, opened 2017-04-26
feat(provider/kubernetes): Enable deployment of baked docker containers

brujoand:
This aims to enable the deployment of baked docker containers by making two changes:
- The bake stages `refId` is added to the `deploymentDetails`
- The parsed docker image name is added to the `imageDetails`

This PR is a requirement for https://github.com/spinnaker/deck/pull/3588 to work and these two PRs should therefor be seen in unison. Like I mentioned in the Deck PR, I'm open to corrections on this approach, as I've merely made this work, and there might be a different preferred path.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
@@ -118,11 +118,12 @@ class BakeStage implements BranchingStageDefinitionBuilder, RestartableStage {
           it.type == PIPELINE_CONFIG_TYPE && bakeInitializationStages*.id.contains(it.parentStageId) && (it.context.ami || it.context.imageId)
         }.collect { Stage bakeStage ->
           def deploymentDetails = [:]
-          ["ami", "imageId", "amiSuffix", "baseLabel", "baseOs", "storeType", "vmType", "region", "package", "cloudProviderType", "cloudProvider"].each {
+          ["ami", "imageId", "amiSuffix", "baseLabel", "baseOs", "refId", "storeType", "vmType", "region", "package", "cloudProviderType", "cloudProvider"].each {
             if (bakeStage.context.containsKey(it)) {
               deploymentDetails.put(it, bakeStage.context.get(it))
             }
           }
+
           return deploymentDetails
         }
       ]
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesContainerFinder.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesContainerFinder.groovy
@@ -20,6 +20,37 @@ import com.netflix.spinnaker.orca.pipeline.model.Pipeline
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 
 class KubernetesContainerFinder {
+  static Map parseContainerPartsFrom(String containerName) {
+    String tag = ""
+    String registry = ""
+
+    String[] parts = containerName.split('/')
+
+    String head = parts.first()
+
+    // Get the registry
+    if (head.contains('.') || head.startsWith('localhost')) {
+      registry = head
+      parts = parts.drop(1)
+    } else {
+      throw new IllegalStateException("Could not parse a registry from the provided docker container reference")
+    }
+
+    // Get the tag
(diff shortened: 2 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesContainerFinder.groovy
```diff
@@ -20,6 +20,33 @@ import com.netflix.spinnaker.orca.pipeline.model.Pipeline
...
+  static Map parseContainerPartsFrom(String containerName) {
+    String tag = ""
+    String registry = ""
+
+    String[] parts = containerName.split('/')
+
+    String head = parts.first()
+
+    // Get the registry
+    if (head.contains('.') || head.startsWith('localhost')) {
```
lwander: If this condition isn't met - what's the registry? I think k8s assumes gcr.io

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -118,11 +118,12 @@ class BakeStage implements BranchingStageDefinitionBuilder, RestartableStage {
           it.type == PIPELINE_CONFIG_TYPE && bakeInitializationStages*.id.contains(it.parentStageId) && (it.context.ami || it.context.imageId)
         }.collect { Stage bakeStage ->
           def deploymentDetails = [:]
-          ["ami", "imageId", "amiSuffix", "baseLabel", "baseOs", "storeType", "vmType", "region", "package", "cloudProviderType", "cloudProvider"].each {
+          ["ami", "imageId", "amiSuffix", "baseLabel", "baseOs", "refId", "storeType", "vmType", "region", "package", "cloudProviderType", "cloudProvider"].each {
```
lwander: Nice, does this play into @gardleopard's artifact refactoring as well?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesContainerFinder.groovy
```diff
@@ -20,6 +20,33 @@ import com.netflix.spinnaker.orca.pipeline.model.Pipeline
...
+      registry = head
+      parts = parts.drop(1)
+    }
+
+    // Get the tag
+    if (parts.last().contains(':')) {
```
lwander: If this condition isn't met, should we default to :latest?

lwander:
Looks good, just a few questions

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesContainerFinder.groovy
brujoand: Good question. My thoughts were that if no registry is defined the k8s kluster should chose a default. 

In retrospect however I'm entirely sure. I think that assuming some default here is hard, so we would either have to use a registry defined for this account or fail if we don't allow this to be empty. Wdyt?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesContainerFinder.groovy
brujoand: Yes, that sounds good to me

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesContainerFinder.groovy
lwander: Failing seems like a good idea, it's best to be explicit

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
brujoand: No this PR has been forgotten in the back of my machine for a couple of weeks, so I haven't taken the Artefact decoration into account.

brujoand:
Thanks, I've addressed the comments and gotten back in sync with master.

PR #1306, opened 2017-04-28
fix(queue): limit re-deliveries of failed messages

robfletcher:
Once a failed message has been tried 10 times it gets passed to a handler that will terminate the associated pipeline

### DIFF
--- a/orca-queue-sqs/src/main/kotlin/com/netflix/spinnaker/config/SqsQueueConfiguration.kt
+++ b/orca-queue-sqs/src/main/kotlin/com/netflix/spinnaker/config/SqsQueueConfiguration.kt
@@ -19,12 +19,14 @@ import com.amazonaws.auth.AWSCredentialsProvider
 import com.amazonaws.services.sqs.AmazonSQS
 import com.amazonaws.services.sqs.AmazonSQSClient
 import com.netflix.spinnaker.clouddriver.aws.bastion.BastionConfig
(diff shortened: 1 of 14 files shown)


robfletcher:
PTAL @spinnaker/reviewers

orca-queue-sqs/src/main/kotlin/com/netflix/spinnaker/config/SqsQueueConfiguration.kt
```diff
@@ -33,14 +35,21 @@ import org.springframework.context.annotation.*
...
 
-  @Bean
-  open fun amazonSqsClient(awsCredentialsProvider: AWSCredentialsProvider, sqsProperties: SqsProperties): AmazonSQS {
-    return AmazonSQSClient.builder()
+  @Bean open fun amazonSqsClient(
+    awsCredentialsProvider: AWSCredentialsProvider,
+    sqsProperties: SqsProperties
+  ) =
+    AmazonSQSClient
+      .builder()
```
robzienert: Is this formatting an IntelliJ setting?

robzienert:
👍

orca-queue-sqs/src/main/kotlin/com/netflix/spinnaker/config/SqsQueueConfiguration.kt
robfletcher: It's me being a prick

PR #1307, opened 2017-04-28
fix(queue): don't just assume task ids are valid

robfletcher:
Because of concurrency issues in `JedisExecutionRepository` it's pretty easy to end up with invalid task ids. Obviously, that needs fixing but it did expose that they weren't getting handled well.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/Executions.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/Executions.kt
@@ -81,13 +81,6 @@ fun Stage<out Execution<*>>.nextTask(task: Task) =
   }
 
 /**
- * @return the task with the specified id.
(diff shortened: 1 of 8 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/Executions.kt
```diff
@@ -85,7 +85,7 @@ fun Stage<out Execution<*>>.nextTask(task: Task) =
  * @throws IllegalArgumentException if there is no such task.
  */
 fun Stage<out Execution<*>>.task(taskId: String) =
-  getTasks().find { it.id == taskId } ?: throw IllegalArgumentException("No such task")
+  getTasks().find { it.id == taskId }
```
robfletcher: Just by removing this I'm then forced by the compiler to handle `null` everywhere this method is called. Kotlin is awesome.

PR #1311, opened 2017-04-29
fix(queue): add synthetic stages in a concurrency-safe way

robfletcher:
instead of adding stages into the pipeline and then saving the entire thing `StartStageHandler` now uses a new method on `ExecutionRepository` to add an individual stage. This should be concurrency safe so if two `StartStage` messages for the same pipeline are processed simultaneously they won't overwrite each other's modifications.


orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepositoryTck.groovy
```diff
@@ -700,4 +705,78 @@ class JedisExecutionRepositorySpec extends ExecutionRepositoryTck<JedisExecution
...
+    def startLatch = new CountDownLatch(1)
+    def doneLatch = new CountDownLatch(3)
+    [stage1, stage2, stage3].each { stage ->
+      Thread.start {
+        startLatch.await(1, SECONDS)
+        repository.addStage(stage)
+        doneLatch.countDown()
+      }
+    }
+    startLatch.countDown()
```
robfletcher: This should prove we can add multiple stages to the same pipeline simultaneously without issues

PR #1316, opened 2017-05-02
feat(pipelinetemplate): Add support for list injection for Before and After stanzas

jeyrschabu:
- Before and After injections now support a list of target stage Ids

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
@@ -31,6 +31,8 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
+import java.util.SortedSet;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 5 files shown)


jeyrschabu:
@robzienert PTAL whenever you get back :)

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
```diff
@@ -117,11 +119,22 @@ private static void dfs(String stageId,
 
     state.put(stageId, Status.VISITING);
     StageDefinition stage = graph.get(stageId);
+    if (stage == null) {
```
robzienert: This happens? Is it worthwhile logging this?

robzienert:
👍 Looks right to me, one minor question - but can be merged without after rebase.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
jeyrschabu: You know what ? I think that's me being lazy on rebuilding the graph after it has been mutated. I think I can get rid of that null check and handle rebuilding the graph. Keeps the dfs method clean that way

PR #1319, opened 2017-05-04
fix(queue): canaries were not attaching their deploy stages

robfletcher:


orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/pipeline/DeployCanaryStage.groovy
```diff
@@ -152,7 +152,7 @@ class DeployCanaryStage extends ParallelDeployStage implements CloudProviderAwar
       def context = stage.context
       def allStages = stage.execution.stages
       def deployStages = allStages.findAll {
-        it.parentStageId == stage.id && it.type == ParallelDeployStage.PIPELINE_CONFIG_TYPE
+        it.parentStageId == stage.id
```
robfletcher: The only child stage of a deploy canary is the deploy stage so the qualification is unnecessary and breaks nü-orca where the type will be `createServerGroup`.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/StageDefinitionBuilders.kt
```diff
@@ -32,7 +32,7 @@ import com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner.STAGE_BEFOR
  */
 fun StageDefinitionBuilder.buildTasks(stage: Stage<*>) {
   val taskGraph =
-    if (this is BranchingStageDefinitionBuilder && stage.getParentStageId() == null) {
+    if (this is BranchingStageDefinitionBuilder && (stage.getParentStageId() == null || stage.parent().getType() != stage.getType())) {
```
robfletcher: Canaries _do_ have parent stages but are also parallel so I had to modify this condition. If the parent type is the same then we're in a situation like bake where the stage does double duty as a parallel fork _and_ the branches that fork creates.

cfieber:
👍

PR #1331, opened 2017-05-09
fix(upsert_lb_task): Fix up context name defaulting.

jtk54:
Prior to this PR, defaulting to `clusterName` without null-checking would cause the context name to default to 'null-frontend' in some cases when Orca asked Clouddriver for a cache refresh, which would cause failures that were subtle and hard to debug without `FULL` debug levels on Orca and Clouddriver. This way is safer and clearer in case of defaulting failures.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/loadbalancer/UpsertLoadBalancerTask.groovy
```diff
@@ -53,7 +53,14 @@ class UpsertLoadBalancerTask extends AbstractCloudProviderAwareTask implements R
     String account = getCredentials(stage)
 
     def context = new HashMap(stage.context)
-    context.name = context.name ?: "${stage.context.clusterName}-frontend"
+    if (!context.name && context.clusterName) {
```
duftler: Looks like `!context.name` can be factored out to an enclosing check.

PR #1334, opened 2017-05-11
feat(dcd) support setting templates from trigger context using spel

gardleopard:
This enables us to use spel to override what template source to use. This means you can upload a template in your ci build and use that one in the pipeline execution. This means we can have pipeline definitions following the same flow as other code and still be able to rollback easily.

```
{
  "appConfig": {},
  "config": {
    "pipeline": {
      "application": "gardtest",
      "name": "My wait pipeline",
      "pipelineConfigId": "eebaa78d-ecc2-4638-86aa-1afa9defebec",
      "template": {
        "source": "file:///opt/orca/templates/${trigger.parameters.template}"
      },
      "variables": {}
    },
    "schema": "1"
  },
  "executionEngine": "v2",
  "lastModifiedBy": "gard.rimestad@schibsted.com",
  "limitConcurrent": true,
  "parallel": true,
  "parameterConfig": [
    {
      "default": "test.yml",
      "name": "template"
    }
  ],
  "stages": [],
  "type": "templatedPipeline",
  "updateTs": "1494415568000"
}

```

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessor.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessor.java
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.orca.pipelinetemplate;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.orca.extensionpoint.pipeline.PipelinePreprocessor;
+import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor;
 import com.netflix.spinnaker.orca.pipelinetemplate.exceptions.IllegalTemplateConfigurationException;
 import com.netflix.spinnaker.orca.pipelinetemplate.exceptions.TemplateLoaderException;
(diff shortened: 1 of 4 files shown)


gardleopard:
Ok, makes sense, I will give jinja a go then.

robzienert:
I see how you can be confused. For the sake of keeping domains separated in my mind, I've been using terms "compile time" and "runtime", although that's not strictly speaking valid. Compile time encapsulates the entirety of the `PipelineTemplatePipelinePreprocessor`, which takes a non-valid pipeline JSON structure and produces a valid pipeline as output for execution (runtime).

For anything done in `PipelineTemplatePipelinePreprocessor`, we should be using Jinja. Running SpEL inside of the Preprocessor may (probably will) cause unexpected errors. A contrived example:

```yaml
pipeline:
  template:
    source: spinnaker://${trigger.parameters.template}
stages:
- id: wait
  config:
    waitTime: "${trigger.parameters.waitTime}"
```

The expression defined in the stage should be evaluated after the generated pipeline has been passed into the rest of Orca for execution, since SpEL has access to the greater execution context. Using Jinja will ensure we can do templating without worrying about accidentally rendering something meant for execution-time.

Does this make sense?

gardleopard:
@robzienert Thanks a lot for explaining this to me. I have created a PR https://github.com/spinnaker/orca/pull/1343 to do this with jinja. I have explicitly done this only to modify the source field because the others will be resolved later in the chain I assume.

PR #1335, opened 2017-05-11
fix(expressions): further restrict object traversal in expression processing

cfieber:
the existing restrictions weren't being applied when the object graph was traversed using property notation

additionally this change removes all the types from the expression root object (it will just be a JSON object of Map<String, Object> where as before it was actually a reference to the execution object

finally this converts ExecutionStatus from groovy to java which had a minor ripple into some of the kotlin code (@robfletcher might have a suggestion on naming the property acccessor differently on ExecutionStatus?)

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/ExecutionStatus.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/ExecutionStatus.java
@@ -14,12 +14,13 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.orca
+package com.netflix.spinnaker.orca;
 
-import groovy.transform.CompileStatic
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 
-@CompileStatic
-enum ExecutionStatus {
+public enum ExecutionStatus {
   /**
    * The task has yet to start.
    */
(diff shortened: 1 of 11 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/ExecutionStatus.java
```diff
@@ -79,31 +80,40 @@
...
 
   /**
    * Indicates an abnormal completion so nothing downstream should run afterward.
    */
-  final boolean halt
+  public final boolean isHalt() {
+    return halt;
+  }
+
+  private static final Collection<ExecutionStatus> SUCCESSFUL = Collections.unmodifiableList(Arrays.asList(SUCCEEDED, STOPPED, SKIPPED));
```
robfletcher: Shouldn't this also include `FAILED_CONTINUE`?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CancelStageHandler.kt
```diff
@@ -40,7 +40,7 @@ open class CancelStageHandler
 
   override fun handle(message: CancelStage) {
     message.withStage { stage ->
-      if (stage.getStatus().halt) {
+      if (stage.getStatus().isHalt) {
```
robfletcher: I've run into this before with Kotlin and I'm confused why it does this but I don't think there's a way round it without violating expected naming conventions for these kind of methods in Java.

tomaslin:
I am worried about this exploit also working on fields in the stage context since that is touched by a whole swarth of groovy classes as well. Does this fix cast all the stage context values into a map as well?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/ExecutionStatus.java
cfieber: possibly, probably?

I just groovy->java'd this so I wasn't really looking at the semantics of this class

I don't really want to add that change into this PR since it feels unrelated

cfieber:
@tomaslin This bit [here](https://github.com/spinnaker/orca/pull/1335/files#diff-94611e064f852dac5ca04357f3576f20R61) untypifies the whole root object by serializing it out to JSON then reading it back in as Map<String, Object> which will lose all the previous types

tomaslin:
LGTM

PR #1336, opened 2017-05-11
feat(pipeline): endpoint for evaluating pipeline expressions

danielpeach:
@duftler please review.


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.controllers
 
+import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor
```
duftler: Sort order is off here.

PR #1341, opened 2017-05-15
A couple of test run fixes

robfletcher:
@duftler this will fix the problem you reported with `SpringIntegrationTest` not running with Gradle.


robfletcher:
The integration test seems to be unreliable on Travis. Looking into it.

PR #1342, opened 2017-05-16
feat(pipelinetemplate): Pipeline Template Partials support

robzienert:
Implements the common cases for [Template Partials](https://github.com/spinnaker/dcd-spec/blob/master/PIPELINE_TEMPLATES.md#partials). There's some cleanup necessary (some fairly rough code in here), as well as adding support for recursive partials before this can go in.

I also departed from the spec in some ways that I'll need to go back and update. Most notably, instead of adding a `ref` field into `StageDefinition`, I rolled it up into the `type`:

```yaml
stages:
- id: foo
  type: partial:myPartial
partials:
- id: myPartial
```

@jeyrschabu PTAL

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
@@ -17,6 +17,7 @@ package com.netflix.spinnaker.orca.pipelinetemplate.v1schema.graph.transform;
 
 import com.netflix.spinnaker.orca.pipelinetemplate.exceptions.IllegalTemplateConfigurationException;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.PipelineTemplateVisitor;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.PartialDefinition;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.PipelineTemplate;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.StageDefinition;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.StageDefinition.InjectionRule;
@@ -55,6 +56,7 @@ public class ConfigStageInjectionTransform implements PipelineTemplateVisitor {
 
   @Override
   public void visitPipelineTemplate(PipelineTemplate pipelineTemplate) {
+    expandStagePartials(pipelineTemplate);
     replaceStages(pipelineTemplate);
     injectStages(pipelineTemplate);
     trimConditionals(pipelineTemplate);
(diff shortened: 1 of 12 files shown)


robzienert:
Travis not picking up the build.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
```diff
@@ -72,6 +74,63 @@ private void replaceStages(PipelineTemplate pipelineTemplate) {
...
+      graph.keySet().forEach(k -> outOrder.put(k, 0));
+      graph.forEach((k, v) -> dfs(k, sorted, state, graph, outOrder));
+      Set<String> leafNodes = sorted
+        .stream()
+        .filter(i -> outOrder.get(i.getId()) == 0)
+        .map(StageDefinition::getId)
+        .collect(Collectors.toSet());
+
+      // assign root nodes to placeholder's dependsOn value
+      stages.stream()
```
jeyrschabu: This logic seems a bit suspect

jeyrschabu: Oh I see what you are doing. Just doesn't look obvious at first glance. Maybe an opportunity to make it readable.
IIRC, you are extracting all partial stages, then find where they fit in in the overall stage graph ?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/StageDefinition.java
```diff
@@ -17,29 +17,31 @@
...
   private String comments;
-  private List<String> when;
+  private List<String> when = new ArrayList<>();
   private InheritanceControl inheritanceControl;
   private Boolean removed = false;
 
   private Set<String> requisiteStageRefIds = new LinkedHashSet<>();
 
-  public static class InjectionRule {
+  public static class InjectionRule implements Cloneable {
```
jeyrschabu: Guessing this is because you want to be able to deep copy the stage definition

jeyrschabu:
A couple of questions around the injection bits.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/StageDefinition.java
robzienert: That's correct.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/ConfigStageInjectionTransform.java
robzienert: That's correct. I think I'll break this method up a little bit so it's more descriptive of what's going, but essentially:

`stage1 -> partial(stage3, stage4) -> stage2`

needs to expand to:

`stage1 -> stage3 -> stage4 -> stage2`

So, I build the graph, then assign stage3's dependsOn to the value of `partial1`, then stage2's dependsOn to the leaf nodes of the partial's internal graph. Once I've done those operations, I add the partial's stages into the template and remove the `partial1` stage from the graph. Make sense?

PR #1343, opened 2017-05-16
feat(dcd): support setting templates from trigger context using jinja

gardleopard:
This enables us to use jinja to override what template source to use. This means you can upload a template in your ci build and use that one in the pipeline execution. This means we can have pipeline definitions following the same flow as other code and still be able to rollback easily.
```
{
  "appConfig": {},
  "config": {
    "pipeline": {
      "application": "gardtest",
      "name": "My wait pipeline",
      "pipelineConfigId": "eebaa78d-ecc2-4638-86aa-1afa9defebec",
      "template": {
        "source": "file:///opt/orca/templates/{{trigger.parameters.template}}"
      },
      "variables": {}
    },
    "schema": "1"
  },
  "executionEngine": "v2",
  "lastModifiedBy": "gard.rimestad@schibsted.com",
  "limitConcurrent": true,
  "parallel": true,
  "parameterConfig": [
    {
      "default": "test.yml",
      "name": "template"
    }
  ],
  "stages": [],
  "type": "templatedPipeline",
  "updateTs": "1494415568000"
}
```


robzienert:
👍 Thanks!

PR #1354, opened 2017-05-23
fix(expressions): control allowed types in object traversal

cfieber:
Adds restriction around allowed types in property notation navigation in expressions.

This is another take on #1335 - turned out that one broke too many cases and stripping the typing out of the expression processing context was causing other problems


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
```diff
@@ -177,8 +184,8 @@ class ContextParameterProcessor {
     }
   }
 
-  static class WhitelistTypeLocator implements TypeLocator {
-    private final Set<Class<?>> allowedTypes = Collections.unmodifiableSet([
+  static class TypeWhitelist {
```
cfieber: now differentiated between what can be instantiated (allowedTypes) and what return types are allowed (allowedReturnTypes)

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
```diff
@@ -232,10 +275,43 @@ class ContextParameterProcessor {
...
+  static class FilteredPropertyAccessor extends ReflectivePropertyAccessor {
+    @Override
+    protected Method findGetterForProperty(String propertyName, Class<?> clazz, boolean mustBeStatic) {
+      Method getter = super.findGetterForProperty(propertyName, clazz, mustBeStatic)
+      if (getter && TypeWhitelist.isAllowedForReturn(getter.returnType)) {
+        return getter
+      }
+
+      if (getter) {
+        log.info("found getter for requested $propertyName but rejected due to return type $getter.returnType")
```
cfieber: if this turns out to be too verbose it could get lowered to debug but may be useful in helping people troubleshoot if they have crafted invalid expressions that are failing in a non obvious way

PR #1364, opened 2017-05-25
Fix occasional stalled pipelines

robfletcher:
This aims to prevent 2 scenarios:

1. two parallel stages complete essentially at the same time and CompleteStageHandler thinks the other one is still running so it no-ops.
2. two branches complete at the same time and CompleteExecutionHandler makes the same mistake.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/Message.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/Message.kt
@@ -133,6 +133,16 @@ data class StartStage(
     this(source.getExecution().javaClass, source.getExecution().getId(), source.getExecution().getApplication(), source.getId())
 }
 
+data class ContinueParentStage(
+  override val executionType: Class<out Execution<*>>,
+  override val executionId: String,
+  override val application: String,
+  override val stageId: String
+) : Message(), StageLevel {
+  constructor(source: Stage<*>) :
+    this(source.getExecution().javaClass, source.getExecution().getId(), source.getExecution().getApplication(), source.getId())
+}
+
 data class CompleteStage(
   override val executionType: Class<out Execution<*>>,
   override val executionId: String,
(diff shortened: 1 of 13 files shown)


robfletcher:
@cfieber PTAL, fix for this morning's issue with parallel bakes

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/Message.kt
```diff
@@ -130,6 +133,16 @@ data class StartStage(
     this(source.getExecution().javaClass, source.getExecution().getId(), source.getExecution().getApplication(), source.getId())
 }
 
+data class ContinueParentStage(
```
robfletcher: Introduced a new message triggered when a BEFORE stage with no downstream BEFORE stages completes (can also be parallel stages)

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ContinueParentStageHandler.kt
```diff
@@ -0,0 +1,80 @@
...
+import com.netflix.spinnaker.orca.q.*
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+import java.time.Duration
+
+@Component
+open class ContinueParentStageHandler
```
robfletcher: New message is handled and will result in parent stage tasks firing

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ContinueParentStageHandler.kt
```diff
@@ -0,0 +1,80 @@
...
+        if (stage.hasTasks()) {
+          stage.runFirstTask()
+        } else if (stage.hasAfterStages()) {
+          stage.runAfterStages()
+        } else {
+          queue.push(CompleteStage(stage, SUCCEEDED))
+        }
+      } else {
+        log.warn("Re-queuing $message as other BEFORE stages are still running")
+        queue.push(message, retryDelay)
```
robfletcher: If all BEFORE stage branches (or parallel branches) don't appear to be complete it will try again later

PR #1372, opened 2017-05-30
fix(queue): allow tasks to refer to global context keys seamlessly

robfletcher:


robfletcher:
@tomaslin this fixes the spindemotitus pipeline whose deploy task was unable to find `deploymentDetails` because it wasn't reading thru from global context.

PR #1380, opened 2017-06-02
fix(expressions): fixes an issue with toJson on expressions.

cfieber:
calling toJson on an object with expressions in it (execution, stage, stage.context) returns
a String containing an expression which is then reevaluated. This string typically contains
the orignal toJson expression and the result is infinitely evaluating the expression and hanging
the execution

this rejects a toJson call that returns a String that still contains an expression

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
@@ -48,7 +48,7 @@ import org.springframework.expression.spel.support.StandardTypeLocator
 class ContextParameterProcessor {
 
   // uses $ instead of  #
-  private final ParserContext parserContext = new TemplateParserContext('${', '}')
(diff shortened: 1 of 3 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
```diff
@@ -329,12 +329,18 @@ abstract class ContextUtilities {
...
+
   static String alphanumerical(String str) {
     str.replaceAll('[^A-Za-z0-9]', '')
   }
 
   static String toJson(Object o) {
-    new ObjectMapper().writeValueAsString(o)
+    String converted = mapper.writeValueAsString(o)
+    if (ContextParameterProcessor.containsExpression(converted)) {
+      return null
```
robfletcher: does this risk an NPE anywhere?

cfieber: the expression processing handles null (by returning the source string)

PR #1383, opened 2017-06-03
feat(queue): Redis-backed throttles

robzienert:
Adding a RedisRateLimiterBackend, and updating the traffic shaping queue
to support global message capacity rate limits, as well as app-specific.

For now I've just implemented a similar rate limiting backend to what we have in gate, except nu-orca's implementation does not offer configuration of the window size (hardcoded at 1000ms). 

Added a global rate limiter should help in broad-strokes capacity management. Maybe a stage type limiter could be handy (e.g. "allow only 100 simultaneous deploys").


orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/TrafficShapingQueueSpec.kt
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  beforeGroup {
+    reset(queueImplCallback)
+    subject.push(StartStage(Pipeline::class.java, "1", "foo", "1"))
+  }
+
+  describe("when no interceptors are triggered") {
+    whenever(interceptor.interceptPoll()) doReturn false
+    whenever(interceptor.interceptMessage(any())).doReturn(null as TrafficShapingInterceptorCallback?)
```
robfletcher: this needs to be done inside a `beforeGroup`. Otherwise it's evaluated when Spek runs through and builds the tree—before any of the tests start running. It's okay to set constants up like that but not something with mutable state like a mock.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/TrafficShapingQueueSpec.kt
```diff
@@ -0,0 +1,85 @@
...
+  val registry: Registry = NoopRegistry()
+  val interceptor: TrafficShapingInterceptor = mock()
+
+  subject(CachingMode.GROUP) {
+    whenever(interceptor.supports(any())) doReturn true
+    TrafficShapingQueue(queueImpl, registry, listOf(interceptor))
+  }
+
+  beforeGroup {
+    reset(queueImplCallback)
```
robfletcher: I think this should probably be done in an `afterGroup` inside of each `describe`. The way you have it now you'll reset the mocks once before _all_ the tests run.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/TrafficShapingQueueSpec.kt
robzienert: Ah, I did not understand it to work that way - this makes things a lot clearer. Thanks.

PR #1390, opened 2017-06-07
fix(orca/find-mutiple-images) : fixed issue to run parallel multiple …

jatinbajaj1:
…find images to run with Deploy job stage. 

With Reference to issue no: 
  https://github.com/spinnaker/spinnaker/issues/1724

Find images and deploy stage , if run in parallel with more than one time results in failure , as   parallel find images works in thread , and the last thread save the common context. 

We introduced a new variable in stage   i.e.  parallelContext which is Map of Maps. 
so now result would be stored like this: 
[stageid1 : [DeployementDetails] ,  stageid2 : [DeploymentDetails]]

and used the above variable to  fetch  the deployement details based on the  stage id. 
Above changes are made only for kubernetes cloud-provider. and for Find images and deploy stages only.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesContainerFinder.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesContainerFinder.groovy
@@ -63,6 +63,15 @@ class KubernetesContainerFinder {
 
     containers.forEach { container ->
       if (container.imageDescription.fromContext) {
+	//  If there are parallel Find images stages , then deploymentDetails would be fetched from it. 
+	// e.g.  parallelContext = [stageid:[deployDetails],stageId2:[deployDetails]]
+	def paraContext = (stage.parallelContext ? stage.parallelContext.find {it.key == container.imageDescription.stageId}?.value : null)
+
+	if (paraContext) {
+           stage.context.deploymentDetails = paraContext.deploymentDetails
+           deploymentDetails = (paraContext.deploymentDetails ?: []) as List<Map>
+        }
+
         def image = deploymentDetails.find {
           // stageId is used here to match the source of the image to the find image stage specified by the user.
           // Previously, this was done by matching the pattern used to the pattern selected in the deploy stage, but
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
@@ -47,6 +47,15 @@ public class Stage<T extends Execution<T>> implements Serializable {
     this(execution, type, null, context);
   }
 
+  @SuppressWarnings("unchecked")
+  public Stage(T execution, String type, String name, Map<String, Object> context, Map<String,  Map<String, Object>> parallelContext) {
+    this.execution = execution;
+    this.type = type;
+    this.name = name;
+    this.parallelContext = parallelContext;
+    this.context = context;
+  }
+
   public Stage(T execution, String type) {
     this(execution, type, emptyMap());
   }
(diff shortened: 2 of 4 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesContainerFinder.groovy
```diff
@@ -63,6 +63,15 @@ class KubernetesContainerFinder {
 
     containers.forEach { container ->
       if (container.imageDescription.fromContext) {
+	//  If there are parallel Find images stages , then deploymentDetails would be fetched from it.
```
ajordens: Can you not just look upstream to all parent/ancestor `Find Image` stages until you find the one that found the image you were looking for?

I do not believe adding another style of context is the correct way to address this.

jatinbajaj1: @ajordens  -   Actually  in the current scenario, we are having the single context variable , when we have the parallel Find images  ,   running in two separate threads , the last thread just updates or overwrite the existing result   , so that was the reason to introduce new variable.
 So looking upstream to all parent/ancestor  Find image would only give the  result output saved by the last thread.
 
Can you suggest/guide us  if we can modify the existing context  in a way  that threads should not overwrite the result of find images . 

Regards,
Jatin

ajordens: Can you include a screenshot of your pipeline?

Each `FindImage` stage should write to it's local context  as well as the global context. I could see the global context getting overridden but IMO there is likely value in moving away from that where possible.

For example the `Find Image` stage should have `amiDetails` in it's immediate stage context.

So it's possible we still need a fix here but it'd be nice if we err'd on the side of looking at individual stage contexts in addition to the global context.

cc @robfletcher @duftler

jatinbajaj1: @ajordens : 

Pipeline i used:  two parallel find images with  deploy as intermediate stage.
image

here pattern is     .* 

Find image save the stage context using  ExecutionContextManager
and it is having 
 static class DelegatingHashMap implements Map<String, Object> 


and find image write to local , as well as global context. 
using these lines 

// fetch the current stage (w/ global context merged in)
        stage = currentStage(chunkContext, true)

I would investigate  further according to the above comments  to look the area where global context is overridden. 
Regards,
Jatin bajaj

jatinbajaj1: @ajordens   

Hi , after debugging this one , Seems like  Multiple threads are competing with each other to write to a global context  and yes for every Find stage you  have the amiDetails for each stage context  which is written to global context based on race condition winner of two threads (or multiple) .   so without the support  of different variable , how we can populate the global context where second thread has already overwritten the context. 
Please can you  guide me for the above 


cc @robfletcher  @duftler

ajordens: I believe the correct fix for this is to support (or favor) using the non-global context when looking for `amiDetails`. 

In the short-term, I suspect you can work around this by adding a small wait stage (~30s) to one of your branches. That should add enough buffer for the first deploy to start before the second `Find Image` overwrites the global context.

emptywee: Ran into this today too. Figured out the workaround with a wait stage myself and then bumped into this PR. This needs to be fixed, IMHO.

robzienert:
Closing - if you'd like to continue resolving this issue, feel free to re-open and address the comments.

PR #1392, opened 2017-06-08
fix(expressions): retry within fromUrl expression helper

jeyrschabu:
- Added a fixed retry logic in fromUrl

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
@@ -325,6 +325,7 @@ class ContextFunctionConfiguration {
   }
 }
 
+@Slf4j
 abstract class ContextUtilities {
 
   @PackageScope static final AtomicReference<ContextFunctionConfiguration> contextFunctionConfiguration = new AtomicReference<>(new ContextFunctionConfiguration())
@@ -350,8 +351,13 @@ abstract class ContextUtilities {
   }
 
   static String fromUrl(String url) {
-    URL u = contextFunctionConfiguration.get().urlRestrictions.validateURI(url).toURL()
-    return u.getText()
+    try {
+      URL u = contextFunctionConfiguration.get().urlRestrictions.validateURI(url).toURL()
+      return HttpClientUtils.httpGetAsString(u.toString())
+    } catch (Exception e) {
+      log.error("Failed getting content from url {}", url, e)
+      throw e
+    }
   }
 
   static Object readJson(String text) {
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.pipeline.util;
+
+import com.google.common.io.CharStreams;
+import com.google.common.util.concurrent.Uninterruptibles;
+import org.apache.http.HttpResponse;
+import org.apache.http.HttpStatus;
+import org.apache.http.client.ServiceUnavailableRetryStrategy;
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.protocol.HttpContext;
+import org.apache.http.protocol.HttpCoreContext;
+import org.slf4j.Logger;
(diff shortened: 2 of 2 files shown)


jeyrschabu:
Retrying fromUrl on HTTP codes: 429, 500 & 503

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -0,0 +1,101 @@
...
+  private static List<Integer> RETRYABLE_500_HTTP_STATUS_CODES =Arrays.asList(
+      HttpStatus.SC_SERVICE_UNAVAILABLE,
+      HttpStatus.SC_INTERNAL_SERVER_ERROR
+  );
+
+  private static CloseableHttpClient httpClient = httpClientWithServiceUnavailableRetryStrategy();
+
+  private static CloseableHttpClient httpClientWithServiceUnavailableRetryStrategy() {
+    HttpClientBuilder httpClientBuilder =  HttpClients.custom()
+      .setServiceUnavailableRetryStrategy(new ServiceUnavailableRetryStrategy() {
```
cfieber: should their also be a retry on network errors?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -0,0 +1,101 @@
...
+        .setConnectionRequestTimeout(TIMEOUT_MILLIS)
+        .setConnectTimeout(TIMEOUT_MILLIS)
+        .build()
+    );
+
+    return httpClientBuilder.build();
+  }
+
+  static String httpGetAsString(String url) throws IOException {
+    try (CloseableHttpClient client = httpClient) {
```
cfieber: this is going to call close on the client after the first fromUrl happens - since the client is static does that leave it in an unusable state for any other callers?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
```diff
@@ -351,7 +352,14 @@ abstract class ContextUtilities {
 
   static String fromUrl(String url) {
     URL u = contextFunctionConfiguration.get().urlRestrictions.validateURI(url).toURL()
-    return u.getText()
+    String response = ""
```
cfieber: we typically throw an exception when one of these methods fails which ultimately results in the expression remaining as is in the context

in this case I _think_ I'm ok with empty stringing out after we've retried sufficient times otherwise we may continue to try to hit that URL

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
jeyrschabu: Good catch! the pool would be closed. Fix coming up

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
jeyrschabu: This is now retrying on 500 & 503 for server errors. Should we handle some other 5xx codes?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
jeyrschabu: I actually like the idea of throwing an Exception here.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
```diff
@@ -350,8 +351,13 @@ abstract class ContextUtilities {
   }
 
   static String fromUrl(String url) {
-    URL u = contextFunctionConfiguration.get().urlRestrictions.validateURI(url).toURL()
-    return u.getText()
+    try {
+      URL u = contextFunctionConfiguration.get().urlRestrictions.validateURI(url).toURL()
+      return HttpClientUtils.httpGetAsString(u.toString())
+    } catch (Exception e) {
```
jeyrschabu: Opted to throw an exception here

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -0,0 +1,98 @@
...
+        .setConnectionRequestTimeout(TIMEOUT_MILLIS)
+        .setConnectTimeout(TIMEOUT_MILLIS)
+        .build()
+    );
+
+    return httpClientBuilder.build();
+  }
+
+  static String httpGetAsString(String url) throws IOException {
+    try(CloseableHttpResponse response = httpClient.execute(new HttpGet(url))) {
```
jeyrschabu: Reusing the same client now.

jeyrschabu:
@cfieber PTAL

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
cfieber: Is there a retry strategy for IOExceptions we could also add (where we didn't get an HTTP response back to determine the code)

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
cfieber: 👍

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -0,0 +1,111 @@
...
+        .setConnectionRequestTimeout(TIMEOUT_MILLIS)
+        .setConnectTimeout(TIMEOUT_MILLIS)
+        .setSocketTimeout(TIMEOUT_MILLIS)
+        .build()
+    );
+
+    return httpClientBuilder.build();
+  }
+
+  static String httpGetAsString(String url) throws Exception {
```
jeyrschabu: Added a retry logic when we get an IO exception and didnt get a response code

PR #1393, opened 2017-06-08
improved read lock mechanism

robfletcher:


robzienert:
👍 I like the part where it's better.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
```diff
@@ -197,27 +196,22 @@ class RedisQueue(
...
-        expire("$locksKey:$it", lockTtlSeconds)
-        move(from, to, delay, it)
-      }
-      ?.firstOrNull()
+      .firstOrNull()
+      ?.takeIf { id -> acquireLock(id) }
+      ?.also { id -> move(from, to, delay, id) }
+
+  private fun Jedis.acquireLock(id: String) =
+    set("$locksKey:$id", "\uD83D\uDD12", "NX", "EX", lockTtlSeconds) == "OK"
```
cfieber: 🔒

PR #1396, opened 2017-06-09
fix(queue): prevent duplicate messages getting pushed to the queue

robfletcher:


orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
```diff
@@ -58,17 +56,19 @@ class RedisQueue(
...
-            readMessage(id) { payload ->
-              callback.invoke(payload) {
-                ack(id)
-                fire<MessageAcknowledged>()
+        zrangeByScore(queueKey, 0.0, score(), 0, 1)
+          .firstOrNull()
+          ?.takeIf { id -> acquireLock(id) }
+          ?.also { id ->
+            readMessage(id) { message ->
+              callback.invoke(message) {
```
cfieber: this looks to me like the redis connection is held open while the callback handles the message, is that expected?

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
```diff
@@ -79,13 +79,15 @@ class RedisQueue(
 
   override fun push(message: Message, delay: TemporalAmount) {
     pool.resource.use { redis ->
-      val id = randomUUID().toString()
-      redis.multi {
-        hset(messagesKey, id, mapper.writeValueAsString(message))
-        zadd(queueKey, score(delay), id)
+      val messageHash = message.hashCode().toString()
```
cfieber: I think you need something other than hashCode here or you will get collisions

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
robzienert: Good catch. Looks that way to me as well.

robfletcher: That entire lambda ``` callback.invoke(message) {
  ackMessage(id)
}``` is called here https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueProcessor.kt#L51-L65 so whatever the handler does is off in another thread.

I need to give this some thought.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
robfletcher: Looking into using murmur

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
```diff
@@ -232,4 +250,13 @@ class RedisQueue(
...
     }
+
+  private fun JedisCommands.hgetInt(key: String, field: String, default: Int = 0) =
+    hget(key, field)?.toInt() ?: default
+
+  private fun Message.hash() =
+    Hashing
+      .murmur3_32()
+      .hashString(toString(), Charset.defaultCharset())
+      .toString()
```
robfletcher: Added a simple hash method. Could, of course, just use the `toString` value itself but this is simple enough.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
```diff
@@ -58,66 +58,74 @@ class RedisQueue(
...
-              }
-            }
+      redis.zrangeByScore(queueKey, 0.0, score(), 0, 1)
+        .firstOrNull()
+        ?.takeIf { id -> redis.acquireLock(id) }
+        ?.also { id ->
+          val ack = this::ackMessage.curry(id)
+          redis.readMessage(id) { message ->
+            callback(message, ack)
           }
```
robfletcher: I think this clarifies how the callbacks work a little, at least I hope so.

`callback` is the lambda `QueueProcessor` uses to find a handler, dispatch the message (on another thread) and then call the `ack` function.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
robfletcher: For context the Kotlin data class `toString` contains all property values so it's unique.

PR #1405, opened 2017-06-13
chore(expressions): retry within fromUrl expression helper on network errors as well

jeyrschabu:
- Added a handler for IOException
- throwing an exception when a request isn't retryable

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
@@ -32,6 +32,8 @@ import org.apache.http.protocol.HttpContext;
 import org.apache.http.protocol.HttpCoreContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.util.Arrays;
@@ -42,7 +44,7 @@ import java.util.concurrent.TimeUnit;
 public class HttpClientUtils {
   private static final Logger LOGGER = LoggerFactory.getLogger(HttpClientUtils.class);
   private static final int MAX_RETRIES = 5;
-  private static final int RETRY_INTERVAL = 3000;
+  private static final int RETRY_INTERVAL = 5000;
   private static final int TIMEOUT_MILLIS = 30000;
   private static List<Integer> RETRYABLE_500_HTTP_STATUS_CODES = Arrays.asList(
     HttpStatus.SC_SERVICE_UNAVAILABLE,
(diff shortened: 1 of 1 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -77,6 +78,13 @@ public long getRetryInterval() {
       }
     );
 
+    httpClientBuilder.setRetryHandler((exception, executionCount, context) -> {
```
jeyrschabu: The exception here is an IOException. Also adding a wait otherwise, it will retry right away

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -58,16 +60,15 @@ private static CloseableHttpClient httpClientWithServiceUnavailableRetryStrategy
...
           int statusCode = response.getStatusLine().getStatusCode();
-          HttpUriRequest currentReq = (HttpUriRequest) context.getAttribute(
-            HttpCoreContext.HTTP_REQUEST);
-
+          HttpUriRequest currentReq = (HttpUriRequest) context.getAttribute(HttpCoreContext.HTTP_REQUEST);
           LOGGER.info("Response code {} for {}", statusCode, currentReq.getURI());
           boolean shouldRetry = (statusCode == 429 || RETRYABLE_500_HTTP_STATUS_CODES.contains(statusCode)) && executionCount <= MAX_RETRIES;
-          if (shouldRetry) {
-            LOGGER.error("Retrying request on response status {}. Count {} Max is {}", statusCode, executionCount, MAX_RETRIES);
+          if (!shouldRetry) {
```
jeyrschabu: Throwing an exception here because the the handler doesn't throw an exception explicitly when we aren't retrying a request.

jeyrschabu:
@cfieber PTAL

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -77,6 +78,13 @@ public long getRetryInterval() {
       }
     );
 
+    httpClientBuilder.setRetryHandler((exception, executionCount, context) -> {
+      HttpUriRequest currentReq = (HttpUriRequest) context.getAttribute(HttpCoreContext.HTTP_REQUEST);
+      Uninterruptibles.sleepUninterruptibly(RETRY_INTERVAL, TimeUnit.MILLISECONDS);
```
cfieber: do we want to verify the type of exception here?

I'd think we would only retry on some subclass of  IOException

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
cfieber: cool, makes sense

cfieber:
LGTM

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
jeyrschabu: Yup, the exception here is of type IOException (not obvious since it's in a lambda). 
The API on that method is `boolean retryRequest(IOException exception, int executionCount, HttpContext context);`

We could potentially exclude certain types of IOException i.e `SSLHandshakeException`

PR #1408, opened 2017-06-14
fix(expressions): should not throw an exception if request was successful

jeyrschabu:
- Handling 200 - 299 response code in fromUrl


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -62,6 +62,11 @@ public boolean retryRequest(HttpResponse response, int executionCount, HttpConte
           int statusCode = response.getStatusLine().getStatusCode();
           HttpUriRequest currentReq = (HttpUriRequest) context.getAttribute(HttpCoreContext.HTTP_REQUEST);
           LOGGER.info("Response code {} for {}", statusCode, currentReq.getURI());
+
+          if (statusCode == HttpStatus.SC_OK) {
+            return false;
+          }
```
robfletcher: Anything 200-299?

robzienert:
👍

PR #1413, opened 2017-06-16
fix(pipelines): remove the generic signature of ExceptionHandler

cfieber:
The consumer of an exception handler is always working in the context of the base exception
class and uses the handles method to determine the whether an ExceptionHandler supports a
particular exception type.


orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/exceptions/ExceptionHandler.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public Date getTimestamp() {
+      return timestamp;
+    }
+
+    public String getExceptionType() {
+      return exceptionType;
+    }
+
+    public void setExceptionType(String exceptionType) {
```
robfletcher: Do we need these to be public setters? Seems like this would be an immutable type.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/MessageHandler.kt
```diff
@@ -37,6 +38,11 @@ interface MessageHandler<M : Message> : (Message) -> Unit {
       throw IllegalArgumentException("Unsupported message type ${message.javaClass.simpleName}")
     }
 
+  fun Collection<ExceptionHandler>.shouldRetry(ex: Exception, taskName: String?): ExceptionHandler.Response? {
```
robfletcher: I was just looking at this & I don't think `taskName` can ever be null. IntelliJ probably just made it that way when I extracted the method because the parameter is not annotated.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/exceptions/ExceptionHandler.java
cfieber: it was `@Canonical` in the groovy world and I didn't want to necessarily chase down all usages (though I actually probably should just do that and make this immutable)

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/MessageHandler.kt
cfieber: I had to change it this way because of some failing tests - I think somewhere is using task as a mock and doesn't have a name?

robfletcher: yeah, I think there's some stupid scenario where it couldn't find the task model too.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/batch/exceptions/ExceptionHandler.java
robfletcher: I would have thought it's only ever constructed internally to the `ExceptionHandler` impls

cfieber: that and tests...

pushed an update that immutablizes it

PR #1421, opened 2017-06-22
feat(redis/metrics): adds metrics for redis connection pools.

cfieber:
adds a dedicated pool for v3 queue processing to differentiate pool usage between queue and execution repository


orca-queue-redis/src/test/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueueIntegrationTest.kt
```diff
@@ -31,6 +36,14 @@ import org.springframework.test.context.junit4.SpringJUnit4ClassRunner
...
-  RedisQueueConfiguration::class
+  RedisQueueConfiguration::class,
+  RedisQueuePoolFixery::class
 ))
 class RedisQueueIntegrationTest : QueueIntegrationTest()
+
+
+
+@Configuration
+open class RedisQueuePoolFixery {
```
robzienert: A+++++ naming.

PR #1429, opened 2017-06-27
feat(aws/commits): add build info to execution

icfantv:
* add build info to pipeline execution.
* currently only adds prior/current build if there are commit diffs

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
@@ -67,9 +67,8 @@ class GetCommitsTask implements DiffTask {
     }
 
     Map repoInfo = [:]
-    String sourceCommit
-    String targetCommit
-    List commitsList
+    Map sourceInfo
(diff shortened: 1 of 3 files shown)


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
```diff
@@ -95,31 +94,35 @@ class GetCommitsTask implements DiffTask {
       String targetAmi = getTargetAmi(stage.context, region)
 
       //get commits from igor
-      sourceCommit = resolveCommitFromAmi(ancestorAmi, account, region)
-      targetCommit = resolveCommitFromAmi(targetAmi, account, region)
-
-      commitsList = []
+      sourceInfo = resolveInfoFromAmi(ancestorAmi, account, region)
```
icfantv: We already get the build information when we call `getByAmiId` below but we'd previously thrown everything away after extracting the commit hash from `appversion`.  I'm using the data returned by this call to also retrieve the build number.

icfantv:
screen shot 2017-06-27 at 12 11 16 pm

PR #1432, opened 2017-06-28
feat(pipelinetemplate): Adds ability to resolve a pipeline id by application and pipeline name

jeyrschabu:
- Added support for resolving a pipeline by name and application

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/tags/PipelineIdTag.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/tags/PipelineIdTag.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.pipelinetemplate.v1schema.render.tags;
+import com.hubspot.jinjava.interpret.*;
+import com.hubspot.jinjava.lib.tag.Tag;
+import com.hubspot.jinjava.tree.TagNode;
+import com.hubspot.jinjava.util.HelperStringTokenizer;
+import com.netflix.spinnaker.orca.front50.Front50Service;
+import com.netflix.spinnaker.orca.pipelinetemplate.exceptions.TemplateRenderException;
+import org.apache.commons.lang.StringUtils;
+import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
+import java.util.Optional;
+import java.util.Collections;
+
+public class PipelineIdTag implements Tag {
+  private static final String APPLICATION = "application";
+  private static final String NAME = "name";
+
+  private final Front50Service front50Service;
+
+  public PipelineIdTag(Front50Service front50Service) {
+    this.front50Service = front50Service;
+  }
+
+  @Override
+  public String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
+    List<String> helper = new HelperStringTokenizer(tagNode.getHelpers()).splitComma(true).allTokens();
+    if (helper.isEmpty()) {
+      throw new TemplateSyntaxException(tagNode.getMaster().getImage(), "Tag 'pipelineId' expects at least a pipeline name: " + helper, tagNode.getLineNumber());
+    }
+
+    Map<String, String> paramPairs = new HashMap<>();
+    helper.forEach(p -> {
+      String[] parts = p.split("=");
+      if (parts.length != 2) {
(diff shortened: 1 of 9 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/tags/PipelineIdTag.java
```diff
@@ -0,0 +1,85 @@
...
+    Context context = interpreter.getContext();
+    String application = paramPairs.getOrDefault("application", (String) context.get("application"));
+
+    PipelineTemplate template = (PipelineTemplate) context.get("pipelineTemplate");
+    if (template == null) {
+      throw new TemplateStateException("Pipeline template missing from jinja context");
+    }
+
+    PipelineId pipeline = template.getPipelineId();
+    List<Map<String, Object>> pipelines = front50Service.getPipelines(pipeline.getApplication());
```
jeyrschabu: I should take another pass at this once this is merged. Maybe cache the result of `front50Service.getPipelines(_)` with a TTL

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/model/PipelineTemplate.java
```diff
@@ -37,6 +37,7 @@
   private List<StageDefinition> stages;
   private List<TemplateModule> modules;
   private List<PartialDefinition> partials = new ArrayList<>();
+  private PipelineId pipelineId;
```
jeyrschabu: I seem to be losing the `pipeline` after the merge of templates and after `pipelineId` tag has resolbed. I think I should carry the value from one of the transform? @robzienert

jeyrschabu:
[WIP]: @robzienert PTAL. I have one outstanding item mentioned in one of the comment.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/tags/PipelineIdTag.java
```diff
@@ -0,0 +1,85 @@
...
+
+    Context context = interpreter.getContext();
+    String application = paramPairs.getOrDefault("application", (String) context.get("application"));
+
+    PipelineTemplate template = (PipelineTemplate) context.get("pipelineTemplate");
+    if (template == null) {
+      throw new TemplateStateException("Pipeline template missing from jinja context");
+    }
+
+    PipelineId pipeline = template.getPipelineId();
```
robzienert: There shouldn't be any need to modify the schema, or pass state around via the template objects to do this. When a user calls `{% pipelineId application=foo, pipeline="my pipeline" %}` these two args will be available in the `paramPairs` map you created above.

You should verify that these two keys have been defined, then use the `application` value to make the call into Front50, and subsequently filtered by `name`.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/tags/PipelineIdTag.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  @Override
+  public String getEndTagName() {
+    return null;
+  }
+
+  @Override
+  public String getName() {
+    return "pipelineid";
```
robzienert: `pipelineId`

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/tags/PipelineIdTag.java
robzienert: We'll probably be fine to not cache this, after all.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/tags/PipelineIdTag.java
jeyrschabu: Jinja wants it lowercase otherwise the tag never registers
```
  /**
   * @return Get name of end tag lowerCase Null if it's a single tag without content.
   */
  String getEndTagName();
```

jeyrschabu:
@robzienert PTAL.
Per our chat earlier, I was totally overthinking what this tag should do.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/tags/PipelineIdTag.java
```diff
@@ -0,0 +1,80 @@
...
+      String[] parts = p.split("=");
+      if (parts.length != 2) {
+        throw new TemplateSyntaxException(tagNode.getMaster().getImage(), "Tag 'pipelineId' expects parameters to be in a 'key=value' format: " + helper, tagNode.getLineNumber());
+      }
+
+      paramPairs.put(parts[0], parts[1]);
+    });
+
+    Context context = interpreter.getContext();
+    String application = paramPairs.getOrDefault("application", (String) context.get("appliacation"));
```
robzienert: Typo in `application`

jeyrschabu: Good catch sir

jeyrschabu:
Updated

PR #1434, opened 2017-06-28
fix(orca) track task invocations as per v2

robfletcher:


robfletcher:
Feels like this could be a little neater than having to call `trackStatus` in 5 different places but the logic is a little convoluted.

PR #1441, opened 2017-07-10
feat(orca) Place produced artifacts in global context

brujoand:
The idea here is to place artifacts from rosco in the global context along with artifacts from `findImageFromTag` and `findImageFromCluster` so that it's in place when we want to start working on the feature toggle to enable deployment of artifacts. I was a bit unsure on the naming and placement of this object `Artifacts` so feel free to suggest alternative approaches.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/Bake.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/Bake.groovy
@@ -18,6 +18,7 @@
 
 package com.netflix.spinnaker.orca.bakery.api
 
+import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import groovy.transform.CompileStatic
 import groovy.transform.EqualsAndHashCode
 import groovy.transform.ToString
@@ -35,5 +36,6 @@ class Bake {
   String ami
   String amiName
   String imageName
+  Artifact artifact
   // TODO(duftler): Add a cloudProviderType property here? Will be straightforward once rosco is backed by redis.
 }
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
@@ -127,10 +127,12 @@ class BakeStage implements StageDefinitionBuilder, RestartableStage {
         it.parentStageId == stage.parentStageId && it.status == ExecutionStatus.RUNNING
       }
 
+      def relatedBakeStages = stage.execution.stages.findAll {
+        it.type == PIPELINE_CONFIG_TYPE && bakeInitializationStages*.id.contains(it.parentStageId)
+      }
+
       def globalContext = [
-        deploymentDetails: stage.execution.stages.findAll {
-          it.type == PIPELINE_CONFIG_TYPE && bakeInitializationStages*.id.contains(it.parentStageId) && (it.context.ami || it.context.imageId)
-        }.collect { Stage bakeStage ->
+        deploymentDetails: relatedBakeStages.findAll{it.context.ami || it.context.imageId}.collect { Stage bakeStage ->
           def deploymentDetails = [:]
           ["ami", "imageId", "amiSuffix", "baseLabel", "baseOs", "refId", "storeType", "vmType", "region", "package", "cloudProviderType", "cloudProvider"].each {
             if (bakeStage.context.containsKey(it)) {
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.orca.clouddriver.tasks.cluster
 import com.fasterxml.jackson.core.type.TypeReference
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.frigga.Names
+import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import com.netflix.spinnaker.moniker.Moniker
 import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.RetryableTask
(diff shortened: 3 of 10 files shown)


jervi:
@spinnaker/reviewers PTAL

icfantv:
@robfletcher making sure you saw this w/ your global context deprecation work.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -112,19 +112,25 @@ class BakeStage implements BranchingStageDefinitionBuilder, RestartableStage {
...
             if (bakeStage.context.containsKey(it)) {
               deploymentDetails.put(it, bakeStage.context.get(it))
             }
           }
-
           return deploymentDetails
+        },
+        artifacts: imageProducingStages.collect { Stage bakeStage ->
+          def artifact = [:]
+          if (bakeStage.context.containsKey("artifact")) {
```
brujoand: I've heard this said before, and I have to admit that I don't understand how that can happen. Could you elaborate?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -227,10 +228,23 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
       }
     }.flatten()
 
+    List<Artifact> artifacts = deploymentDetails.collect {
+      new Artifact(
+        name: it.get("imageName"),
+        type: it.get("cloudProvider"),
```
brujoand: Good point. This PR was submitted before I went on vacation, and I'm seeing that I need to adjust it somewhat to fit the design document. I'll try ta adapt this.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
brujoand: ah, maybe if there are multiple regions in an aws bake?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
lwander: No worries, I didn't see the PR had been out as long as it has, I assumed it was submitted around the time gard pinged us.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
lwander: Yup, or if you tag multiple docker images in your build (one per release track or various global registries).

robfletcher:
I think it should be fine with the recent changes to global context usage. The artifacts will end up in the `outputs` field of the stage where downstream stages should look for them. They'll also be in global context for the time being, but we're tracking usage of keys there and trying to eliminate places where global context gets used.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -227,10 +228,23 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
...
+          "sourceServerGroup": it.get("sourceServerGroup")
+        ]
+      )
+    }
     return new TaskResult(ExecutionStatus.SUCCEEDED, [
       amiDetails: deploymentDetails
     ], [
-      deploymentDetails: deploymentDetails
+      deploymentDetails: deploymentDetails,
+      artifacts: artifacts
```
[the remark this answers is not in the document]
brujoand: Yes, this will leave the latest artifacts only in the global context, but all artifacts will be available in their respective stage.

brujoand:
Sorry for the on/off activity, I've been a bit all over the place lately. I've gotten back in touch with master, addressed the comments and added a PR to adapt Rosco's artifact creation to this standard. https://github.com/spinnaker/rosco/pull/222.

@robfletcher since these artifacts are being added without being used (intended for the next PR with feature toggle), should I just avoid putting them in the global context at all?

lwander:
Yes I would rely on the stage outputs. @jtk54 & I put some work into that (for artifacts specifically, but works for any stage output) to aggregate & collect outputs from triggers as well: https://github.com/spinnaker/orca/pull/1659 & https://github.com/spinnaker/orca/pull/1655.

brujoand:
Roger that. I'm keeping the artifacts in their respective contexts now.

brujoand:
@lwander & @robfletcher anything else we need to merge this?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -229,8 +230,45 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
       }
     }.flatten()
 
+    List<Artifact> artifacts = imageSummaries.collect { placement, summaries ->
```
lwander: Is this meant to be an interim solution to make it appear like all stages emit artifacts rather than image summaries?

brujoand: Yes. The idea is to do this in two takes. This first PR to make artifacts available where we produce/find images. Then the next iteration will be to actually use these artifacts for deployments which will in turn make them even more available.

brujoand:
Yes, Rosco already emits Artifacts along with ami references. My pr spinnaker/rosco#222 is to make that artifact compatible with the one used in this PR.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/Bake.groovy
```diff
@@ -35,5 +36,6 @@ class Bake {
   String ami
   String amiName
   String imageName
+  Artifact artifact
```
lwander: This should be a list, right? We can produce multiple images per bake.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -229,8 +230,45 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
...
+        }
+
+        artifact.metadata = metadata
+        artifact.name = summary.imageName
+        artifact.location = location
+        artifact.type = "${cloudProvider}/image"
+        artifact.reference = "${summary.imageId}"
+        artifact.uuid = UUID.randomUUID().toString()
+      }
+      return artifact
```
lwander: This seems a little odd - instead of assigning to `artifact` created in the parent scope, I would create artifact here, and then return that. Then assign the result of `findResults {..}.flatten` to something like `List<Artifact> artifacts`

PR #1442, opened 2017-07-10
feat(pipelines): Add support for task-backed pipeline config saves

robzienert:
Merging this won't physically change any functionality. A followup PR in gate will move the save path from `gate -> front50` to `gate -> orca -> front50`.

This PR includes a mutator that will allow templated pipelines to correctly save params, triggers, etc to where the rest of spinnaker expects them to be when saved into front50.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/Front50Service.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/Front50Service.groovy
@@ -52,6 +52,9 @@ interface Front50Service {
   @POST("/pipelines")
   Response savePipeline(@Body Map pipeline)
 
+  @PUT("/pipelines/{pipelineId}")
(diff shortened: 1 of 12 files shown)


jeyrschabu:
Needs a rebase too but you know that

robzienert:
The format is defined by the gate request that initiates these tasks - this is done in a CRUD workflow (when a user creates or updates a pipeline via the UI or API, for example), as opposed to actually executing the pipeline. The gate code will look sorta like this...

```groovy
// PipelineController.groovy
  void savePipeline(@RequestBody Map pipeline) {
    def operation = [
      description: "Save pipeline '${pipeline.get("name") ?: "Unknown"}'",
      application: pipeline.get('application'),
      job: [
        [
          type: "savePipeline",
          pipeline: pipeline
        ]
      ]
    ]
    // ...
  }
```
Where `pipeline` is the full pipeline JSON definition you'd see if you clicked `Edit Pipeline JSON`.

PR #1449, opened 2017-07-12
fix(core)- optional stage support cannot handle expressions with ${}

tomaslin:
There were changes to the order in which expressions get validated for the v3 engine. This makes sure that expressions both with ${} and without are evaluated correctly in the `conditional on expression` field for stages.


ajordens:
👍 Seems reasonable to me.

PR #1450, opened 2017-07-12
chore(dependencies): Bump spinnaker-dependencies to 0.102.0

ajordens:
The generic exception handling from `kork-web` allows us to remove all
custom @ExceptionHandler's.


ajordens:
PTAL @robzienert ... primarily affects the pipeline template exception handling.

Response status codes should be unchanged but payloads are slightly different now:

```
{
    "timestamp": 1499820272440,
    "status": 404,
    "error": "Not Found",
    "exception": "...",
    "message": "...",
}
```

instead of

```
{
    "status": "BAD_REQUEST",
    "message": "..."
}
```

Any issue?

robzienert:
That'll be fine.

PR #1451, opened 2017-07-12
fix(pipelinetemplate): Prefer pipeline ID over templated value; map on save

robzienert:
@danielpeach Sorta what I'm thinking to fix that TODO in roer. I think the mutator work is superfluous to the reordering in the JSON generator, though. I'd be interested to hear what version of orca they're running that they don't see the auto-mapping already working - maybe this reordering will do the trick alone.


danielpeach:
Works for me. Thanks!

PR #1453, opened 2017-07-13
fix(queue): don't pull messages from the queue when no threads available

cfieber:
do not merge: I've not actually tried this out other than the QueueIntegrationTest


robzienert:
Looks sane. Probably perfect.

PR #1458, opened 2017-07-14
chore(pipelinetemplate): Refactor renderer to take a list of tags

jeyrschabu:
- renderer takes a list of tags that it registers

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/JinjaRenderer.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/JinjaRenderer.java
@@ -27,6 +27,7 @@ import com.hubspot.jinjava.interpret.TemplateError.ErrorItem;
 import com.hubspot.jinjava.interpret.TemplateError.ErrorReason;
 import com.hubspot.jinjava.interpret.TemplateError.ErrorType;
 import com.hubspot.jinjava.interpret.errorcategory.BasicTemplateErrorCategory;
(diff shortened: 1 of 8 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/JinjaRenderer.java
```diff
@@ -67,6 +69,10 @@ public JinjaRenderer(RenderedValueConverter renderedValueConverter, ObjectMapper
     jinja.setResourceLocator(new NoopResourceLocator());
     jinja.getGlobalContext().registerTag(new ModuleTag(this, pipelineTemplateObjectMapper));
```
jeyrschabu: Opted to leave these tags as defaults. A little bit of smell with `ModuleTag` as it has a circular dependency with the renderer itself. That made it hard to extract into its own component without having to do some Spring handwaving

jeyrschabu:
This paves the way for adding & registering jinja tags with the renderer. 
Allows us to add custom Netflix tags with ease

jeyrschabu:
@spinnaker/netflix-reviewers or just @robzienert PTAL

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/JinjaRenderer.java
robzienert: Module having a circular dependency on Renderer is sort of an inevitability for that individual tag, but I agree that these tags should always be there, since they are key to implementing the spec.

PR #1459, opened 2017-07-14
fix(queue): Sets boundary on thread pool work queue and aborts work when full

robzienert:
The current messageHandler thread pool never expands its capacity because the queue never fills (default capacity is `Integer.MAX_VALUE`) so the new `QueueExecutor` will never stop polling the redis queue.

@cfieber @robfletcher PTAL

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/config/QueueConfiguration.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/config/QueueConfiguration.kt
@@ -34,7 +34,7 @@ import org.springframework.context.event.SimpleApplicationEventMulticaster
 import org.springframework.scheduling.annotation.EnableScheduling
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor
 import java.time.Clock
-import java.util.concurrent.Executor
+import java.util.concurrent.ThreadPoolExecutor
 
 @Configuration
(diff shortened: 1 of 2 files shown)


robzienert:
Alternative is to force use of a fixed thread pool size.

orca-queue/src/main/kotlin/com/netflix/spinnaker/config/QueueConfiguration.kt
```diff
@@ -61,6 +61,8 @@ open class QueueConfiguration {
       ThreadPoolTaskExecutor().apply {
         corePoolSize = 20
         maxPoolSize = 150
+        setQueueCapacity(20)
```
cfieber: should queue capacity be 0? I think we don't want to queue anything. or if this sizes the BlockingQueue then maybe set it to 1 instead of 20?

robzienert: Chose that number sort of arbitrarily, I suppose keeping a work queue at all is what's causing processing anomalies - now that you mention that, fixed thread pool is probably the best way forward.

robzienert:
Updated to have queue size 0 and fixed pool size: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/SynchronousQueue.html

PR #1463, opened 2017-07-17
fix(pipeline_templates): require scopes

danielpeach:
@robzienert 
Changed my mind from this morning, since templates without scopes don't show up under `{front50}/pipelineTemplates`, which is pretty confusing. Also the dcd-spec says they're required.


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/SavePipelineTemplateTask.java
```diff
@@ -35,6 +35,9 @@ default void validate(PipelineTemplate template) {
     if (template.getMetadata().getDescription() == null || "".equalsIgnoreCase(template.getMetadata().getDescription())) {
       missingFields.add("metadata.description");
     }
+    if (template.getMetadata().getScopes() == null || template.getMetadata().getScopes().size() == 0) {
```
robzienert: `isEmpty()` is a better option over `size()`.

robzienert:
Works for me.

PR #1466, opened 2017-07-18
fix(queue): Support for deserializing messages that may be wrapped in…

ajordens:
… an Envelope

Also increase the requeue delay from 5s to 60s in `StartStageHandler`.

### DIFF
--- a/orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
+++ b/orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.orca.q.redis
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.KotlinModule
+import com.fasterxml.jackson.module.kotlin.readValue
 import com.google.common.hash.Hashing
 import com.netflix.spinnaker.orca.q.Message
 import com.netflix.spinnaker.orca.q.Queue
@@ -61,6 +62,18 @@ class RedisQueue(
   private val locksKey = "$queueName.locks"
   private val hashKey = "$queueName.hash"
   private val hashesKey = "$queueName.hashes"
+
+  companion object {
+    fun convertToMessage(json: String, mapper: ObjectMapper): Message {
+      val messageMap = mapper.readValue<Map<String, Any>>(json)
+
+      return if (messageMap.containsKey("payload")) {
+        mapper.convertValue(messageMap.get("payload"), Message::class.java)
+      } else {
+        mapper.readValue<Message>(json)
+      }
+    }
+  }
 
   override fun poll(callback: (Message, () -> Unit) -> Unit) {
     pool.resource.use { redis ->
@@ -215,11 +228,11 @@ class RedisQueue(
         removeMessage(id)
       } else {
         try {
-          val message = mapper.readValue<Message>(json)
+          val message = convertToMessage(json, mapper)
           block.invoke(message)
         } catch(e: IOException) {
-          log.error("Failed to read message $id", e)
-          removeMessage(id)
+          log.error("Failed to read message $id, requeuing...", e)
+          requeueMessage(id)
         }
       }
     }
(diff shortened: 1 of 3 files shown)


ajordens:
First step before introducing the `Envelope` across the board (changing `Queue` and the callback interfaces).

PTAL @robzienert @robfletcher

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
```diff
@@ -62,6 +63,18 @@ class RedisQueue(
   private val hashKey = "$queueName.hash"
   private val hashesKey = "$queueName.hashes"
 
+  companion object {
+    fun convertToMessage(json: String, mapper: ObjectMapper): Message {
+      val messageMap = mapper.readValue<Map<String, Any>>(json)
+
+      return if (messageMap.containsKey("payload")) {
```
ajordens: Point is to try and be backwards compatible when introducing an envelope (provided that envelope contains the original `Message` as a `payload` attribute).

Once the old orca is no longer running, this change can be reverted in favor of using a yet-to-be-defined `Envelope` across the board.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
```diff
@@ -215,11 +228,11 @@ class RedisQueue(
...
       } else {
         try {
-          val message = mapper.readValue<Message>(json)
+          val message = convertToMessage(json, mapper)
           block.invoke(message)
         } catch(e: IOException) {
-          log.error("Failed to read message $id", e)
-          removeMessage(id)
+          log.error("Failed to read message $id, requeuing...", e)
+          requeueMessage(id)
```
ajordens: Perhaps less important with the support for deserializing a nested `Message` .... wdyt?

orca-queue-redis/src/test/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueueSpec.kt
```diff
@@ -50,3 +59,29 @@ private val createQueue = { clock: Clock,
 private fun shutdownCallback() {
   redis?.destroy()
 }
+
+class ConvertToMessageSpec : Spek({
+  describe("should support deserializing a nested message") {
```
ajordens: I bet @robfletcher could tell me how to add this only to the `RedisQueueSpec` (that's extending from `QueueSpec`).

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
```diff
@@ -44,7 +44,7 @@ open class StartStageHandler
   private val exceptionHandlers: List<ExceptionHandler>,
   private val clock: Clock,
   private val contextParameterProcessor: ContextParameterProcessor,
-  @Value("\${queue.retry.delay.ms:5000}") retryDelayMs: Long
+  @Value("\${queue.retry.delay.ms:60000}") retryDelayMs: Long
```
ajordens: In the event that we do still get hit with a looper, at least delay it a bit.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
robzienert: Should be fine, I think.

robfletcher: That should never happen unless we change the format. I don't think retrying will help

orca-queue-redis/src/test/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueueSpec.kt
robfletcher: Add it to `QueueSpec` and it will run as part of `RedisQueueSpec`. Am I missing something?

ajordens:
We will be changing the format when the message envelope is introduced.

Didn't add it to QueueSpec as the other queue providers don't necessary
deal with deserialization the same way.

robfletcher:
I'm still trying to figure out a way to augment inherited specs like that. Seems to be a bit of a blind spot in Spek.

ajordens:
IMO we should just use Spock for testing ... not sure if/what Spek buys us?

PR #1467, opened 2017-07-18
fix(stages/webhook) ensure buildInfo is a Map

erran:
Fix spinnaker/spinnaker#1842 — Wrap the response body under the key responseBody so we don't cause an exception in the ContextParameterProcessor.

This code change was initially made before realising this only affected JSON string/array bodies. It should be a taken as a grain of salt and is meant to start additional feedback around spinnaker/spinnaker#1842.

**NOTE**: This will change the value of `buildInfo` to be a map for people so is a breaking change to users who use the `stage.context.buildInfo` pipeline expression for a webhook configuration stage.

Another route that may make more sense is doing a type check inside of `ContextParameterProcessor` before calling `scm` on `buildInfo`. I initially opted against that before discovering this only affected non-JSON Object types. Performing a type check keeps backwards compatibility afaik though. 🙆‍♂️

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
@@ -65,7 +65,7 @@ class CreateWebhookTask implements RetryableTask {
     def outputs = [:]
     outputs << [statusCode: statusCode]
     if (response.body) {
-      outputs << [buildInfo: response.body]
+      outputs << [buildInfo: [responseBody: response.body]]
     }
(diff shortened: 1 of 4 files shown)


erran-r7:
Another route that may make more sense is doing a type check inside of `ContextParameterProcessor` before calling `scm` on `buildInfo`.

robzienert:
@jervi Can you review this PR please?

jervi:
I'm sceptical about this, as it is a potential breaking change. And in my quick review of the linked issue, it might not be required either, as I was not able to reproduce it. It might already be solved in later versions of Spinnaker.
Can you confirm, @erran?
webhook

erran:
@jervi I'll be back at work from paternity in two weeks. I'll try to remember to confirm then. 👶 🍼 

I agree about the potentially breaking change. I really just wanted to document the issue and submit a potential fix. 🙆‍♂️

PR #1476, opened 2017-07-20
fix(pipeline_templates): preprocess configs in dependent pipeline sta…

danielpeach:
…rter

@robzienert please review.

Couldn't start a templated pipeline from another pipeline, or use a pipeline trigger for a templated pipeline.


robzienert:
💯

PR #1477, opened 2017-07-21
allow optional stage expressions to reference global context

robfletcher:


tomaslin:
should revert this change as part of this pr - https://github.com/spinnaker/orca/pull/1449/files

tomaslin:
LGTM otherwise

PR #1480, opened 2017-07-21
fix(script, jenkins): put results of properties files in its own map

tomaslin:
This writes out the contents of a properties file to a map called propertiesFileContent. This will allow us to render it in the UI and potentially use it to reconstruct a global context for restarts.


tomaslin:
no, right now we only allow one file but nothing stops that from a list

ajordens:
If we think this might be applicable to global context, thoughts on naming
the map key more generically than propertyFileContent?

Could use an additional attribute to denote what came from a property file
if need be?

tomaslin:
not too attached to the name, but chose it over something more generic like properties to prevent too many pipelines breaking if they're already using global context mechanism. 

Feel like we want to have these variables outside of the context object in the stage, but didn't feel that was too easy to do with our existing architecture

ajordens:
As long as nobody ever references the name in an expression, it's simple
enough to migrate away from on the future so I'm not too concerned.

tomaslin:
any more comments on this @ajordens @robfletcher ?

PR #1485, opened 2017-07-25
feat(queue): Make instance status externally settable

lwander:
Note this only works when the instance is not managed by eureka.

### DIFF
--- a/orca-eureka/src/main/java/com/netflix/spinnaker/orca/eureka/NoDiscoveryApplicationStatusPublisher.java
+++ b/orca-eureka/src/main/java/com/netflix/spinnaker/orca/eureka/NoDiscoveryApplicationStatusPublisher.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.eureka;
 
+import com.netflix.appinfo.InstanceInfo.InstanceStatus;
 import com.netflix.discovery.StatusChangeEvent;
 import com.netflix.spinnaker.kork.eureka.RemoteStatusChangedEvent;
 import org.slf4j.Logger;
@@ -24,6 +25,7 @@ import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ApplicationListener;
 import org.springframework.context.event.ContextRefreshedEvent;
 
+import static com.netflix.appinfo.InstanceInfo.InstanceStatus.OUT_OF_SERVICE;
 import static com.netflix.appinfo.InstanceInfo.InstanceStatus.UNKNOWN;
 import static com.netflix.appinfo.InstanceInfo.InstanceStatus.UP;
 
@@ -32,7 +34,7 @@ public class NoDiscoveryApplicationStatusPublisher implements ApplicationListene
   private final ApplicationEventPublisher publisher;
   private final Logger log = LoggerFactory.getLogger(NoDiscoveryApplicationStatusPublisher.class);
 
-  private static final RemoteStatusChangedEvent DEFAULT_UP_EVENT = new RemoteStatusChangedEvent(new StatusChangeEvent(UNKNOWN, UP));
+  private static InstanceStatus instanceStatus = UNKNOWN;
 
   public NoDiscoveryApplicationStatusPublisher(ApplicationEventPublisher publisher) {
     this.publisher = publisher;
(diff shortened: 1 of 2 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -232,4 +238,25 @@ class OperationsController {
...
+      return
+    }
+
+    if (!discoveryStatusPoller instanceof  NoDiscoveryApplicationStatusPublisher) {
+      throw new IllegalStateException("Discovery status cannot be overwritten - there is an existing discovery client managing orca")
+    }
+
+    NoDiscoveryApplicationStatusPublisher noDiscoveryApplicationStatusPublisher = (NoDiscoveryApplicationStatusPublisher) discoveryStatusPoller;
+
+    boolean enabled = enabledWrapper.get("enabled")
```
robfletcher: Why not just a boolean param?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -232,4 +238,25 @@ class OperationsController {
...
+  @RequestMapping(value = "/instance/enabled", method = RequestMethod.POST)
+  void setInstanceStatus(@RequestBody Map<String, Boolean> enabledWrapper) {
+    if (!discoveryStatusPoller) {
+      log.warn("No configured discovery status poller")
+      return
+    }
+
+    if (!discoveryStatusPoller instanceof  NoDiscoveryApplicationStatusPublisher) {
+      throw new IllegalStateException("Discovery status cannot be overwritten - there is an existing discovery client managing orca")
+    }
```
robfletcher: what HTTP status will this result in? Would make sense to return 501, maybe?

lwander: 500 - 501 seems reasonable

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
lwander: A boolean param alone wasn't being deserialized correctly/I don't know what content type makes sense there - do you have an example of this working?

robfletcher: oh I see, maybe I've done it with a query string before. Doesn't really matter

robzienert:
👍

lwander:
Sorry for the wait, was OOO, will set the error type to 501 now

lwander:
Actually, 405 (method not allowed) kind of makes more sense since we're rejecting the user's attempted state change. Any objections?

lwander:
It is a POST - and having a special error code makes it easier for Halyard as a client to figure out why it can't change discovery status. In this case the error means all is well, orca will deregister itself once disabled.

lwander:
Oh you mean rejecting POST, PATCH, or PUT as in "don't use these", not, "don't use these like that"? Hmm then I should find something different.

lwander:
Maybe 501 does make the most sense from the perspective of "at some point we can flip eureka status from orca"

ajordens:
Is this `POST` going to get fronted by `gate`? ... if so I suspect the exception handling there is going to catch this and return it as a `500` anyhow.

Also 501 does have a generally intended meaning that the HTTP method you supplied (HTTP POST) is not implemented on the server.

If you don't want to differentiate by message, would returning an additional attribute in the error payload let Halyard do what it needs to do?

See the `HasAdditionalAttributes` interface that an exception can implement.

```
@ResponseStatus(HttpStatus.NOT_IMPLEMENTED)
 private static class DiscoveryUnchangeableException extends IllegalStateException implements HasAdditionalAttributes {
   Map<String, Object> additionalAttributes = new HashMap<>();

   DiscoveryUnchangeableException(String message, int errorCode) {
     super(message)
     additionalAttributes.put("errorCode", errorCode);
   }
  
   Map<String, Object> getAdditionalAttributes() { return additionalAttributes; }
```

Something like the above should work.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -232,4 +236,32 @@ class OperationsController {
     def pipeline = orchestrationLauncher.start(json)
     [ref: "/tasks/${pipeline.id}".toString()]
   }
+
+  @RequestMapping(value = "/instance/enabled", method = RequestMethod.POST)
```
ajordens: Thinking to the future ... any value in rooting these sort of admin-like operations under a `/admin` path?

(I wouldn't be surprised if someday we support manipulating queue throttles via an admin API as well)

Might make it easier to specifically attach additional auth configurations (for example).

lwander: Oh I like that - I'll create a separate controller for that.

lwander:
Halyard will be opening an ssh tunnel & talking directly to orca for this, so we don't have to worry about gate squashing the error.

I remember spring only copying over the error message & class when returning to a client - let me check that the additional attributes are preserved. If so, that sounds like a good way to deal with this.

lwander:
Moved into a new admin controller, and took the `HasAdditionalAttributes` approach

PR #1490, opened 2017-07-25
fix(pipeline_templates): do type conversion during pipeline save

danielpeach:
@robzienert please review.

An update fails if the pipeline config uses a template and has parameters/notifications/triggers.


robzienert:
👍

PR #1492, opened 2017-07-25
chore(queue) remove appConfig field from executions

robfletcher:
This no longer gets used anywhere (that I can see).


robzienert:
👍

PR #1497, opened 2017-07-26
fix(queue): consider throttle time before timeout

icfantv:
* add new message attribute for keeping track of the total amount of
time a message is throttled.
* modify the task handler to deduct the total throttled time in the
calculation of how long a task has been executing before timing out.


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/Message.kt
```diff
@@ -37,6 +37,12 @@ interface Attribute {
 data class MaxAttemptsAttribute(val maxAttempts: Int = -1) : Attribute {
 }
 
+data class TotalThrottleTimeAttribute(var totalThrottleTimeMs: Long = 0) : Attribute {
```
robfletcher: I'm actually surprised the compiler lets you use `var` on a data class.

PR #1503, opened 2017-07-27
chore(queue): update timeout message

icfantv:
* update task timeout message to include metrics used for timeout
calculation.


robzienert:
👍

PR #1512, opened 2017-07-31
feat(task): Tasks can specify a dynamic `backoffPeriod`

ajordens:
This PR also introduces a few new metrics:
- `executions.started`: A counter for all started pipelines/orchestrations
  including application name.
- `task.invocations.duration`: A distribution summary for tasks grouped by
  duration (lt5m, gt5m, gt15m, gt30m, gt60m).

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.instance
 
+import java.time.Duration
 import java.util.concurrent.TimeUnit
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.frigga.Names
@@ -69,6 +70,19 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
   // For the other uses of this task, we will fail if the server group doesn't exist.
   boolean waitForUpServerGroup() {
     return false
+  }
+
+  @Override
+  long getDynamicBackoffPeriod(Duration taskDuration) {
+    if (taskDuration.toMillis() > TimeUnit.MINUTES.toMillis(60)) {
+      // task has been running > 60min, drop retry interval to every 2 minutes
+      return Math.max(backoffPeriod, TimeUnit.SECONDS.toMillis(120))
+    } else if (taskDuration.toMillis() > TimeUnit.MINUTES.toMillis(30)) {
+      // task has been running > 30min, drop retry interval to every 60s
+      return Math.max(backoffPeriod, TimeUnit.SECONDS.toMillis(60))
+    }
+
+    return backoffPeriod
   }
 
   @Override
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/RetryableTask.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/RetryableTask.java
@@ -1,5 +1,7 @@
 package com.netflix.spinnaker.orca;
 
+import java.time.Duration;
+
 /**
  * A retryable task defines its backoff period (the period between delays) and its timeout (the total period of the task)
  */
@@ -7,4 +9,8 @@ public interface RetryableTask extends Task {
   long getBackoffPeriod();
 
   long getTimeout();
+
+  default long getDynamicBackoffPeriod(Duration taskDuration) {
+    return getBackoffPeriod();
+  }
 }
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.orca.q.handler
 
 import com.netflix.spectator.api.Registry
+import com.netflix.spectator.api.histogram.BucketDistributionSummary
 import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.ExecutionStatus.*
 import com.netflix.spinnaker.orca.RetryableTask
@@ -42,6 +43,7 @@ import java.time.Duration
 import java.time.Duration.ZERO
 import java.time.Instant
 import java.time.temporal.TemporalAmount
+import java.util.concurrent.TimeUnit
 
 @Component
 open class RunTaskHandler
@@ -75,17 +77,17 @@ open class RunTaskHandler
               stage.processTaskOutput(result)
               when (result.status) {
                 RUNNING -> {
-                  queue.push(message, task.backoffPeriod())
-                  trackResult(stage, task.javaClass, result.status)
+                  queue.push(message, task.backoffPeriod(taskModel))
+                  trackResult(stage, taskModel, result.status)
                 }
                 SUCCEEDED, REDIRECT, FAILED_CONTINUE -> {
                   queue.push(CompleteTask(message, result.status))
-                  trackResult(stage, task.javaClass, result.status)
+                  trackResult(stage, taskModel, result.status)
                 }
                 TERMINAL, CANCELED -> {
                   val status = stage.failureStatus(default = result.status)
                   queue.push(CompleteTask(message, status))
-                  trackResult(stage, task.javaClass, status)
+                  trackResult(stage, taskModel, status)
                 }
                 else ->
                   TODO("Unhandled task status ${result.status}")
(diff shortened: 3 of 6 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -97,35 +98,54 @@ open class RunTaskHandler
...
-  private fun trackResult(stage: Stage<*>, taskType: Class<Task>, status: ExecutionStatus) {
+  private fun trackResult(stage: Stage<*>, taskModel: com.netflix.spinnaker.orca.pipeline.model.Task, status: ExecutionStatus) {
     val id = registry.createId("task.invocations")
       .withTag("status", status.toString())
       .withTag("executionType", stage.getExecution().javaClass.simpleName)
-      .withTag("stageType", stage.getType())
-      .withTag("taskType", taskType.simpleName)
       .withTag("isComplete", status.isComplete.toString())
-      .withTag("sourceApplication", stage.getExecution().getApplication())
+      .withTag("application", stage.getExecution().getApplication())
```
ajordens: Will require some minor updates to the existing dashboards.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -97,35 +98,54 @@ open class RunTaskHandler
...
       .let { id ->
         stage.getContext()["cloudProvider"]?.let {
           id.withTag("cloudProvider", it.toString())
         } ?: id
       }
     registry.counter(id).increment()
+
+    registry.counter(
+      // a per duration bucket counter (5m, 15m, 30m, 60m)
+      registry.createId("task.invocations.${bucketDuration(taskModel)}").withTags(id.tags())
```
ajordens: Should this be a tag on the `task.invocations` metric ... was cognizant of not wanting to blow that tag space out.

In reality probably don't even need the same tags, most concerned with `application`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
```diff
@@ -72,6 +73,19 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
   }
 
   @Override
+  long getExponentialBackoffPeriod(Duration taskDuration) {
```
ajordens: This is the only hierarchy of task that actually specifies a dynamic backoff period ... encompasses most of the WaitForXXX instance tasks.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -143,9 +163,11 @@ open class RunTaskHandler
         }
     }
 
-  private fun Task.backoffPeriod(): TemporalAmount =
+  private fun Task.backoffPeriod(taskModel: com.netflix.spinnaker.orca.pipeline.model.Task): TemporalAmount =
     when (this) {
-      is RetryableTask -> Duration.ofMillis(backoffPeriod)
+      is RetryableTask -> Duration.ofMillis(
+        getExponentialBackoffPeriod(Duration.ofMillis(System.currentTimeMillis() - taskModel.startTime))
+      )
```
robfletcher: Why not just have that `getExponentialBackoffPeriod` return a `Duration`?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/RetryableTask.java
```diff
@@ -1,10 +1,16 @@
...
+
 /**
  * A retryable task defines its backoff period (the period between delays) and its timeout (the total period of the task)
  */
 public interface RetryableTask extends Task {
   long getBackoffPeriod();
 
   long getTimeout();
+
+  default long getExponentialBackoffPeriod(Duration taskDuration) {
```
robfletcher: Why "exponential"? It doesn't have to be, right?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
```diff
@@ -72,6 +73,19 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
...
+    if (taskDuration.toMillis() > TimeUnit.MINUTES.toMillis(60)) {
+      // task has been running > 60min, drop retry interval to every 2 minutes
+      return TimeUnit.SECONDS.toMillis(120)
```
cfieber: maybe `Math.max(backoffPeriod, TimeUnit.SECONDS.toMillis(120))` here (and below)

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
robfletcher: Yeah, I was about to ask why this isn't just a tag

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -97,35 +98,54 @@ open class RunTaskHandler
...
     }
   }
 
-  private fun trackResult(stage: Stage<*>, taskType: Class<Task>, status: ExecutionStatus) {
+  private fun trackResult(stage: Stage<*>, taskModel: com.netflix.spinnaker.orca.pipeline.model.Task, status: ExecutionStatus) {
     val id = registry.createId("task.invocations")
       .withTag("status", status.toString())
       .withTag("executionType", stage.getExecution().javaClass.simpleName)
-      .withTag("stageType", stage.getType())
-      .withTag("taskType", taskType.simpleName)
```
robzienert: Do you not think that tracking invocations by stage / task type is valuable?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
ajordens: Just kept it the same as the existing `getBackoffPeriod()`

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
robzienert: I'm ok with them being different IDs. That seems to be consistent with what libraries like dyno do. ¯\\_(ツ)\_/¯

orca-core/src/main/groovy/com/netflix/spinnaker/orca/RetryableTask.java
ajordens: You're right (was originally thinking of doing something more generic across the board but pulled back) ... perhaps `getDynamicBackoffPeriod()` is better?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
ajordens: Not hugely ... and it cuts down on the amount of time we can analyze metrics.

Once we know the hot application we can just look at it's tasks/pipelines for details?

robfletcher: We have a graph by stage type on our dashboard right now

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -97,35 +98,54 @@ open class RunTaskHandler
...
-      .withTag("sourceApplication", stage.getExecution().getApplication())
+      .withTag("application", stage.getExecution().getApplication())
       .let { id ->
         stage.getContext()["cloudProvider"]?.let {
           id.withTag("cloudProvider", it.toString())
         } ?: id
       }
     registry.counter(id).increment()
+
+    registry.counter(
```
cfieber: This feels like a [BucketDistributionSummary](http://netflix.github.io/spectator/en/latest/javadoc/spectator-api/com/netflix/spectator/api/histogram/BucketDistributionSummary.html#get-com.netflix.spectator.api.Registry-com.netflix.spectator.api.Id-java.util.function.LongFunction-) - would be worth playing with that and seeing how it ends up recorded

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
ajordens: The larger the tag space, the shorter duration that tags can be analyzed over ... unsure if have a tag with 6 possible values would cause us grief.

Maybe I'll flip it to a tag and we can run it for a bit.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/RetryableTask.java
robfletcher: We could replace `getBackoffPeriod`. Have the hard-coded versions be defaults?

PR #1514, opened 2017-07-31
fix(orca) prevent bean override

robfletcher:
Commit 09e2933c185871c4ac91168f6f816394da3ce84f fails on startup as there are > 1 `ApplicationListener` beans in play.


ajordens:
🚢

lwander:
Sorry, thanks!

robfletcher:
No problem. Not sure this is the ideal way to fix it but it works.

PR #1521, opened 2017-08-02
feat(clouddriver): Support multiple read-only clouddrivers

ajordens:
Criteria can be applied to each clouddriver to filter the types of requests that will be sent to it.

This PR supports execution type and application (regex) criteria.

If no criteria is specified, all requests are eligible to be sent to the particular clouddriver.

Example Configuration:

```
clouddriver:
  baseUrl: https://my-writeable-clouddriver.com
  readonly:
    baseUrls:
    - baseUrl: https://clouddriver-for-orchestrations.com
      priority: 5
      config:
        selectorClass: com.netflix.spinnaker.orca.clouddriver.config.ByExecutionTypeServiceSelector
        executionTypes:
          - orchestration
    - baseUrl: https://clouddriver-for-spindemo.comm
      priority: 10
      config:
        selectorClass: com.netflix.spinnaker.orca.clouddriver.config.ByApplicationServiceSelector
        applicationPattern: .*spindemo.*|.*cdspindemo.*
    - baseUrl: https://clouddriver-for-everything-else.com
      priority: 1
```

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingClouddriverService.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingClouddriverService.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver;
+
+import com.netflix.spinnaker.orca.ExecutionContext;
+import com.netflix.spinnaker.orca.clouddriver.config.SelectableService;
+
+class DelegatingClouddriverService<T> {
+  private final SelectableService selectableService;
+
+  DelegatingClouddriverService(SelectableService selectableService) {
+    this.selectableService = selectableService;
+  }
+
+  T getService() {
+    SelectableService.Criteria criteria = new SelectableService.Criteria(null, null);
+
+    ExecutionContext executionContext = ExecutionContext.get();
+    if (executionContext != null) {
+      criteria = new SelectableService.Criteria(
+        executionContext.getApplication(),
+        executionContext.getExecutionType(),
+        executionContext.getExecutionId()
+      );
+    }
+
+    return (T) selectableService.getService(criteria);
+  }
+}
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.groovy
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.orca.clouddriver.*
 import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper
 import com.netflix.spinnaker.orca.retrofit.RetrofitConfiguration
 import com.netflix.spinnaker.orca.retrofit.logging.RetrofitSlf4jLog
+import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
@@ -33,6 +34,9 @@ import retrofit.RequestInterceptor
 import retrofit.RestAdapter
 import retrofit.client.Client
 import retrofit.converter.JacksonConverter
+
+import java.util.regex.Pattern
+
 import static retrofit.Endpoints.newFixedEndpoint
 
 @Configuration
@@ -101,18 +105,32 @@ class CloudDriverConfiguration {
           .create(type)
     }
 
-    public <T> T buildReadOnlyService(Class<T> type) {
-      if (cloudDriverConfigurationProperties.cloudDriverReadOnlyBaseUrl == cloudDriverConfigurationProperties.cloudDriverBaseUrl) {
+    @CompileDynamic
+    private <T> SelectableService buildReadOnlyService(Class<T> type) {
+      if (cloudDriverConfigurationProperties.cloudDriverReadOnlyBaseUrls*.baseUrl == [cloudDriverConfigurationProperties.cloudDriverBaseUrl]) {
         log.info("readonly URL not configured for clouddriver, using writeable clouddriver $cloudDriverConfigurationProperties.cloudDriverBaseUrl for $type.simpleName")
       }
 
-      return buildService(type, cloudDriverConfigurationProperties.cloudDriverReadOnlyBaseUrl)
+      return new SelectableService(
+        cloudDriverConfigurationProperties.cloudDriverReadOnlyBaseUrls.collect {
+          def selector = new DefaultServiceSelector(buildService(type, it.baseUrl), it.priority, it.config)
+
+          def selectorClass = it.config?.selectorClass as Class<ServiceSelector>
+          if (selectorClass) {
+            selector = selectorClass.getConstructors()[0].newInstance(
+              selector.service, it.priority, it.config
+            )
+          }
+
+          selector
+        }
+      )
     }
   }
 
   @Bean
   MortService mortDeployService(ClouddriverRetrofitBuilder builder) {
-    return builder.buildReadOnlyService(MortService)
+    return new DelegatingMortService(builder.buildReadOnlyService(MortService))
   }
 
   @Bean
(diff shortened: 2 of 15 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationProperties.groovy
```diff
@@ -25,12 +25,18 @@ class CloudDriverConfigurationProperties {
...
+  static class MultiBaseUrl {
+    String baseUrl
+    List<BaseUrl> baseUrls
   }
 
   @Canonical
   static class CloudDriver {
     String baseUrl;
-    BaseUrl readonly;
+    MultiBaseUrl readonly;
```
ajordens: I'll double check again but this is expected to preserve the existing way of configuring a single read-only clouddriver.

```
clouddriver:
  baseUrl: http://something.com
  readonly:
    baseUrl: http://something-readonly.com
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationProperties.groovy
```diff
@@ -42,7 +48,17 @@ class CloudDriverConfigurationProperties {
     clouddriver?.baseUrl ?: kato?.baseUrl ?: oort?.baseUrl ?: mort?.baseUrl
   }
 
-  String getCloudDriverReadOnlyBaseUrl() {
-    clouddriver?.readonly?.baseUrl ?: getCloudDriverBaseUrl()
+  List<BaseUrl> getCloudDriverReadOnlyBaseUrls() {
```
ajordens: Again, this should preserve the existing configuration.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/SelectableService.java
```diff
@@ -0,0 +1,94 @@
...
+
+  public SelectableService(List<ServiceSelector> serviceSelectors) {
+    this.serviceSelectors = serviceSelectors;
+  }
+
+  public Object getService(Criteria criteria) {
+    Assert.notNull(criteria);
+
+    // look for a service selector that explicit supports this application (cannot be the default service selector)
+    Criteria applicationCriteria = new Criteria(
```
ajordens: Put another way ... application criteria overrides executionType criteria.

This allows you to pin pipelines/orchestrations for an individual application.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
```diff
@@ -39,6 +40,16 @@ interface AuthenticationAware {
       allowedAccounts = getExecution().getAuthentication()?.allowedAccounts
     }
 
-    AuthenticatedRequest.propagate(block, false, currentUser).call()
+    try {
+      ExecutionContext.set(ExecutionContext(
+        getExecution().getApplication(),
+        getExecution().javaClass.simpleName.toLowerCase(),
+        getExecution().getId(),
```
ajordens: executionId/stageId are not currently used but nevertheless added to the execution context.

ajordens:
This is what I was playing around with today @cfieber / @robzienert.

Need to give it another whirl of live testing (refactored a few things) but it should work as intended and allow us to pin applications/orchestrations/pipelines to specific read-only clouddrivers.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingClouddriverService.java
```diff
@@ -0,0 +1,40 @@
...
+
+class DelegatingClouddriverService<T> {
+  private final SelectableService selectableService;
+
+  DelegatingClouddriverService(SelectableService selectableService) {
+    this.selectableService = selectableService;
+  }
+
+  T getService() {
+    ExecutionContext executionContext = ExecutionContext.get();
```
cfieber: maybe worth handling the case of no threadlocal info here in case this gets called outside of a context where it is populated?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/ServiceSelector.java
```diff
@@ -0,0 +1,91 @@
...
+}
+
+class ByExecutionTypeServiceSelector implements ServiceSelector {
+  private final Object service;
+  private final Set<String> executionTypes;
+
+  ByExecutionTypeServiceSelector(Object service, Map<String, Object> config) {
+    this.service = service;
+    this.executionTypes = new HashSet(
+      ((Map<String, String>) config.get("executionTypes")).values()
```
cfieber: `<String, Collection<String>>` ?

cfieber: I think this needs to be either:

`new HashSet(((Map<String, Collection<String>>) config).get("executionTypes"))`
or
`new HashSet((Collection<String>) config.get("executionTypes"))`
?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/config/SelectableServiceSpec.groovy
```diff
@@ -0,0 +1,61 @@
...
+  @Shared
+  def katoService = Mock(KatoService)
+
+  @Unroll
+  def "should lookup service by application or executionType"() {
+    given:
+    def selectableService = new SelectableService(
+      [
+        new ByApplicationServiceSelector(mortService, ["applicationPattern": ".*spindemo.*"]),
+        new ByExecutionTypeServiceSelector(oortService, ["executionTypes": [0: "orchestration"]]),
```
cfieber: is the [0: "orchestration"] really how the yaml deserializes?

I'd expect
```
config:
  executionTypes:
    - orchestration
```
to deserialize to [config: executionTypes: ['orchestration']]`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/SelectableService.java
cfieber: from the discussion in slack, the prioritization could be achieved also by listing the different replicas in priority order with their matching rules and picking the first match

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.groovy
```diff
@@ -101,18 +105,32 @@ class CloudDriverConfiguration {
...
+      if (cloudDriverConfigurationProperties.cloudDriverReadOnlyBaseUrls*.baseUrl == [cloudDriverConfigurationProperties.cloudDriverBaseUrl]) {
         log.info("readonly URL not configured for clouddriver, using writeable clouddriver $cloudDriverConfigurationProperties.cloudDriverBaseUrl for $type.simpleName")
       }
 
-      return buildService(type, cloudDriverConfigurationProperties.cloudDriverReadOnlyBaseUrl)
+      return new SelectableService(
+        serviceSelectors: cloudDriverConfigurationProperties.cloudDriverReadOnlyBaseUrls.collect {
+          def selector = new DefaultServiceSelector(buildService(type, it.baseUrl), it.config)
+
+          def selectorClass = it.config?.selectorClass as Class<ServiceSelector>
```
cfieber: worth having a list of selectors here that all have to match? If I wanted to specifically route the orchestrations from a particular app somewhere that would allow it

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/SelectableService.java
cfieber: or a priority number

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
cfieber: worth adding stage type ? or alternatively the whole execution so the selectors can pull what they are interested in?

cfieber: not super in favor of the whole execution given it is a mutable thing though

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/ServiceSelector.java
ajordens: I blame teh spring!

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
ajordens: actually going to remove stageId instead ... agree with the sentiment that we might have problems if we flop between replicas between stages.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/SelectableService.java
ajordens: ended up going with a priority number ... defaulting to 1 (low priority)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingClouddriverService.java
ajordens: Done ... defaults to a `Criteria` with no application, executionType or executionId.

In such a scenario the `DefaultServiceSelector` would pick it up. If there is no `DefaultServiceSelector`, the first read-only clouddriver will be used.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.groovy
ajordens: At this point, not convinced it's useful.

Should be easy enough to support should the need come up.

ajordens:
Thanks for the feedback, updates have been made (ended up going with `priority`, thanks @robzienert).

I've verified that the previous behavior continues to work but I'm content to wait until I'm back in the office to merge this down.

PR #1522, opened 2017-08-03
fix(pipeline_templates) ignore yaml keywords in rendered value converter

danielpeach:
@robzienert please review.

If your template had
```
keyA: yes
keyB: 'yes'
```
Orca was mapping it in the rendered pipeline to

```
{
  "keyA": true,
  "keyB": true
}
```

Should now map to 

```
{
  "keyA": true,
  "keyB": "yes"
}
```

Open to suggestions for a better approach.

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/YamlRenderedValueConverter.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/YamlRenderedValueConverter.java
@@ -16,6 +16,8 @@
 package com.netflix.spinnaker.orca.pipelinetemplate.v1schema.render;
 
 import com.netflix.spinnaker.orca.pipelinetemplate.exceptions.TemplateRenderException;
+import java.util.Arrays;
+import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.yaml.snakeyaml.Yaml;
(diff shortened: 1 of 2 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/YamlRenderedValueConverter.java
```diff
@@ -55,4 +59,13 @@ public Object convertRenderedValue(String renderedValue) {
   private static boolean containsEL(String renderedValue) {
     return renderedValue.trim().startsWith("${");
   }
+
+  private static boolean isYamlKeyword(String renderedValue) {
+    for (String keyword: yamlKeywords) {
```
ttomsu: This could be condensed to:
```
return yamlKeywords.contains(renderedValue.toLowerCase())
```

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/YamlRenderedValueConverter.java
```diff
@@ -24,6 +26,8 @@
 
 public class YamlRenderedValueConverter implements RenderedValueConverter {
 
+  private static final List<String> yamlKeywords = Arrays.asList("yes", "no", "on", "off");
```
duftler: `YAML_KEYWORKS` would be more in keeping with naming conventions for `static final` fields.

robzienert:
👍 Nice catch.

PR #1524, opened 2017-08-03
Initial cut of removing global context from executions

robfletcher:
`StageContext` is a new map implementation with transparent read-thru of properties from upstream stages if they are not present in the stage's own context.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/TaskResult.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/TaskResult.java
@@ -14,20 +14,14 @@ public final class TaskResult {
 
   ExecutionStatus status;
   ImmutableMap<String, ?> stageOutputs;
-  ImmutableMap<String, ?> globalOutputs;
 
   public TaskResult(ExecutionStatus status) {
-    this(status, emptyMap(), emptyMap());
-  }
-
-  public TaskResult(ExecutionStatus status, Map<String, ?> stageOutputs, Map<String, ?> globalOutputs) {
-    this.status = status;
-    this.stageOutputs = ImmutableMap.copyOf(stageOutputs);
-    this.globalOutputs = ImmutableMap.copyOf(globalOutputs);
+    this(status, emptyMap());
   }
 
   public TaskResult(ExecutionStatus status, Map<String, ?> stageOutputs) {
-    this(status, stageOutputs, emptyMap());
+    this.status = status;
+    this.stageOutputs = ImmutableMap.copyOf(stageOutputs);
   }
 
   @Deprecated
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
@@ -18,7 +18,6 @@ import lombok.Data;
 import org.codehaus.groovy.runtime.ReverseListIterator;
 import static com.netflix.spinnaker.orca.ExecutionStatus.NOT_STARTED;
 import static java.lang.String.format;
-import static java.util.Arrays.asList;
 import static java.util.Collections.*;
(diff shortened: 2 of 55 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/TaskResult.java
```diff
@@ -14,20 +14,14 @@
 
   ExecutionStatus status;
   ImmutableMap<String, ?> stageOutputs;
-  ImmutableMap<String, ?> globalOutputs;
```
robfletcher: Removed the `globalOutputs` property from `TaskResult`. Any tasks that were using this are now putting the same values in the stage context instead (most of them were duplicating anyway).

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -94,7 +91,7 @@ public String getName() {
   /**
    * The context driving this stage. Provides inputs necessary to component steps
    */
-  private Map<String, Object> context = new HashMap<>();
+  private Map<String, Object> context = new StageContext(this);
```
robfletcher: `Stage.context` has its own type.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package com.netflix.spinnaker.orca.pipeline.model;
+
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+import com.google.common.collect.ForwardingMap;
+
+public class StageContext extends ForwardingMap<String, Object> {
```
robfletcher: This class is the key part of the change. It's a map implementation whose `get` method will look back using `Stage.ancestors()` for any missing keys.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
```diff
@@ -1,26 +1,21 @@
...
-import java.util.Collection;
-import java.util.Map;
-
 public interface ExecutionRepository {
   void store(@Nonnull Orchestration orchestration);
 
   void store(@Nonnull Pipeline pipeline);
 
-  void storeExecutionContext(
-    @Nonnull String id, @Nonnull Map<String, Object> context);
```
robfletcher: This method is gone which theoretically unblocks dynomite

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/model/StageContextSpec.groovy
```diff
@@ -0,0 +1,128 @@
...
+ */
+
+package com.netflix.spinnaker.orca.pipeline.model
+
+import spock.lang.Specification
+import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionEngine.v3
+import static com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner.STAGE_AFTER
+import static java.lang.System.currentTimeMillis
+
+class StageContextSpec extends Specification {
```
robfletcher: Test for the `StageContext` class which should clarify what it does.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
robfletcher: This class should probably handle expressions too in the long term but I didn't want to bite off any more for a first cut.

robfletcher:
Going to revisit this in a way that falls back to existing global context to minimize disruption

PR #1526, opened 2017-08-04
feat(execution): Add origin field to executions

robzienert:
Depends on https://github.com/spinnaker/kork/pull/90, https://github.com/spinnaker/gate/pull/439

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.java
@@ -16,13 +16,20 @@
 
 package com.netflix.spinnaker.orca.pipeline.model;
 
-import java.io.Serializable;
-import java.util.*;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.netflix.spinnaker.orca.ExecutionStatus;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import com.netflix.spinnaker.security.User;
 import lombok.Data;
+
+import java.io.Serializable;
+import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -227,9 +228,16 @@ class OperationsController {
...
   private Map<String, String> startTask(Map config) {
     convertLinearToParallel(config)
+    injectExecutionOrigin(config)
     def json = objectMapper.writeValueAsString(config)
     log.info('requested task:{}', json)
     def pipeline = orchestrationLauncher.start(json)
     [ref: "/tasks/${pipeline.id}".toString()]
   }
+
+  private void injectExecutionOrigin(Map pipeline) {
```
ajordens: s/pipeline/execution for slightly more clarity

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.java
```diff
@@ -53,6 +60,7 @@
   AuthenticationDetails authentication;
   PausedDetails paused;
   ExecutionEngine executionEngine = DEFAULT_EXECUTION_ENGINE;
+  String origin;
```
ajordens: Anything needed to ensure this gets persisted correctly? Don't recall if we need to do this in the `ExecutionRepository` for first level attributes.

robzienert: Good call - forgot about that spot. Updated.

PR #1531, opened 2017-08-07
fix(queue): Inherit timeout value from top-level stage

robzienert:

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
@@ -1,9 +1,5 @@
 package com.netflix.spinnaker.orca.pipeline.model;
 
-import java.io.IOException;
-import java.io.Serializable;
-import java.util.*;
-import javax.annotation.Nonnull;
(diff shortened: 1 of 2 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -328,6 +344,31 @@ public void resolveStrategyParams() {
...
+   */
+  @JsonIgnore public Optional<Long> getTopLevelTimeout() {
+    Stage<T> topLevelStage = this;
+    while (topLevelStage.parentStageId != null) {
+      String sid = topLevelStage.parentStageId;
+      Optional<Stage<T>> stage = execution.getStages().stream().filter(s -> s.id.equals(sid)).findFirst();
+      if (stage.isPresent()) {
+        topLevelStage = stage.get();
+      } else {
+        return Optional.empty();
```
robfletcher: Isn't this an error state?

robzienert: Yes, true. Fixed.

PR #1534, opened 2017-08-08
fix(queue): Increase hashing bits to decrease message collisions

robzienert:
I'll followup in a separate PR with some better error condition logging / handling, I'd like to grab the message by the hash and log that out as well so we can have better insight / capability to recover from actual collisions. I'd really like to find a strategy to just handle it and not require manual intervention.

### DIFF
--- a/orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
+++ b/orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
@@ -21,7 +21,10 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.KotlinModule
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.google.common.hash.Hashing
-import com.netflix.spinnaker.orca.q.*
+import com.netflix.spinnaker.orca.q.AttemptsAttribute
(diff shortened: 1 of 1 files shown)


robzienert:
Going to update the PR to include a migration of existing hashes.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisQueue.kt
```diff
@@ -104,8 +107,13 @@ class RedisQueue(
   override fun push(message: Message, delay: TemporalAmount) {
     pool.resource.use { redis ->
       val messageHash = message.hash()
-      if (redis.sismember(hashesKey, messageHash)) {
-        log.warn("Ignoring message as an identical one is already on the queue: $message")
```
ajordens: We say "identical" but ... is it possible to get the current message from the queue and log it? I don't think we index things that way tho.

ajordens: I'm also curious what code is pushing the duplicate message ... dumping a stack trace seems heavyweight tho.

robzienert: I was going to handle getting better logging in place after getting this in. If we want to just knock it out now, I'm ok with that.

PR #1537, opened 2017-08-09
feat(sharding): Support for execution type restriction in `ByOriginSelector`

ajordens:
```
clouddriver:
  readonly:
    baseUrls:
    - baseUrl: https://clouddriver-deck-only.com
      priority: 20
      config:
        selectorClass: com.netflix.spinnaker.orca.clouddriver.config.ByOriginServiceSelector
        origin: deck
        executionTypes:
          - orchestration
```

Also adjusted logging around the `CompleteExecution` message.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
@@ -36,7 +36,7 @@ open class CompleteExecutionHandler
   override val queue: Queue,
   override val repository: ExecutionRepository,
   private val publisher: ApplicationEventPublisher,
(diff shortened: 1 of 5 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
```diff
@@ -36,7 +36,7 @@ open class CompleteExecutionHandler
   override val queue: Queue,
   override val repository: ExecutionRepository,
   private val publisher: ApplicationEventPublisher,
-  @Value("\${queue.retry.delay.ms:5000}") retryDelayMs: Long
+  @Value("\${queue.retry.delay.ms:30000}") retryDelayMs: Long
```
robzienert: 🎉

PR #1538, opened 2017-08-09
fix(queue): Set upper bounds on the # of re-queue attempts for `Compl…

ajordens:
…eteExecution` messages


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
```diff
@@ -76,7 +76,11 @@ open class CompleteExecutionHandler
       } else if (stages.any { it.getStatus() == STOPPED } && !stages.otherBranchesIncomplete()) {
         block.invoke(if (execution.shouldOverrideSuccess()) TERMINAL else SUCCEEDED)
       } else {
-        log.warn("Re-queuing $this as the execution is not yet complete")
+        val attempts = getAttribute<AttemptsAttribute>()?.attempts ?: 0
+        log.warn("Re-queuing $this as the execution is not yet complete (attempts: $attempts)")
+
+        // no need to re-queue continuously as the RUNNING stages will also emit a `CompleteExecution` message
+        setAttribute(MaxAttemptsAttribute(10))
```
robfletcher: Where does this get evaluated? Is that something that already exists?

PR #1540, opened 2017-08-10
fix(pipeline_templates): fail if template has undefined variable

danielpeach:
@robzienert 

Template rendering should fail if the template includes, e.g., `{{region}}` and `region` hasn't been defined.


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/JinjaRenderer.java
```diff
@@ -175,4 +176,15 @@ private static Errors unwrapJinjaTemplateErrorException(FatalTemplateErrorsExcep
     return errors;
   }
 
+  private static JinjavaConfig buildJinjavaConfig() {
+    JinjavaConfig.Builder configBuilder = JinjavaConfig.newBuilder();
+
+    configBuilder.withFailOnUnknownTokens(true);
```
robzienert: Cool, I didn't know this was a method.

robzienert:
👍

PR #1544, opened 2017-08-11
fix(orca) stop just re-queueing maxed out CompleteExecution messages

robfletcher:


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/DeadMessageHandler.kt
```diff
@@ -17,16 +17,19 @@
 package com.netflix.spinnaker.orca.q.handler
 
 import com.netflix.spinnaker.orca.ExecutionStatus.TERMINAL
+import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
 import com.netflix.spinnaker.orca.q.*
 import org.slf4j.LoggerFactory
-import org.springframework.stereotype.Component
 
-@Component open class DeadMessageHandler {
+open class DeadMessageHandler(
```
robfletcher: I don't think it was a good idea for this to be a `@Component` since it was overridden. I've made that explicit in the configuration class instead.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/DeadMessageHandler.kt
```diff
@@ -17,16 +17,19 @@
...
+  private val executionRepository: ExecutionRepository
+) {
   private val log = LoggerFactory.getLogger(javaClass)
 
   open fun handle(queue: Queue, message: Message) {
     log.error("Dead message: $message")
     when (message) {
+      is CompleteExecution -> executionRepository.updateStatus(message.executionId, TERMINAL)
```
robfletcher: It's possible we need some other conditions here but I think it would be nice to get this fix in to reduce churn in the metrics we're seeing currently.

PR #1549, opened 2017-08-15
feat(pipeline_template): Prettify optional stage names

robzienert:
Small quality of life change, will now convert a camel-cased stage ID to title-cased words for stages that do not define a name.

@t0mpson FYI


robzienert:
This actually causes stage names to be rendered rather strangely in some cases (especially in partials), so closing for now.

PR #1553, opened 2017-08-16
fix(bake): root_volume_size should be ignored if null

jeyrschabu:
- root_volume_size should be ignored if null


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/BakeRequest.groovy
```diff
@@ -56,7 +56,7 @@ class BakeRequest {
   Boolean enhancedNetworking
   String amiName
   String amiSuffix
-  int rootVolumeSize
+  int rootVolumeSize = 10 //Bakery default
```
ajordens: This is the netflix bakery default ... not sure if there's anything to do with `rosco`? (cc: @duftler)

Seems like the sort of thing that could be configured externally in `orca.yml`?

(or alternatively use an `Integer` and double check that we don't send null values over to the bakery?)

jeyrschabu: Good point. Going to make a change to ignore that field if null

ajordens:
LGTM

PR #1554, opened 2017-08-16
move stage timeout and failure override settings out of context

robfletcher:
Stage gets 2 new fields:

* `Long timeoutMs` — `null` if default timeout should apply.
* `FailurePolicy onFailure` — enum replacing 3 booleans `failPipeline`, `continuePipeline` and `completeOtherBranchesThenFail` that were in context.

Both will require migrations in front50 and latter will need corresponding change in deck.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AlertOnAccessMap.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AlertOnAccessMap.java
@@ -25,18 +25,14 @@ import com.netflix.spectator.api.Registry;
 
 class AlertOnAccessMap<E extends Execution<E>> extends ForwardingMap<String, Object> {
 
+  private final Execution<E> execution;
   private final Map<String, Object> delegate;
   private final Registry registry;
-  private final Id counterId;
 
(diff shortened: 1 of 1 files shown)


robzienert:
What's the fate of existing configurations that have `stageTimeoutMs` set?

robfletcher:
They'll need migrating. I've already written 2 migrators for front50

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
```diff
@@ -583,6 +583,8 @@ class JedisExecutionRepository implements ExecutionRepository {
     map["stage.${stage.id}.parentStageId".toString()] = stage.parentStageId
     map["stage.${stage.id}.requisiteStageRefIds".toString()] = stage.requisiteStageRefIds?.join(",")
     map["stage.${stage.id}.scheduledTime".toString()] = String.valueOf(stage.scheduledTime)
+    map["stage.${stage.id}.timeoutMs".toString()] = stage.timeoutMs?.toString()
```
robzienert: Why is this getting converted to a String when just below we're keeping it a Long?

PR #1557, opened 2017-08-17
tweak(queue): Add a few more attempts when a stage fails to plan

ajordens:
Will now retry for 10 minutes with a 15s delay between attempts.

Improved logging in `TargetServerGroupResolver` in the event that retries
are exhausted. The exception will be included which should better allow
for the caller to be identified.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupResolver.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupResolver.groovy
@@ -58,18 +58,23 @@ class TargetServerGroupResolver {
   }
 
   private TargetServerGroup resolveByTarget(TargetServerGroup.Params params, Location location) {
-    Map tsgMap = fetchWithRetries(Map) {
-      oortService.getTargetServerGroup(params.app,
-        params.credentials,
-        params.cluster,
-        params.cloudProvider,
-        location.value,
-        params.target.name())
+    try {
+      Map tsgMap = fetchWithRetries(Map) {
+        oortService.getTargetServerGroup(params.app,
+          params.credentials,
+          params.cluster,
+          params.cloudProvider,
+          location.value,
+          params.target.name())
+      }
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupResolver.groovy
```diff
@@ -168,7 +168,14 @@ class TargetServerGroupResolver {
           throw RetrofitError.conversionError(response.url, response, converter, responseType, ce)
         }
       } catch (RetrofitError re) {
-        if (re.kind == RetrofitError.Kind.NETWORK && i < maxRetries) {
+        if (i > maxRetries) {
+          log.error("Max retries exhausted", re)
```
ajordens: Really just want the exception to be logged so we can identifier callers if this ever happens

robfletcher: How can it get there if `i > maxRetries`? The loop condition at line 154 will prevent it, surely.

ajordens: Lol ... I literally just pulled that up from the other conditional because I didn't want to duplicate it, but I reckon you're correct.

ajordens: I'll fix that now.

robfletcher: That's what happens when you don't write a test ¬_¬

ajordens: Truth

PR #1559, opened 2017-08-19
perf(*): WaitForXXX instance checks should favor loading server group…

ajordens:
… over cluster

Most WaitForXXX instance checks operate on a single server group
but pay a performance penalty by loading a cluster and filtering
to a specific server group client-side.

In particular, we noticed that clouddriver response times significant
degrade when fetching clusters containing dozens of server groups and
thousands of instances.

This PR is a targeted at `AbstractInstancesCheckTask`, there has been
no effort to improve other callers of `getCluster()` and
`getServerGroupFromCluster()` (future improvement).

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingOortService.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingOortService.java
@@ -36,8 +36,13 @@ public class DelegatingOortService
   }
 
   @Override
-  public Response getServerGroup(String app, String account, String cluster, String serverGroup, String region, String cloudProvider) {
-    return getService().getServerGroup(app, account, cluster, serverGroup, region, cloudProvider);
+  public Response getServerGroupFromCluster(String app, String account, String cluster, String serverGroup, String region, String cloudProvider) {
+    return getService().getServerGroupFromCluster(app, account, cluster, serverGroup, region, cloudProvider);
+  }
+
+  @Override
+  public Response getServerGroup(String app, String account, String region, String serverGroup) {
+    return getService().getServerGroup(app, account, region, serverGroup);
   }
 
   @Override
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/OortService.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/OortService.groovy
@@ -31,12 +31,18 @@ interface OortService {
                       @Path("cloudProvider") String cloudProvider)
 
   @GET("/applications/{app}/clusters/{account}/{cluster}/{cloudProvider}/serverGroups/{serverGroup}")
+  Response getServerGroupFromCluster(@Path("app") String app,
+                                     @Path("account") String account,
+                                     @Path("cluster") String cluster,
+                                     @Path("serverGroup") String serverGroup,
+                                     @Query("region") String region,
+                                     @Path("cloudProvider") String cloudProvider)
+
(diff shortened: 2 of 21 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingOortService.java
```diff
@@ -36,8 +36,13 @@ public Response getCluster(String app, String account, String cluster, String cl
   }
 
   @Override
-  public Response getServerGroup(String app, String account, String cluster, String serverGroup, String region, String cloudProvider) {
-    return getService().getServerGroup(app, account, cluster, serverGroup, region, cloudProvider);
+  public Response getServerGroupFromCluster(String app, String account, String cluster, String serverGroup, String region, String cloudProvider) {
```
ajordens: `getServerGroupFromCluster()` was previously known as `getServerGroup()` and is rather inefficient (loads the cluster and then filters).

I opt'd to not chase down callers and see which could be switched to the more optimal `getServerGroup()`, this can be done later. Most should be easily converted but there are a couple situations where `region` is not available.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/OortService.groovy
```diff
@@ -31,12 +31,18 @@ interface OortService {
                       @Path("cloudProvider") String cloudProvider)
 
   @GET("/applications/{app}/clusters/{account}/{cluster}/{cloudProvider}/serverGroups/{serverGroup}")
+  Response getServerGroupFromCluster(@Path("app") String app,
```
ajordens: The worst part of this particular API is that the Response could be:

- a Map if the `region` query parameter is supplied
- a Collection if the `region` query parameter is NOT supplied

Both forms are being called.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
```diff
@@ -199,6 +195,21 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
     }
   }
 
+  private List<Map> fetchServerGroups(String account, String cloudProvider, Map<String, List<String>> serverGroupsByRegion) {
+    Names names = Names.parseName(serverGroupsByRegion.values().flatten()[0])
+
+    if (serverGroupsByRegion.values().flatten().size() > 1) {
+      def response = oortService.getCluster(names.app, account, names.cluster, cloudProvider)
```
ajordens: Very rare (if at all)  that this is called with > 1 server groups.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
```diff
@@ -199,6 +195,21 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
...
+      def cluster = objectMapper.readValue(response.body.in().text, Map)
+      return cluster.serverGroups ?: []
+    } else {
+      def region = serverGroupsByRegion.keySet()[0]
+      def response = oortService.getServerGroup(names.app, account, region, names.group)
+      def serverGroup = objectMapper.readValue(response.body.in().text, Map)
+      return [serverGroup]
```
ajordens: Payload should be identical regardless of whether the server group was loaded directly or fetched via a cluster.

PR #1560, opened 2017-08-21
feat(pipelineexpressions): Adding a v2 expressions evaluator

jeyrschabu:
- Extracted expressions evaluation to new package
- Exposing expression failures via a detailed summary
- Added support for flagging a pipeline to be processed with v2

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.groovy
@@ -1,0 +1,164 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.pipeline.expressions
+
+import groovy.util.logging.Slf4j
+import org.springframework.expression.EvaluationContext
+import org.springframework.expression.Expression
+import org.springframework.expression.ExpressionParser
+import org.springframework.expression.ParserContext
+import org.springframework.expression.common.CompositeStringExpression
+
+@Slf4j
+class ExpressionTransform {
+  private static final List<String> EXECUTION_AWARE_FUNCTIONS = ["judgment", "judgement", "stage"]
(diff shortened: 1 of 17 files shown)


jeyrschabu:
The goal of this PR is to introduce a v2 for expression evaluation.
This change has three parts: 
  
1. Expose localized evalution failures:
In addition to error logging in case of a failure, a block like this will appear in the stage execution json
```
"expressionEvaluationSummary": {
  "{#stage('Wait')['context']['test'] == 3": [
    {
      "description": "Expression '{#stage( #root.execution, 'Wait')['context']['test'] == 3' @ 0: No ending suffix '}' for expression starting at character 0: {#stage( #root.execution, 'Wait')['context']['test'] == 3",
      "exceptionType": "org.springframework.expression.ParseException",
      "level": "ERROR",
      "timestamp": 1503293885728
    }
  ]
},
```
 A map where the key is the failed expression and the value is detailed information about the failure

 2. Add support for flagging a pipeline to use the v2 evaluator
 add this `"spel-evaluator" : "v2"` to the pipeline json

 3. Refactor by moving expressions related code to a new package

 Next, lining up a PR that will get rid of the multipass evaluation approach by moving the evaluation at `StageStart`. This will remove the ambiguity on when a given expression is processed. Finally, looking to improve UI support to provide feedback on syntax errors

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupport.java
```diff
@@ -0,0 +1,342 @@
...
+  public static Boolean toBoolean(String str) {
+    return Boolean.valueOf(str);
+  }
+
+  /**
+   * Encodes a string to base64
+   * @param text plain string
+   * @return converted string
+   */
+
```
robzienert: These extra newlines bother me.

jeyrschabu: Curious to know why? I don't have a problem removing them.

robzienert: Primarily from a consistency perspective: It's very atypical; certainly not something we do anywhere else in our projects.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.groovy
```diff
@@ -0,0 +1,168 @@
...
+      T result
+      Expression exp
+      String escapedExpressionString = null
+      Throwable exception = null
+      try {
+        exp = parser.parseExpression(literalExpression, parserContext)
+        escapedExpressionString = escapeExpression(exp)
+        result = exp.getValue(evaluationContext) as T
+      } catch (Exception e) {
+        log.info("Failed to resolve $source, returning raw value {}", e?.getMessage())
```
robzienert: I don't think there's a condition where `e` would actually be null here?

jeyrschabu: leftover. removing that

PR #1561, opened 2017-08-21
Deprecate use of the global context

robfletcher:
Instead of reading from global context, the stage context tries to resolve missing values from the outputs of prior stages before falling back to the global context. We'll get an Atlas counter incremented if the global context is accessed so we can start to track down usages.

### DIFF
--- a/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/model/StageContextSpec.groovy
+++ b/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/model/StageContextSpec.groovy
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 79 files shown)


orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/model/StageContextSpec.groovy
```diff
@@ -0,0 +1,138 @@
...
+  def "if all else fails will read from global context"() {
+    expect:
+    pipeline.stageByRef("3>1").context.fnord == "global-fnord"
+  }
+
+  private static PipelineInit pipeline(
+    @DelegatesTo(Pipeline) Closure init = {}) {
+    def pipeline = new Pipeline()
+    pipeline.id = UUID.randomUUID().toString()
+    pipeline.executionEngine = v3
```
robzienert: Why define this? Shouldn't it already be defaulting to v3?

robfletcher: I have no idea why I did this

PR #1567, opened 2017-08-22
fix(clouddriver): support retries from multiple MonitorKatoTasks in a stage.

cfieber:
Outputs a sentinel value after successfully retrieving the task info from clouddriver so that a subsequent instance of MonitorKatoTask in the same stage doesn't inherit the state tracking from the original MonitorKatoTask.

Sentinal value because you can't delete things from the stage context via task results due to use of ImmutableMap.


ajordens:
lgtm

PR #1570, opened 2017-08-23
Re-introduce the stage outputs change

robfletcher:


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AlertOnAccessMap.java
```diff
@@ -0,0 +1,66 @@
...
+      key,
+      execution.getApplication(),
+      execution.getClass().getSimpleName(),
+      execution.getId(),
+      execution.getName()
+    );
+    Id counterId = registry
+      .createId("global.context.access")
+      .withTag("application", execution.getApplication())
+      .withTag("key", String.valueOf(key));
```
robfletcher: This tag look sane @ajordens?

PR #1571, opened 2017-08-24
feat(pipeline_templates): Support jinja expressions in template variables

ajordens:
Imagine the following managed pipeline that is triggered with:
```
shouldWait: true
   regions: us-west-2,us-east-1
```
```
---
schema: "1"
id: waitChain
variables:
- name: shouldWait
  defaultValue: "{{ trigger.parameters.shouldWait == true }}"
- name: regions
  defaultValue: "{{ trigger.parameters.regions | split(',') }}"
stages:
- id: wait1
  type: wait
  config:
    waitTime: 1
  when:
  - "{{ shouldWait == true }}"
- id: bake
  type: bake
  dependsOn: [wait1]
  config:
    ...
    regions: |
      {% for region in regions %}
      - "{{ region }}"
      {% endfor %}
  when:
  - "{{ regions | length > 0 }}"
```


ajordens:
I'm thinking it might be nice to _not_ have the `trigger` available to anything outside of variables. This would force anyone that wanted to have an trigger parameter-based expression to first identify it as a variable.

Is this overkill?

robzienert:
@gardleopard has been using trigger data in `TemplateLoader`s. This bit excluded, I'm unaware of people using trigger data - I think your suggestion is a good path forward.

PR #1580, opened 2017-08-29
chore(orca): Kotlin enhancements for Spring 1.5

robfletcher:
- No longer need to use `open` and `@Autowired constructor` on most `@Component` classes.
- Trying to quiet some of the logging from the queue integration tests.

### DIFF
--- a/orca-queue/orca-queue.gradle
+++ b/orca-queue/orca-queue.gradle
@@ -26,6 +26,7 @@ dependencies {
   compile "org.threeten:threeten-extra:1.0"
   compile "org.funktionale:funktionale-partials:1.1"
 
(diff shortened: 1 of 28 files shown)


orca-queue/orca-queue.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
   compile "org.threeten:threeten-extra:1.0"
   compile "org.funktionale:funktionale-partials:1.1"
 
+  testCompile "org.springframework.boot:spring-boot-test:1.5.4.RELEASE"
```
robfletcher: @cfieber does this seem like the right way to pull this dependency in? Do we have a constant for the Spring version somewhere? I'm wary of this getting out of sync.

PR #1586, opened 2017-08-30
feat(pipeline_templates): Support jinja expressions in source field

jervi:
Add support for recursively supporting jinja expressions in the source field of DCD templates. Currently, only the top level template definition supports jinja in the source field.


jervi:
Upstream PR: https://github.com/spinnaker/orca/pull/1586

jervi:
@spinnaker/reviewers PTAL

robzienert:
I've mixed feelings on this. By introducing `Renderer` into the template loader process, we're tightly coupling a v1 schema concept (jinja rendering) to `TemplateLoader`, which may not be a concept in future schema implementations.

jervi:
Do you have an idea of how this can be solved without using the renderer inside `TemplateLoader`? Or is dynamic sources in templates just not supposed to be supported?

robzienert:
Dynamic sources were never intended to be supported. Dynamicism in this regard can nerf the ability for the publish template stage to perform pre-emptive plans as well as cascading pipeline saves (necessary for propagating some configs defined in templates to their concrete configs).

What's your use case that you need to have such dynamic behavior in templates?

gardleopard:
We will experiment with shim pipelines to see if that can solve this for us. We will test that on Monday.

PR #1587, opened 2017-08-30
feat(strategies): wait period before disable for rolling push and rolling red black

tomaslin:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
```diff
@@ -80,6 +84,18 @@ class RedBlackStrategy implements Strategy, ApplicationContextAware {
       )
     }
 
+    if(stageData?.waitTime) {
+      def waitContext = [waitTime: stageData?.waitTime]
```
ajordens: `waitTime` seems like a possibly poorly named variable (as opposed to a name indicating it's a delayBeforeDisable or something?), I assume that's what deck is putting in there in rolling push and this is consistent with that?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
```diff
@@ -80,6 +84,18 @@ class RedBlackStrategy implements Strategy, ApplicationContextAware {
...
+      stages << newStage(
+        stage.execution,
+        waitStage.type,
+        "wait",
+        waitContext,
+        stage,
+        SyntheticStageOwner.STAGE_AFTER
```
ajordens: I believe we don't let you shrinkCluster to 1 (otherwise it's just highlander) so this seems reasonable to have injected after the <optional> shrinkCluster.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
tomaslin: it's consistent with the naming in the wait stage

ajordens:
My point was only that stageData.waitTime could be more descriptive in the
context of a deploy stage.

No big deal, just makes it slightly easier when looking at a stage context
json what is being used for what.

tomaslin:
makes sense, renamed to `delayBeforeDisableSec`

PR #1590, opened 2017-08-30
fix(canary-v2): Do not propagate result strategy to cana...

duftler:
…ry judge.


duftler:
Companion orca pr: https://github.com/spinnaker/orca/pull/1590

PR #1591, opened 2017-08-30
fix(core): handle incoming non-parallelized pipelines

robfletcher:


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -87,8 +86,8 @@ class OperationsController {
       throw new InvalidRequestException("Pipeline is disabled and cannot be started.")
     }
 
-    def parallel = pipeline.parallel as Boolean
-    if (!parallel) {
+    def linear = pipeline.stages.every { it.refId == null }
+    if (linear) {
```
robfletcher: This is the key bit that was breaking things. It's bad that we do this evaluation before converting the JSON to an actual object IMO.

PR #1593, opened 2017-08-31
feat(canary-v2) Add cancellation support.

duftler:


orca-kayenta/src/main/groovy/com/netflix/spinnaker/orca/kayenta/pipeline/RunCanaryPipelineStage.groovy
```diff
@@ -40,4 +48,20 @@ class RunCanaryPipelineStage implements StageDefinitionBuilder {
...
+
+    log.info("Cancelling stage (stageId: $stage.id: executionId: $stage.execution.id, canaryPipelineExecutionId: $canaryPipelineExecutionId, context: $stage.context)")
+
+    try {
+      kayentaService.cancelPipelineExecution(canaryPipelineExecutionId, "")
+    } catch (Exception e) {
+      log.info("Failed to cancel stage (stageId: $stage.id, executionId: $stage.execution.id), e: $e.message")
+    }
+
+    return new CancellableStage.Result(stage, [:])
```
ewiseblatt: Is it normal for a result to look like this, with only the stage and no surrounding context? Do you get success/failure from looking at whether the stage was cancelled? I guess there is no error handling so reasons for failure to cancel dont matter.

If the caller just calls cancel and doesnt care about what happens after that then this is ok. Otherwise it seems odd.

LGTM otherwise.

orca-kayenta/src/main/groovy/com/netflix/spinnaker/orca/kayenta/pipeline/RunCanaryPipelineStage.groovy
```diff
@@ -40,4 +48,20 @@ class RunCanaryPipelineStage implements StageDefinitionBuilder {
...
+  CancellableStage.Result cancel(Stage stage) {
+    Map<String, Object> context = stage.getContext()
+    String canaryPipelineExecutionId = (String)context.get("canaryPipelineExecutionId")
+
+    log.info("Cancelling stage (stageId: $stage.id: executionId: $stage.execution.id, canaryPipelineExecutionId: $canaryPipelineExecutionId, context: $stage.context)")
+
+    try {
+      kayentaService.cancelPipelineExecution(canaryPipelineExecutionId, "")
+    } catch (Exception e) {
+      log.info("Failed to cancel stage (stageId: $stage.id, executionId: $stage.execution.id), e: $e.message")
```
ajordens: Probably worth `log.error` and including the actual exception vs just the message

orca-kayenta/src/main/groovy/com/netflix/spinnaker/orca/kayenta/pipeline/RunCanaryPipelineStage.groovy
duftler: Yeah, and I took this style from an existing cancellable stage.

orca-kayenta/src/main/groovy/com/netflix/spinnaker/orca/kayenta/pipeline/RunCanaryPipelineStage.groovy
duftler: Makes sense; I'll do that. Btw, this line is verbatim from `JenkinsStage` and `PipelineStage`, so we may want to make the same change there.

ajordens: That's fair.

orca-kayenta/src/main/groovy/com/netflix/spinnaker/orca/kayenta/pipeline/RunCanaryPipelineStage.groovy
duftler: Seems like the majority of cancellable stages don't return any additional context.

orca-kayenta/src/main/groovy/com/netflix/spinnaker/orca/kayenta/pipeline/RunCanaryPipelineStage.groovy
duftler: Updated as discussed, and changed those other 2 stages as well.

PR #1598, opened 2017-09-06
feat(web) disabling front50 refresh when orca fetches pipeline

emjburns:
This should prevent orca from triggering a front50 refresh when fetching pipelines. This call can be slow, so we want to pull the cached version from front50 (no more than a minute old). 

@ajordens PTAL


ajordens:
LGTM

Do not see a strong need for orca -> front50 to wait for a pipeline refresh, seems reasonable to grab recently cached (no more than 60s out of date).

PR #1601, opened 2017-09-06
feat(pipeline_template): Plan all dependent pipelines before updating a template

robzienert:
This change will now perform best-effort plans against all resolvable
dependents of a template. Pipelines that resolve their templates
dynamically will not be checked for compatibility while performing a
template save operation.

On a failure, the stage outputs look sorta like this:

```
{
  # ...
  "outputs": {
    "notification.type": "plantemplatedependents",
    "pipelineTemplate.allDependentPipelines": [
      "d8a7e9ad-93f2-4273-9873-6ea4674bafa1"
    ],
    "pipelineTemplate.dependentErrors": {
      "d8a7e9ad-93f2-4273-9873-6ea4674bafa1": [
        {
          "location": "configuration:stages.foo",
          "message": "Stage is missing type",
          "severity": "FATAL"
        },
        {
          "location": "configuration:stages.foo",
          "message": "Stage configuration is unset",
          "severity": "FATAL"
        },
        {
          "location": "configuration:stages.foo",
          "message": "A configuration-defined stage should have either dependsOn or an inject rule defined",
          "severity": "WARN"
        }
      ]
    },
    "pipelineTemplate.id": "wait-child"
  }
  # ...
}
```

My biggest concern with this change is the performance impact of templates that everyone under the sun inherits from. We don't have any like this yet, but I suspect it could become an issue in the future. That said, I think the safety this change provides is worth it.

@danielpeach @emjburns @ajordens @cfieber PTAL

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.pipelinetemplate.tasks;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.orca.ExecutionStatus;
+import com.netflix.spinnaker.orca.RetryableTask;
+import com.netflix.spinnaker.orca.TaskResult;
+import com.netflix.spinnaker.orca.front50.Front50Service;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import com.netflix.spinnaker.orca.pipelinetemplate.PipelineTemplatePipelinePreprocessor;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.PipelineTemplate;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+@Component
+public class PlanTemplateDependentsTask implements RetryableTask {
+
+  @Autowired(required = false)
+  private Front50Service front50Service;
+
+  @Autowired
+  private ObjectMapper pipelineTemplateObjectMapper;
+
+  @Autowired
+  private PipelineTemplatePipelinePreprocessor pipelineTemplatePipelinePreprocessor;
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    if (front50Service == null) {
+      throw new UnsupportedOperationException("Front50 is not enabled. Fix this by setting front50.enabled: true");
+    }
+
+    if (!stage.getContext().containsKey("pipelineTemplate")) {
+      throw new IllegalArgumentException("Missing required task parameter (pipelineTemplate)");
+    }
+
+    if (!(stage.getContext().get("pipelineTemplate") instanceof String)) {
+      throw new IllegalArgumentException("'pipelineTemplate' context key must be a base64-encoded string: Ensure you're on the most recent version of gate");
+    }
+
+    PipelineTemplate pipelineTemplate = (PipelineTemplate) stage.decodeBase64(
+      "/pipelineTemplate",
+      PipelineTemplate.class,
+      pipelineTemplateObjectMapper
+    );
+
+    List<Map<String, Object>> dependentPipelines = front50Service.getPipelineTemplateDependents(pipelineTemplate.getId(), false);
+
+    Map<String, Object> errorResponses = new HashMap<>();
+    for (Map<String, Object> dependentPipeline : dependentPipelines) {
+      Map<String, Object> request = new HashMap<>();
+      request.put("type", "templatedPipeline");
+      request.put("trigger", new HashMap<>());
+      request.put("config", dependentPipeline.get("config"));
+      request.put("template", pipelineTemplate);
+      request.put("plan", true);
+
+      Map<String, Object> response = pipelineTemplatePipelinePreprocessor.process(request);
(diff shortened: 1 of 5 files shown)


ajordens:
A strategy around versioning would be an alternative to re-planning everything?

Thinking that nothing would pickup an update until it's been planned, but the planning could happen asynchronously with notifications going back to the author on failure (and perhaps flagging the template as bad or something).

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
```diff
@@ -0,0 +1,112 @@
...
+  private ObjectMapper pipelineTemplateObjectMapper;
+
+  @Autowired
+  private PipelineTemplatePipelinePreprocessor pipelineTemplatePipelinePreprocessor;
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    if (front50Service == null) {
+      throw new UnsupportedOperationException("Front50 is not enabled, no way to fetch pager duty. Fix this by setting front50.enabled: true");
```
ajordens: uhhh?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+@Component
+public class PlanTemplateDependentsTask implements RetryableTask {
+
+  @Autowired(required = false)
```
ajordens: not sure you would ever have an orca w/o front50?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
```diff
@@ -0,0 +1,112 @@
...
+    Map<String, Object> errorResponses = new HashMap<>();
+    for (Map<String, Object> dependentPipeline : dependentPipelines) {
+      Map<String, Object> request = new HashMap<>();
+      request.put("type", "templatedPipeline");
+      request.put("trigger", new HashMap<>());
+      request.put("config", dependentPipeline.get("config"));
+      request.put("template", pipelineTemplate);
+      request.put("plan", true);
+
+      Map<String, Object> response = pipelineTemplatePipelinePreprocessor.process(request);
```
ajordens: try/catch around this or confident that no exceptions will come out?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
```diff
@@ -0,0 +1,112 @@
...
+    outputs.put("pipelineTemplate.allDependentPipelines", dependentPipelines.stream().map(it -> it.get("id")).collect(Collectors.toList()));
+
+    if (!errorResponses.isEmpty()) {
+      outputs.put("pipelineTemplate.dependentErrors", errorResponses);
+    }
+
+    return new TaskResult(
+      errorResponses.isEmpty() ? ExecutionStatus.SUCCEEDED : ExecutionStatus.TERMINAL,
+      Collections.emptyMap(),
+      outputs
```
ajordens: you expecting these outputs to be global / consumable downstream? I would have expected them to be exported into the current context.

ajordens: or maybe just need to clarify the intent of what `outputs` means ... historically that was for the global context and generally speaking not everything needs to be global.

ajordens:
Not anti doing it this way for now, with probably at least a conversation had around template versioning and how that might factor in to this.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
danielpeach: If the feature flag for pipeline templates is ever removed, requiring front50 here would break how Halyard bootstraps a Spinnaker installation (https://github.com/spinnaker/orca/pull/1231).

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
```diff
@@ -0,0 +1,112 @@
...
+
+    List<Map<String, Object>> dependentPipelines = front50Service.getPipelineTemplateDependents(pipelineTemplate.getId());
+
+    Map<String, Object> errorResponses = new HashMap<>();
+    for (Map<String, Object> dependentPipeline : dependentPipelines) {
+      Map<String, Object> request = new HashMap<>();
+      request.put("type", "templatedPipeline");
+      request.put("trigger", new HashMap<>());
+      request.put("config", dependentPipeline.get("config"));
+      request.put("template", pipelineTemplate);
```
danielpeach: This only verifies that a config and a root template work together, right? i.e., you won't verify that `pipeline -> childTemplate -> ... -> rootTemplate` all work together.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
robzienert: No exceptions will come out of this method. It's pretty heavily wrapped internally, which converts exceptions into a standard error format for returning via the API.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
robzienert: I was a little fuzzy on this - context is definitely what I want.

robzienert:
@ajordens Agreed, this isn't the nicest way forward, however getting this and a cascade pipeline re-save implemented on template update is necessary for unblocking some internal teams for their Q3 objectives. I'd like to loop back and build something more elegant, as well as start investigating a versioning strategy. I do like your idea of async planning & notifications - I was going to start going down the path of an internal pipeline that would do something similar.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/PlanTemplateDependentsTask.java
robzienert: Hmm, good point. Front50 returns all dependents, even of child templates, but obviously that's not going to work too hot. I think I'll double back and just plan direct descendants. WDYT?

robzienert:
Travis is all messed up.

PR #1602, opened 2017-09-07
fix(pipeline_template): Build partials, execute conditionals after and trim conditions as the last step

PerGon:
When using partials and conditionals stages, the evaluation of the `when` tag and the removal of stages that evaluate to `false` must be done in the end - when we are sure that the partials have been expanded and added to the stage tree.


robzienert:
Cool thanks!

PR #1605, opened 2017-09-07
fix(timeout): stage timeout overrides cumulative task durations

emjburns:
Previously overriding the stage timeout did not take into account the cumulative time the task were running/had run. This changes the stage timeout behavior so that when a user overrides the timeout, the stage times out after that value.

This change respects pausing the stage but NOT tasks getting throttled. If we want to take throttling time into account, we need to revisit the throttling strategy (as it's not easy to correctly calculate the stage throttle duration). Right now, we are assuming that users will enter pipeline timeout overrides have enough margin for throttling.

@ajordens @robfletcher @robzienert PTAL

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
@@ -395,9 +395,9 @@ public class Stage<T extends Execution<T>> implements Serializable {
   }
 
   /**
-   * Returns the top-most stage timeout value if present.
+   * Returns the top-most stage.
    */
-  @JsonIgnore public Optional<Long> getTopLevelTimeout() {
+  @JsonIgnore public Stage<T> getTopLevelStage() {
     Stage<T> topLevelStage = this;
     while (topLevelStage.parentStageId != null) {
       String sid = topLevelStage.parentStageId;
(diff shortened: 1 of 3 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -419,6 +427,14 @@ public void resolveStrategyParams() {
     return Optional.empty();
   }
 
+  /**
+   * Returns the top-most stage start time.
+   */
+  @JsonIgnore public Long getTopLevelStartTime() {
```
ajordens: Do we need this as a dedicated method? Looks like it's only called in one spot and could just as well be ...

`val startTime = stage.getTopLevelStage().getStartTime().toInstant()`

?

ajordens:
seems reasonable to me

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
emjburns: Changing, I agree.

emjburns:
@ajordens yes, that seems like the impact. Hopefully this change makes the flow make more sense (what you described to me would be weird behavior from a user side)

PR #1606, opened 2017-09-08
feat(pipeline_template): Re-save dependent pipelines on template save

robzienert:
Allows updates to templates that define triggers, params, notifications
or concurrency options to cascade changes to downstream pipelines.

We'll selectively re-save pipelines only if:

1. The template actually defines an inheritable config (triggers, params, notifications, concurrency options) **AND**
2. A pipeline explicitly defines that it wants to inherit any of these configs.

@danielpeach @ajordens @emjburns PTAL

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/pipeline/UpdatePipelineTemplateStage.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/pipeline/UpdatePipelineTemplateStage.java
@@ -15,16 +15,41 @@
  */
 package com.netflix.spinnaker.orca.pipelinetemplate.pipeline;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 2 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/pipeline/UpdatePipelineTemplateStage.java
```diff
@@ -35,4 +60,61 @@
...
+    }
+
+    PipelineTemplate pipelineTemplate = stage.decodeBase64(
+      "/pipelineTemplate",
+      PipelineTemplate.class,
+      pipelineTemplateObjectMapper
+    );
+
+    if (!pipelineTemplate.getConfiguration().hasInheritableConfigs()) {
+      // Pipeline templates that don't have inheritable configurations don't need to cascade save child pipelines.
```
danielpeach: Isn't this something you would want to propagate? e.g., if the change is that the template no longer has triggers, you'll want to remove them from its dependent pipelines.

PR #1615, opened 2017-09-14
feat(pipeline_template): Convert to pipeline template endpoint

robzienert:
This is step one of getting `Convert to Pipeline Template` functionality that's in roer and https://github.com/robzienert/spin-dcd-converter. This will get exposed via gate, then hopefully get added to deck as well.

@emjburns @danielpeach


orca-pipelinetemplate/src/main/resources/pipelineTemplateHeader.txt
```diff
@@ -0,0 +1,18 @@
+# GENERATED BY spinnaker
+#
+# The template generated below should be used as a base for further modifications.
+# It does not mae assumptions as to what things can be made into variables,
```
danielpeach: s/mae/make

robzienert: Mae is a good band, though.

emjburns:
LGTM

PR #1621, opened 2017-09-16
refactor(pipeline_template): Support multiple MPT schema versions

robzienert:
Adds the ability to parse different schema versions. This PR doesn't
actually change any functionality, except for switching to the new
handler API. All existing v1 schema code works the same way it did.

Additionally introduced Kotlin as the new language default.

Paves the way for introducing v2 schema, which I'm targeting as jsonnet. Certain concepts, such as DAG manipulation and so-on I'm planning to carry forward. There's no current plans to remove v1 schema, but will become deprecated in the sense of no more major feature development.

@spinnaker/netflix-reviewers


emjburns:
LGTM

PR #1629, opened 2017-09-19
fix(managed_pipeline): validation of variable types, integration test…

emjburns:
…s for invalid pipelines
Validation of variable types, and integration tests around this.
Also adding integration tests for invalid stage formats, bad versions.

@robzienert PTAL

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/DefaultVariableAssignmentTransform.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/DefaultVariableAssignmentTransform.java
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.PipelineTempla
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.TemplateConfiguration;
 import org.apache.commons.lang3.StringUtils;
 
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -37,7 +38,8 @@ public class DefaultVariableAssignmentTransform implements PipelineTemplateVisit
 
   @Override
   public void visitPipelineTemplate(PipelineTemplate pipelineTemplate) {
-    if (pipelineTemplate.getVariables() == null || pipelineTemplate.getVariables().isEmpty()) {
+    List<Variable> pipelineTemplateVariables = pipelineTemplate.getVariables();
+    if (pipelineTemplateVariables == null || pipelineTemplateVariables.isEmpty()) {
       return;
     }
 
@@ -46,11 +48,11 @@ public class DefaultVariableAssignmentTransform implements PipelineTemplateVisit
       : new HashMap<>();
 
     // if the config is missing vars and the template defines a default value, assign those values from the config
-    pipelineTemplate.getVariables().stream()
+    pipelineTemplateVariables.stream()
       .filter(templateVar -> !configVars.containsKey(templateVar.getName()) && templateVar.hasDefaultValue())
       .forEach(templateVar -> configVars.put(templateVar.getName(), templateVar.getDefaultValue()));
 
-    List<String> missingVariables = pipelineTemplate.getVariables().stream()
+    List<String> missingVariables = pipelineTemplateVariables.stream()
       .filter(templateVar -> !configVars.containsKey(templateVar.getName()))
       .map(Variable::getName)
       .collect(Collectors.toList());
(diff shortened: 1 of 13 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/DefaultVariableAssignmentTransform.java
```diff
@@ -59,6 +60,36 @@ public void visitPipelineTemplate(PipelineTemplate pipelineTemplate) {
       throw new IllegalTemplateConfigurationException("Missing variable values for: " + StringUtils.join(missingVariables, ", "));
     }
 
-    // TODO rz - validate variable values match the defined variable type
+    // collect variables where value type doesn't match the required type
+    List<String> wrongTypeErrorMessages = pipelineTemplateVariables.stream()
+      .filter(templateVar -> {
+        if (missingVariables.contains(templateVar.getName())) {
```
robzienert: This condition will never evaluate to true.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/DefaultVariableAssignmentTransform.java
```diff
@@ -59,6 +60,36 @@ public void visitPipelineTemplate(PipelineTemplate pipelineTemplate) {
...
+
+        if (expectedType.equals("int") && actualType.equals("Integer")){
+          return false;
+        } else if (expectedType.equals("list") && actualType.equals("ArrayList")) {
+          return false;
+        } else if (expectedType.equals(actualType.toLowerCase())) {
+          return false;
+        } else if (expectedType.equals(actualType)){
+          return false; // types match
+        }
```
robzienert: Lets change these all to `equalsIgnoreCase`, then we can get rid of the last conditional.

Further, I'm kinda wondering if matching against the actual type might be a better way forward, so we don't run into any funny business (esp around lists).

```java
String expectedType = templateVar.getType();
if (expectedType.equalsIgnoreCase("object") {
  return false;
}

Class<?> actualType = configVars.get(templateVar.getName()).getClass();

return !((expectedType.equalsIgnoreCase("int") && actualType instanceof Integer.class) ||
  (expectedType.equalsIgnoreCase("bool") && actualType instanceof Boolean) ||
  (expectedType.equalsIgnoreCase("list") && actualType instanceof Collection) ||
  (expectedType.equalsIgnoreCase("string") && actualType instanceof CharSequence.class) ||
  (expectedType.equalsIgnoreCase(actualType.getSimpleName()));
```

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/DefaultVariableAssignmentTransform.java
```diff
@@ -59,6 +60,36 @@ public void visitPipelineTemplate(PipelineTemplate pipelineTemplate) {
...
+        return true; //types don't match, add to list
+      })
+       .map(var -> var.getName() + " (expected type '" + var.getType() + "' found type '" + configVars.get(var.getName()).getClass().getSimpleName() + "')")
```
robzienert: We've already got the actualType, we can use that here rather than getting again.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/DefaultVariableAssignmentTransform.java
emjburns: Removing.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/DefaultVariableAssignmentTransform.java
emjburns: Changing.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/graph/transform/DefaultVariableAssignmentTransform.java
emjburns: Can we access variables from filter in a map stage? They're local variables.

PR #1632, opened 2017-09-20
feat(pipeline_template): Allow granular inheritance control on params, triggers, notifications

robzienert:
Internal teams would like the ability to have some level of control over how params, triggers and notifications are inherited. Adds two new flags, `remove` and `merge`, which are booleans and will control whether or not the config option is either removed from the final pipeline, or its values are merged into with the template values.


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/TemplateMerge.java
```diff
@@ -89,7 +89,13 @@ private static void mergeConfiguration(PipelineTemplate a, PipelineTemplate b) {
       boolean updated = false;
       for (int i = 0; i < merged.size(); i++) {
         if (merged.get(i).getName().equals(bNode.getName())) {
-          merged.set(i, bNode);
+          if (bNode.isRemove()) {
```
danielpeach: I see how this handles inheritance between parent and child templates, but does this bit also handle merge/remove as defined by the template config? Or am I missing something below?

PR #1633, opened 2017-09-20
perf(rollingpush): Avoid unnecessarily searching for instance ids

ajordens:
If a `serverGroupName` or `asgName` is available on the stage context,
there is no need to lookup each instance id individually.

Also, `getServerGroup()` is more efficient than `getServerGroupFromCluster()`
when dealing with clusters containing multiple server groups.


ajordens:
Looking to tune the rolling push and remove some unnecessary lookups.

@duftler / @ttomsu: Not sure how much you guys use rolling push, I don't believe there should be a significant difference (or any at all) between what was coming back in `getServerGroupFromCluster()` vs `getServerGroup()`. 

The way the `Response` was being explicitly converted to a `Map` indicates that `region` was always supplied. If `region` was not supplied, `getServerGroupFromCluster()` would return a `List` as per https://github.com/spinnaker/clouddriver/blob/master/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy#L165.

ajordens:
@anotherchrisberry You originally wrote `TrafficGuard` and I changed it slightly ... hence you being on this PR.

PR #1635, opened 2017-09-20
fix(stages/bakery): Fix bake stage, ensure deploymentDetails is added to context

christopherthielen:
The `deploymentDetails` field in the stage output was not created properly because #1597 changed the stage types for the parallel stages.  This reverts the names of the parallel stages to the basic name (without `.parallel`)

The `Parallel` static Stage Def classes are removed in favor of inline conditionals


jrsquared:
sure

robfletcher:
This puts the actual stage types back as they were before but retains the streamlining of how parallel stages are defined from https://github.com/spinnaker/orca/pull/1597.

ajordens:
Assuming that we sanity checked the UI behaves correctly now?

christopherthielen:
Sanity checked UI based on node quark's use

PR #1636, opened 2017-09-20
feat(pipeline_template): Support template-less configurations

robzienert:
Depends on #1621.

Adds support for creating pipelines with the same syntax as MPT, but without the template. This was spawned by the desire to express early concepts more simply in the new docs that are being written. As an added bonus, this should help enable people to onboard pipelines that simply don't need to have a parent template.


t0mpson:
@rz +1, this is very useful for some production scenarios I encountered and can now simplify thanks to this PR! Thank you, go team pipeline templates! 🍺

PR #1640, opened 2017-09-22
fix(timeouts): some tasks inherit stage timeout override

emjburns:
Added a new type of task `OverridableTimeoutRetryableTask` to indicate tasks (like waiting and monitoring) that want to have the same timeout value as the stage. This came about because the bake stage with an overridden timeout of 2.5 hours failed, because the `monitorBakeTask` has a default timeout value of 1h. 

With this change, some waiting/monitoring stages will not use their default timeout when there is an overridden timeout. I think I grabbed all the `tasks` that make sense, but PTAL @robzienert @robfletcher.

@jrsquared this (should) fix the bake issue.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/MonitorBakeTask.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/MonitorBakeTask.groovy
@@ -17,7 +17,7 @@
 package com.netflix.spinnaker.orca.bakery.tasks
 
 import com.netflix.spinnaker.orca.ExecutionStatus
-import com.netflix.spinnaker.orca.RetryableTask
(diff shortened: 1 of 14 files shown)


ajordens:
I’m not too sure about picking and choosing individual tasks. 

Agree that this should be the default behavior, with the only change that the entire stage will timeout once the stageTimeoutMs has been exhausted. 

Any reason not to do this?

emjburns:
@ajordens I'm not sure about it either, but there are some tasks that it feels like we want to have time out way before the stage times out (the tasks that indicate something is wrong with another service). For example, FindImage or StartBake or any AbstractServerGroup task - if these timeout, it's probably indicative of another problem that the person who controls the pipeline can't solve.

The other way I can think to address it is to take the default timeout values out of the tasks in question, so that they will never time out and the stage timeout will stop the tasks. This seems like a more risky refactor because for all retryable tasks they must have a timeout value. 

Do you have any better ideas on how to solve this?

PR #1644, opened 2017-09-25
fix(pipeline_template): Regression in deserializing tempalted pipeline requests

robzienert:
@jtk54


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/TemplatedPipelineRequest.java
```diff
@@ -32,6 +32,9 @@
 
   @JsonProperty("config")
   private void unpackConfig(Map<String, Object> config) {
+    if (config == null) {
```
robzienert: This is the actual fix.

PR #1646, opened 2017-09-26
fix(web): Enforce limit of pipelines when using previous redis conn

robzienert:
If we define a previous Redis connection, the task repository will get `n` executions from each redis. So, if you limit by 2, 4 results will be returned. This change will enforce that the limit is accurate on the way out, after sorting by start time.

cc @anotherchrisberry @emjburns


asher:
lgtm

emjburns:
LGTM

PR #1648, opened 2017-09-26
feat(artifacts): Add receivedArtifacts to Pipeline model.

jtk54:


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Pipeline.java
```diff
@@ -42,6 +43,16 @@ public Pipeline(
 
   private String pipelineConfigId;
 
+  private List<Artifact> actualArtifacts = new ArrayList<>();
```
lwander: What does this name mean?

jtk54: Distinct from `expectedArtifacts`.

lwander: Ah so these are the ones that are received from the trigger?

lwander: Or do they also include artifacts resolved via the replacement policy?

jtk54: Will add javadoc comment.

jtk54:
Depends on https://github.com/spinnaker/echo/pull/174.

robfletcher:
My one question here would bet whether artifacts belongs as part of the trigger. I realize that's probably a more extensive change as it would require defining the trigger as a type of its own.

Is there an expectation that artifacts will _always_ come from a trigger? Can stages generate artifacts?

lwander:
Yes good point, stages can create artifacts - @jtk54 does that throw a wrench in this?

jtk54:
As of right now, we have only `receivedArtifacts`, which should come from triggers only. We'll add the following in later:

Before the pipeline starts, we'll resolve any missing artifacts and merge them with `receivedArtifacts` into a third set just called `artifacts`. After the pipeline starts, we could have stages append artifacts they produce into the third resolved `artifacts` set. Thoughts?

lwander:
Yup sounds good.

robfletcher:
I'm worried we're reinventing the global context here, which we're trying to move away from. If artifacts come from a trigger, I think they should be in the trigger and if they're created by stages they should be in that stage's `outputs`. What's consuming these artifacts? If it's downstream stages then they are already automatically able to resolve things from upstream stages' `outputs`.

One of the problems we're trying to address in deprecating use of the global context is that any stage can write into it at any time and there's a serious risk of concurrent updates corrupting each other. Although the likelihood is lower with artifacts it's still structured in such a way that I worry concurrent writes to that collection could be a problem.

Stages are guaranteed to be updated in a single thread on a single Orca instance but multiple stages can be running concurrently, likely on multiple instances.

We're already tracking hits on the global context and my next upcoming piece of work is going to be to track those down and standardize them on using the stage outputs mechanism, ultimately getting rid of the global context altogether. I'd really like to go further and have the inputs and outputs of different stage types strongly typed.

jtk54:
Ok sorry for the quick merge -- we should sync offline to discuss this moving forward.

robfletcher:
No worries. I'm not intending to torpedo your effort here but I think it does align with stuff we're thinking about and co-ordinating would be great.

PR #1649, opened 2017-09-27
feat(job): decrease job timout and make overridable

emjburns:
Decreasing the timeout in run job, allowing timeout to be overridden.
@lwander @tomaslin PTAL


lwander:
https://github.com/spinnaker/spinnaker/issues/1961

tomaslin:
LGTM

PR #1655, opened 2017-09-28
feat(stage context): Lookup all stage outputs matching a key

lwander:
I was going to add the trigger output support first, but got stuck on a question about executions that I pinged you about.


lwander:
If you think this is too generic, maybe I can implement `getAll(key)` instead that just aggregates all existing outputs results in a list.

lwander:
fyi @jtk54

lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1915

lwander:
I thought about it and figured that the `getAll()` approach is more sensible, since it's easier to use, and you can still reduce over keys if you need to.

lwander:
ping

PR #1657, opened 2017-09-29
feat(queue): Redis dead letter redrive admin command

robzienert:
⚠️ Not super confident this does all the right behavior WRT changing statuses. Is NOT_STARTED what we always want? I feel like we'll need to alter the state of the execution more meticulously based on the specific message.

### DIFF
--- a/orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisDeadLetterRedriveCommand.kt
+++ b/orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisDeadLetterRedriveCommand.kt
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisDeadLetterRedriveCommand.kt
```diff
@@ -0,0 +1,108 @@
...
+
+  private val mapper = ObjectMapper().apply {
+    registerModule(KotlinModule())
+  }
+
+  override fun redrive(after: Long?, before: Long?) {
+    pool.resource.use { redis ->
+      val redrivenMessages = getMessages(redis, after, before)
+
+      log.warn("Redriving ${redrivenMessages.size} dead lettered messages between ${after?:"-inf"} and ${before?:"+inf"}")
```
asher: I think at least one of [before,after] should be a required argument instead of defaulting to *. Redriving overly old messages could have unintended negative consequences to production services managed by Spinnaker. I think that justifies forcing the operator to consider the time range of DLQ'd messages to redrive.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
```diff
@@ -57,6 +66,15 @@ class AdminController {
     noDiscoveryApplicationStatusPublisher.setInstanceEnabled(enabled)
   }
 
+  @RequestMapping(value = "/queue/redrive", method = RequestMethod.POST)
```
asher: /queue/dlq/redrive? This PR is a good start for additional /queue/dlq/* commands.

PR #1659, opened 2017-10-02
feat(context): Adds trigger contents to stage context

lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1915


lwander:
Merging since this is blocking a few customers, open to change if you take a look later Rob1 or Rob2

PR #1660, opened 2017-10-02
feat(pause): pausing immidiately instead of polling

emjburns:
Pushing pause messages instead of updating the context then pausing on next poll of tasks.
Cascading pause message from execution -> running stages -> running tasks, instead of previous organization (update execution -> next time a task is run & execution paused, pause task -> pause stage tasks belongs to).

@robfletcher PTAL

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
@@ -57,7 +57,7 @@ class MonitorKatoTask implements RetryableTask {
     this.clock = clock
   }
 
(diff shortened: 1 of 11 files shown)


robfletcher:
This all looks very good and soundly implemented.

My only concern is whether there's a possibility of a race condition between the `PauseTaskHandler` and activity going on in `RunTaskHandler`. Consider, if `RunTaskHandler` has started doing something and got past the check for `PAUSED` status but concurrently another Orca instance processes a `PauseTask` message, the `PAUSED` status may subsequently get overwritten when the `RunTaskHandler` completes and sends a `CompleteTask` message.

It would not be an issue if `RunTaskHandler` re-queued its message, only if it pushes a `CompleteTask` message.

I'm not sure what the best solution is, though.

emjburns:
Closing because of the possibility of race conditions. Instead, opening https://github.com/spinnaker/orca/pull/1676 to implement this desired behavior.

PR #1661, opened 2017-10-02
feat(polling): remove polling for waiting stages

emjburns:
Removed polling for wait, manual judgment, and execution window stages.
Message sent to reevaluate the stage at the time it would complete or timeout.
A PATCH to /pipelines/{executionId}/stages/{stageId} triggers a reevaluation of the running tasks for the stage in question.
Dynamic backoffs also implemented in these tasks so that if the task is not finished at the expected time polling will resume on the old backoff schedule (in case of throttling or pausing, for example).

@robfletcher PTAL - specifically at how I'm using the messages and pushing to the queue.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/RestrictExecutionDuringTimeWindow.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/RestrictExecutionDuringTimeWindow.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.pipeline
 
+import java.time.Duration
 import java.util.concurrent.TimeUnit
 import com.google.common.annotations.VisibleForTesting
 import com.netflix.spinnaker.orca.ExecutionStatus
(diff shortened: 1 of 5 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ReevaluateStageHandler.kt
```diff
@@ -0,0 +1,48 @@
...
+
+  val log: Logger
+    get() = LoggerFactory.getLogger(javaClass)
+
+  override val messageType = ReevaluateStage::class.java
+
+  @Suppress("UNCHECKED_CAST")
+  override fun handle(message: ReevaluateStage) {
+    message.withStage { stage ->
+      log.info("Reevaluating stage " + stage.getId() + " (" + stage.getName() + ")")
```
robfletcher: You don't need to concatenate here, just use `${}` inside the string.

emjburns:
Don't merge, I want to wait on https://github.com/spinnaker/orca/pull/1676/files so that I can reuse the `reschedule` method.

PR #1662, opened 2017-10-02
feat(qa): dry run pipelines

robfletcher:
Allow pipelines to run in "dry-run" mode. Stages are proxied and simply test correct evaluation of expressions, optionality, ordering and consistent results.


orca-dry-run/src/main/kotlin/com/netflix/spinnaker/orca/dryrun/DryRunTask.kt
```diff
@@ -0,0 +1,99 @@
...
+      .let { execution ->
+        when (execution) {
+          is Pipeline -> {
+            realStage(execution, stage)
+              .let { realStage ->
+                stage.evaluateStage(realStage)
+              }
+          }
+          else -> {
+            log.error("Dry run is only supported for pipelines")
```
robzienert: Feel like we could use more context on this log.

PR #1664, opened 2017-10-03
feat(moniker): Use a cluster's moniker if available.

andrewbackes:
Deducing the application name of a cluster is handled in two ways:
1. For newer pipeline configs: a moniker will be used to determine a cluster's application.
2. Older pipeline configs: since there isn't a moniker we need to fall back on frigga.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.orca.clouddriver.tasks.cluster
 import com.fasterxml.jackson.core.type.TypeReference
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.frigga.Names
+import com.netflix.spinnaker.moniker.Moniker
 import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.RetryableTask
 import com.netflix.spinnaker.orca.TaskResult
@@ -80,6 +81,7 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
   @Canonical
   static class FindImageConfiguration {
     String cluster
+    Moniker moniker
     List<String> regions
     List<String> zones
     List<String> namespaces
@@ -89,7 +91,7 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
     String imageNamePattern
 
     String getApplication() {
-      Names.parseName(cluster).app
+      moniker?.app ?: Names.parseName(cluster).app
     }
 
     Set<Location> getRequiredLocations() {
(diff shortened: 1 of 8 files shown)


orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTaskSpec.groovy
```diff
@@ -142,6 +143,25 @@ class AbstractClusterWideClouddriverTaskSpec extends Specification {
...
+      moniker    : moniker,
+    ])
+    when:
+    ClusterSelection selection = stage.mapTo(ClusterSelection)
+
+    then:
+    selection.getApplication() == expected
+
+    where:
+    cluster | moniker | expected
```
lwander: I would let intellij auto-align these pipe characters to make this more readable

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTaskSpec.groovy
```diff
@@ -142,6 +143,25 @@ class AbstractClusterWideClouddriverTaskSpec extends Specification {
 
   }
 
+  def 'cluster wide task should use moniker if available or else fallback on frigga'() {
```
lwander: If you attach `@Unroll` here each `where` is given its own test. Can help with making a failing test easier to debug.

lwander: You can also embed `cluster` & `expected` into the test name like this: http://spockframework.org/spock/javadoc/1.0/spock/lang/Unroll.html

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -80,6 +81,7 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
   @Canonical
   static class FindImageConfiguration {
     String cluster
```
lwander: Can we drop cluster? Or are you keeping it around to avoid breaking things for now?

andrewbackes: Yeah, I left that in there to avoid breaking things. This is going to be a pattern throughout the whole process of migrating to Moniker. Most stages/tasks classes already have some subset of the fields that are in Moniker (app, stack, detail, cluster). We should probably have a larger discussion about how we want to handle those in the future. For now I am viewing that optimization as a separate task from the migration.

andrewbackes:
The moniker will be added to the task in Deck. A PR should go up today for it, but you can get an idea by looking at this branch comparison:

https://github.com/spinnaker/deck/compare/master...armory-io:cluster-ops-moniker

Here is an example of a cluster operation generated by Deck:
```
{
  "cloudProvider": "aws",
  "cloudProviderType": "aws",
  "cluster": "nginx",
  "credentials": "aws-dev",
  "name": "Disable Cluster",
  "preferLargerOverNewer": "false",
  "regions": [
    "us-west-2"
  ],
  "remainingEnabledServerGroups": 1,
  "type": "disableCluster"
}
```

When the deck changes are made it will look like this:
```
{
  "cloudProvider": "aws",
  "cloudProviderType": "aws",
  "cluster": "nginx",
  "credentials": "aws-dev",
  "moniker": {
    "app": "nginx",
    "cluster": "nginx"
  },
  "name": "Disable Cluster",
  "preferLargerOverNewer": "false",
  "regions": [
    "us-west-2"
  ],
  "remainingEnabledServerGroups": 1,
  "type": "disableCluster"
}
```

PR #1665, opened 2017-10-04
feat(pipeline_template): Add marker support to disable rendered value expansion

robzienert:
A user needs the ability to pass JSON into EC2 tags while using pipeline templates. The Renderer will automatically attempt to expand json values into objects, but in this case we want the raw value. Adds a rudimentary marker `noexpand:` that can be used to skip the Converter.

cc @jonsie @emjburns


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/YamlRenderedValueConverter.java
```diff
@@ -67,4 +70,12 @@ private static boolean containsEL(String renderedValue) {
   private static boolean isYamlKeyword(String renderedValue) {
     return YAML_KEYWORDS.contains(renderedValue.toLowerCase());
   }
+
+  private static boolean containsNoConvertMarker(String renderedValue) {
+    return renderedValue.startsWith("noconvert:");
```
jonsie: Do you want the marker to be `noconvert` or `noexpand` ?  You have `noconvert` here but your PR comment uses `noexpand`.

robzienert: Whoops. Which do you prefer?

jonsie:
It's a bit of an edge case, but because the Jinja `{% raw %}` block is essentially unusable in the case of JSON I think it's nice to have.  Thanks.

👍

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/YamlRenderedValueConverter.java
jonsie: `noexpand`

PR #1667, opened 2017-10-05
feat(pipeline): Resolve received and expected artifacts in trigger.

jtk54:
In addition, adds artifact resolution to the GAE server group creator. This change should be backwards-compatible.

@ rz and fletch: I couldn't figure out the StageContext issues, so we went with reading the trigger for a first rough cut of this feature.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
@@ -16,8 +16,13 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.providers.appengine
 
+import com.netflix.spinnaker.kork.artifacts.model.Artifact
+import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.ServerGroupCreator
+import com.netflix.spinnaker.orca.pipeline.model.Execution
+import com.netflix.spinnaker.orca.pipeline.model.Pipeline
 import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.pipeline.util.ArtifactResolver
 import groovy.util.logging.Slf4j
 import org.springframework.stereotype.Component
 
@@ -38,6 +43,7 @@ class AppEngineServerGroupCreator implements ServerGroupCreator {
       operation.putAll(stage.context)
     }
 
+    resolveArtifacts(stage, operation)
     operation.branch = AppEngineBranchFinder.findInStage(operation, stage) ?: operation.branch
 
     return [[(OPERATION): operation]]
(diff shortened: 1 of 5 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.groovy
```diff
@@ -0,0 +1,111 @@
...
+                      "Received artifact value did not match expected artifact for fieldName: ${field.getFieldName()}")
+                  break
+                case ExpectedArtifact.ArtifactField.FieldType.FIND_IF_MISSING:
+                  switch (field.getMissingPolicy()) {
+                    case ExpectedArtifact.ArtifactField.MissingPolicy.FAIL_PIPELINE:
+                      throw new ArtifactResolutionException(
+                          "Received artifact value missing, failing pipeline")
+                      break
+                    case ExpectedArtifact.ArtifactField.MissingPolicy.EXPRESSION:
+                      if (!field.getExpression()) {
```
lwander: Shouldn't this be handled already by the expected artifact's validate method?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.groovy
```diff
@@ -0,0 +1,111 @@
...
+      hasMatch?.each { e -> resolvedArtifacts << (receivedArtifacts.find { a -> expectedMatch((ExpectedArtifact) e, (Artifact) a) }) }
+
+      // Otherwise, resolve artifact based on the missing fieldType and missingPolicy.
+      needsLookup?.each { e ->
+        receivedArtifacts.each { a ->
+          e.fields?.each { field ->
+            if (!lookupField((ExpectedArtifact.ArtifactField) field, (Artifact) a)) {
+              switch (field.getFieldType()) {
+                case ExpectedArtifact.ArtifactField.FieldType.MUST_MATCH:
+                  throw new ArtifactResolutionException(
```
lwander: Are you using the exception here to indicate that this particular artifact didn't match?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
```diff
@@ -47,4 +53,43 @@ class AppEngineServerGroupCreator implements ServerGroupCreator {
...
+      if (execution instanceof Pipeline) {
+        trigger = ((Pipeline) execution).getTrigger()
+      }
+
+      List<Map> artifacts = (List<Map>) trigger.artifacts
+      Artifact artifact = (Artifact) artifacts.find { a -> ArtifactResolver.expectedMatch((ExpectedArtifact) expectedArtifact, (Artifact) a) }
+      if (artifact?.reference) {
+        String repositoryUrl = ''
+        switch (artifact.type) {
+          // TODO(jacobkiefer): These object types are pretty fragile, we need to harden this somehow.
```
lwander: I'm guessing this will all be moved to clouddriver? e.g. this this passes the artifact along w/o much question

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.groovy
jtk54: Yes.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
jtk54: Yes, we can do that refactor in the future.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.groovy
jtk54: Yes, should I do a `NotImplemented` or something similar here instead?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.groovy
lwander: I think you should probably not use an exception, or at least one that's different from the exceptions that indicates that something is really wrong. This can happen all the time, right?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.groovy
lwander: Well I don't think you need to do any of the validation work here, instead you can assume validation passed and treat it like that

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
lwander: Sounds good

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.groovy
jtk54: Removed spurious check, this section just handles whether the `Artifact` matches or not.

PR #1672, opened 2017-10-06
(DO NOT MERGE) feat(expressions): Expose getTargetServerGroup() as an expression fn

ajordens:
This proof of concept exposes an extension point through which custom
expression functions can be registered.

The provided example maps `OortService.getServerGroups()` to an
expression function such that the following precondition could be
written:

```
#api_v1_getServerGroups('my_app').?[
  accountName == 'my_account' &&
  cluster == 'my_cluster' &&
  image.imageId == 'ami-xxxxx'
].size() == 0
```

`getServerGroups()` is probably not the right granularity of API to
expose as it can be expensive for large applications.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingOortService.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingOortService.java
@@ -41,8 +41,8 @@ public class DelegatingOortService
   }
 
   @Override
-  public Response getServerGroups(String app) {
-    return getService().getServerGroups(app);
+  public Response getServerGroups(String app, boolean expand) {
+    return getService().getServerGroups(app, expand);
   }
 
   @Override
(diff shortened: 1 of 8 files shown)


ajordens:
This was super quickly put together to see what was possible.

jeyrschabu:
A couple of comments but the approach is very sound. I dig it

ajordens:
@jeyrschabu I'll revise this over the weekend and expose something along the lines of `getTargetServerGroup()` which is what most uses of `jsonFromUrl('http://spinnaker...')` are doing. 

We can mull that over and see if it's something we'd like to merge.

We could also have a newer `utils_v1_jsonFromUrl` (or similar?) that does what I believe you were thinking about wrt to more robust error handling ... I was thinking a response like:

```
{
  status: xxx,
  result: {
  },
  results: [
  ]
}
```

The caller would need to know whether their url returned a map or list but in the event of an error, could still check status and/or deal with an empty response (rather than the whole thing blowing up as it does now)?

Thoughts?

Maybe there's a better way of namespacing functions. It wasn't exactly what I was hoping for but SPeL is a tad restrictive from what I can see.

PR #1680, opened 2017-10-10
Oort get server groups

andrewbackes:
A new `getServerGroup` endpoint that does not require application name was added to clouddriver here: https://github.com/spinnaker/clouddriver/pull/1967

This PR adds it to the Oort service. It also updates the OortHelper to use the new endpoint.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/OortService.groovy
```diff
@@ -43,9 +43,14 @@ interface OortService {
...
   Response getServerGroup(@Path("app") String app,
-                     @Path("account") String account,
-                     @Path("region") String region,
-                     @Path("serverGroup") String serverGroup)
+                          @Path("account") String account,
+                          @Path("region") String region,
+                          @Path("serverGroup") String serverGroup)
+
+  @GET("/serverGroups/{account}/{region}/{serverGroup}")
+  Response getServerGroupByName(@Path("account") String account,
```
lwander: I think it'd be OK to overload the method name here. Also then you could mark the other `getServerGroup` as `@Deprecated`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingOortService.java
```diff
@@ -51,6 +51,11 @@ public Response getServerGroup(String app, String account, String region, String
   }
 
   @Override
+  public Response getServerGroupByName(String account, String serverGroup, String region) {
```
lwander: Same comment here as below

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/DelegatingOortService.java
```diff
@@ -51,6 +51,11 @@ public Response getServerGroup(String app, String account, String region, String
   }
 
   @Override
+  public Response getServerGroup(String account, String serverGroup, String region) {
```
lwander: Maybe flag the prior function with  `@Deprecated` as well here

andrewbackes: Good find. Missed that one. Made the change here: https://github.com/spinnaker/orca/pull/1680/files#diff-044981bc49bd67385ca8beec17533501R48

lwander: Sweet let's merge

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/WaitForNewUpInstancesLaunchTask.groovy
```diff
@@ -45,7 +45,6 @@ class WaitForNewUpInstancesLaunchTask implements RetryableTask {
 
     // similar check in `AbstractInstancesCheckTask`
     def response = oortService.getServerGroup(
-      stageData.application,
```
jonsie: This is backwards incompatible as it breaks the API contract with CloudDriver (unless you update CloudDriver, too).  In my opinion this should not have been merged as written.

Why not try the application specific endpoint after a 404?  The app specific endpoint is marked as deprecated (thank you) and can be logged out as such.  That would give folks some time to update CloudDriver (perhaps as part of a blessed release).

I will probably put up a PR for that if it seems reasonable.

lwander: You should (ideally) be installing Spinnaker from a list of [verified versions](https://www.spinnaker.io/community/releases/versions/). We publish those to avoid having to add all the complexity of supporting multiple endpoints for several releases.

jonsie: We've been trying to stick with the verified versions, but due to various issues across Spinnaker services we inevitably need to update to newer versions while we keep other services set at the verified version.  I agree that ideally we could always stick with the verified versions, but as of yet that has not been the case.

lwander: When issues come up can you file them in spinnaker/spinnaker/issues? This way we can help keep Spinnaker as up-to-date as possible by publishing verified cross-component releases.

jonsie: Sure.  In that same note - how do we prod you guys for a new verified release?  Currently 1.4.2 features a number of bugs around Orca pipeline templates and we can't make any progress until we update both Orca and CloudDriver, which we are hesitant to due because we continue to bump into problems (like this one) in our testing environment.

The fundamental problem I see with trying to bundle Spinnaker verified releases is that you're treating it like a monolithic release, whereas Spinnaker is really a set of microservices.  If semver would be taken a bit more seriously, problems like this wouldn't come about and we could treat Spinnaker like a set of microservices (with confidence) rather than waiting on the "verified release".

ethanfrogers: @jonsie here's a good overview from @jtk54 about the release process for verified releases if you're interested: https://www.youtube.com/watch?v=aQ1lhi1l18s

PR #1681, opened 2017-10-11
fix(rrb): RRB should determine resize percentages at runtime

ajordens:
This PR introduces the ability to specify a `scalePct` when scaling
based on the capacity of another server group (scale_to_server_group).

Previously all RRB capacity adjustments on the target server group were
determined at build time.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
@@ -20,7 +20,7 @@ import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerG
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.DetermineTargetServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroupResolver
-import com.netflix.spinnaker.orca.kato.pipeline.support.SourceResolver
+import com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategy
 import com.netflix.spinnaker.orca.pipeline.WaitStage
 import com.netflix.spinnaker.orca.pipeline.model.Execution
 import com.netflix.spinnaker.orca.pipeline.model.Stage
@@ -65,24 +65,11 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
         cloudProvider                          : cleanupConfig.cloudProvider,
     ]
 
-    Map originalCapacity = stageData.originalCapacity ?: stageData.capacity
-    if (!originalCapacity) {
-      originalCapacity = [
-          min: stageData.targetSize,
-          max: stageData.targetSize,
-          desired: stageData.targetSize
-      ]
-    }
-
     if (stageData.targetSize) {
       stage.context.targetSize = 0
     }
 
     if (stage.context.useSourceCapacity) {
-      List<TargetServerGroup> target = targetServerGroupResolver.resolve(new Stage(null, null, null, baseContext + [target: TargetServerGroup.Params.Target.current_asg_dynamic]))
-      if (target.size() > 0) {
-        originalCapacity = target.get(0).capacity
-      }
       stage.context.useSourceCapacity = false
     }
 
@@ -116,8 +103,11 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
     targetPercentages.forEach({ p ->
       def resizeContext = baseContext + [
           target: TargetServerGroup.Params.Target.current_asg_dynamic,
+          action: ResizeStrategy.ResizeAction.scale_to_server_group,
+          source: getSource(targetServerGroupResolver, stageData, baseContext),
           targetLocation: cleanupConfig.location,
-          capacity: makeIncrementalCapacity(originalCapacity, p)
+          scalePct: p,
+          pinCapacity: p < 100 // if p = 100, capacity should be unpinned
       ]
 
       stages << newStage(
@@ -147,6 +137,7 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
         )
       }
 
+      // TODO-AJ this has questionable behavior w/ > 1 old server group
       stages << newStage(
           stage.execution,
           disableClusterStage.type,
@@ -161,18 +152,33 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
     return stages
   }
 
-  private Map makeIncrementalCapacity(Map originalCapacity, Integer p) {
-
-    if (p == 100) {
-      return originalCapacity
+  static ResizeStrategy.Source getSource(TargetServerGroupResolver targetServerGroupResolver,
+                                         RollingRedBlackStageData stageData,
+                                         Map baseContext) {
+    if (stageData.source) {
+      return new ResizeStrategy.Source(
+        region: stageData.source.region,
+        serverGroupName: stageData.source.serverGroupName ?: stageData.source.asgName,
+        credentials: stageData.credentials ?: stageData.account,
+        cloudProvider: stageData.cloudProvider
+      )
     }
 
-    def desired = (Integer) originalCapacity.desired * (p / 100d)
-    return [
-        min: desired,
-        max: desired,
-        desired: desired
-    ]
+    // no source server group specified, lookup current server group
+    TargetServerGroup target = targetServerGroupResolver.resolve(
+      new Stage(null, null, null, baseContext + [target: TargetServerGroup.Params.Target.current_asg_dynamic])
+    )?.get(0)
+
+    if (!target) {
+      throw new IllegalStateException("No target server groups found (${baseContext})")
+    }
+
+    return new ResizeStrategy.Source(
+      region: target.getLocation().value,
+      serverGroupName: target.getName(),
+      credentials: stageData.credentials ?: stageData.account,
+      cloudProvider: stageData.cloudProvider
+    )
   }
 
   ApplicationContext applicationContext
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -65,24 +65,11 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
         cloudProvider                          : cleanupConfig.cloudProvider,
     ]
 
-    Map originalCapacity = stageData.originalCapacity ?: stageData.capacity
```
ajordens: Realized that this could be problematic if someone ever RRB'd w/o `use source capacity` ... this seems like an abnormal use case. 

Is there a strong desire to maintain this?

RRB'ing to a fixed capacity doesn't seem to jive with disabling certain percentages of an existing server group / cluster?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -147,6 +137,7 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
         )
       }
 
+      // TODO-AJ this has questionable behavior w/ > 1 old server group
```
ajordens: This is a cluster-wide operation whereas you're resizing based on a single server group. 

I would think that we would be disabling percentages of a specific server group and not the overall cluster. Specifically noticed this getting weird when I had more then one active server group. 

Am I missing something?

ajordens:
Not sure it's valid to determine resize capacities at build time, how's this look for moving it to a runtime calculation?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: Oh I think this is here because for a reason I could never find - the strategy is evaluated twice with the same stage data. I could be wrong about this being here for that reason but I would double check.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -116,8 +103,11 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
     targetPercentages.forEach({ p ->
       def resizeContext = baseContext + [
           target: TargetServerGroup.Params.Target.current_asg_dynamic,
+          action: ResizeStrategy.ResizeAction.scale_to_server_group,
+          source: getSource(stageData, baseContext),
           targetLocation: cleanupConfig.location,
-          capacity: makeIncrementalCapacity(originalCapacity, p)
+          scalePct: p,
+          pinCapacity: p < 100 // if p = 100, capacity should be unpinned
```
lwander: What does it mean to pin the capacity?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: I figured we'd want this to disable everything but what we want to deploy. Similar to how red/black uses a cluster wide disable. Does that make sense?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -161,18 +152,33 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
...
+    if (stageData.source) {
+      return new ResizeStrategy.Source(
+        region: stageData.source.region,
+        serverGroupName: stageData.source.serverGroupName ?: stageData.source.asgName,
+        credentials: stageData.credentials ?: stageData.account,
+        cloudProvider: stageData.cloudProvider
+      )
+    }
+
+    // no source server group specified, lookup current at build time
```
lwander: I don't follow the build time comment - do you mean it's build time because we're inserting a dynamic stage now?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
ajordens: I'm not sure we ever need to care about capacity stored on the stage data ... I reckon the UI would let you set an explicit capacity right now but that doesn't make a lot of sense to me? 

Would be problematic if any autoscaling had occurred (which is the motivation behind making all this runtime / dynamic).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
ajordens: min = max = desired

`scale_to_server_group` already supported pinning min = desired, this just builds on that to min, max and desired.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
ajordens: I think the bit that confused me is that I saw (or at least I thought I saw) it happened after the first % rollout completed. That struck me as a little weird is all.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
ajordens: I need to actually rejig this further ... and move the source determination to something along the lines of ancestor + `DetermineTargetServerGroup` and make it determined at runtime when the deploy stage starts.

It's build time right now because this is all happening when the pipeline / stages are constructed.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: I think the capacity is coming from whatever the capacity is when the stage starts - but you're right - it should be updated to reflect any scaling

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: Huh do you mean it disabled something incorrectly after the new SG grew for the first time?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: Ah sounds good

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
ajordens: This PR reflects that.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
ajordens: Hrmph. Just ignore me ... at this point not seeing any premature disables, so that's good.

Every pre-existing server group gets the disable percentage treatment , that seems alright as well.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
ajordens: Turns out v3 makes this a-ok ... the determination isn't made until the deploy stage is constructed.

I'm otherwise happy with this PR, alright to merge? 

We've had a request to do something other than a wait between rollouts, so I'll start looking at supporting pipelines next.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: heh yup was agreeing with you

PR #1683, opened 2017-10-11
chore(loadbalancer): Include application in upsert load balancer request

jeyrschabu:
- Adding the application name in the request to clouddriver


ajordens:
Can you shed light on why we need to send application down? Is not not derivable from the load balancer name?

It looked like `application` would have previously been sent down if it was present in the stage context.

jeyrschabu:
Looking back, deck might be a better place to include `application` in the request.

@ajordens I opted not to derive the application from the elb name with frigga because the elb name might be arbitrary.

jeyrschabu:
Closing this one. Incorrect assumptions were made.

PR #1688, opened 2017-10-12
feat(pipeline_template): PipelineIdTag checks context for variables defining application and name

jonsie:
@robzienert


orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/tags/PipelineIdTagSpec.groovy
```diff
@@ -77,7 +78,26 @@ class PipelineIdTagSpec extends Specification {
     '{% pipelineId name="Rob\'s great pipeline" %}'          || '1685429e-beb1-4d35-963c-123456789012'
   }
 
-  def 'should handle missing input params'() {
```
robzienert: Can you reformat this so it's consistent with the rest of the file?

jonsie: Should be better now.

PR #1689, opened 2017-10-12
feat(moniker): Use moniker for instance check task

andrewbackes:
First try to use the moniker for server group and cluster names. If a moniker can not be found, resort to using frigga.


andrewbackes:
@lwander FYI

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
```diff
@@ -195,16 +197,18 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
     }
   }
 
-  private List<Map> fetchServerGroups(String account, String cloudProvider, Map<String, List<String>> serverGroupsByRegion) {
-    Names names = Names.parseName(serverGroupsByRegion.values().flatten()[0])
-
+  private List<Map> fetchServerGroups(String account, String cloudProvider, Map<String, List<String>> serverGroupsByRegion, Map<String, String> moniker) {
```
lwander: Why not use the proper `Moniker` class here?

PR #1690, opened 2017-10-12
Remove unused frigga import

andrewbackes:
This was missed when removing frigga from the cluster based tasks in a prior PR.


andrewbackes:
@lwander FYI

PR #1693, opened 2017-10-13
feat(moniker): Use monikers within server-group tasks

andrewbackes:
Use monikers in server group tasks if available. If not available, fall back on frigga.


andrewbackes:
@lwander FYI

PR #1695, opened 2017-10-13
fix(pipeline_template): Add back required field check in PipelineIdTag

jonsie:
@robzienert


jonsie:
Closing since this condition will always evaluate to true, thus the exception will never be thrown.  Additionally, there is a test that validates a `TemplateRenderException` is thrown if the name parameter is missing.

PR #1696, opened 2017-10-13
feat(artifacts): simplify artifact matching

lwander:


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -147,6 +147,7 @@ class OperationsController {
         pipeline.trigger.parameters[it.name] = pipeline.trigger.parameters.containsKey(it.name) ? pipeline.trigger.parameters[it.name] : it.default
       }
     }
+
```
jtk54: Inadvertent?

lwander: whoopsie

PR #1699, opened 2017-10-14
feat(moniker): Use moniker for Job stages.

andrewbackes:
First try to use a moniker, then fallback on the application name, then fallback on frigga.

There will be a Deck change to add the moniker to the stage JSON.


andrewbackes:
@lwander FYI

PR #1702, opened 2017-10-16
feat(moniker): Use moniker in DetermineHealthProvidersTask

andrewbackes:
When application name can't be found in the context, try to use the moniker before frigga.


andrewbackes:
@lwander FYI

cfieber:
this doesn't work - a stage from the UI that includes a moniker payload (Shrink Cluster for example) fails with 
```
java.lang.ClassCastException: java.util.LinkedHashMap cannot be cast to com.netflix.spinnaker.moniker.Moniker
	at com.netflix.spinnaker.orca.clouddriver.tasks.DetermineHealthProvidersTask.execute(DetermineHealthProvidersTask.java:101)
```

lwander:
Sorry for the trouble - fix incoming

PR #1703, opened 2017-10-16
feat(moniker): Use moniker for Rollingpush tasks.

andrewbackes:
Similarly to how I've done the previous PRs for monikers, first try to use the moniker then fallback on frigga.


andrewbackes:
@lwander FYI

PR #1708, opened 2017-10-17
fix(moniker): Check if moniker exists before casting

lwander:
https://github.com/spinnaker/orca/pull/1702

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
@@ -22,6 +22,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.frigga.Names;
 import com.netflix.spinnaker.moniker.Moniker;
 import com.netflix.spinnaker.orca.ExecutionStatus;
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
```diff
@@ -98,7 +105,11 @@ public TaskResult execute(Stage stage) {
 
     try {
       String applicationName = (String) stage.getContext().get("application");
-      Moniker moniker = (Moniker) stage.getContext().get("moniker");
+      Moniker moniker = null;
+      if (stage.getContext().containsKey("moniker")) {
+        moniker = objectMapper.convertValue(stage.getContext().get("moniker"), Moniker.class);
```
ajordens: The `stage.mapTo("/moniker", Moniker.class)` is more consistent with other uses IMO ... also does not require explicit injection of an `objectMapper`.

Any reason it didn't work for you?

cfieber: `stage.mapTo("/moniker", Moniker.class)`

lwander: Oh neat, will do that

lwander: I didn't know it existed - will switch

lwander:
@andrewbackes you probably have a better handle of all the places we try to pull moniker out of the stage context - do want to take this from here?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/DetermineHealthProvidersTask.java
ajordens: 👍

andrewbackes:
Yeah

lwander:
Cool thanks Andrew

PR #1711, opened 2017-10-18
fix(canary): gentle canary cleanup

asher:
When canaries finish successfully: disable canaryAsg + waitTask(2min) + disable baselineAsg + waitTask(2min) + cache refresh, then proceed with deletion of both canary and baseline as before.

This addresses issues resulting from both canary and baseline asg's getting force deleted (which deletes asg lifecycle hooks) with simultaneous instance terminations (which also bypasses lifecycle hooks) while still serving traffic.

When canaries are terminated early and CanaryStage.cancel() invoked, asg's are also disabled prior to deletion.

### DIFF
--- a/orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/pipeline/CanaryStage.groovy
+++ b/orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/pipeline/CanaryStage.groovy
@@ -17,8 +17,10 @@
 package com.netflix.spinnaker.orca.mine.pipeline
 
 import com.netflix.spinnaker.orca.CancellableStage.Result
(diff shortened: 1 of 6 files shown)


robzienert:
I feel like we should.make the wait duration configurable via the stage context. Default to 2 minutes, but if 2 minutes is overkill for someone, or not enough, itd be nice to make that adjustment without a redeploy.

ajordens:
Seems reasonable to me ... would be nice to remove the initial `Thread.sleep(2 minutes)` in `cancel()`. That is only present to handle situations where the stage failed while deploying the clusters ... if the clusters were deployed, there's no need to wait 2 minutes before disabling.

Even better would be to have these sorts of operations run as new stages in the pipeline itself to provide some additional context to the user, but we're probably a ways away from supporting that.

PR #1713, opened 2017-10-18
fix(polling): update time left calculation

emjburns:
Account for pausing in time calculation.

@robfletcher @robzienert PTAL


robzienert:
Not sure I understand what's being fixed here?

emjburns:
@robzienert Should have provided more explanation. Any time run task is called, if the task is not supposed to be finished, it will calculate the end time not taking into account how long the task has been running. So, if you pause a pipeline during a wait task, and then resume it before it is done waiting, it will put another 'runTask' message on the queue for the entire time it is supposed to be waiting.

This fixes it so that it will deliver that runTask message at the right time (waitTime - timeAlreadyWaited + 0.5 seconds)

PR #1716, opened 2017-10-19
feat(rollback): Support rolling back to a server group that no longer…

ajordens:
… exists

This PR provides a rollback strategy that will clone forward with the
image details that were tagged as part of spinnaker/orca#1705.

It also supports the `imageName` being explicitly provided, but that's
an exceptional case that would not be supported in the UI.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/PreviousImageRollback.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/PreviousImageRollback.groovy
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


ajordens:
Now that we've got > 1 rollback strategy, I took the opportunity to refactor the `Rollback` implementations and tests slightly.

They've been extracted from `RollbackServerGroupStage`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/PreviousImageRollback.groovy
```diff
@@ -0,0 +1,144 @@
...
+import com.netflix.spinnaker.orca.clouddriver.OortService
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.CloneServerGroupStage
+import com.netflix.spinnaker.orca.pipeline.model.Execution
+import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner
+import org.springframework.beans.factory.annotation.Autowired;
+
+import static com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder.newStage
+
+class PreviousImageRollback implements Rollback {
```
ajordens: This is the new bit

PR #1718, opened 2017-10-19
feat(pipeline_template): Render configuration for templated pipelines with dynamic source

jervi:
When you try to edit a pipeline based on a template with dynamic source, you're greeted with the 
following message in Deck:

skjermbilde 2017-10-19 kl 15 38 23

This commit makes Orca try to fetch ~the last~ any previously ran pipeline execution and use the context of that to render the source field, thus getting rid of the above error. If you press the "Configure Template" button, you're able to edit the variables. This commit also fix so that the graph renders correctly, going from this:
skjermbilde 2017-10-19 kl 15 39 38

To this:
skjermbilde 2017-10-31 kl 13 35 22


jervi:
@spinnaker/reviewers PTAL

PR #1727, opened 2017-10-20
feat(moniker): Use moniker in TrafficGuard.

andrewbackes:
This PR will remove the use of frigga from `TrafficGuard`.

### Flow:
First, `Map<String, Object> operation` is generated from a `Stage` object in `TargetServerGroup.toClouddriverOperationPayload()` and `AbstractServerGroupTask.convert()`. 
`Stage` object may or may not have a moniker. If it is an older pipeline config, then it will not have a moniker.
`operation` is used by `validateClusterStatus()` in 
  - `TerminateInstanceAndDecrementServerGroupTask.groovy`
  - `TerminateInstancesTask.groovy`
  - `BulkDestroyServerGroupTask.java`
  - `BulkDisableServerGroupTask.java`
  - `DestroyServerGroupTask.groovy`
  - `DisableServerGroupTask.groovy`
  - `ResizeServerGroupTask.groovy`

When calling `TrafficGuard.verifyTrafficRemoval()` a moniker is generated (via MonikerHelper and frigga) if one can't be found in the `operation`. Once we transition old pipeline configs to all use monikers, we can remove generating the moniker. In the case of the new k8s provider, a moniker is always expected to be present.

### Testing

To test, I made an app in Spinnaker and set Traffic Guards. Then I made a pipeline that tries to disable, resize, and destroy the guarded server group. Then I made a copy of the pipeline and removed the monikers. 

You can find the pipeline configurations here:
- https://github.com/Armory/spinnaker-test-pipelines/blob/master/json/trafficguard-moniker-test.json
- https://github.com/Armory/spinnaker-test-pipelines/blob/master/json/trafficguard-no-moniker-test.json

After running the pipelines, I got the expected results:

Pipeline Execution History

Note: These pipelines will be added to our internal test suite that runs against our installation of Spinnaker.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
@@ -82,7 +82,7 @@ class TargetServerGroup {
     // serverGroup.moniker is a Map type, but Groovy is able to convert it to a Moniker type.
     return serverGroup.moniker
   }
-  
+
   Map toClouddriverOperationPayload(String account) {
     //TODO(cfieber) - add an endpoint on Clouddriver to do provider appropriate conversion of a TargetServerGroup
     def op = [
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support
 
+import com.netflix.spinnaker.orca.clouddriver.utils.MonikerHelper
 import groovy.transform.InheritConstructors
 import groovy.transform.ToString
 import groovy.util.logging.Slf4j
@@ -78,6 +79,7 @@ class TargetServerGroup {
    * Used in TrafficGuard, which is Java, which doesn't play nice with @Delegate
    */
   Moniker getMoniker() {
+    // serverGroup.moniker is a Map type, but Groovy is able to convert it to a Moniker type.
     return serverGroup.moniker
   }
 
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
@@ -82,8 +82,7 @@ class TargetServerGroup {
     // serverGroup.moniker is a Map type, but Groovy is able to convert it to a Moniker type.
     return serverGroup.moniker
   }
-
-
+  
   Map toClouddriverOperationPayload(String account) {
     //TODO(cfieber) - add an endpoint on Clouddriver to do provider appropriate conversion of a TargetServerGroup
     def op = [
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
@@ -29,7 +29,8 @@ import com.netflix.spinnaker.orca.pipeline.model.Stage
  */
 class TargetServerGroup {
   // Delegates all Map interface calls to this object.
-  @Delegate private final Map<String, Object> serverGroup
+  @Delegate
+  private final Map<String, Object> serverGroup
 
   TargetServerGroup(Map<String, Object> serverGroupData) {
     if (serverGroupData.instances && serverGroupData.instances instanceof Collection) {
@@ -73,12 +74,21 @@ class TargetServerGroup {
     return (serverGroup.instances ?: []) as List<Map>
   }
 
+  /**
+   * Used in TrafficGuard, which is Java, which doesn't play nice with @Delegate
+   */
+  Moniker getMoniker() {
+    return serverGroup.moniker
+  }
+
+
   Map toClouddriverOperationPayload(String account) {
     //TODO(cfieber) - add an endpoint on Clouddriver to do provider appropriate conversion of a TargetServerGroup
     def op = [
       credentials    : account,
       accountName    : account,
       serverGroupName: serverGroup.name,
+      moniker        : serverGroup.moniker,
       asgName        : serverGroup.name,
       cloudProvider  : serverGroup.cloudProvider ?: serverGroup.type
     ]
@@ -173,19 +183,21 @@ class TargetServerGroup {
       /**
        * "Previous Server Group"
        */
-      ancestor_asg_dynamic,
+        ancestor_asg_dynamic,
       /**
        * "Oldest Server Group"
        */
-      oldest_asg_dynamic,
+        oldest_asg_dynamic,
       /**
        * "(Deprecated) Current Server Group"
        */
-      @Deprecated current_asg,
+        @Deprecated
+        current_asg,
       /**
        * "(Deprecated) Last Server Group"
        */
-      @Deprecated ancestor_asg,
+        @Deprecated
+        ancestor_asg,
 
       boolean isDynamic() {
         return this.name().endsWith("dynamic")
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.orca.clouddriver.utils.MonikerHelper
 import groovy.transform.InheritConstructors
 import groovy.transform.ToString
@@ -79,8 +80,8 @@ class TargetServerGroup {
    * Used in TrafficGuard, which is Java, which doesn't play nice with @Delegate
    */
   Moniker getMoniker() {
-    // serverGroup.moniker is a Map type, but Groovy is able to convert it to a Moniker type.
-    return serverGroup.moniker
+    ObjectMapper objectMapper = new ObjectMapper();
+    return objectMapper.convertValue(serverGroup.moniker, Moniker);
   }
 
   Map toClouddriverOperationPayload(String account) {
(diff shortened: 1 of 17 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/MonikerHelper.java
```diff
@@ -58,4 +59,20 @@ static public Moniker monikerFromStage(Stage stage) {
       return null;
     }
   }
+
+  static public Moniker monikerFromStage(Stage stage, String fallbackFriggaName) {
+    Moniker moniker = monikerFromStage(stage);
```
ajordens: Q: what happens when you have a stage that's dealing with potentially > 1 server group or cluster? 

`stage.context.moniker` seems to imply there only ever being one, and nothing in the context key indicates it's intent.

The rollback stages are a current example with > 1 server group specified.

andrewbackes: So in the case of TrafficGuard, the moniker:
```
"moniker" : {
      "app" : "nginx",
      "cluster" : "nginx-trafficguards",
      "detail" : null,
      "sequence" : 11,
      "stack" : "trafficguards"
    }
```
would be passed through. In TrafficGuard itself only `moniker.cluster` is used. In `ClusterMatcher` only `moniker.stack` and `moniker.detail` are used. `moniker.sequence` is not used. So the moniker is more of a cluster moniker than a server group moniker.

andrewbackes: @ajordens Do you know an example of a stage that uses multiple clusters?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/BulkDisableServerGroupTask.java
```diff
@@ -37,6 +38,7 @@ String getClouddriverOperation() {
   @Override
   void validateClusterStatus(Map<String, Object> operation) {
     trafficGuard.verifyTrafficRemoval((String) operation.get("serverGroupName"),
+      (Moniker) operation.get("moniker"),
```
lwander: Does this require an `ObjectMapper`? I remember that biting us in a prior PR

lwander: (there's a few other places in this PR that does the same)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/MonikerHelper.java
```diff
@@ -58,4 +59,28 @@ static public Moniker monikerFromStage(Stage stage) {
...
+    return monikerOrFrigga(moniker, fallbackFriggaName);
+  }
+
+  static public Moniker monikerOrFrigga(Moniker moniker, String friggaName) {
+    if (moniker == null) {
```
ajordens: If interested, a one liner equivalent: `Optional.ofNullable(moniker).orElse(friggerToMoniker(friggaName)`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -91,22 +90,19 @@ public void verifyInstanceTermination(String serverGroupNameFromStage,
...
     }
 
-    verifyOtherServerGroupsAreTakingTraffic(serverGroupName, location, account, cloudProvider, operationDescriptor);
+    verifyOtherServerGroupsAreTakingTraffic(serverGroupName, serverGroupMoniker,location, account, cloudProvider, operationDescriptor);
   }
 
-  private void verifyOtherServerGroupsAreTakingTraffic(String serverGroupName, Location location, String account, String cloudProvider, String operationDescriptor) {
-    Names names = Names.parseName(serverGroupName);
-    Optional<Map> cluster = oortHelper.getCluster(names.getApp(), account, names.getCluster(), cloudProvider);
+  private void verifyOtherServerGroupsAreTakingTraffic(String serverGroupName, Moniker serverGroupMoniker, Location location, String account, String cloudProvider, String operationDescriptor) {
```
ajordens: Do we still need to pass around `serverGroupName` if we've got a moniker? Maybe we do now as a transition but couldn't you compare moniker's instead of `!serverGroupName.equals(tsg.getName()) &&`?

andrewbackes: The server group name isn't contained in th moniker object (only app, stack, detail, cluster, sequence). In the case of the Frigga naming convention, you can reconstruct the name using the moniker object. But that reconstruction won't work for the manifest based k8s provider.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/MonikerHelper.java
lwander: Or `moniker == null ? friggaToMoniker(friggaName) : moniker`

ajordens: That’s even better.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
ajordens: Is it silly to ask why we couldn’t have it also contain the serverGroupName?

ajordens:
LGTM otherwise. 

@lwander my original comment about moniker being singular was rooted in the fact that there is no constraint that a stage only operates on a single cluster / server group. I would not be surprised if this causes some grief with some future stage impl, or hinders being able to use monikers everywhere that we would otherwise use Frigga.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
andrewbackes: @lwander can probably give you a more thorough explanation. From my POV, I think that it isn't necessary to facilitate the migration away from frigga. It would also require a significant amount of effort. If you search for `serverGroupName` in Orca there are 362 occurrences.

FYI: This is related to another point. There is now a lot of redundancy with application/stack/detail in the stage data payload. In many places, those fields are already included at the root level of the stage and are now in the moniker as well. A follow up task will probably be to consolidate those.

lwander: No that's a good question. We don't want the moniker to be the identifier for the object since the point of the moniker is to decouple the name from the relationships (app, cluster) that frigga presented us. The separation of concerns is: `name` is a way to identify an object and `moniker` is to decide which cluster & application an object belongs to. 

When you assign or derive a moniker from an object, that object doesn't even have to provide us with a name, (e.g. ec2 tags), so the idea is really to keep these concepts separate.

ajordens: My point about including the name was somewhat predicated in that when I hear `Moniker`, I think of name. 

The relationships are really part of the identity of an object, which is why we need to pass application/cluster/region/name/etc. around everywhere.

My biggest grief is that we're slowly developing more and more ways to specify the coordinates for an object: serverGroupName, asgName, regions, region, zone, zones, moniker, etc. 

In the case of moniker it's only dealing with names, so we can ignore regions/region/zone/zones and think about it only as a partial coordinate.

I hold out hope that there will be some consolidation, as it's difficult and error prone to keep everything straight.

Just my 2c. 

I think it would be cleaner and possible to pass a Moniker into validateClusterStatus but otherwise this PR fits the existing patterns and can be merged.

lwander: My hope is this kind of clean up/refactor will point out more places where we can consolidate the many names/coordinates you point out. We'll keep this in mind going forward.

andrewbackes:
@ajordens I believe I have addressed all of your prior comments. If you wouldn't mind give it another look to make sure my changes fall in line with what you had in mind, I'd appreciate it.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
```diff
@@ -73,6 +76,14 @@ class TargetServerGroup {
     return (serverGroup.instances ?: []) as List<Map>
   }
 
+  /**
+   * Used in TrafficGuard, which is Java, which doesn't play nice with @Delegate
+   */
+  Moniker getMoniker() {
+    ObjectMapper objectMapper = new ObjectMapper();
```
ajordens: Should inject an `ObjectMapper` into the constructor (it'll be autowired) vs explicitly creating one here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/AbstractBulkServerGroupTask.java
```diff
@@ -104,8 +104,9 @@ public TaskResult execute(Stage stage) {
     targetServerGroups.forEach( targetServerGroup -> {
       Map<String , Map> tmp = new HashMap<>();
       Map operation = targetServerGroup.toClouddriverOperationPayload(request.getCredentials());
-
-      validateClusterStatus(operation);
+      Moniker moniker = targetServerGroup.getMoniker() == null ?
```
ajordens: Slightly cleaner alternative to a somewhat ugly multiline ternary.

```
Moniker moniker = targetServerGroup.getMoniker();
if (moniker == null) {
  moniker = MonikerHelper.friggaToMoniker(targetServerGroup.getName())
}
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/AbstractServerGroupTask.groovy
```diff
@@ -108,6 +110,17 @@ abstract class AbstractServerGroupTask extends AbstractCloudProviderAwareTask im
     operation
   }
 
+  Moniker convertMoniker(Stage stage) {
+    if (TargetServerGroup.isDynamicallyBound(stage)) {
+      TargetServerGroup tsg = TargetServerGroupResolver.fromPreviousStage(stage);
+      return  tsg.getMoniker() == null ?  MonikerHelper.friggaToMoniker(tsg.getName()) : tsg.getMoniker();
+    } else {
```
ajordens: don't need the `else {`

and since this is groovy, we can do:

`return MonikerHelper.monikerFromStage(stage, serverGroupName ?: asgName)`

andrewbackes:
@ajordens I took care of the style changes. I'm not really sure what to do about autowiring `ObjectMapper` into `TargetServerGroup` since it isn't a bean. I might be missing something here. For now I moved `objectMapper` to a `final static` member.

ajordens:
odd ... I don't know wth I was looking at to think it was a `Component`. 

I've updated the branch and will merge when the build passes.

PR #1732, opened 2017-10-24
feat(moniker): Pass moniker to cleanup stages.

andrewbackes:
Right now Deck will generate something like this when a user creates a new deploy stage in a pipeline:

```
{
  "clusters": [
    {
      "account": "aws-dev",
      "application": "nginx",
      "associatePublicIpAddress": true,
      "availabilityZones": {
        "us-west-2": [
          "us-west-2a"
        ]
      },
      "cloudProvider": "aws",
      "freeFormDetails": "",
      "instanceType": "t2.micro",
      "moniker": {
        "app": "nginx",
        "cluster": "nginx-cluster1",
        "detail": "",
        "stack": "cluster1"
      },
      "provider": "aws",
      "stack": "cluster1",
      "strategy": "redblack",
      "subnetType": "external (spinnaker)",
      "terminationPolicies": [
        "Default"
      ]
    }
  ],
  "name": "Deploy",
  "type": "deploy"
}
```

A deploy stage (redblack/highlander/rolling/etc) will generate several sub-stages. This PR will allow the cluster's moniker to be passed to the cleanup sub-stages (shrink/destroy/disable). Other methods require the moniker down stream (`AbstractClusterWideClouddriverTask.ClusterSelection.getApplication()` for example.)

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/AbstractDeployStrategyStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/AbstractDeployStrategyStage.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
 
+import com.netflix.spinnaker.moniker.Moniker
 import com.netflix.spinnaker.orca.clouddriver.pipeline.AbstractCloudProviderAwareStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.Location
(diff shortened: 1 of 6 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/AbstractDeployStrategyStage.groovy
```diff
@@ -153,10 +154,10 @@ abstract class AbstractDeployStrategyStage extends AbstractCloudProviderAwareSta
     return className
   }
 
-  @Immutable
```
andrewbackes: Moniker is not supported by Immutable.

PR #1734, opened 2017-10-24
feat(provider/kubernetes): support to orchestrate update and delete o…

chlung:
About this PR,

- orchestrate update strategies operation for statefulset and daemonset.
Add a new strategy (recreateupdate) for supporting OnDelete update in order to handle resize statefulset and daemonset pod(s) to 0 before update controller definition.
- orchestrate delete operation for statefulset and daemonset.
Remove this operation from original PR and leverage current orca task flow.  Disable server group logic, which is not applicable,  will be controlled in clouddriver.

Thanks.
…perations for new controllers.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2017 Cisco, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
+
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerGroupStage
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.DetermineTargetServerGroupStage
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroupResolver
+import com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategy
+import com.netflix.spinnaker.orca.pipeline.model.Execution
+import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner
+import groovy.util.logging.Slf4j
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.ApplicationContext
+import org.springframework.context.ApplicationContextAware
+import org.springframework.stereotype.Component
+
+import static com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder.newStage
+
+@Slf4j
+@Component
+class RecreateUpdateStrategy implements Strategy, ApplicationContextAware {
+  final String name = "recreateupdate"
+
+  @Autowired
+  ResizeServerGroupStage resizeServerGroupStage
+
+  @Autowired
+  DetermineTargetServerGroupStage determineTargetServerGroupStage
+
+  @Autowired
+  TargetServerGroupResolver targetServerGroupResolver
+
+  @Override
+  List<Stage> composeFlow(Stage stage) {
+    def stages = []
+    def stageData = stage.mapTo(RecreateUpdateStrategyData)
+    def cleanupConfig = AbstractDeployStrategyStage.CleanupConfig.fromStage(stage)
+
+    Map baseContext = [
+      (cleanupConfig.location.singularType()): cleanupConfig.location.value,
+      cluster                                : cleanupConfig.cluster,
+      credentials                            : cleanupConfig.account,
+      cloudProvider                          : cleanupConfig.cloudProvider,
+    ]
+
+    stage.context.capacity = [
+      min    : 0,
+      max    : 0,
+      desired: 0
+    ]
+
+    def p = 0
+    def findContext = baseContext + [
+      target        : TargetServerGroup.Params.Target.current_asg_dynamic,
+      targetLocation: cleanupConfig.location,
+    ]
+
+    stages << newStage(
+      stage.execution,
+      determineTargetServerGroupStage.type,
+      "Determine Deployed Server Group",
+      findContext,
+      stage,
+      SyntheticStageOwner.STAGE_BEFORE
+    )
+
+    def source
+    try {
+      source = getSource(targetServerGroupResolver, stageData, baseContext)
+    } catch (Exception e) {
+      stages = []
+      return stages
+    }
+
+    def resizeContext = baseContext + [
+      target        : TargetServerGroup.Params.Target.current_asg_dynamic,
+      action        : ResizeStrategy.ResizeAction.scale_exact,
+      source        : source,
+      targetLocation: cleanupConfig.location,
+      kind          : stageData.kind,
+      capacity      : stage.context.capacity,
+      resizeType    : "exact",
+      scaleNum      : 0
+    ]
+
+    def resizeStage = newStage(
+      stage.execution,
+      resizeServerGroupStage.type,
+      "Scale down to $p% Desired Size(0)",
+      resizeContext,
+      stage,
+      SyntheticStageOwner.STAGE_BEFORE
+    )
+    stages << resizeStage
+
+    return stages
+  }
+
+  static ResizeStrategy.Source getSource(TargetServerGroupResolver targetServerGroupResolver,
+                                         RecreateUpdateStrategyData stageData,
+                                         Map baseContext) {
+    if (stageData.source) {
+      return new ResizeStrategy.Source(
+        region: stageData.source.region,
+        serverGroupName: stageData.source.serverGroupName ?: stageData.source.asgName,
+        credentials: stageData.credentials ?: stageData.account,
+        cloudProvider: stageData.cloudProvider
+      )
+    }
+
+    // no source server group specified, lookup current server group
+    TargetServerGroup target = targetServerGroupResolver.resolve(
+      new Stage(null, null, null, baseContext + [target: TargetServerGroup.Params.Target.current_asg_dynamic])
+    )?.get(0)
+
+    return new ResizeStrategy.Source(
+      region: target.getLocation().value,
+      serverGroupName: target.getName(),
+      credentials: stageData.credentials ?: stageData.account,
+      cloudProvider: stageData.cloudProvider
+    )
+  }
+  ApplicationContext applicationContext
+}
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategyData.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategyData.groovy
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2017 Cisco, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
+
+import com.netflix.spinnaker.orca.kato.pipeline.support.StageData
+
+class RecreateUpdateStrategyData extends StageData {
+  String kind
+  Map capacity
+  Integer targetSize
+}
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/strategy/Strategy.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/strategy/Strategy.java
@@ -25,6 +25,7 @@ public enum Strategy implements StrategyFlowComposer{
   RED_BLACK("redblack"),
   HIGHLANDER("highlander"),
   ROLLING_PUSH("rollingpush"),
+  RECREATE_UPDATE("recreateupdate"),
   CUSTOM("custom"),
   NONE("none");
 
@@ -49,7 +50,7 @@ public enum Strategy implements StrategyFlowComposer{
 
   @Override
   public boolean replacesBasicSteps() {
-    return this == ROLLING_PUSH || this == CUSTOM;
+    return this == ROLLING_PUSH || this == CUSTOM || this == RECREATE_UPDATE;
   }
 
   @Override
@@ -61,6 +62,8 @@ public enum Strategy implements StrategyFlowComposer{
         return builder.composeHighlanderFlow(stage);
       case ROLLING_PUSH:
         return builder.composeRollingPushFlow(stage);
+      case RECREATE_UPDATE:
+        return builder.composeRecreateUpdateFlow(stage);
       case CUSTOM:
         return builder.composeCustomFlow(stage);
     }
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/strategy/DeployStrategyStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/strategy/DeployStrategyStage.groovy
@@ -210,6 +210,10 @@ abstract class DeployStrategyStage extends AbstractCloudProviderAwareStage {
     return stages
   }
 
+  protected List<Stage> composeRecreateUpdateFlow(Stage stage) {
+    return []
+  }
+
   protected List<Stage> composeCustomFlow(Stage stage) {
     def stages = []
     def cleanupConfig = determineClusterForCleanup(stage)


chlung:
@lwander Please review.  Thanks.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
```diff
@@ -43,11 +48,28 @@ class CreateServerGroupStage extends AbstractDeployStrategyStage {
...
 
-    def tasks = [
-      TaskNode.task("createServerGroup", CreateServerGroupTask),
-      TaskNode.task("monitorDeploy", MonitorKatoTask),
-      TaskNode.task("forceCacheRefresh", ServerGroupCacheForceRefreshTask),
-    ]
+    def tasks = []
+    String kind = stage.context?.get("kind")
+
+    if (stage.context?.get("kind").toString() == KUBERNETES_STATEFULSET_CONTROLLER_TYPE ||
```
lwander: I don't think that this is the way to handle this. These stages are meant to be generic, and putting top level references to k8s resources makes this code fairly messy.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DestroyServerGroupStage.groovy
```diff
@@ -29,21 +29,35 @@ import org.springframework.stereotype.Component
...
     try {
-      builder
-        .withTask("disableServerGroup", DisableServerGroupTask)
-        .withTask("monitorServerGroup", MonitorKatoTask)
-        .withTask("waitForNotUpInstances", WaitForAllInstancesNotUpTask)
-        .withTask("forceCacheRefresh", ServerGroupCacheForceRefreshTask)
-        .withTask("destroyServerGroup", DestroyServerGroupTask)
-        .withTask("monitorServerGroup", MonitorKatoTask)
-        .withTask("waitForDestroyedServerGroup", WaitForDestroyedServerGroupTask)
+      if (stage.context.get("kind")?.toString() == KUBERNETES_STATEFULSET_CONTROLLER_TYPE ||
```
lwander: Same comment here

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ResizeServerGroupAllInstnacesTerminatedTask.groovy
```diff
@@ -0,0 +1,120 @@
...
+import com.netflix.spinnaker.orca.clouddriver.utils.TrafficGuard
+import com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategy
+import com.netflix.spinnaker.orca.pipeline.model.Stage
+import groovy.util.logging.Slf4j
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Component
+
+@Component
+@Slf4j
+class ResizeServerGroupAllInstnacesTerminatedTask extends AbstractServerGroupTask {
```
lwander: Typo in `Instnaces`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ResizeServerGroupAllInstnacesTerminatedTask.groovy
```diff
@@ -0,0 +1,120 @@
...
+    def operation = new HashMap(stage.context)
+    String serverGroupName = (operation.serverGroupName ?: operation.source?.asgName) as String
+
+    if (!operation.region) {
+      operation.region = operation.namespace
+    }
+
+    if (!serverGroupName) {
+      //asgName and serverGroupName is not always presented.  Add the following to build serverGroupName.
+      if (stage.type == PIPELINE_CONFIG_CREATE_SERVERGROUP_TYPE) {
```
lwander: Why would a stage that resizes a server group to 0 instances have type `createServerGroup`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ResizeServerGroupAllInstnacesTerminatedTask.groovy
```diff
@@ -0,0 +1,120 @@
...
+    }
+    if (operation.capacity.desired == 0) {
+      trafficGuard.verifyTrafficRemoval(operation.serverGroupName as String,
+        getCredentials(operation),
+        getLocation(operation),
+        getCloudProvider(operation), "Removal of all instances in ")
+    }
+  }
+
+  private String generateServerGroupName(String application, String stack, String details, Integer sequence) {
```
lwander: This should probably be made `moniker` friendly

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ResizeServerGroupAllInstnacesTerminatedTask.groovy
chlung: For statefulset and daemonset, we must remove pods (only for OnDelete) before we can ensure the new pod is running the new spec.template definition.  Are you suggest that user will handle resize stage in pipeline.  I am ok with that.

chlung:
@lwander I rollbacked createServerGroupStage and destroyServerGroupStage changes.  Per your comments, I added strategy to handle resize for createServerGroup.   I still need a way to inject pre-stage in destryServerGroupStage. Please review.  thanks.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategyData.groovy
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
+
+import com.netflix.spinnaker.orca.kato.pipeline.support.StageData
+
+class RecreateUpdateStrategyData extends RollingRedBlackStageData {
```
lwander: Just curious, why extend rrb?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
```diff
@@ -0,0 +1,220 @@
...
+        stages = []
+        return stages
+      }
+      def resizeContext = baseContext + [
+        target        : TargetServerGroup.Params.Target.current_asg_dynamic,
+        action        : ResizeStrategy.ResizeAction.scale_exact,
+        source        : source,
+        targetLocation: cleanupConfig.location,
+        scalePct      : p,
+        pinCapacity   : p < 100, // if p = 100, capacity should be unpinned
```
lwander: Do we need all this resize/scaling logic if we're just recreating?

chlung: I can remove the scaling because I pass the exactly size.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
```diff
@@ -0,0 +1,215 @@
...
+            serverGroupName: source.serverGroupName
+          ],
+          "percentageComplete" : percentageComplete
+        ]
+      ]
+
+      stages << newStage(
+        parentStage.execution,
+        pipelineStage.type,
+        "Run Validation Pipeline",
```
lwander: Does this need a validation pipeline?

chlung: I guess not, but I don't understand what is the purpose for validate pipeline and when I need to call it.  Can you explain?

chlung:
highlander will create another copy controller and disable and delete the old one, but statefulset and daemonset always keep the original one because the name constraint. Seen the same behavior in kubectl.

chlung:
@lwander updated per your feedback.  Please review.  thanks.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
```diff
@@ -0,0 +1,206 @@
...
+        targetSize    : 0,
+        capacity      : stage.context.capacity,
+        resizeType    : "incr",
+        scaleNum      : 0
+      ]
+
+      def resizeStage = newStage(
+        stage.execution,
+        resizeServerGroupStage.type,
+        "Scale down to $p% Desired Size",
```
lwander: Where is `p` defined?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
```diff
@@ -0,0 +1,206 @@
...
+      }
+      def resizeContext = baseContext + [
+        target        : TargetServerGroup.Params.Target.current_asg_dynamic,
+        action        : ResizeStrategy.ResizeAction.scale_exact,
+        source        : source,
+        targetLocation: cleanupConfig.location,
+        kind          : stageData.kind,
+        targetSize    : 0,
+        capacity      : stage.context.capacity,
+        resizeType    : "incr",
```
lwander: `incr` seems wrong when this is a scale down

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    stage.context.capacity = [
+      min    : 0,
+      max    : 0,
+      desired: 0
+    ]
+
+
+    def targetPercentages = stageData.getTargetPercentages()
```
lwander: These are for a list of sequential scale downs, I don't think this is needed

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
```diff
@@ -0,0 +1,206 @@
...
+    if (!stageData.delayBeforeCleanup) {
+      stageData.delayBeforeCleanup = DEFAULT_TERMINATION_WAIT_TIME_IN_SECONDS
+    }
+
+    stage.context.capacity = [
+      min    : 0,
+      max    : 0,
+      desired: 0
+    ]
+
```
lwander: Extra space

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  List<Stage> getBeforeCleanupStages(Stage parentStage,
+                                     RecreateUpdateStrategyData stageData,
+                                     ResizeStrategy.Source source,
+                                     String deployedServerGroupName,
+                                     int percentageComplete) {
+    def stages = []
+
+    def waitContext = [waitTime: stageData.delayBeforeCleanup]
```
lwander: Is this wait necessary here? What's it for?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategyData.groovy
lwander: Still wondering

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/strategy/Strategy.java
```diff
@@ -62,6 +63,8 @@ public boolean replacesBasicSteps() {
         return builder.composeHighlanderFlow(stage);
       case ROLLING_PUSH:
         return builder.composeRollingPushFlow(stage);
+      case RECREATE_UPDATE:
+        return builder.composeRollingPushFlow(stage);
```
lwander: This looks like a typo

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
```diff
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.Ab
 import com.netflix.spinnaker.orca.clouddriver.tasks.MonitorKatoTask
 import com.netflix.spinnaker.orca.clouddriver.tasks.instance.WaitForUpInstancesTask
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.CreateServerGroupTask
+
```
lwander: Extra newline

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/strategy/Strategy.java
chlung: It doesn't have major differences between OnDeleteUpdate and RolllingRedBlack strategies.  It is just resize happen before or after the update the controller definition.  OnDeleteUpdate also require to provide `kind`.  Do you want me to re-use RollingRedBlack strategy?

chlung:
@lwander updated per your feedback.  please review.  I would like to keep the wait to ensure the cache sync without any delay.  thanks.

lwander:
Do you mean the force cache refresh? It's typical for that to take time - I don't think adding a wait to add extra time is a good idea there.

chlung:
@lwander updated per your feedback.  Please review.  I also update the PR description.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
```diff
@@ -0,0 +1,152 @@
...
+@Component
+class RecreateUpdateStrategy implements Strategy, ApplicationContextAware {
+  final String name = "recreateupdate"
+  final static int DEFAULT_TERMINATION_WAIT_TIME_IN_SECONDS = 30
+
+  @Autowired
+  ResizeServerGroupStage resizeServerGroupStage
+
+  @Autowired
+  WaitStage waitStage
```
lwander: Is this still needed? I think you removed the wait

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
```diff
@@ -0,0 +1,152 @@
...
+
+    Map baseContext = [
+      (cleanupConfig.location.singularType()): cleanupConfig.location.value,
+      cluster                                : cleanupConfig.cluster,
+      credentials                            : cleanupConfig.account,
+      cloudProvider                          : cleanupConfig.cloudProvider,
+    ]
+
+    if (!stageData.delayBeforeCleanup) {
+      stageData.delayBeforeCleanup = DEFAULT_TERMINATION_WAIT_TIME_IN_SECONDS
```
lwander: Is this still needed?

chlung:
@lwander I cleanup those obsoleted objects and variable.  please review.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
```diff
@@ -0,0 +1,139 @@
...
+      stage.execution,
+      resizeServerGroupStage.type,
+      "Scale down to $p% Desired Size(0)",
+      resizeContext,
+      stage,
+      SyntheticStageOwner.STAGE_BEFORE
+    )
+    stages << resizeStage
+
+    return stages
```
lwander: I'm still not sure what this stage is doing or how it's supposed to work - all that's being inserted is a single resize to 0 stage. If this behaves like other deployment strategies it will deploy the new server group, then immediately scale it down to 0 instances. Is that the goal?

chlung: In order to update the controller, it needs to terminate the existing pods.  Resize is to terminate running pods and allow k8s to recreate with new pods with new definition.

lwander: Which kind is this for?

chlung: https://github.com/chlung/orca/blob/2d0c83393e389a3a036706957232a1d37b115fd4/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy#L95

lwander: Well that's any kind, which one in particular? StatefulSet, DaemonSet, Job, etc...

chlung: StatefulSet and DaemonSet will need it to use this strategy.  This is only need it for k8s OnDelete stratey.  RollingUpdate strategy doesn't require it.

lwander: Wait so this is only for OnDelete? RollingUpdate I don't think is supported for k8s at all

chlung: that is correct.  RollingUpdate is all controlled by k8s.  It does not depend on OnDelete event to trigger pod creation.  Sorry, I need to update spec.
RollingUpdate:
DaemonSet : 1.6 and after (https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/)
StatefulSet: 1.7 and after (http://blog.kubernetes.io/2017/09/kubernetes-statefulsets-daemonsets.html)

chlung:
@lwander I fixed conflicts issue.  please review.  thanks.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RecreateUpdateStrategy.groovy
lwander: So if we don't need to wait for pod deletion, we don't need this strategy right?

chlung: there is two use cases.
1. OnDelete - use delete pod to trigger create new pod. without deletion, the pod will still running old definition.
2. RollingUodate - no action required.

Is this what you asking?

lwander: So when a user, without spinnaker, updates a statefulset/daemonset, they have to manually resize it down to 0 pods?

lwander: If so - doesn't this stage need to resize the controller back to the original capacity? And why would someone want to use this if they can just rely on RollingUpdate? I think the OnDelete is meant for a different scenario...

chlung: Resize is just an option.  The purpose is to delete these running pods in order to launch new pods.  Why user want to use OnDelete opetion?  It just another option that k8s provides, and we support it as well. Especially, those people running older version of k8s.  This is another option.  I believe that RollingUpdate  will be used in most people as long as they are running the up-to-date k8s version.

chlung: We are starting with k8s 1.5 that onDelete is only available for us to use at that time.

chlung: Yes.  The controller will resize back to whatever the new defined size that all pods will be launched with new definition.

lwander: Ahh OK I understand, this makes sense now. I think you need a second resize stage after the first to ensure that you are back at full capacity after the strategy finishes, right?

chlung: I did not use patch library instead that I replace the whole controller.spec.  So, it can cover the replicas and resize according to the new controller definition.  The pipeline should have all controller definition because this is a deploy stage.  If you see any issue not patch library, please share with me. thanks.

lwander: What do you mean by "patch library"? I'm worried that this pipeline will always leave a server group at size 0 when it finishes.

chlung: I don't really compare any changes and patch the controller.  I basically replace the controller.spec and make replace api call.  Ideally, the controller will automatically launch pods.  The only potential problem only if controller failed during update and leave the resize to 0.  If you feel that the extra resize will help user.  I can add it.  Please let me know.

lwander: Well as-is, this stage will always leave it scaled down to 0, right? If instead that only happens when the user makes an error, that seems better, right?

chlung: I can include extra resize after update finish.  I do want to walk through my flow one more time with you.
Define statefulset with replicas set to 4 for update.
1. Pre-update: set the replicas to 0.
2. Deploy the update.  The replicas will be 4.  It will launch 4 pods after successful deployment.
***3. Now, add another resize to ensure 2 failed and pod will remain the original size.  If 2 is completed successfully, k8s will ignore this resize because pod size is same as replicas defined.

If we all agree, I will implement accordingly.  thanks.

lwander: If the only way these can be updated is by destroying, why can't we use the highlander strategy?

chlung: If I am not wrong, the highlander will dlete controller and create a new copy.  This is not the purpose of OnDelete.  We only replace the original controller definition and launch new pods.

lwander: These deployment strategies always start by creating a new server group though - have you disabled that here?

chlung: Yes.  I don't use deployment.  I don't use these labels to control cluster disable as well.

lwander: I'm not sure what you mean - can you show me a screenshot of how the clusters panel looks while this is running?

chlung:
@lwander I closed this PR because we agree not to implement OnDelete in orca.  Thanks for your efforts in cod review and implementation discussion.

PR #1736, opened 2017-10-24
feat(moniker): Use moniker for app name over frigga in flex

andrewbackes:
Similar to the other PRs for monikers, first try to use the moniker from the stage (which is added by deck). Then fall back on frigga.


andrewbackes:
@lwander FYI

PR #1742, opened 2017-10-26
fix(deploy): avoid crossing the streams in parallelized deploys

robfletcher:
Ensure a deploy stage is getting deployment details from its ancestor and not from global context.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AlertOnAccessMap.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AlertOnAccessMap.java
@@ -46,21 +46,24 @@ public class AlertOnAccessMap<E extends Execution<E>> extends ForwardingMap<Stri
   @Override protected Map<String, Object> delegate() { return delegate; }
 
   @Override public Object get(@Nullable Object key) {
-    log.warn(
-      "Global context key \"{}\" accessed by {} {}[{}] \"{}\"",
-      key,
-      execution.getApplication(),
-      execution.getClass().getSimpleName(),
-      execution.getId(),
-      execution.getName()
-    );
-    Id counterId = registry
-      .createId("global.context.access")
-      .withTag("application", execution.getApplication())
-      .withTag("key", String.valueOf(key));
-    registry
-      .counter(counterId)
-      .increment();
-    return super.get(key);
+    Object value = super.get(key);
+    if (value != null) {
+      log.warn(
+        "Global context key \"{}\" accessed by {} {}[{}] \"{}\"",
+        key,
+        execution.getApplication(),
+        execution.getClass().getSimpleName(),
+        execution.getId(),
+        execution.getName()
+      );
+      Id counterId = registry
+        .createId("global.context.access")
+        .withTag("application", execution.getApplication())
+        .withTag("key", String.valueOf(key));
+      registry
+        .counter(counterId)
+        .increment();
+    }
+    return value;
   }
 }
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
@@ -175,7 +175,11 @@ public class Stage<T extends Execution<T>> implements Serializable {
   }
 
   public void setContext(@Nonnull Map<String, Object> context) {
-    this.context = context;
+    if (context instanceof StageContext) {
+      this.context = context;
+    } else {
+      this.context = new StageContext(this, context);
+    }
   }
 
   /**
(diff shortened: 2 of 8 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/AlertOnAccessMap.java
```diff
@@ -46,21 +46,24 @@ public AlertOnAccessMap(Execution<E> execution, Registry registry) {
...
+        execution.getName()
+      );
+      Id counterId = registry
+        .createId("global.context.access")
+        .withTag("application", execution.getApplication())
+        .withTag("key", String.valueOf(key));
+      registry
+        .counter(counterId)
+        .increment();
+    }
```
robfletcher: Only alert on actually finding something in the global context.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
```diff
@@ -639,7 +639,7 @@ class JedisExecutionRepository implements ExecutionRepository {
         stage.parentStageId = map["stage.${stageId}.parentStageId".toString()]
         stage.requisiteStageRefIds = map["stage.${stageId}.requisiteStageRefIds".toString()]?.tokenize(",") ?: emptySet()
         stage.scheduledTime = map["stage.${stageId}.scheduledTime".toString()]?.toLong()
-        stage.context = map["stage.${stageId}.context".toString()] ? mapper.readValue(map["stage.${stageId}.context".toString()], MAP_STRING_TO_OBJECT) : emptyMap()
+        stage.context = map["stage.${stageId}.context".toString()] ? new StageContext(stage, mapper.readValue(map["stage.${stageId}.context".toString()], MAP_STRING_TO_OBJECT)) : new StageContext(stage)
```
robfletcher: Repository needs to construct a `StageContext` so that ancestor chaining works on retrieved stages.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
```diff
@@ -147,13 +145,15 @@ class ContextParameterProcessor {
...
+          def deployDetails = [
             account    : deployStage.context.account,
             capacity   : deployStage.context.capacity,
             parentStage: deployStage.parentStageId,
             region     : deployStage.context.region ?: deployStage.context.availabilityZones.keySet().first(),
           ]
-          deployDetails.putAll(context.execution?.context?.deploymentDetails?.find { it.region == deployDetails.region } ?: [:])
+          deployDetails.putAll(deployStage.context.deploymentDetails?.find {
+            it.region == deployDetails.region
+          } ?: [:])
```
robfletcher: Use the deploy stage's context to get `deploymentDetails` as it will chain back to the relevant find image/bake.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
```diff
@@ -49,7 +50,7 @@ interface ExpressionAware {
           }
         }
 
-        val result = processed[key] ?: execution.getContext()[key]
+        val result = processed[key]
```
robfletcher: No need to do this as `processed` is now a `StageContext` so it will do it anyway.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
```diff
@@ -89,7 +90,9 @@ interface ExpressionAware {
       stage.getContext(),
       stage.getContext().augmentContext(stage.getExecution()),
       true
-    )
+    ).let {
+      StageContext(stage, it)
+    }
```
robfletcher: Need chaining to work in expressions so convert map to a `StageContext`.

robfletcher:
Not super happy with the integration with expressions but that code is so opaque I'm not sure of a better approach right now.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -175,7 +175,11 @@ public void setStatus(@Nonnull ExecutionStatus status) {
   }
 
   public void setContext(@Nonnull Map<String, Object> context) {
-    this.context = context;
+    if (context instanceof StageContext) {
+      this.context = context;
+    } else {
+      this.context = new StageContext(this, context);
+    }
```
robfletcher: Ensure the context object held by a `Stage` is always a `StageContext`.

PR #1745, opened 2017-10-27
fix(security): Prevent webhook users from spoofing authed user

andrewbackes:
The webhook stage runs from Orca, so it has direct access to Clouddriver, Front50, etc. A user can create a request to, say, Clouddriver with a `X-SPINNAKER-USER` header of a user with sufficient permissions and effectively bypass all security checks.

This PR strips the `X-SPINNAKER-USER` and `X-SPINNAKER-ACCOUNT`  headers before the request is created.

@ttomsu

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
@@ -30,6 +30,9 @@ import org.springframework.web.client.RestTemplate
 @Service
 class WebhookService {
 
+  // These headers create a security vulnerability.
+  static final List<String> headerBlacklist = ["X-SPINNAKER-USER", "X-SPINNAKER-ACCOUNT"];
+
   @Autowired
   private RestTemplate restTemplate
 
@@ -60,6 +63,9 @@ class WebhookService {
   private static HttpHeaders buildHttpHeaders(Object customHeaders) {
     HttpHeaders headers = new HttpHeaders()
     customHeaders?.each { key, value ->
+      if (headerBlacklist.contains(key.toUpperCase())) {
+        return;
+      }
       if (value instanceof List<String>) {
         headers.put(key as String, value as List<String>)
       } else {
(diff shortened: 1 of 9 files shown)


orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
```diff
@@ -30,6 +30,9 @@ import org.springframework.web.client.RestTemplate
 @Service
 class WebhookService {
 
+  // These headers create a security vulnerability.
+  static final List<String> headerBlacklist = ["x-spinnaker-user", "x-spinnaker-account"];
```
ttomsu: It looks like this list has expanded (my fault for telling you `account` when it should be `accounts` also). You can probably just reference these constants instead of redeclaring:

```
  public static final String SPINNAKER_USER = "X-SPINNAKER-USER";
  public static final String SPINNAKER_ACCOUNTS = "X-SPINNAKER-ACCOUNTS";
  public static final String SPINNAKER_USER_ORIGIN = "X-SPINNAKER-USER-ORIGIN";
  public static final String SPINNAKER_REQUEST_ID = "X-SPINNAKER-REQUEST-ID";
  public static final String SPINNAKER_EXECUTION_ID = "X-SPINNAKER-EXECUTION-ID";
```

Source: https://github.com/spinnaker/kork/blob/master/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java#L29-L33

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
```diff
@@ -60,10 +63,12 @@ class WebhookService {
   private static HttpHeaders buildHttpHeaders(Object customHeaders) {
     HttpHeaders headers = new HttpHeaders()
     customHeaders?.each { key, value ->
-      if (value instanceof List<String>) {
-        headers.put(key as String, value as List<String>)
-      } else {
-        headers.add(key as String, value as String)
+      if (!headerBlacklist.contains(key.toLowerCase())) {
```
ttomsu: Style nit: it's easier to read an early escaping block, so something like:

```
if (headerBlacklist.contains(key.toUpperCase()) {
  continue; // or return? Whatever compiles.
}

if (value instanceof ...) {
...

```

PR #1746, opened 2017-10-27
fix(security): Prevent webhook users from spoofing authed user

andrewbackes:
The webhook stage runs from Orca, so it has direct access to Clouddriver, Front50, etc. A user can create a request to, say, Clouddriver with a X-SPINNAKER-USER header of a user with sufficient permissions and effectively bypass all security checks.

This PR strips the X-SPINNAKER-USER and X-SPINNAKER-ACCOUNT headers before the request is created.

@ttomsu


ttomsu:
LGTM

cfieber:
definitely highlights the need to run each of the services with mutual TLS for a strong security story (I think this only applies if you are running all the services bound to localhost on a single box)

PR #1748, opened 2017-10-27
feat(core): Add correlation ids to orchestrations

robzienert:
Allows us to repeatedly send orchestrations with the same correlation id and only running a single one. Pre-req for keel.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
@@ -16,14 +16,18 @@
 
 package com.netflix.spinnaker.orca.pipeline.persistence.jedis
 
-import java.util.concurrent.Executors
-import java.util.function.Function
 import com.fasterxml.jackson.core.type.TypeReference
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spectator.api.Registry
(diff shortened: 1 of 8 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.groovy
```diff
@@ -509,6 +520,25 @@ class JedisExecutionRepository implements ExecutionRepository {
...
+      if (orchestrationId != null) {
+        def orchestration = withJedis(getJedisPoolForId(orchestrationId)) { Jedis jedis ->
+          retrieveInternal(jedis, Orchestration, orchestrationId)
+        }
+        if (!orchestration.status.isComplete()) {
+          return orchestration
+        }
+        correlationJedis.del(key)
+      }
+      throw new ExecutionNotFoundException("No Orchestration found for correlation ID $correlationId")
```
robfletcher: Is this really what we want to do rather than returning `null`?

robzienert: I was keeping pattern with the other find methods. I'm totally down with (and would prefer) using null, but throwing an exception seemed consistent with the rest of the repository.

robfletcher: Fair enough

PR #1750, opened 2017-10-27
fix(moniker): hotfix canary deploy stages

cfieber:
moniker objects are incorrect on canary deploy stages, this removes those objects for now pending a fix of the source problem


asher:
lgtm

cfieber:
@andrewbackes @lwander FYI the moniker object in a canary deploy stage is incorrect (I expect this is in deck stage builder code somewhere), this is a quick workaround for that

lwander:
Sorry for the trouble, thanks @cfieber

PR #1751, opened 2017-10-27
feat(expressions): Adding the ability to specify a runtime spel modifier class

jeyrschabu:
- added runtime expression updaters
- parameters.foo -> trigger.parameters.foo after parameters have been processed once

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.orca.pipeline.util
 
+import com.netflix.spinnaker.orca.pipeline.expressions.RuntimeExpressionUpdater
+
 import java.lang.reflect.Field
 import java.lang.reflect.Method
 import java.text.SimpleDateFormat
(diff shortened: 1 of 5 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.groovy
```diff
@@ -449,6 +458,25 @@ abstract class ContextUtilities {
 
 }
 
+/**
+ * Include the execution object (#root.execution) for Stage locating functions at runtime
+ * @param expression #stage('property') becomes #stage(#root.execution, 'property')
+ * @return an execution aware helper function (only limited to judg(e?)ment & stage)
+ */
+class ExecutionAwareExpressionUpdater implements RuntimeExpressionUpdater {
```
jeyrschabu: Externalized this fit into the pattern.  Was previously a standalone method

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
```diff
@@ -62,6 +63,21 @@ interface ExpressionAware {
     return this
   }
 
+  /**
```
jeyrschabu: The idea here is that once `parameters` have been processed, they are inserted into `trigger`.
This helps route the call to the right place using the `RuntimeExpressionUpdater` pattern.

jeyrschabu:
The second part of this PR is to route parameter accessing expressions to the right place once they have been processed. Once parameters have been processed, they are inserted in `trigger`.
This helps route the call to the right place using the RuntimeExpressionUpdater pattern.

jeyrschabu:

robzienert:
@jeyrschabu Should we close?

jeyrschabu:
Yeah. closing.

PR #1758, opened 2017-10-30
fix(titus): Tag titus server groups with previous image metadata

ajordens:


ajordens:
PTAL @tomaslin 

Crux of this issue is to enable metadata tagging for Titus and to check both `NEWEST` and `ANCESTOR`.

ajordens:
If/when Titus does force cache refresh, I can take out the titus-specific bits.

tomaslin:
LGTM

PR #1760, opened 2017-10-31
fix(front50): Keep front50 optional

lwander:
As we're trying to get our nightlies to pass we found that front50 has become required for orca again, breaking some halyard flows.


ajordens:
Any chance to have some config that you guys could set to use a `Front50` mock rather than expecting it to be optional and null checked everywhere?

lwander:
Halyard creates an orca node that doesn't require front50 to run pipelines that update spinnaker - so I don't think that works there.

We could try to create a NoOp bean for the front50 service, but it's tricky to handle because in some scenarios (traffic guards) it's totally acceptable for the front50 service to do nothing (we assume nothing has a guard) where as in others (pipeline templates) the front50 service should throw an error if it's a NoOp bean.

PR #1763, opened 2017-11-01
feat(artifacts): Resolve expectedArtifact by ids in trigger.

jtk54:


jtk54:
Depends on https://github.com/spinnaker/deck/pull/4349 and https://github.com/spinnaker/echo/pull/194.

PR #1767, opened 2017-11-02
fix(rollingpush): Ensure `waitTaskState` is cleared between iterations

ajordens:


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/tasks/WaitTask.groovy
```diff
@@ -43,7 +43,9 @@ class WaitTask implements RetryableTask {
     def waitTimeMs = TimeUnit.MILLISECONDS.convert(waitTime, TimeUnit.SECONDS)
     def now = timeProvider.millis
 
-    if (!stage.context.containsKey("waitTaskState") || !stage.context.waitTaskState instanceof Map) {
+
+    def waitTaskState = stage.context.waitTaskState
+    if (!waitTaskState || !waitTaskState instanceof Map) {
```
ajordens: `waitTaskState` can now be an empty map so checking not null and not empty.

(would be slick if we had a consistent pattern for removing keys from the context, perhaps another attribute of the `TaskResult`)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/CheckForRemainingTerminationsTask.groovy
```diff
@@ -33,6 +33,9 @@ class CheckForRemainingTerminationsTask implements Task {
       return TaskResult.SUCCEEDED
     }
 
-    return new TaskResult(ExecutionStatus.REDIRECT, [skipRemainingWait: false])
+    return new TaskResult(ExecutionStatus.REDIRECT, [
+      skipRemainingWait: false,
+      waitTaskState: [:]
```
ajordens: Not sure how this would have worked previously ... without reseting `waitTaskState` the wait task was just auto-completing after the first cycle.

PR #1769, opened 2017-11-03
chore(gradle): Avoid running tests on master/release branch

ajordens:


ajordens:
This the right way to do it @cfieber? Checked build logs on travis and looked like this would stop running tests on master / release branches.

I don't think `assemble` runs tests but put `-x test` there anyhow.

PR #1771, opened 2017-11-03
fix(canary): fix wait task after baseline asg disable

asher:


orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/DisableCanaryTask.groovy
```diff
@@ -47,12 +47,12 @@ class DisableCanaryTask extends AbstractCloudProviderAwareTask implements Task {
...
     def taskId = katoService.requestOperations(cloudProvider, ops).toBlocking().first()
 
-    stage.context.remove('waitTaskState')
     return new TaskResult(ExecutionStatus.SUCCEEDED, [
       'kato.last.task.id'    : taskId,
       'deploy.server.groups' : dSG,
       disabledCluster        : selector,
-      waitTime               : waitTime
+      waitTime               : waitTime,
+      waitTaskState          : ""
```
ajordens: When I did this in another task I set it to `[:]` ... either should technically work, just a little weird to see something you know should be a map become a string?

It'd be nice to have a way to explicitly remove things from the context short of `stage.context.remove()` from a Task (in theory the `WaitTask` itself could clean itself up upon completion since `waitTaskState` is really only an internal structure that doesn't need to live beyond the wait stage).

asher: WaitTask includes `!waitTaskState instanceof Map` as a condition to set a new start time, so I thought why not :) I prefer the idea of waitTask cleaning up waitTaskState though - going to do that.

ajordens: The `!waitTaskState` will also catch an empty `[:]`

ajordens: I don't know what compelled us to do the `waitTaskState instanceof Map` check but we must of had a good reason once upon a time.

PR #1772, opened 2017-11-03
Use StageContext consistently without breaking strategies

robfletcher:
Restores commit f49653055f614cb38110e4e7924cedf923f0fdb5 and fixes the issue in parsing parent Pipeline from trigger that meant the `execution` back-ref was not set on stages.


ajordens:
Tests look out of wack, but otherwise I'm going to trust in the magic of `@JsonManagedReference`.

It does seem like this is a slight change in behavior between looking upstream / looking at trigger.

Probably worth a conversation. 

Personally, I like the idea of looking upstream but there are likely some expectations now from users of what's going to be available on the strategies trigger for expressions.

robfletcher:
@ajordens I think some code is explicitly looking in the trigger.

PR #1773, opened 2017-11-03
fix(pipeline_template): Deal with whitespace in jinja module kv pairs

robzienert:
cc @jeyrschabu @emjburns


emjburns:
Excellent use of commenting 🥇

PR #1777, opened 2017-11-06
Fixes restarting ACA task stages

robfletcher:
When an ACA task stage is restarted it attempts to cancel the previous canary. However, if enough time has passed the canary will have ended already. This case wasn't being handled. Also `canaryDeployments` got removed from context which makes sense for "full" canaries as they'll do a fresh deploy but not for ACA tasks.


ajordens:
LGTM

I'd maybe wire up `mine.cancelCanary()` with RetrySupport to handle a network blip (we weren't doing that before).

PR #1780, opened 2017-11-07
fix(log): clarify null custom strategy error

asher:
If deployment strategy == custom and the resulting strategy dropdown remains blank (as may happen if custom strategy was selected by accident) the resulting `Unable to located referenced pipeline null` error message doesn't make sense to users unfamiliar with custom strategy configuration.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/StartPipelineTask.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/StartPipelineTask.groovy
@@ -50,7 +50,7 @@ class StartPipelineTask implements Task {
     Map<String, Object> pipelineConfig = pipelines.find { it.id == pipelineId }
 
     if (!pipelineConfig) {
-      throw new IllegalArgumentException("Unable to locate referenced pipeline $pipelineId.")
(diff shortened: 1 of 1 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/StartPipelineTask.groovy
```diff
@@ -50,7 +50,8 @@ class StartPipelineTask implements Task {
     Map<String, Object> pipelineConfig = pipelines.find { it.id == pipelineId }
 
     if (!pipelineConfig) {
-      throw new IllegalArgumentException("Unable to locate referenced pipeline $pipelineId.")
+      throw new IllegalArgumentException("${isStrategy ? 'Custom strategy selected but ' : ''}" +
```
ajordens: Will this not render:

`Custom strategy selected but Unable to locate referenced pipeline $pipelineId.`

That still doesn't seem quite right.

`The referenced ${isStrategy ? 'custom strategy' : 'pipeline'} cannot be located (${pipelineId})`

asher: Much clearer, thanks. Updated.

ajordens:
You have an extra `be` in your `be be` :)

ajordens:
You shall speak good

PR #1781, opened 2017-11-07
fix(trafficguards): Fix Moniker usage in instance termination

ajordens:
The instance termination task does not normally have a `serverGroupName`
explicitly specified.

Without such, the moniker used to check traffic guards is invalid.


ajordens:
FYI @andrewbackes @lwander

ajordens:
Fall back to frigga parsing the server group name as there's nothing better available in instance search results.

PR #1786, opened 2017-11-08
fix(web): Reverting needless method sig refactor from #1718

robzienert:
Was causing 500 errors because the method couldn't be found in other places in the Controller. Thanks, Groovy.

@spinnaker/netflix-reviewers @jervi FYI


jervi:
Oh, sorry about that, @robzienert! Seems like I messed up during a refactor...

PR #1788, opened 2017-11-09
fix(queue): custom de/serializer so we can migrate queue values slowly

robfletcher:


ajordens:
Seems reasonable.

Appropriate test would be to roll forward and then roll backward?

robzienert:
Nice.

robfletcher:
It's writing same values now so it shouldn't need anything to roll back. We'll do a phase 2 once happy with the Orca 4 release.

PR #1794, opened 2017-11-09
upgrade dependencies and fix tag cleanup

robfletcher:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/CleanUpTagsTask.java
```diff
@@ -87,14 +87,14 @@ public TaskResult execute(Stage stage) {
 
       List<String> tagsToDelete = tags.stream()
         .flatMap(entityTag -> ((List<Map>) entityTag.get("tags")).stream())
+        .filter(tag -> "astrid_rules".equals(tag.get("namespace")))
```
ajordens: One reason to maybe not do this is that `astrid_rules` means nothing in OSS-land.

We might need/want to revisit this at some point and tidy it up a little bit (perhaps making the tags to clean up a little more configurable).

ajordens: That said, we're the only people using entity tags.

PR #1797, opened 2017-11-13
feat(pagerduty): Support multiple applications and keys directly

jrsquared:
Related echo PR: https://github.com/spinnaker/echo/pull/195

### DIFF
--- a/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
+++ b/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
@@ -17,6 +17,8 @@
 
 package com.netflix.spinnaker.orca.echo.tasks
 
+import com.fasterxml.jackson.databind.ObjectMapper
+
 import java.util.concurrent.TimeUnit
 import com.netflix.spinnaker.orca.RetryableTask
 import com.netflix.spinnaker.orca.TaskResult
@@ -47,27 +49,46 @@ class PageApplicationOwnerTask implements RetryableTask {
     if (!front50Service) {
       throw new UnsupportedOperationException("Front50 is not enabled, no way to fetch pager duty. Fix this by setting front50.enabled: true");
     }
-    def application = stage.context.application as String
-    def applicationPagerDutyKey = fetchApplicationPagerDutyKey(application)
-    if (!applicationPagerDutyKey) {
-      throw new IllegalStateException("${application} does not have a pager duty service key!")
+    def applications = stage.context.applications as List<String>
+    def incomingPagerDutyKeys = stage.context.keys as List<String>
+    def allPagerDutyKeys = []
+
+    if (incomingPagerDutyKeys) {
+      // Add all arbitrary keys to the pager duty key list
+      allPagerDutyKeys.addAll(incomingPagerDutyKeys)
     }
 
+    // Add applications to the pager duty key list if they have a valid key
+    if (applications) {
+      def invalidApplications = []
+      applications.each { application ->
+        def applicationPagerDutyKey = fetchApplicationPagerDutyKey(application)
+        if (!applicationPagerDutyKey) {
+          invalidApplications.push(application)
+        } else {
+          allPagerDutyKeys.push(applicationPagerDutyKey)
+        }
+      }
+
+      if (invalidApplications.size() > 0) {
+        throw new IllegalStateException("${invalidApplications.join(", ")} ${invalidApplications.size() > 1 ? "do" : "does"} not have a pager duty service key.")
+      }
+    }
+
+    // echo service will not send a proper response back if there is an error, it will just throw an exception
     echoService.create(
       new EchoService.Notification(
-        to: [
-          applicationPagerDutyKey
-        ],
+        to: allPagerDutyKeys,
         notificationType: EchoService.Notification.Type.PAGER_DUTY,
         source: new EchoService.Notification.Source(user: stage.execution.authentication.user),
         additionalContext: [
-          message: stage.context.message
+          message: stage.context.message,
+          details: stage.context.details
         ]
       )
     )
 
-    log.info("Sent page (application: ${application}, message: '${stage.context.message}')")
-
+    log.info("Sent page (application(s): ${applications.join(", ")}, message: '${stage.context.message}')")
     return new TaskResult(SUCCEEDED)
   }
 


orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
```diff
@@ -42,32 +44,54 @@ class PageApplicationOwnerTask implements RetryableTask {
...
   TaskResult execute(Stage stage) {
     if (!front50Service) {
       throw new UnsupportedOperationException("Front50 is not enabled, no way to fetch pager duty. Fix this by setting front50.enabled: true");
     }
-    def application = stage.context.application as String
-    def applicationPagerDutyKey = fetchApplicationPagerDutyKey(application)
-    if (!applicationPagerDutyKey) {
-      throw new IllegalStateException("${application} does not have a pager duty service key!")
+    def applications = stage.context.applications as List<String>
+    def keys = stage.context.keys as List<String>
```
ajordens: Would it be clearer if `context.keys` were `context.pagerDutyKeys`?

Either way you can do `def pagerDutyKeys = stage.context.keys as List<String>`

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
```diff
@@ -42,32 +44,54 @@ class PageApplicationOwnerTask implements RetryableTask {
...
+    def pagerDutyKeys = []
+
+    if (keys) {
+      // Add all arbitrary keys to the pager duty key list
+      pagerDutyKeys.addAll(keys)
     }
 
+    // Add applications to the pager duty key list if they have a valid key
+    if (applications && applications.size() > 0) {
```
ajordens: `if (applications)`

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
```diff
@@ -42,32 +44,54 @@ class PageApplicationOwnerTask implements RetryableTask {
...
-          message: stage.context.message
+          message: stage.context.message,
+          details: stage.context.details
         ]
       )
     )
 
-    log.info("Sent page (application: ${application}, message: '${stage.context.message}')")
-
+    // If an error happened, an exception was thrown so this won't get hit
```
ajordens: the javadoc is probably unnecessary ...

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
```diff
@@ -42,32 +44,54 @@ class PageApplicationOwnerTask implements RetryableTask {
...
+      applications.each { application ->
+        def applicationPagerDutyKey = fetchApplicationPagerDutyKey(application)
+        if (!applicationPagerDutyKey) {
+          invalidApplications.push(application)
+        } else {
+          pagerDutyKeys.push(applicationPagerDutyKey)
+        }
+      }
+
+      if (invalidApplications.size() > 0) {
```
ajordens: `if (invalidApplications)`

Q: if one of the applications doesn't have a pager duty key, do you want to fail everything and not send to other valid applications?

Maybe a moot point and something that won't really happen in reality?

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
```diff
@@ -42,32 +44,54 @@ class PageApplicationOwnerTask implements RetryableTask {
   @Autowired(required = false)
   Front50Service front50Service
 
+  private final ObjectMapper mapper = new ObjectMapper()
```
ajordens: Again, should be able to `@Autowired` this in here _but_ I'm not sure you even need it. Doesn't appear to be used anywhere.

jrsquared: Yep, leftover from a refactor. Good call.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
jrsquared: Since it's a pager duty task, I think calling it pagerDutyKeys is superfluous, but totally agree on the `def pagerDutyKeys` part.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
jrsquared: True, very good catch. I'll fix it to not fail entirely if one does not have a pd key. Very nice.

jrsquared: Actually, I'll figure this out in another iteration. Moot point for our specific use case, but something very important to figure out....

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
```diff
@@ -47,27 +49,46 @@ class PageApplicationOwnerTask implements RetryableTask {
     if (!front50Service) {
       throw new UnsupportedOperationException("Front50 is not enabled, no way to fetch pager duty. Fix this by setting front50.enabled: true");
     }
-    def application = stage.context.application as String
-    def applicationPagerDutyKey = fetchApplicationPagerDutyKey(application)
-    if (!applicationPagerDutyKey) {
-      throw new IllegalStateException("${application} does not have a pager duty service key!")
+    def applications = stage.context.applications as List<String>
+    def incomingPagerDutyKeys = stage.context.keys as List<String>
+    def allPagerDutyKeys = []
```
ajordens: You can still simplify this further ...

```
def allPagerDutyKeys = (stage.context.keys as List<String>) ?: []
```

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/tasks/PageApplicationOwnerTask.groovy
```diff
@@ -47,27 +49,46 @@ class PageApplicationOwnerTask implements RetryableTask {
...
+          allPagerDutyKeys.push(applicationPagerDutyKey)
+        }
+      }
+
+      if (invalidApplications.size() > 0) {
+        throw new IllegalStateException("${invalidApplications.join(", ")} ${invalidApplications.size() > 1 ? "do" : "does"} not have a pager duty service key.")
+      }
     }
 
+    // echo service will not send a proper response back if there is an error, it will just throw an exception
```
ajordens: Generally speaking this is common knowledge ... not sure it's worth javadoc?

PR #1805, opened 2017-11-15
fix(moniker): Use the correct moniker when applying source server group

andrewbackes:
`ApplySourceServerGroupCapacityTask` overrides `convert()`. `convertMoniker()` also needs to be overridden. The moniker used when verifying the cluster should be the same as the server group in the operation.

Everything seems to be working just fine for AWS, but the k8s rollback stage is currently failing. There is a bigger problem that @lwander is working as to why k8s is using this stage that is in the aws provider.

@ajordens - Would you mind taking a look at this for me? This task seems to work a lot differently than the others that extend `AbstractServerGroupTask` and based on the blame you have the most experience with it.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ApplySourceServerGroupCapacityTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ApplySourceServerGroupCapacityTask.groovy
@@ -20,6 +20,7 @@ package com.netflix.spinnaker.orca.clouddriver.pipeline.providers.aws
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.moniker.Moniker
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.AbstractServerGroupTask
 import com.netflix.spinnaker.orca.clouddriver.utils.OortHelper
@@ -78,6 +79,12 @@ class ApplySourceServerGroupCapacityTask extends AbstractServerGroupTask {
       log.error("Unable to apply source server group capacity (executionId: ${stage.execution.id})", e)
       return null
     }
+  }
+
+  @Override
+  Moniker convertMoniker(Stage stage) {
+    // Used in AbstractServerGroupTask.execute() but not needed here.
+    return null;
   }
 
   /**


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ApplySourceServerGroupCapacityTask.groovy
```diff
@@ -80,6 +82,26 @@ class ApplySourceServerGroupCapacityTask extends AbstractServerGroupTask {
     }
   }
 
+  @Override
+  Moniker convertMoniker(Stage stage) {
```
ajordens: Confused as to how this actually works ... you've got two `return` statements in the same scope?

Anyhow ... this stage should be applying the previously captured capacity to the newly created server group. 

Generally speaking, I believe it should be applicable across cloud providers and certainly must be used if `CaptureSourceServerGroupCapacityTask` is used (as this task pins min=desired capacity).

andrewbackes: Ah, thank you for pointing that out. The first `return` was left over from debugging. It would still work, but is not the intended behavior. 

The intended behavior is to grab the target sever group that will be used in the resize operation. Then return its moniker. Since we are getting the target server group from Clouddriver, it is guaranteed to have a moniker.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ApplySourceServerGroupCapacityTask.groovy
```diff
@@ -80,6 +82,25 @@ class ApplySourceServerGroupCapacityTask extends AbstractServerGroupTask {
...
+      def targetServerGroup = oortHelper.getTargetServerGroup(
+        context.credentials as String,
+        context.serverGroupName as String,
+        context.region as String,
+        context.cloudProvider as String
+      ).get()
+
+      return targetServerGroup.getMoniker()
+    } catch (Exception e) {
+      log.error("Unable to apply source server group capacity (executionId: ${stage.execution.id})", e)
```
ajordens: Please add the offending server group name / region / account to this error message.

Rather than returning null on an exception, could a reasonable fallback be to construct a moniker via frigga.

The duplicate call to `clouddriver` is definitely sub-optimal given that (a) it fetches far more data than just a moniker and (b) it literally happened in the immediately preceding call (and would have also happened previously since we do have the serverGroupName in the context.

In the short-term this is probably alright, but I would like an alternative to retrieving full server groups just to get the moniker. This may end up being a _simple_ `/moniker` endpoint in clouddriver taking a few query parameters?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/AbstractServerGroupTask.groovy
```diff
@@ -74,7 +74,7 @@ abstract class AbstractServerGroupTask extends AbstractCloudProviderAwareTask im
       return new TaskResult(ExecutionStatus.SUCCEEDED)
     }
 
-    def taskId = kato.requestOperations(cloudProvider, [[(serverGroupAction): operation]])
+    def taskId = kato.requestOperations(operation.cloudProvider, [[(serverGroupAction): operation]])
```
andrewbackes: @ajordens This fixes the problem in the k8s rollback stage. `getCloudProvider(stage)` was defaulting to `aws`. Changing the clouddriver request to use the operation's `cloudProvider` makes a lot of sense to me conceptually. But it will have an impact on a lot of other stages and I'd love your eyes on it.

You can see the original problem here:
k8s-v1 deploy problem

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ApplySourceServerGroupCapacityTask.groovy
andrewbackes: I broke up the exception handling into more specific cases. I think you are right, in the case that the moniker is in the wrong place in the payload, malformed, etc. then we should fall back on frigga. In the case that there is a clouddriver problem or something else unexpected, I think falling back on frigga might obfuscate the true error downstream (this is more of an inssue for non-frigga providers.)

I agree that the double call to `clouddriver` is undesirable. I'd also love to avoid it. There are a few options here:
1. Store the response as a class member.
2. Rework `execute()` in the base class.
  a. Combine `convert()` and `converMoniker()` in 
  b. Return different types from `convert()` and use it in `convertMoniker()`
3. Put a lighter weight call into clouddriver (As you pointed out.)

I figured duplicating what was in `convert()` introduced the least amount of risk. Albeit, at the cost of a double call.

As an aside, there is a certain amount of urgency to this PR since k8s rollbacks currently do not work.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/AbstractServerGroupTask.groovy
lwander: worth noting the breaks all non-aws platforms too

lwander: We have a slightly cleaner fix andrew is patching up in a minute

lwander:
Any concerns with merging @ajordens? This fixes rollback for all non-aws platforms.

ajordens:
The `cloudProvider` bit in the rollback looks reasonable to me and is orthogonal to the other moniker changes.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ApplySourceServerGroupCapacityTask.groovy
ajordens: There are no retries around the call to clouddriver, and the impact of ever having a null `Moniker` is likely to be a NPE.

In this particular task, nothing actually appears to use the moniker (`validateClusterStatus()` isn't overridden). 

I'm calling these sorts of issues out because we've run into a number of issues lately around network failures between orca <-> clouddriver (or other services) without retries.

Similarly, at scale, we've noticed unnecessary overhead from inefficient calls to fetch data from clouddriver. A lot of these calls have been optimized for `aws` but we still try and avoid them where not strictly necessary.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/AbstractServerGroupTask.groovy
ajordens: Hah, yup ... that's an oldie right there!

I wonder if we shouldn't at least visibly log when a `cloudProvider` hasn't been specified and we're falling back to `aws`.

andrewbackes:
Split out the `cloudProvider` changes into a different PR: https://github.com/spinnaker/orca/pull/1810

andrewbackes:
Yeah, these other changes are still needed. Without it you get this:

Moniker error

This fails because the stage is using the `convertMoniker()` method from its base class. The base class method does a `stage.mapTo` which fails because it can not find the moniker field.

ajordens:
My earlier point is that this `Moniker` isn't actually used for anything (there's no traffic guard check) ... so we could just as easily `return null` from here and avoid the overhead until such time as it's needed.

Right?

andrewbackes:
Oh man, that is a great observation. I didn't realize there wasn't a trafficguard check here.

You're right, I tested it for k8s and aws and returning `null` is sufficient.

ajordens:
👍

ajordens:
Will merge after Travis

PR #1809, opened 2017-11-17
feat(manifest): more robust status handling

lwander:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/UpdateManifestWaitForStableTask.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Autowired
+  OortService oortService;
+
+  @Autowired
+  ObjectMapper objectMapper;
+
+  @Override
+  public long getBackoffPeriod() {
+    return TimeUnit.SECONDS.toMillis(1);
```
ajordens: This seems a little overly aggressive ... for most of the AWS-operations we've backed this off to 5-10s.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/UpdateManifestWaitForStableTask.java
```diff
@@ -0,0 +1,104 @@
...
+  public TaskResult execute(Stage stage) {
+    String account = getCredentials(stage);
+    String name = (String) stage.getContext().get("manifest.name");
+    String location = (String) stage.getContext().get("manifest.location");
+    String id = String.format("'%s' in '%s' for account %s", name, location, account);
+    List<String> messages = new ArrayList<>();
+
+    Response response;
+    try {
+      response = oortService.getManifest(account, location, name);
```
ajordens: Is there a reason you're returning `Response` from `getManifest` vs. `Manifest`? 

I can't see a need for Response unless you were wanting to stream a response or inspect headers. There's a lot of legacy using `Response` to no great benefit IMO.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/UpdateManifestWaitForStableTask.java
```diff
@@ -0,0 +1,104 @@
...
+    } catch (RetrofitError e) {
+      log.warn("Unable to read manifest {}", id, e);
+      return new TaskResult(ExecutionStatus.RUNNING, new HashMap<>(), new HashMap<>());
```
ajordens: The `new HashMap<>()` should be redundant, there's a status-only constructor.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/UpdateManifestWaitForStableTask.java
```diff
@@ -0,0 +1,104 @@
...
+    if (response.getStatus() != 200) {
+      return new TaskResult(ExecutionStatus.RUNNING, new HashMap<>(), new HashMap<>());
+    }
+
+    Manifest manifest;
+    try {
+      manifest = objectMapper.readValue(response.getBody().in(), Manifest.class);
+    } catch (IOException e) {
+      log.error("Failed to read " + id, e);
+      throw new RuntimeException(e);
```
ajordens: Why not throw exception w/ message and cause? At this point, you're likely to get the stack logged twice ... one for the explicit log here, and another when the upstream catches the `RuntimeException` and logs.

You may also want to consider including the `executionId` in the error message, we've found it difficult to debug issues in general without it.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/UpdateManifestWaitForStableTask.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    Manifest.Status status = manifest.getStatus();
+    if (status.getStable() == null || !status.getStable().isState()) {
+      messages.add(status.getStable().getMessage());
+      Map<String, Object> context = new ImmutableMap.Builder<String, Object>()
+          .put("stableMessages", messages)
+          .build();
+
+      return new TaskResult(ExecutionStatus.RUNNING, context, new HashMap<>());
```
ajordens: This is going to overwrite any `stableMessages` that were added previously, is that likely to be an issue in reality?

lwander: That won't be a problem - however - 

I thought that with the new non-global context that @robfletcher has been working on migrating to we could look up any stage's outputs, and even aggregate across all stages that emit a certain key?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/UpdateManifestWaitForStableTask.java
lwander: Sounds good

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/UpdateManifestWaitForStableTask.java
lwander: Ah neat

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/UpdateManifestWaitForStableTask.java
lwander: Didn't know that was option, `Manifest` sound cleaner.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/UpdateManifestWaitForStableTask.java
lwander: Ok - will do

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/UpdateManifestWaitForStableTask.java
danielpeach: I think he might be asking about the sequence of `stableMessages` generated by retries of this task. Which might be interesting.

lwander: Ahh, that's definitely worth overwriting

PR #1810, opened 2017-11-17
fix(kubernetes/rollback): Pass cloudProvider to downstream tasks.

andrewbackes:
`ApplySourceServerGroupCapacityTask` is expecting to have the `cloudProvider` field. If it doesn't, then it defaults to AWS. This is causing k8s rollbacks to fail.

This PR combined with https://github.com/spinnaker/orca/pull/1805 will fix k8s rollbacks (and I suspect some other providers as well.)


ajordens:
Do you also need this in `buildCaptureSourceServerGroupCapacityStage()` ... I suspect it's also going to default to `aws`.

andrewbackes:
`CaptureSourceServerGroupCapacityTask` has its own `execute()` which doesn't work the same way. It doesn't inherit from `AbstractServerGroupTask` like `ApplySourceServerGroupCapacityTask` does.

ajordens:
That's fair ... LGTM.

`CaptureSourceServerGroupCapacityTask` does default to `aws` but the call to `oortHelper.getTargetServerGroup()` just throws away the `cloudProvider` it's passed so no harm.

Thanks for the diligence on this.

PR #1814, opened 2017-11-20
fix(trafficguards): debug logging when no enabled asgs found

asher:
Occasionally seeing TrafficGuard prevent disabling the old servergroup during redblack and highlander deploys, when the new servergroup has UP instances. Adding additional logging to help troubleshoot.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
@@ -23,6 +23,7 @@ import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.Locat
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup;
 import com.netflix.spinnaker.orca.front50.Front50Service;
 import com.netflix.spinnaker.orca.front50.model.Application;
+import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -116,11 +116,20 @@ private void verifyOtherServerGroupsAreTakingTraffic(String serverGroupName, Mon
       .filter(tsg -> location.equals(tsg.getLocation()))
       .collect(Collectors.toList());
 
+    List<TargetServerGroup> otherServerGroups = targetServerGroups.stream().filter( tsg -> !tsg.getName().equalsIgnoreCase(serverGroupName)).collect(Collectors.toList());
```
ajordens: Something like this might be cleaner:

```
  private void verifyOtherServerGroupsAreTakingTraffic(String serverGroupName,
                                                       Moniker serverGroupMoniker,
                                                       Location location,
                                                       String account,
                                                       String cloudProvider,
                                                       String operationDescriptor) {
    Optional<Map> cluster = oortHelper.getCluster(serverGroupMoniker.getApp(), account, serverGroupMoniker.getCluster(), cloudProvider);

    if (!cluster.isPresent()) {
      throw new IllegalStateException(format("Could not find cluster '%s' in %s/%s with traffic guard configured.", serverGroupMoniker.getCluster(), account, location.getValue()));
    }
    List<TargetServerGroup> targetServerGroups = ((List<Map<String, Object>>) cluster.get().get("serverGroups"))
      .stream()
      .map(TargetServerGroup::new)
      .filter(tsg -> location.equals(tsg.getLocation()))

    List<TargetServerGroup> otherTargetServerGroups = targetServerGroups
      .stream()
      .filter(tsg -> serverGroupName.equalsIgnoreCase(tsg.getName()))

    boolean hasOtherEnabledServerGroups = otherTargetServerGroups.stream().anyMatch(tsg ->
      (tsg.getInstances().stream().filter(i -> "Up".equals(i.get("healthState"))).count()) > 0
    );

    if (!hasOtherEnabledServerGroups) {
      List<Map> context = otherTargetServerGroups.stream().map(tsg -> ImmutableMap.builder()
        .put("name", tsg.getName())
        .put("disabled", tsg.isDisabled())
        .put("instances", tsg.getInstances())
        .build()
      ).collect(Collectors.toList());

      log.debug(
        "Traffic guard check has failed (cluster: {}, account: {}, context: {})",
        serverGroupMoniker.getCluster(),
        account,
        context
      );

        format(
          "This cluster ('%s' in %s/%s) has traffic guards enabled. %s %s would leave the cluster with no instances taking traffic.",
          serverGroupMoniker.getCluster(),
          account,
          location.getValue(),
          operationDescriptor,
          serverGroupName
        )
      );
    }
  }
```

My change above should log a single line with the server group details and line wraps a little nicer. I've not actually tested that it gives reasonable output, _I think it should tho_ although `.instances` is probably _very_ verbose ... we might be able to flatten that down to just including instanceId and health.

It would be really nice if the executionId were propagated down (or included in the mdc context) as it's annoying to have to correlate logs based on cluster name.

PR #1815, opened 2017-11-21
fix(provider/kubernetes): add more checking for verifying runing instances.

chlung:
About this PR,

Add a new onDemand type for controller force refresh.

Thanks.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
@@ -36,7 +36,6 @@ import static com.netflix.spinnaker.orca.ExecutionStatus.SUCCEEDED
 @Slf4j
 class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask implements RetryableTask {
   static final String REFRESH_TYPE = "ServerGroup"
-  static final String REFRESH_CONTROLLERS_TYPE = "ControllerSet"
 
   long backoffPeriod = TimeUnit.SECONDS.toMillis(10)
   long timeout = TimeUnit.MINUTES.toMillis(15)
@@ -120,12 +119,6 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
           status = SUCCEEDED
         }
 
-        response = cacheService.forceCacheUpdate(cloudProvider, REFRESH_CONTROLLERS_TYPE, model)
-        if (response.status == HttpURLConnection.HTTP_OK) {
-           //cache update was applied immediately, no need to poll for completion
-           status = SUCCEEDED
-        }
-
         stageData.refreshedServerGroups << model
       } catch (e) {
         stageData.errors << e.message
@@ -148,21 +141,8 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
                                                   String cloudProvider,
                                                   StageData stageData,
                                                   Long startTime) {
-    boolean serverGroupRefreshStatus = processPendingForceCacheUpdates(account, cloudProvider, stageData, startTime, REFRESH_TYPE) ?: false
-    boolean controllerRefreshStatus = processPendingForceCacheUpdates(account, cloudProvider, stageData, startTime, REFRESH_CONTROLLERS_TYPE) ?: false
 
-    if (serverGroupRefreshStatus && controllerRefreshStatus) {
-      return true
-    } else {
-      return false
-    }
-  }
-  private boolean processPendingForceCacheUpdates(String account,
-                                                  String cloudProvider,
-                                                  StageData stageData,
-                                                  Long startTime,
-                                                  String refreshType) {
-    def pendingForceCacheUpdates = cacheStatusService.pendingForceCacheUpdates(cloudProvider, refreshType)
+    def pendingForceCacheUpdates = cacheStatusService.pendingForceCacheUpdates(cloudProvider, REFRESH_TYPE)
 
     boolean finishedProcessing = true
     stageData.deployServerGroups.each { String region, Set<String> serverGroups ->
@@ -179,7 +159,11 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
         }
 
         def forceCacheUpdate = pendingForceCacheUpdates.find {
-          (it.details as Map<String, String>).intersect(model) == model
+          if (it) {
+            (it.details as Map<String, String>).intersect(model) == model
+          } else {
+            return false
+          }
         }
 
         if (!forceCacheUpdate) {
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
@@ -36,6 +36,7 @@ import static com.netflix.spinnaker.orca.ExecutionStatus.SUCCEEDED
 @Slf4j
 class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask implements RetryableTask {
   static final String REFRESH_TYPE = "ServerGroup"
+  static final String REFRESH_CONTROLLERS_TYPE = "ControllerSet"
 
   long backoffPeriod = TimeUnit.SECONDS.toMillis(10)
   long timeout = TimeUnit.MINUTES.toMillis(15)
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -36,6 +36,7 @@ import static com.netflix.spinnaker.orca.ExecutionStatus.SUCCEEDED
 @Slf4j
 class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask implements RetryableTask {
   static final String REFRESH_TYPE = "ServerGroup"
+  static final String REFRESH_CONTROLLERS_TYPE = "ControllerSet"
```
lwander: Why can't the "controllerset" type accept refreshes for server groups instead of adding this new type here?

chlung: We talked abou this.  I thought that you would like to break into two different request call.

lwander: No maybe we miscommunicated, clouddriver should handle the single on demand update for any controller that can, which may be more than one.

chlung:
@lwander updated with your feedback. Please review.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy
```diff
@@ -43,6 +43,11 @@ class WaitForCapacityMatchTask extends AbstractInstancesCheckTask {
     if (!serverGroup.capacity || serverGroup.capacity.desired != instances.size()) {
       return false
     }
+
+    if (stage.context.get("capacity")?.get("desired") != instances.size()) {
```
lwander: If this key is missing, the stage will fail every time

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -158,7 +159,11 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
         }
 
         def forceCacheUpdate = pendingForceCacheUpdates.find {
-          (it.details as Map<String, String>).intersect(model) == model
+          if (it) {
```
lwander: Why is this needed?

chlung: It is possible to send null to here because both agent is send data here here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy
chlung: you mean context?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy
```diff
@@ -43,6 +43,13 @@ class WaitForCapacityMatchTask extends AbstractInstancesCheckTask {
     if (!serverGroup.capacity || serverGroup.capacity.desired != instances.size()) {
       return false
     }
+
+    if (stage.context.get("capacity") && stage.context.get("capacity").get("desired")) {
```
lwander: Why doesn't the above check work? Shouldn't the server group have the capacity set?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTaskSpec.groovy
```diff
@@ -49,7 +49,7 @@ class WaitForCapacityMatchTaskSpec extends Specification {
       oort = Stub(OortService)
       oort.getServerGroup("test", "us-east-1", "kato-main-v000") >> { new Response('kato', 200, 'ok', [], new TypedString(mapper.writeValueAsString(serverGroup))) }
       task.oortService = oort
-      def context = [account: "test", "deploy.server.groups": ["us-east-1": ["kato-main-v000"]]]
+      def context = [account: "test", "deploy.server.groups": ["us-east-1": ["kato-main-v000"]], capacity: [desired: 3]]
```
lwander: Are these changes still needed with the above null check?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy
chlung: This is possible my use case.  First, I have stage is set size to 0.  I think that it is timing during the resizing.  The above condition will satisfy and return true and it won't checking the size anymore and launch next stage.  So, some of the pod might not shutdown before remove the statefulset.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTaskSpec.groovy
chlung: I will check

chlung:
@lwander  I rollback unit test changes because null checking can resolve the exception.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy
lwander: @ajordens what are your thoughts on this change? I'm worried it'll have some weird consequences

chlung: stage should have the desired instance.  return false will ask to recheck.  I can move the code the end before return true.  That might reduce some impact.  How do you think?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
ajordens: That seems like an error on the cloud provider agent, I can't think of a good reason why `pendingForceCacheUpdates` would contain a `null` value.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy
ajordens: I'm not sure what we're checking here ... it looks like this change is relying on groovy truth and offering up some different behavior when `capacity.desired == 0`?

I don't believe we should short-circuit this task until the actual instance count matches desired.

Is that problem that you don't want to wait after you have resized to zero?

chlung: My test case is for statefulset and daemon controllers which don't need disable serverGroup.
- create a pipeline
- create resize server group stage and set to 0 ( currently running 4 pods)
- create a destroy server group stage to delete controller.

First onDemand call,
(!serverGroup.capacity || serverGroup.capacity.desired != instances.size())  ==> true (4=4)

Now, we reach the following line.  serverGroup.disabled = false.  Now, it will return true and stop request any update and wait.
!serverGroup.disabled ?
      WaitForUpInstancesTask.allInstancesMatch(stage, serverGroup, instances, interestingHealthProviderNames) :
      true

chlung: My goal is make sure to wait.  If I delete controller before pods terminated.  It will have orphan pods.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
chlung: I will take a look if it is possible filter in clouddriverside  if null.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy
ajordens: If you've resized to zero, `capacity.desired` would not equal `instances.count()` ... 0 != 4.

I must be missing something?

chlung:
@lwander amd @ajordens - I found a bug in ControllerAgent(clouddriver side).  I will close this PR for now.  Thanks for your helps.

PR #1817, opened 2017-11-22
refactor(expressions): Remove v1 SPEL code

jeyrschabu:
- Making v2 as the default engine for expressions
- More improvements on the way


Nagarajj:
@jeyrschabu We plan to use Pipeline expressions to evaluate dynamic code(make REST API calls) to build custom headers as part of webhook stage. 

But in the current state CPP only allows expressions defined in "WhitelistTypeLocator" to be executed. Could you please guide on how can custom code be executed as part of pipeline stage into spinnaker.

jeyrschabu:
@Nagarajj Expressions are meant to be used for simple evaluations and not for heavy handed scenarios. A `Jenkins` stage or `Script` stage might be more appropriate for your use case. Feel free to open up an issue with some details about what you are trying to accomplish and tag me in it. We can continue the conversation there.

PR #1819, opened 2017-11-27
Xenial builds

brandonnelson3:
Added systemd configuration for orca


orca-web/lib/systemd/system/orca.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Spinnaker Orca
+PartOf=spinnaker.service
+
+[Service]
+ExecStart=/opt/orca/bin/orca 2>&1 /var/log/spinnaker/orca/orca.log
+WorkingDirectory=/opt/orca/bin
+SuccessExitStatus=143
```
lwander: heh

brandonnelson3: Nothing suspicious here :)

Just so when someone in the future comes looking:

https://docs.spring.io/spring-boot/docs/current/reference/html/deployment-install.html#deployment-systemd-service

PR #1822, opened 2017-11-29
fix(templates): Tolerate all failures on execution lookup.

jtk54:


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -123,7 +123,7 @@ class OperationsController {
           def previousExecution = pipelineTemplateService.retrievePipelineOrNewestExecution(pipeline.executionId, pipeline.id)
           pipeline.trigger = previousExecution.trigger
           pipeline.executionId = previousExecution.id
-        } catch (ExecutionNotFoundException ignore) {
+        } catch (def ignore) {
```
ajordens: Normally you would just `Exception ignored` this ... the `def ignore` is a pattern I've not seen a lot of.

Is this exception truly ignorable? Even an `ExecutionNotFoundException` probably warrants some sort of an error message in the log?

jtk54: Yeah even `Exception` is too broad here, this block is best-effort to initialize the pipeline config from a previous execution, but I didn't realize that the `retrieve...` actually declares what it will throw: https://github.com/spinnaker/orca/blob/23802efc3be8e1f2c43d2a845ec7c170239eefd6/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplateService.java#L76.

I'll just catch those two and add a `log.info`.

PR #1827, opened 2017-11-30
feat(artifacts): support 'use prior execution'

lwander:


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.groovy
```diff
@@ -16,52 +16,107 @@
...
+  static class ResolveResult {
+    Set<Artifact> resolvedArtifacts = new HashSet<>()
+    Set<ExpectedArtifact> unresolvedExpectedArtifacts = new HashSet<>();
+  }
+
+  private static Closure startTimeOrId = { a, b ->
+    def aStartTime = a.startTime ?: 0
+    def bStartTime = b.startTime ?: 0
+
+    return bStartTime <=> aStartTime ?: b.id <=> a.id
```
jtk54: Are the `id`s incremental?

lwander: tie breaker

lwander: ( to group pipelines together )

PR #1832, opened 2017-12-05
fix(moniker): fix cluster if detail is set to empty via SpEL

asher:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/MonikerHelper.java
```diff
@@ -56,7 +57,11 @@ public String getClusterNameFromStage(Stage stage, String fallbackFriggaName) {
 
   static public Moniker monikerFromStage(Stage stage) {
     if (stage.getContext().containsKey("moniker")) {
-      return (Moniker) stage.mapTo("/moniker", Moniker.class);
+      Moniker moniker = stage.mapTo("/moniker", Moniker.class);
+      if (moniker.getCluster().endsWith("-")) {
```
ajordens: This doesn't seem quite right ... the Moniker shouldn't have an invalid cluster in the first place.

Can we fix it at Moniker creation time vs. fixing it on read afterwards?

asher: This issue in question is triggered when a pipeline.json contains something like: 
```
          "moniker": {
            "app": "spintest",
            "cluster": "spintest-asher-${parameters.foo ?: ''}",
            "detail": "${parameters.foo ?: ''}",
            "stack": "asher"
          },
```
Unless we don't want to support an optional parameterized cluster detail, seems like something needs to happen after SpEL evaluation in orca.

asher: @lwander @andrewbackes Is it the expected behavior for `deck` to pre-construct a cluster of the form `application-stack-details`?

This happens in a deploy stage if you use a SPeL expression for details.

andrewbackes: @asher - Deck will not always constructs a cluster name of the form `app-stack-detail`. Only providers still using frigga follow that scheme. The other `Namers` treat cluster as a free form field. Like @ajordens pointed out, I think it would be safest to handle this when the moniker is created. Something like:
```
          "moniker": {
            "app": "spintest",
            "cluster": "spintest-asher${parameters.foo ? '-' + parameters.foo : ''}",
            "detail": "${parameters.foo ?: ''}",
            "stack": "asher"
          },
```
There may be some unexpected behavior with this PR because users can still set the cluster value to whatever they want. Clouddriver does not reject cluster names that end in hyphen. But anything in Orca that uses the `monikerFromStage()` helper method will be stripping the hyphen. The monikers will not match. I'm not sure what all of the downstream consequences will be, but I think there will be some. TrafficGuard comes to mind.

asher: @andrewbackes the example you give would work with this PR reverted, but to automatically create that would require deck to parse and modify user provided SPeL expressions when generating a moniker. I'm hesitant to go down that path.

Can we avoid pre-constructing moniker.cluster entirely if using frigga naming?

andrewbackes: @asher I'm not sure I understand the use case you are describing. Where is the moniker being automatically generated? Is it somewhere in Deck?

andrewbackes: Also, what is the underlying issue you are trying to solve?

asher: screen shot 2017-12-06 at 2 24 31 pm

@andrewbackes - The underlying problem is that this generates a broken `moniker.cluster` in the pipeline json, as in my first comment here, which breaks deploys when the detail parameter is unset.

anotherchrisberry: In the case of a pipeline stage where the user selects a cluster via the cluster selector, it's not that uncommon for the user to select a cluster, then edit the `cluster` field in the JSON.

This change seems like a reasonable safeguard against a number of scenarios where we'd end up with one or two trailing dashes due to parameters.

ajordens: i'm actually less concerned about that, because it's up to the user to build a properly formed cluster.

The issue here is that we're concatenating app, stack and details, and it's not the user telling us explicitly what the cluster is.

I don't actually know that we need to send a `cluster` across as part of deploy stage, pretty sure we weren't do that previously were we? I don't actually know what the UI looks like for a provider that isn't frigga-based and how such a thing is even built up.

anotherchrisberry: I don't think this is from a deploy stage. It's from anything that extends `AbstractClusterWideClouddriverTask` - we've always sent `cluster` as part of that operation

andrewbackes: In the spirit of sharing information, here is a screenshot of the k8s v2 deploy wizard.

k8s-v2 deploy stage

Disclaimer: This is an early draft and is will probably change a few times.

andrewbackes: My only concern is around Orca chopping off the hyphen but clouddriver not. Seems like that would be something very hard to debug if it ever came up for somebody. I'd prefer consistency.

asher: I manually edited my test redblack pipeline that contained the following (auto generated via the aws deploy wizard):
```
          "moniker": {
            "app": "spintest",
            "cluster": "spintest-asher-${parameters.foo ?: ''}",
            "detail": "${parameters.foo ?: ''}",
            "stack": "asher"
          },
```
And removed the pre-generated cluster to get:
```
          "moniker": {
            "app": "spintest",
            "detail": "${parameters.foo ?: ''}",
            "stack": "asher"
          },
```
And, oops - got a NPE from `monikerFromStage()` thanks to this very PR. Fixed the NPE and everything worked as it would have pre-moniker with an empty `parameters.foo`.

So, how about if instead of this PR, we modify deck to stop pre-constructing `moniker.cluster` in the form of `application-stack-details`? My test obviously wasn't exhaustive in confirming that there aren't dependencies on the presence of a `moniker.cluster` if `moniker` exists, but it feels like the right direction to go in.

lwander: Sorry i've been OOO - what you're suggesting sounds best to me asher, and this is the direction the k8s provider takes moniker.

andrewbackes: @asher - I think that sounds like a good approach.

asher:
Thanks, all. Made the deck change @ https://github.com/spinnaker/deck/pull/4570 and reverting this @ https://github.com/spinnaker/orca/pull/1849.

PR #1836, opened 2017-12-06
fix(runJob): retry fetching property files from CI if not found

xabierlaiseca:
We have seen a few cases where no properties were returned by
igor even the CI tool (Travis in our case) printed the magic json
search string in logs.

Somehow related to https://github.com/spinnaker/orca/pull/1721


duftler:
This looks ok to me.
@robzienert @ajordens @tomaslin Any concerns?

PR #1837, opened 2017-12-06
Deployment details no longer accessed from global context

robfletcher:
This ensures deployment details are retrieved from upstream stages, trigger or parent pipeline without using global context

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
@@ -55,7 +55,6 @@ class PackageInfo {
   public Map findTargetPackage(boolean allowMissingPackageInstallation) {
     Map requestMap = [:]
     // copy the context since we may modify it in createAugmentedRequest
-    requestMap.putAll(stage.execution.context)
     requestMap.putAll(stage.context)
 
     if (stage.execution.type == PIPELINE) {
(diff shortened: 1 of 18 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
```diff
@@ -224,14 +224,16 @@ class PackageInfo {
...
-      artifactMatch(it.context.buildInfo?.artifacts as List<Map<String, String>>, packageFilePattern)
+      artifactMatch(it.outputs.buildInfo?.artifacts as List<Map<String, String>>, packageFilePattern)
     }
-    return upstreamStage ? upstreamStage.context.buildInfo as Map : null
+    return upstreamStage ? upstreamStage.outputs.buildInfo as Map : null
   }
 
-  private static boolean artifactMatch(List<Map<String, String>> artifacts, Pattern pattern) {
+  private
+  static boolean artifactMatch(List<Map<String, String>> artifacts, Pattern pattern) {
```
ajordens: mr line breaky break

ajordens:
Are all the stages putting the right bits into `outputs`? Guessing you had changed that previously?

ajordens:
Worth sanity checking that deployment strategies still work with both a bake and find image stage in the parent stage (verify both independently).

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy
robfletcher: Ugh.

PR #1839, opened 2017-12-07
fix(mdc): `executionId` was not propagated across threads in a few spots

ajordens:


ajordens:
@asher This fixes a few spots I noticed when running locally

@robfletcher Did I use the kotlin extension fn's correctly?

PR #1840, opened 2017-12-07
feat(expressions): deployed server groups helper functions & misc fixes

jeyrschabu:
- added helper function for deployedServerGroups

- allows specifying an optional name/id for the stage to filter on

- added alias to deployedServerGroups


jeyrschabu:
Previously, deployedServerGroups was added to the global context during expression processing.
With global context out of the picture, `deployedServerGroups` was no longer available.

This PR adds two things:
- A new SPEL helper function to get `deployedServerGroups`. This function accepts an optional list of stage names/id for filtering.
- An alias to the `deployedServerGroups` helper function. This is for backward compatibility and will be removed in the future

Though the old functionality is still supported, the new recommended way to use `deployedServerGroups` is with the helper function:
```
${#deployedServerGroups()}
or
${#deployedServerGroups("Deploy in us-east-1")}
```

PR #1842, opened 2017-12-08
fix(expressions): process notification content

jeyrschabu:
- explicitly processing expressions in notification content

### DIFF
--- a/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingExecutionListener.groovy
+++ b/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingExecutionListener.groovy
@@ -65,10 +65,7 @@ class EchoNotifyingExecutionListener implements ExecutionListener {
             type       : "orca:${execution.type}:starting".toString(),
             application: execution.application,
           ],
-          content: [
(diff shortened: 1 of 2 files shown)


orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingExecutionListener.groovy
```diff
@@ -59,16 +59,18 @@ class EchoNotifyingExecutionListener implements ExecutionListener {
...
           details: [
             source     : "orca",
             type       : "orca:${execution.type}:starting".toString(),
             application: execution.application,
           ],
-          content: [
-            execution  : execution,
-            executionId: execution.id
-          ]
+          content: contextParameterProcessor.process([execution: execution] as Map<String, Object>, content)
```
ajordens: Not sure if it's worth cutting down the duplication with ...

```
private Map<String, Object> buildContent(Execution execution) {
  return contextParameterProcessor.process(
    [execution: execution] as Map<String, Object>,
    [
      execution: execution, 
      executionId: execution.id
    ]
  )
}
```

```
echoService.recordEvent(
  ...
  content: buildContent(execution)
)
```
?

PR #1843, opened 2017-12-08
feat(core): AFTER_STAGE stages can now be planned post task completion

ajordens:
This allows for interesting scenarios where a stage could have a `Task`
responsible for building the context that is subsequently used to build
a dynamic set of synthetic stages.

Currently this logic must reside entirely in `aroundStages()` and you
lose the ability to arbitrarily retry with backoff's, etc. that normal
`Task` execution provides.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/cluster/RollbackClusterStage.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/cluster/RollbackClusterStage.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline.cluster;
+
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.RollbackServerGroupStage;
+import com.netflix.spinnaker.orca.clouddriver.tasks.cluster.DetermineRollbackCandidatesTask;
+import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder;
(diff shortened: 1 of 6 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/cluster/RollbackClusterStage.java
```diff
@@ -0,0 +1,99 @@
...
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder.newStage;
+
+@Component
+public class RollbackClusterStage implements StageDefinitionBuilder {
```
ajordens: Consider this a proof of concept that takes advantage of the new `afterStages()` support.

The implementation of `DetermineRollbackCandidatesTask` is not entirely complete.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandlerTest.kt
```diff
@@ -42,9 +43,10 @@ object CompleteStageHandlerTest : SubjectSpek<CompleteStageHandler>({
   val publisher: ApplicationEventPublisher = mock()
   val clock = fixedClock()
   val contextParameterProcessor: ContextParameterProcessor = mock()
+  val stageDefinitionBuilderFactory: StageDefinitionBuilderFactory = mock()
 
   subject(GROUP) {
-    CompleteStageHandler(queue, repository, publisher, clock, contextParameterProcessor)
+    CompleteStageHandler(queue, repository, publisher, clock, contextParameterProcessor, stageDefinitionBuilderFactory)
```
ajordens: If we're happy with the approach, I'll add an additional test in here to validate that `planAfterStages()` is invoked as expected.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/StageDefinitionBuilder.java
```diff
@@ -50,6 +50,10 @@ default void taskGraph(
     return emptyList();
   }
 
+  default @Nonnull List<Stage> afterStages(@Nonnull Stage stage) {
```
ajordens: My expectation is that we can subsequently start moving `STAGE_AFTER` definitions from `aroundStages()` to `afterStages()`.

robfletcher: What happens if you have after stages in both methods?

ajordens: Your `afterStages()` would not get planned ... the `CompleteStageHandler` first calls `Stage.firstAfterStages()` to verify that there aren't any synthetic after stages to begin with.

PR #1847, opened 2017-12-08
fix(expressions): Fixing spel in notifications

jeyrschabu:
- processing the notification content


orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingExecutionListener.groovy
```diff
@@ -146,11 +146,11 @@ class EchoNotifyingExecutionListener implements ExecutionListener {
 
   private Map<String, Object> buildContent(Execution execution) {
     return contextParameterProcessor.process(
-      [execution: execution] as Map<String, Object>,
       [
         execution: execution,
         executionId: execution.id
-      ] as Map<String, Object>
+      ] as Map<String, Object>,
+      [execution: execution] as Map<String, Object>
```
christopherthielen: How did this fix it??

PR #1852, opened 2017-12-11
fix(manifests): s/name/manifestName

lwander:
https://github.com/spinnaker/clouddriver/pull/2222


lwander:
@andrewbackes

andrewbackes:
In the case of using `labelSelectors`, `manifestName` will not be present. I think you will get something like this in Deck:

Error

Also, I think name is used here: https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java

lwander:
That's fair, but for a separate PR

lwander:
I'm planning on having clouddriver return every manifest it attempted to modify, that orca can then refresh and monitor all at once

lwander:
https://github.com/spinnaker/spinnaker/issues/2169

PR #1854, opened 2017-12-12
feat(queue): stage execution monitoring

robzienert:
A first step towards better metrics around stage execution information. Records invocations (starts), and finally durations. I'll followup later with a Redis-backed strategy for deriving active stages, as well as better resolution of cloudProvider in other PRs.


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -76,10 +90,42 @@ class CompleteStageHandler(
...
     }
   }
 
+  private fun trackResult(stage: Stage) {
+    // We only want to record durations of parent-level stages; not synthetics.
+    if (stage.parentStageId != null) {
+      return
+    }
+
+    val id = registry.createId("stage.invocations.duration")
```
ajordens: So the only downfall with this is that I reckon the tag space is going to be _gigantic_ and we're not going to be able to look at too much history.

We used to have a similar metric and were pretty much limited to the last 3hrs (which might be perfectly fine!).

PR #1855, opened 2017-12-13
feat(queue): Active executions monitoring

robzienert:
First (well, second) pass at active execution monitoring.


orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisActiveExecutionsMonitor.kt
```diff
@@ -0,0 +1,172 @@
...
+
+  private fun getActiveExecutions() =
+    pool.resource.use { redis ->
+      redis.hgetAll(REDIS_KEY).map { objectMapper.readValue(it.value, ActiveExecution::class.java) }
+    }
+
+  private fun ActiveExecution.getMetricId() =
+    registry.createId("executions.active")
+      .withTag("type", type.toString())
+      .withTag("application", application)
```
ajordens: I don't think we need application here, having it is going to dramatically up the tag space and may cause issues trying to look historically.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisActiveExecutionsMonitor.kt
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  private fun getActiveExecutions() =
+    pool.resource.use { redis ->
+      redis.hgetAll(REDIS_KEY).map { objectMapper.readValue(it.value, ActiveExecution::class.java) }
+    }
+
+  private fun ActiveExecution.getMetricId() =
+    registry.createId("executions.active")
+      .withTag("type", type.toString())
```
ajordens: Maybe call this `executionType` to be consistent with other metrics (`task.invocations` at least)

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisActiveExecutionsMonitor.kt
asher: I'd continue to store application in monitor.activeExecutions[execution.id] though, could be useful for /admin functions.

ajordens: Agreed, that's a great call @asher!

PR #1858, opened 2017-12-13
fix(core): Fix leaky pipeline expression evaluator test

robzienert:
Statics are evil.


robzienert:
Welp, a hack didn't fix it - I'm going to refactor the class so it isn't static at all. Stay tuned.

PR #1862, opened 2017-12-14
feat(helm): Add Helm ServerGroup tasks creator

posox:
Add Helm ServerGroup tasks creator

This patch adds support creation of server groups using Helm clouddriver.


skim1420:
I also feel helm support in Spinnaker needs dedicated discussion in the community prior to doing work in earnest in support of it. I have questions about how safe deployments are supported when doing helm chart deployments (e.g. how do you roll back an update pushed via a helm chart?)

As Lars suggests, perhaps a design doc is a good point to get this discussion started. Let's please avoid carrying the debate here in this PR; if you really can't avoid the temptation, I suggest you start a topic in community.spinnaker.io :)

robzienert:
@skim1420 Do we want to leave this open?

skim1420:
@lwander any further update on this?

lwander:
Let's close this for now as this generally doesn't work with the approach we're taking w/ helm. The consensus seems to be that the best spot for helm integration is to rely on its templating to produce manifests to deploy using the Kubernetes provider, rather than surface helm objects as first-class spinnaker resources. We can always revisit or rethink this.

PR #1863, opened 2017-12-14
fix(provider/kubernetes): correctly handle ops on many manifests

lwander:
https://github.com/spinnaker/spinnaker/issues/2170
closes https://github.com/spinnaker/spinnaker/issues/2169


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeleteManifestStage.java
```diff
@@ -33,6 +33,6 @@
   public void taskGraph(Stage stage, TaskNode.Builder builder) {
     builder.withTask(DeleteManifestTask.TASK_NAME, DeleteManifestTask.class)
         .withTask("monitorDelete", MonitorKatoTask.class)
-        .withTask(DeleteManifestForceCacheRefreshTask.TASK_NAME, DeleteManifestForceCacheRefreshTask.class);
```
danielpeach: Looks like this is the only reference to that task - can it be deleted?

lwander: Good catch

PR #1866, opened 2017-12-15
feat(core): Support for rolling back a cluster (across multiple regions)

ajordens:
This PR is motivated by a desire for a smarter rollback capable of operating on more than one region.

It builds on top of the existing server group rollback capabilities and makes a determination on what to rollback based on a two factors.

Whether or not the current server group has a a `spinnaker:metadata` entity tag indicating the previously deployed image.

If the entity tag exists, we will attempt to (a) enable an existing server group with that image or (b) perform a red/black with the previous image.

If the entity tag does not exist, we will attempt to rollback to the most recent previous server group, regardless of what image it was deployed with.

This is intended to be cross cloud provider but your mileage may vary with anything beyond AWS right now.


ajordens:
Best to look at the `RollbackClusterStage` and `DetermineRollbackCandidateTask` in raw form since there was a skeleton committed previously (diff kinda screws it up!).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/PreviousImageRollbackSupport.groovy
```diff
@@ -0,0 +1,78 @@
+/*
```
robzienert: No javas?

ajordens: That’s fair, this was originally just lifted from an existing  groovy class but I can add a bunch more null checks and do the java conversion.

PR #1869, opened 2017-12-16
feat(artifacts): find artifact from pipeline execution

lwander:


lwander:
https://github.com/spinnaker/spinnaker/issues/2175

PR #1872, opened 2017-12-19
fix(queue): really cancel queued pipelines that were canceled

emjburns:
Looks like we were never canceling pipelines that were canceled while queued. This will not start a pipeline if it is in a canceled state.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartExecutionHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartExecutionHandler.kt
@@ -16,12 +16,18 @@
 
 package com.netflix.spinnaker.orca.q.handler
 
+import com.netflix.spinnaker.orca.ExecutionStatus.CANCELED
+import com.netflix.spinnaker.orca.ExecutionStatus.NOT_STARTED
 import com.netflix.spinnaker.orca.ExecutionStatus.RUNNING
+import com.netflix.spinnaker.orca.events.ExecutionComplete
 import com.netflix.spinnaker.orca.events.ExecutionStarted
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
 import com.netflix.spinnaker.orca.q.*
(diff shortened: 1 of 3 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartExecutionHandler.kt
```diff
@@ -34,15 +36,19 @@ class StartExecutionHandler(
 
   override fun handle(message: StartExecution) {
     message.withExecution { execution ->
-      repository.updateStatus(message.executionId, RUNNING)
+      if (execution.status == CANCELED || execution.isCanceled) {
```
ajordens: Worth checking any other status ... the execution should really be NOT_STARTED?

emjburns: Yup I agree, I had that thought last night - changing!

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartExecutionHandler.kt
```diff
@@ -34,15 +36,20 @@ class StartExecutionHandler(
 
   override fun handle(message: StartExecution) {
     message.withExecution { execution ->
-      repository.updateStatus(message.executionId, RUNNING)
+      // isCanceled flag lets the pipeline clean itself up, and may be true before the status is set to CANCELED
+      if (execution.status != NOT_STARTED || execution.isCanceled) {
+        publisher.publishEvent(ExecutionComplete(this, message.executionType, message.executionId, execution.status))
```
robfletcher: Do we really want to fire this message for _any_ other status? Trying to think what the possible causes are here.

emjburns: Changing as we discussed - not sending message if it's not canceled.

PR #1874, opened 2017-12-19
fix(core): Generate the correct lock key

ajordens:


ajordens:
Make orca great again

PR #1878, opened 2017-12-21
feat(provider/appengine): Allow app engine flex deploys with container image URL

sbwsg:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
```diff
@@ -60,7 +60,7 @@ class AppEngineServerGroupCreator implements ServerGroupCreator {
   }
 
   private void resolveArtifacts(Stage stage, Map operation) {
-    if (operation.repositoryUrl) {
+    if (operation.repositoryUrl || operation.containerImageUrl) {
```
lwander: I think this resolveArtifacts stage should get some love - right now orca is manually parsing & manipulating artifact objects, when ideally it should just be passing them directly through to clouddriver, which (currently) is the only service with integrations to fetch/download/parse artifacts.

lwander: Not something to do in this PR necessarily, but something to be aware of.

danielpeach: Agreed - I think this will be refactored (and all pushed to Clouddriver) when Scott adds artifact support for deploying containers.

sbwsg:
fixed commit message + PR title

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
lwander: :+1:

PR #1879, opened 2018-01-03
feat(provider/kubernetes): find deployed artifacts from context

lwander:


lwander:
https://github.com/spinnaker/spinnaker/issues/2184

PR #1887, opened 2018-01-08
feat(webhooks): support preconfig params

ethanfrogers:
adds support for configuring predefined parameters as part of a
preconfigured webhook stage. this will allows admins to add options for
users to fill out. the values set by users can then be used in the other
options via SPEL. this is a really basic implementation and only
supports string types options.

spinnaker/spinnaker#2224

@robzienert @robfletcher @ajordens could some of you guys take a look at this. it's my first orca PR and _seems_ like it was easier than it should have been. i'm open to having a deeper conversation around it if need be. thanks!

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -222,7 +222,8 @@ class OperationsController {
         type: it.type,
         waitForCompletion: it.waitForCompletion,
         preconfiguredProperties: it.preconfiguredProperties,
-        noUserConfigurableFields: it.noUserConfigurableFields()
+        noUserConfigurableFields: it.noUserConfigurableFields(),
+        parameters: it.parameters,
       ]
     }
   }
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/config/PreconfiguredWebhookProperties.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/config/PreconfiguredWebhookProperties.groovy
@@ -26,7 +26,7 @@ import org.springframework.http.HttpMethod
 class PreconfiguredWebhookProperties {
 
   public static List<String> ALL_FIELDS = PreconfiguredWebhook.declaredFields.findAll {
-    !it.synthetic && !['props', 'enabled', 'label', 'description', 'type'].contains(it.name)
+    !it.synthetic && !['props', 'enabled', 'label', 'description', 'type', 'parameters', 'parameterValues'].contains(it.name)
   }.collect { it.name }
 
   List<PreconfiguredWebhook> preconfigured = []
@@ -37,10 +37,12 @@ class PreconfiguredWebhookProperties {
     String label
     String description
     String type
+    List<WebhookParameter> parameters
 
     // Stage configuration fields (all optional):
     String url
     Map<String, List<String>> customHeaders
+    Map<String, String> parameterValues
     HttpMethod method
     String payload
     Boolean waitForCompletion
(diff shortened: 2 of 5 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -222,7 +222,8 @@ class OperationsController {
         type: it.type,
         waitForCompletion: it.waitForCompletion,
         preconfiguredProperties: it.preconfiguredProperties,
-        noUserConfigurableFields: it.noUserConfigurableFields()
+        noUserConfigurableFields: it.noUserConfigurableFields(),
+        webhookParams: it.webhookParams
```
ajordens: maybe just `parameters` vs `webhookParams`

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/config/PreconfiguredWebhookProperties.groovy
```diff
@@ -67,6 +68,11 @@ class PreconfiguredWebhookProperties {
     }
   }
 
+  static class WebhookParameter {
+    String label
+    String defaultValue
```
ajordens: The other ticket also mentions a `type` ... that's probably also relevant here.

I could imagine drop-down's, etc.

ajordens:
What do you imagine happening for the web hook stage's execution view? Is there any aspect of that that should be customizable?

I suspect there's also a `deck` PR that handles the actual layout and rendering of the parameters?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ethanfrogers: exactly the type of feedback i wanted. i definitely like that better. thanks @ajordens!

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/config/PreconfiguredWebhookProperties.groovy
ethanfrogers: I thought about implementing that in this PR, but wanted to get the basics in first. I suppose I could add that and default to string type. other types could be added later.

ethanfrogers:
I don't think anything in the execution view should be customizable. My first thought was to emulate what happens with Jenkins parameters. Display the value of each parameter the webhook was executed with. 

There _will_ be a deck PR once I have this nailed down. I wanted to make sure I had the structure down before I dove into the UI.

ethanfrogers:
Also, I'll tackle updating that failing test ASAP. I've known about just haven't had the bandwidth to fix it.

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/config/PreconfiguredWebhookProperties.groovy
```diff
@@ -67,6 +69,17 @@ class PreconfiguredWebhookProperties {
...
+    String value
+    ParameterType type = string
+
+    static enum ParameterType {
```
ethanfrogers: @ajordens i added this here. if you can think of a better way to do this, let me know.

ethanfrogers:
i think i may refactor parameters into a `List<WehbhookParameter>` to make it easier to work with. I started working on the Deck side and working with the `Map` is tough.

ajordens:
@ethanfrogers List makes sense to me

ethanfrogers:
@ajordens this is good to go for me. the final config for parameters would look like this:

```yaml
parameters:
        - name: appId
          label: Application ID
          description: The ID of your application in New Relic
        - name: revision
          label: Revision
          description: The revision being deployed
        - name: user
          label: User
          defaultValue: ${execution.authentication.user}
```

the user defined values for these parameters can be accessed via the `parameterValues` context object:

```yaml
payload: |
        {
          "deployment": {
            "revision": "${parameterValues['revision']}",
            "user": "${parameterValues['user']}"
          }
        }
```

PR #1889, opened 2018-01-10
feat(tasks) - adds configuration property to waitForCluster timeouts

imosquera:
Adds a configuration property to allow for longer timeouts for WaitForCluster tasks.   The default is 30 minutes.

Example of having a timeout of 60 minutes:
```
tasks:
  waitForClusterTimeout: 3600000
```


avram:
For context-- this is something that we ran into when using [AWS's autoscaling group lifecycle hooks](https://docs.aws.amazon.com/autoscaling/latest/userguide/AutoScalingGroupLifecycle.html#as-lifecycle-hooks); we use them in conjunction with [shudder](https://github.com/scopely/shudder) to do safe shutdowns of [Airflow](https://github.com/apache/incubator-airflow) workers, but that can mean that instances are in the Terminating:WAIT state for longer than Orca is willing to wait.

PR #1891, opened 2018-01-12
feat(keel): upsert and delete intents through stages

emjburns:
Adding keel module to orca.
Upserting and deleting intents via stages. 

Fixing a pipelinetemplate error message that was bothering me


ajordens:
Since this is a brand new artifact, thoughts on just starting with `kotlin` vs combo `java` and `groovy`?

robzienert:
Agree re: Kotlin.

emjburns:
@ajordens I agree, will change to Kotlin.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/config/KeelConfiguration.kt
```diff
@@ -0,0 +1,66 @@
...
+import retrofit.Endpoints
+import retrofit.RestAdapter
+import retrofit.client.Client
+import retrofit.converter.JacksonConverter
+
+@Configuration
+@ConditionalOnProperty("keel.enabled")
+@ComponentScan(
+  basePackages = [
+    "com.netflix.spinnaker.orca.keel",
```
robzienert: Can remove this one, we've no components on this package path.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/config/KeelConfiguration.kt
```diff
@@ -0,0 +1,66 @@
...
+    "com.netflix.spinnaker.orca.keel.pipeline"
+  ]
+)
+open class KeelConfiguration {
+  @Bean open fun keelEndpoint(@Value("\${keel.baseUrl}") keelBaseUrl: String): Endpoint {
+    return Endpoints.newFixedEndpoint(keelBaseUrl)
+  }
+
+  @Bean open fun keelService(keelEndpoint: Endpoint,
+                  objectMapper: ObjectMapper,
```
robzienert: You'll want this to be `keelObjectMapper: ObjectMapper` so we wire in the correct one.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/DeleteIntentTask.kt
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    val intentId = stage.context["intentId"].toString()
+
+    val status = if (stage.context.containsKey("status")) stage.context["status"].toString() else null
+
+    val response = keelService.deleteIntents(intentId, status)
+
+    val outputs = HashMap<String, Any>()
+    outputs.put("intent.id", intentId)
```
robzienert: You can do: `val outputs = mapOf("intent.id" to intentId)`

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/DeleteIntentTask.kt
```diff
@@ -0,0 +1,65 @@
...
+    )
+  }
+
+  override fun getBackoffPeriod(): Long {
+    return TimeUnit.SECONDS.toMillis(15)
+  }
+
+  override fun getTimeout(): Long {
+    return TimeUnit.MINUTES.toMillis(1)
+  }
```
robzienert: Stylistic choice, but when returning a single value from an expression, I tend to just do something like:

`fun bleh() = TimeUnit.MINUTES.toMillis(1)`

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
```diff
@@ -0,0 +1,88 @@
...
+  private val log = LoggerFactory.getLogger(javaClass)
+
+  override fun execute(stage: Stage) : TaskResult {
+    if (!stage.context.containsKey("intents")) {
+      throw IllegalArgumentException("Missing required task parameter (intents)")
+    }
+
+    if (!stage.context.containsKey("dryRun")) {
+      throw IllegalArgumentException("Missing required task parameter (dryRun)")
+    }
```
robzienert: I would aggregate these required fields together and throw an exception instead. That would allow an end-user to submit once and get all errors without resubmitting and being surprised by another error:

```kotlin
val errors = mutableListOf<String>()
if (!stage.context.containsKey("intents") {
  errors.add("intents")
}
// ...

if (errors.isNotEmpty()) {
  throw IllegalArgumentException("Missing required task parameters (${errors.join(", ")})"
}
```

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
```diff
@@ -0,0 +1,88 @@
...
+
+    val upsertIntentRequest = UpsertIntentRequest(
+      intents = keelObjectMapper.convertValue(stage.context["intents"], object : TypeReference<List<Intent>>() {}),
+      dryRun = java.lang.Boolean.valueOf(stage.context["dryRun"].toString())!!
```
robzienert: Can we not do `stage.context["dryRun"].toString().toBoolean()` instead?

robzienert: I also think avoiding `!!` is a good idea. That's a sure-fire way to get an exception, especially with orca contexts. What we could do is something like... `if (stage.context["dryRun"] == null) false else stage.context["dryRun"].toString().toBoolean()` so we're handling the null case explicitly.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
```diff
@@ -0,0 +1,88 @@
...
+    try {
+      if (upsertIntentRequest.dryRun) {
+        val dryRunResponse = keelObjectMapper.readValue<List<UpsertIntentDryRunResponse>>(response.body.`in`(), object : TypeReference<List<UpsertIntentDryRunResponse>>(){})
+        outputs.put("upsertIntentResponse", dryRunResponse)
+      } else {
+        val upsertResponse = keelObjectMapper.readValue<List<UpsertIntentResponse>>(response.body.`in`(), object : TypeReference<List<UpsertIntentResponse>>(){})
+        outputs.put("upsertIntentResponse", upsertResponse)
+      }
+    } catch (e: Exception) {
+        log.warn("Error processing upsert intent response")
```
robzienert: We should definitely re-throw the exception here so the task fails, and we get the context of the failure in the task / logs.

`throw RuntimeException("Error processing upsert intent response from keel", e)`

orca-web/src/main/groovy/com/netflix/spinnaker/orca/keel/model/OrchestrationViewModel.groovy
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.orca.model
+package com.netflix.spinnaker.orca.keel.model
```
robzienert: I don't think we want this?

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/DeleteIntentTask.kt
emjburns: Thanks for all the kotlin tips and tricks and style choices. Please keep them coming!

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
emjburns: Yes, that was leftover from some intellij magic of pasting java to kotlin. Fixing.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/keel/model/OrchestrationViewModel.groovy
emjburns: Don't know how that snuck in there!! Good catch.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/pipeline/DeleteIntentStage.kt
```diff
@@ -0,0 +1,35 @@
...
+
+@Component
+class DeleteIntentStage() : StageDefinitionBuilder {
+  override fun taskGraph(stage: Stage, builder: TaskNode.Builder) {
+    builder.withTask("deleteIntent", DeleteIntentTask::class)
+  }
+
+  private fun TaskNode.Builder.withTask(name: String, type: KClass<out Task>) =
+    withTask(name, type.java)
+}
```
robfletcher: A common pattern for that kind of thing would be to declare an extension function:

```kotlin
inline fun <reified T: Class<Task>> StageDefinitionBuilder.withTask(name: String) =
  withTask(name, T::class.java)
```

Which you could then call with `builder.withTask<DeleteIntentTask>("deleteIntent")`.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/DeleteIntentTask.kt
```diff
@@ -0,0 +1,59 @@
...
+  private val log = LoggerFactory.getLogger(javaClass)
+
+  override fun execute(stage: Stage) : TaskResult {
+    if (!stage.context.containsKey("intentId")) {
+      throw IllegalArgumentException("Missing required task parameter (intentId)")
+    }
+
+    val intentId = stage.context["intentId"].toString()
+
+    val status = if (stage.context.containsKey("status")) stage.context["status"].toString() else null
```
robfletcher:
```kotlin
val status = stage.context["status"]?.toString()
```
would do the same.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
```diff
@@ -0,0 +1,91 @@
...
+      missingParams.add("dryRun")
+    }
+
+    if (missingParams.isNotEmpty()) {
+      throw IllegalArgumentException("Missing required task parameters (${missingParams.joinToString(",")})")
+    }
+
+    val upsertIntentRequest = UpsertIntentRequest(
+      intents = keelObjectMapper.convertValue(stage.context["intents"], object : TypeReference<List<Intent>>() {}),
+      dryRun = if (stage.context["dryRun"] == null) false else stage.context["dryRun"].toString().toBoolean()
```
robfletcher: do we really need to do `toString().toBoolean()` there? What is the value actually?

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
```diff
@@ -0,0 +1,91 @@
...
+    )
+
+    val response = keelService.upsertIntents(upsertIntentRequest)
+
+    val outputs = HashMap<String, Any>()
```
robfletcher:
```kotlin
val outputs = mutableMapOf<String, Any>()
```

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
```diff
@@ -0,0 +1,91 @@
...
+      if (upsertIntentRequest.dryRun) {
+        val dryRunResponse = keelObjectMapper.readValue<List<UpsertIntentDryRunResponse>>(response.body.`in`(), object : TypeReference<List<UpsertIntentDryRunResponse>>(){})
+        outputs.put("upsertIntentResponse", dryRunResponse)
+      } else {
+        val upsertResponse = keelObjectMapper.readValue<List<UpsertIntentResponse>>(response.body.`in`(), object : TypeReference<List<UpsertIntentResponse>>(){})
+        outputs.put("upsertIntentResponse", upsertResponse)
+      }
+    } catch (e: Exception) {
+        throw RuntimeException("Error processing upsert intent response from keel", e)
+    }
```
robfletcher: Is there a better exception type? Could we just not catch?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/CreatePipelineTemplateTask.java
```diff
@@ -44,7 +44,7 @@
   @Override
   public TaskResult execute(Stage stage) {
     if (front50Service == null) {
-      throw new UnsupportedOperationException("Front50 is not enabled, no way to fetch pager duty. Fix this by setting front50.enabled: true");
+      throw new UnsupportedOperationException("Front50 is not enabled, no way to save pipeline templates. Fix this by setting front50.enabled: true");
```
robfletcher: Ha. Nice bit of copy 'n paste!

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
emjburns: The value _should_ be either `true` or `false`. `.toBoolean()` returns true if the value is equal to "true", ignoring case, and false otherwise. Since context is a map of `<String,Object>` I think we need to do `.toString()` first. 

I can simplify to `dryRun = stage.context["dryRun"].toString().toBoolean()` though.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
emjburns: An error here indicates that we can't understand the response from keel. It might not be a big deal, but for now I want to pass it along I think? I'll throw a better exception though.

emjburns: Actually maybe I"ll just log it for now. Doesn't seem worth stopping everything for.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/pipeline/DeleteIntentStage.kt
emjburns: I copied the style from https://github.com/spinnaker/orca/blob/master/orca-dry-run/src/main/kotlin/com/netflix/spinnaker/orca/dryrun/DryRunStage.kt#L39

I'm gonna ask you to explain this to me offline

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
robfletcher: Is it a _string_ that's `"true"` or `"false"` or is it actually a boolean?

Knowing us it can probably be either!

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
robfletcher: Yeah, I'd log & re-throw

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/UpsertIntentTask.kt
emjburns: Should be an actual boolean (it's typed in keel). Gonna keep it as a boolean in the context to try and help with the untyped garbage.

PR #1892, opened 2018-01-17
fix(spel): Evaluate expressions in synthesized stages.

jtk54:
Expressions were not being evaluated in dynamic stages/tasks added (e.g. expressions in synthesized regional deploys in a deploy stage).


jtk54:
Merging if no complaints @robfletcher.

PR #1893, opened 2018-01-18
fix(core): WaitForNewUpInstancesLaunchTask should support stage-level timeout override

ajordens:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/WaitForNewUpInstancesLaunchTask.groovy
```diff
@@ -30,7 +31,7 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
 @Component
-class WaitForNewUpInstancesLaunchTask implements RetryableTask {
+class WaitForNewUpInstancesLaunchTask implements RetryableTask, OverridableTimeoutRetryableTask {
```
emjburns: It just needs to be `OverridableTimeoutRetryableTask`, that interface just extends `RetryableTask`.

ajordens: 👍

PR #1894, opened 2018-01-18
chore(core): Consistency around `Millis` suffix on `task.*` values

ajordens:
We have a few situatations where task timeouts are overridable:

- tasks.waitForClusterTimeout (value is expected to be milliseconds)
- tasks.waitForUpsertedImageTagsTimeout (value is expected to be seconds)
- tasks.disableClusterMinTimeMillis (value is expected to be milliseconds)

The first two were added in the last 1-2 weeks, the latter was merged in
late November.

My preference would be to align around the suffix for these sort of
properties, even if it might mean a breaking change right now.


ajordens:
FYI @amanya @gardleopard @imosquera @yue9944882 

Do any of the above have an issue with aligning the naming of these timeout overrides? 

My preference would be to align around `Ms` but could do `Millis` given that we have some slightly older prior art.

I'm not aware that there is any usage or documentation of these overrides so I haven't gone down the path of maintaining backwards compatibility quite yet.

Happy to hear opinions.

yue9944882:
+1

robfletcher:
Very much in favor of this. I’m going to add something to contribution guidelines as well.

PR #1897, opened 2018-01-19
fix(queue): Move `RedisActiveExecutionsMonitor` off @Scheduler thread

ajordens:
This PR also fixes the guage creation such that metrics on total number
of active pipelines and orchestrations are correctly reported.

### DIFF
--- a/orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisActiveExecutionsMonitor.kt
+++ b/orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisActiveExecutionsMonitor.kt
@@ -26,12 +26,15 @@ import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionNotFoundExceptio
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.ApplicationListener
-import org.springframework.scheduling.annotation.Scheduled
 import org.springframework.stereotype.Component
 import redis.clients.jedis.Jedis
 import redis.clients.util.Pool
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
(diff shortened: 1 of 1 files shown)


orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisActiveExecutionsMonitor.kt
```diff
@@ -59,15 +64,41 @@ class RedisActiveExecutionsMonitor(
...
+  )
+
+  private val activeOrchestrationCounter = registry.gauge(
+    registry.createId("executions.active").withTag("executionType", Execution.ExecutionType.ORCHESTRATION.toString()),
+    AtomicInteger(0)
+  )
+
+  init {
+    executor.scheduleWithFixedDelay(
+      { refreshGauges() },
```
robfletcher: What happens if this call throws an exception?

ajordens: Good call ... I've now wrapped in a try / catch to prevent the exception from (a) being lost and (b) preventing reschedule.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisActiveExecutionsMonitor.kt
```diff
@@ -45,12 +48,14 @@ import java.util.concurrent.atomic.AtomicLong
...
  */
-
+@Component
 class RedisActiveExecutionsMonitor(
   private val executionRepository: ExecutionRepository,
   @Qualifier("jedisPool") private val pool: Pool<Jedis>,
   private val objectMapper: ObjectMapper,
-  private val registry: Registry
+  private val registry: Registry,
+  @Value("\${queue.monitor.activeExecutions.refresh.frequency.ms:60000}") refreshFrequencyMs: Long,
```
ajordens: @robzienert Changed the config key name here ... I cannot imagine anyone is actually using it so figured it was ok (s/register/refresh)

robzienert: 💯

PR #1901, opened 2018-01-20
fix(expressions): restrict spel fx deployedServerGroups to valid deploy stages

jeyrschabu:
- only match to valid deploy stages


tomaslin:
LGTM

PR #1902, opened 2018-01-20
(DO NOT MERGE) feat(core): Adding Redis Sentinel support

robzienert:
Sentinel support requires client-side modifications. This PR is the start of getting it in place. I think once it works here, I'll work on pulling out all of our Redis configuration code into kork, since it's all mostly the same.

⚠️ Still a WIP. I haven't gotten sentinel working locally yet, so I can't test if this is right yet.

Side note: IntelliJ is having a fit about kotlin 1.2, so I temp-switched to `arrayOf`: Not permanent.

cc @emptywee / https://github.com/spinnaker/spinnaker/issues/2250


emptywee:
I can test this as soon as an image is built and deployed to quay.io, but would like to have a configuration example posted here so I can adjust my orca-local.yml accordingly.

robzienert:
@emptywee Closing this. I've got work in-flight to standardize redis connections, at which point someone who runs a Redis Cluster / Redis Sentinel topology should be able to add support for this relatively easily across the services.

Since we don't run Sentinel, I don't feel comfortable contributing code we won't be able to vet in a real environment.

PR #1903, opened 2018-01-22
refactor(persistence): java rewrite

asher:
- implements RedisClientDelegate for ExecutionRepository
- prep work for Dynomite persistence and a transactionless schema
- removes ability to write stage context/output to global execution context


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/AbstractRedisExecutionRepository.java
```diff
@@ -0,0 +1,345 @@
...
+import org.springframework.stereotype.Component;
+import redis.clients.jedis.JedisCommands;
+import rx.Scheduler;
+
+import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.PIPELINE;
+import static java.lang.System.currentTimeMillis;
+import static java.text.MessageFormat.format;
+
+@Slf4j
+abstract class AbstractRedisExecutionRepository implements ExecutionRepository {
```
asher: This class provides RedisDelegate based implementations of all ExecutionRepository methods that don't currently use redis transactions. JedisExecutionRepository adds on methods that use transactions and multi-key pipeline operations. Because Dynomite does not support redis transactions and requires that pipelined commands share a hash key, the upcoming DynomiteExecutionRepository will also refactor the underlying schema in order to guarantee correctness without the use of transactions.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
```diff
@@ -0,0 +1,649 @@
...
+import static java.text.MessageFormat.format;
+import static java.util.Collections.*;
+import static net.logstash.logback.argument.StructuredArguments.value;
+import static redis.clients.jedis.BinaryClient.LIST_POSITION.AFTER;
+import static redis.clients.jedis.BinaryClient.LIST_POSITION.BEFORE;
+
+@Component
+@Slf4j
+public class JedisExecutionRepository extends AbstractRedisExecutionRepository {
+
```
asher: `JedisExecutionRepository` and `AbstractRedisExecutionRepository` are the most important classes to review in this lengthy PR. All methods in `JedisExecutionRepository` will be refactored away from transactions and multi-key pipelines in an upcoming `DynomiteExecutionRepository`.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/AbstractRedisExecutionRepository.java
```diff
@@ -0,0 +1,343 @@
...
+      if (StringUtils.isNotEmpty(user)) {
+        data.put("canceledBy", user);
+      }
+      if (StringUtils.isNotEmpty(reason)) {
+        data.put("cancellationReason", reason);
+      }
+      ExecutionStatus currentStatus = ExecutionStatus.valueOf(c.hget(key, "status"));
+      if (currentStatus == ExecutionStatus.NOT_STARTED) {
+        data.put("status", ExecutionStatus.CANCELED.name());
+      }
```
robzienert: @robfletcher I'd love to get a summary of why we don't always set the status to `CANCELED` here.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/AbstractRedisExecutionRepository.java
```diff
@@ -0,0 +1,343 @@
...
+    String key = fetchKey(id);
+
+    getRedisDelegateForId(key).withCommandsClient(c -> {
+      ExecutionStatus currentStatus = ExecutionStatus.valueOf(c.hget(key, "status"));
+      if (currentStatus != ExecutionStatus.RUNNING) {
+        throw new IllegalStateException(format(
+          "Unable to pause pipeline that is not RUNNING (executionId: %s, currentStatus: %s)",
+          id,
+          currentStatus.toString()
+        ));
```
robzienert: In the spirit of trying to make more specific exception classes, any feels on a new exception class? `UnpausablePipelineException extends IllegalStateException` or similar?

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/AbstractRedisExecutionRepository.java
```diff
@@ -0,0 +1,343 @@
...
+
+      PausedDetails pausedDetails = new PausedDetails();
+      pausedDetails.setPausedBy(user);
+      pausedDetails.setPauseTime(currentTimeMillis());
+
+      HashMap<String, String> data = new HashMap<>();
+      try {
+        data.put("paused", mapper.writeValueAsString(pausedDetails));
+      } catch (JsonProcessingException e) {
+        throw new RuntimeException("Failed converting pausedDetails to json", e);
```
robzienert: Would definitely like to steer clear of direct usage of `RuntimeException`, too.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/AbstractRedisExecutionRepository.java
```diff
@@ -0,0 +1,343 @@
...
+
+    if (delegate[0] == null && previousRedisClientDelegate.isPresent()) {
+      previousRedisClientDelegate.get().withCommandsClient(c -> {
+        if (c.exists(id)) {
+          delegate[0] = previousRedisClientDelegate.get();
+        }
+      });
+    }
+
+    return (delegate[0] == null) ? redisClientDelegate : delegate[0];
```
robzienert: I would consider changing the `delegate` from a `final` and skipping the array usage. I'm not sure it's buying us anything since the value of `delegate` is always only one, and that value isn't final.

asher: done, will be addressed in the next push.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/AbstractRedisExecutionRepository.java
asher: Status is set to `CANCELLED` (so long as `executionStatus != TERMINAL`) via an `afterExecution` listener in `ExecutionPropagationListener`. I think this makes sense, a task could be executing on an orca instance other than the one handling the cancel request. That task will still complete, so seems correct to set pipeline status to `CANCELED` at that time.

robfletcher: Yeah, the idea is you don't want to interrupt a graceful stop of the pipeline.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/RedisConfiguration.groovy
```diff
@@ -66,6 +66,20 @@ class RedisConfiguration {
...
+  @Primary
+  RedisClientDelegate redisClientDelegate(@Qualifier("jedisPool") Pool<Jedis> jedisPool) {
+    return new JedisClientDelegate(jedisPool)
+  }
+
+  @Bean(name="previousRedisClientDelegate")
+  RedisClientDelegate previousRedisClientDelegate(@Qualifier("jedisPoolPrevious") JedisPool jedisPoolPrevious) {
+    if (jedisPoolPrevious == null) {
+      return null
+    }
```
robfletcher: I think this might be better done using `@ConditionalOnBean`. I'm not sure what the implications are of returning `null` from a `@Bean` method.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
```diff
@@ -0,0 +1,649 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.kork.jedis.RedisClientDelegate;
+import com.netflix.spinnaker.orca.ExecutionStatus;
+import com.netflix.spinnaker.orca.pipeline.model.Execution;
+import com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner;
+import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionNotFoundException;
+import groovy.util.logging.Slf4j;
```
robfletcher: This shouldn't be used in Java (and won't work).

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
```diff
@@ -0,0 +1,649 @@
...
+    String key = format("%s:%s", type.toString(), stage.getExecution().getId());
+    getRedisDelegateForId(key).withTransaction(tx -> {
+      storeStageInternal(tx, type, stage);
+      tx.exec();
+    });
+  }
+
+  @Override
+  public void removeStage(@Nonnull Execution execution, @Nonnull String stageId) {
+    String key = format("%s:%s", execution.getType().toString(), execution.getId());
```
robfletcher: You shouldn't need to specify `.toString()` in these format expressions. The `%s` in the expression does that anyway.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
```diff
@@ -0,0 +1,649 @@
...
+    getRedisDelegateForId(key).withCommandsClient(c -> {
+      List<String> stageIds = new ArrayList<>(Arrays.asList(c.hget(key, "stageIndex").split(",")));
+      stageIds.remove(stageId);
+      boolean hasDiscreteIndex = c.exists(indexKey);
+      List<String> keys = c.hkeys(key).stream()
+        .filter(k -> k.startsWith("stage." + stageId))
+        .collect(Collectors.toList());
+
+      getRedisDelegateForId(key).withTransaction(tx -> {
+        tx.hset(key, "stageIndex", stageIds.stream().map(Object::toString).collect(Collectors.joining(",")));
```
robfletcher: Why do we need `.map(Object::toString)`? They _are_ strings.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
```diff
@@ -0,0 +1,649 @@
...
+    execution.getStages().forEach(s -> map.putAll(serializeStage(s)));
+
+    if (execution.getType() == PIPELINE) {
+      try {
+        map.put("name", execution.getName());
+        map.put("pipelineConfigId", execution.getPipelineConfigId());
+        map.put("notifications", mapper.writeValueAsString(execution.getNotifications()));
+        map.put("initialConfig", mapper.writeValueAsString(execution.getInitialConfig()));
+      } catch (JsonProcessingException e) {
+        throw new RuntimeException("Failed serializing execution", e);
```
robfletcher: Can we use a better exception type? I know this is probably the fault of whoever wrote the old Groovy code…

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
```diff
@@ -0,0 +1,649 @@
...
+    }
+
+    tx.hdel(key, "config");
+    tx.hmset(key, filterValues(map, Objects::nonNull));
+  }
+
+  private void storeStageInternal(Transaction tx, ExecutionType type, Stage stage) {
+    String key = format("%s:%s", type.toString(), stage.getExecution().getId());
+
+    HashMap<String, String> serializedStage = serializeStage(stage);
```
robfletcher: s/HashMap/Map/

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
asher: Fixed

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
asher: 👍

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
asher: Derp, replaced with `stageIds.stream().collect(Collectors.joining(",")))`

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
asher: Will go over all of the exceptions thrown in this PR and implement new types when there isn't already a good specific fit.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/jedis/JedisExecutionRepository.java
asher: Fixed here and throughout

orca-core/src/main/groovy/com/netflix/spinnaker/orca/config/RedisConfiguration.groovy
asher: done, similarly refactoring `jedisPoolPrevious()` in the same class

PR #1910, opened 2018-01-25
fix(front50): save with existing pipeline index

sbwsg:
Pipelines that had been manually sorted in deck would become out of order when roer was used to save over them.  This happened because any existing index property on the saved pipeline was wiped out by the incoming roer pipeline.

This PR checks whether a new pipeline matches application and ID with an existing pipeline and then, if so, copies the existing pipeline's index into the new pipeline so that ordering is maintained.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.orca.TaskResult;
 import com.netflix.spinnaker.orca.front50.Front50Service;
 import com.netflix.spinnaker.orca.front50.PipelineModelMutator;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -75,6 +76,13 @@ public class SavePipelineTask implements RetryableTask {
 
     Map<String, Object> pipeline = (Map<String, Object>) stage.decodeBase64("/pipeline", Map.class);
 
+    if (!pipeline.containsKey("index")) {
+      Map<String, Object> existingPipeline = fetchExistingPipeline(pipeline);
+      if (existingPipeline != null) {
+        pipeline.put("index", existingPipeline.get("index"));
+      }
+    }
+
     pipelineModelMutators.stream().filter(m -> m.supports(pipeline)).forEach(m -> m.mutate(pipeline));
 
     Response response = front50Service.savePipeline(pipeline);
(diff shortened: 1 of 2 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
```diff
@@ -99,4 +106,18 @@ public long getBackoffPeriod() {
   public long getTimeout() {
     return TimeUnit.SECONDS.toMillis(30);
   }
+
+  private Map<String, Object> fetchExistingPipeline(Map<String, Object> newPipeline) {
+    String applicationName = (String) newPipeline.get("application");
+    String newPipelineID = (String) newPipeline.get("id");
+    if (newPipelineID != null && !newPipelineID.equals("")) {
```
jtk54: Could use `isNotEmpty` etc from [apache StringUtils](https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html).

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
```diff
@@ -99,4 +106,18 @@ public long getBackoffPeriod() {
...
+      List<Map<String, Object>> applicationPipelines = front50Service.getPipelines(applicationName);
+      for (Map<String, Object> existingPipeline : applicationPipelines) {
```
jtk54: Could use a java 8 `stream()` with `findFirst()` over a manual loop here.

lwander: The stream api is a little daunting at first:

```java
.stream()
.filter(p -> p.containsKey("id"))
.filter(p -> p.get("id").equals(newPipelineID))
.findFirst().orElse(null)
```

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTaskSpec.groovy
```diff
@@ -65,4 +65,67 @@ class SavePipelineTaskSpec extends Specification {
     ])
   }
 
+  def "should copy existing pipeline index to new pipeline"() {
+    Integer testIndex = 14
```
danielpeach: From my non-existent knowledge of Spock conventions, I would put this (and `existingPipeline` in `when:`) in the `given:` block. Same for the next spec.

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTaskSpec.groovy
```diff
@@ -65,4 +65,67 @@ class SavePipelineTaskSpec extends Specification {
...
+    def stage = new Stage(Execution.newPipeline("orca"), "whatever", [
+      pipeline: Base64.encoder.encodeToString(objectMapper.writeValueAsString(pipeline).bytes)
+    ])
+
+    when:
+    Map<String, Object> existingPipeline = [
+      index: testIndex,
+      id: 'existing-pipeline-id',
+    ]
+    Integer index = 0
```
danielpeach: This spec would read easier to me if you didn't initialize this.

PR #1912, opened 2018-01-25
fix(persistence): remove javafx dep for vanilla openjdk

asher:


cfieber:
wat

also LGTM

also I hope that wasn't me and sorry if it was

PR #1914, opened 2018-01-28
feat(provider/kubernetes): v2 Add find artifacts from resource stage

wjoel:
Can be used for https://github.com/spinnaker/spinnaker/issues/2267


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
```diff
@@ -0,0 +1,103 @@
...
+  @Autowired
+  OortService oortService;
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    StageData stageData = stage.mapTo(StageData.class);
+    String account = getCredentials(stage);
+    Set<String> expectedArtifactIds = new HashSet<>(stageData.expectedArtifactIds);
+    List<ExpectedArtifact> expectedArtifacts = artifactResolver.getExpectedArtifacts(stage);
```
lwander: I don't think you need to be looking up every expected artifact here - from my reading you're looking all the way back into the stage history to find every expected artifact declared in upstream stages.

I would think that all you need are the expected artifacts declared in this stage.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
```diff
@@ -0,0 +1,103 @@
...
+    Map<String, Object> outputs = new HashMap<>();
+
+    List<ExpectedArtifact> unresolvedExpectedArtifacts = expectedArtifacts.stream()
+      .filter(a -> expectedArtifactIds.contains(a.getId()))
+      .filter(a -> a.getBoundArtifact() == null
```
lwander: I don't think this check is necessary - the "boundArtifact" should never be null after a stage has executed.

This means that any expected artifacts you found in upstream stages should already be bound, and the ones declared in this stage need to be bound.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
```diff
@@ -0,0 +1,103 @@
...
+      stageData.manifestName);
+
+    if (manifest != null) {
+      List<Artifact> artifactsFromManifest = manifest.getArtifacts().stream()
+        .filter(a -> unresolvedExpectedArtifactNames.contains(a.getName()))
+        .collect(Collectors.toList());
+      List<ExpectedArtifact> resolvedExpectedArtifacts = new ArrayList<>();
+
+      for (ExpectedArtifact expectedArtifact : unresolvedExpectedArtifacts) {
+        for (Artifact newArtifact : artifactsFromManifest) {
```
lwander: I think we have logic for this already: https://github.com/spinnaker/orca/pull/1914/files#diff-ab3dd9a63d59c02eb0cf7191ce664c72R124

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
wjoel: Maybe, I'm not sure exactly when the binding happen, but I need to get a `List<ExpectedArtifact>` to check if they are bound, and if so, if they are bound to the default artifact. This was the only way I could find.

https://github.com/spinnaker/orca/pull/1914/files#diff-ab3dd9a63d59c02eb0cf7191ce664c72R54 gets me a `List<ExpectedArtifact>` but it takes them from `stage.getContext()).getAll("resolvedExpectedArtifacts")`, while I also want unresolved ones (as in the Docker image from Configuration without a default).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
wjoel: This is likely to often be the first stage after Configuration. Isn't it possible for "boundArtifact" to be null, then?

For example, a pipeline triggered by a GitHub webhook after a manifest has been updated, but where the Docker image being an expected artifact defined in Configuration has no default (and doesn't use the prior execution either)?

That's the use-case I have in mind.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
wjoel: Done.

wjoel:
Perhaps I've been overthinking this, and it can be made simpler, but here are my concerns:

1. "Cold start" where the manifest uses a Docker image name without a tag, but we want to start from a specific tag, so we specify a default artifact for our pipeline. The manifest is deployed for the first time (using the default to replace the image name in the manifest), and creates a deployment. During this first deployment our FindArtifactsFromResource stage fails, but that can be configured to not fail the pipeline.
However, when we later update the manifest (say, change a sidecar), we want the default artifact which gets set when our pipeline is triggered (as I understand it) to be overwritten by the FindArtifactsFromResource stage, which now succeeds.

2. Even if we allow "use prior execution", if the deployment has been updated from outside of Spinnaker or perhaps from another pipeline in Spinnaker, the image found by FindArtifactsFromResource should (optionally, perhaps) take precedence and overwrite the currently bound image.

lwander:
> During this first deployment our FindArtifactsFromResource stage fails, but that can be configured to not fail the pipeline.

Because the manifest doesn't exist? So we rely on the default artifact specified in the FindArtifactsFromResource, which shouldn't fail the pipeline, right?

> However, when we later update the manifest (say, change a sidecar), we want the default artifact which gets set when our pipeline is triggered (as I understand it) to be overwritten by the FindArtifactsFromResource stage, which now succeeds.

That's exactly how the default artifact works in the expected artifact defined in this stage. It'll succeed in binding whatever it found on the manifest, and ignore the default.

> Even if we allow "use prior execution", if the deployment has been updated from outside of Spinnaker or perhaps from another pipeline in Spinnaker, the image found by FindArtifactsFromResource should (optionally, perhaps) take precedence and overwrite the currently bound image.

That's exactly how it already works, it'll use what it binds in the stage, and not default to using a prior execution.

---

I think I just realized the source of our confusion - we don't want to define the "expected artifacts" at the start of the pipeline - we want to define them in this stage. That means picking the kind, name, reference, etc... rather than from a prepopulated list. This way the artifact isn't bound _until_ we reach this stage, where we just rely on how expected artifacts are normally evaluated at the start of a pipeline.

wjoel:
Thank you for the questions and clarifications, @lwander. I had misunderstood some of the concepts around artifacts vs expected artifacts, believing that all artifacts that could be consumed in any stage had to be listed in "expected artifacts", perhaps because I thought "required artifacts" (in the Deploy (manifest) stage) were the only artifacts that could be consumed by that stage.

I went back and re-read the "Spinnaker Artifact Handling" design doc and see that this is not the case, because:

> Currently, every task in Spinnaker outputs “stageOutputs” and “globalOutputs”. The list of artifacts produced by a task will be exactly represented in the list of “stageOutputs”. This allows  other stages to lookup artifacts by stage name. Updates to the list artifacts of “globalOutputs” will override existing artifacts when “type” and “name” of artifacts being added match existing artifacts"

So, if I'm getting this right, I only need to add the artifacts found in the manifest to `stageOutputs` and `globalOutputs` - my concern about default artifacts that this stage should override is already taken care of by existing functionality, as you mentioned.

Any stage can produce any artifacts it wants to, and stages downstream can use any artifacts produced by previous stages, irrelevant of what is defined in the pipeline configuration. The "expected artifacts" is simply a mechanism that stages can choose to use to say "I need the artifacts in this list, or I won't work correctly, so don't even start this stage if they are not present". Is that correct?

If so, I'll make the changes and verify that they work, but it could be a few days before I get to it.

lwander:
> So, if I'm getting this right, I only need to add the artifacts found in the manifest to stageOutputs and globalOutputs - my concern about default artifacts that this stage should override is already taken care of by existing functionality, as you mentioned.

That's right

> Any stage can produce any artifacts it wants to, and stages downstream can use any artifacts produced by previous stages, irrelevant of what is defined in the pipeline configuration. The "expected artifacts" is simply a mechanism that stages can choose to use to say "I need the artifacts in this list, or I won't work correctly, so don't even start this stage if they are not present". Is that correct?

Exactly!

Thanks for doing this :)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
wjoel: As discussed on Slack, this (updating the image from a Docker trigger) should just be a separate pipeline.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
```diff
@@ -0,0 +1,71 @@
...
+    String account = getCredentials(stage);
+    Map<String, Object> outputs = new HashMap<>();
+
+    Manifest manifest = oortService.getManifest(account, stageData.location,
+      stageData.manifestName);
+
+    if (manifest != null) {
+      outputs.put("artifacts", manifest.getArtifacts());
+    } else {
+      log.warn("Manifest " + stageData.manifestName + " not found in "
```
lwander: I think we should fail here, or at least give the option to not fail. If a user doesn't make their expected artifacts explicit, they might (without realizing it) fail to bind any artifacts in a downstream deploy manifest stage because they referenced the wrong manifest here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
```diff
@@ -0,0 +1,71 @@
...
+  OortService oortService;
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    StageData stageData = stage.mapTo(StageData.class);
+    String account = getCredentials(stage);
+    Map<String, Object> outputs = new HashMap<>();
+
+    Manifest manifest = oortService.getManifest(account, stageData.location,
```
lwander: Could you add a retry like shown here? https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java#L88

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
wjoel: Done, throwing an `IllegalArgumentException` instead.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
wjoel: Done.

PR #1919, opened 2018-01-30
fix(queue): synthetic stages aren't restartable

asher:
If RestartStage is called on a synthetic stage, actually restart its parent.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt
@@ -40,11 +40,14 @@ class RestartStageHandler(
 
   override fun handle(message: RestartStage) {
     message.withStage { stage ->
-      if (stage.status.isComplete) {
-        stage.addRestartDetails(message.user)
-        stage.reset()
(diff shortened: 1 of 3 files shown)


orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.java
```diff
@@ -294,6 +294,15 @@ public Stage stageByRef(String refId) {
...
   }
 
+  @Nonnull
+  public Stage parentStageById(String stageId) {
+    Stage stage = stageById(stageId);
+    while (stage.getParentStageId() != null) {
+      stage = stageById(stage.getParentStageId());
+    }
+    return stage;
+  }
```
robfletcher: Couldn't this be a method on `Stage`? Then you don't need the parameter.

ajordens:
Other than @robfletcher's comment about it being an extension of `Stage` ... this seems reasonable to me.

Nice work.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Execution.java
asher: And I just saw that Stage.getTopLevelStage() already does this, somehow missed that yesterday.

PR #1920, opened 2018-01-30
fix(core): ancestors() of a synthetic stage should include `STAGE_BEFORE` siblings

ajordens:
This addresses an issue where a resize operation does not correctly
restore the aws scaling processes that were manipulated by a sibling
`STAGE_BEFORE` stage.

This was previously not an issue due to the global context.

### DIFF
--- a/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
+++ b/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
@@ -357,13 +357,30 @@ public class Stage implements Serializable {
         .addAll(previousStages.stream().flatMap(it -> it.ancestorsOnly().stream()).collect(toList()))
         .build();
     } else if (parentStageId != null) {
-      return execution.getStages().stream().filter(it -> it.id.equals(parentStageId)).findFirst()
-        .<List<Stage>>map(parent -> ImmutableList
-          .<Stage>builder()
-          .add(parent)
-          .addAll(parent.ancestorsOnly())
-          .build())
-        .orElse(emptyList());
+      List<Stage> ancestors = new ArrayList<>();
+      if (getSyntheticStageOwner() == SyntheticStageOwner.STAGE_AFTER) {
+        ancestors.addAll(
+          execution
+            .getStages()
+            .stream()
+            .filter(it ->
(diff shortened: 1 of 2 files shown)


ajordens:
Current behavior:

1) Disable the Launch / Terminate scaling processes on an ASG (or just disable the entire ASG)
2) Resize the ASG
- observe the Launch and Terminate are re-enabled (as expected!)
- observe that the resize happens (as expected)
- observe that Launch and Terminate are left enabled (**not expected!**)

This PR results in Launch and Terminate correctly being suspended following a successful resize.

orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -357,13 +357,30 @@ public Task taskById(String taskId) {
...
     } else if (parentStageId != null) {
-      return execution.getStages().stream().filter(it -> it.id.equals(parentStageId)).findFirst()
-        .<List<Stage>>map(parent -> ImmutableList
-          .<Stage>builder()
-          .add(parent)
-          .addAll(parent.ancestorsOnly())
-          .build())
-        .orElse(emptyList());
+      List<Stage> ancestors = new ArrayList<>();
+      if (getSyntheticStageOwner() == SyntheticStageOwner.STAGE_AFTER) {
```
ajordens: Not sure if there was a good reason to not consider sibling stages when looking for ancestors.

The particular issue in question arises because a `STAGE_BEFORE` produces an output that a `STAGE_AFTER` expects to read.

robfletcher:
This will have the side effect of making expressions chain back to BEFORE stage outputs. Not sure if that's desirable.

ajordens:
@robfletcher I believe that would have been the case previously with the global context, no?

robfletcher:
I guess so. It makes me wonder if `ancestors` should then include the STAGE_AFTER outputs of prior stages?

ajordens:
I think we already are?

robfletcher:
Hmm, yeah. I think you're right. I need to clarify this in my head. It gets really complex when you consider precedence between nested synthetic stages and their parents when queried by things downstream.

ajordens:
Talked with @robfletcher ... going to merge this down once Travis is happy.

PR #1921, opened 2018-01-30
chore(restarts): remove redundant logic from stage restart methods

robfletcher:
Resetting start/end time, status, and tasks is handled from the place `prepareStageForRestart` is called so there's no need for implementations of that method do such things.


robzienert:
Onlookers: Here's where the things are done now: https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt#L63

PR #1924, opened 2018-01-31
More useful dry run

robfletcher:
This allows pipelines with a `dry-run` trigger type to generate dummy data from certain stage types. The plan is to add a "dry run" checkbox to the manual trigger dialog to allow users to test out pipelines verifying expressions, optional stages, functionality based on parameters, etc.

### DIFF
--- a/orca-dry-run/src/main/kotlin/com/netflix/spinnaker/orca/dryrun/stub/BakeOutputStub.kt
+++ b/orca-dry-run/src/main/kotlin/com/netflix/spinnaker/orca/dryrun/stub/BakeOutputStub.kt
@@ -47,8 +47,4 @@ class BakeOutputStub : OutputStub {
         }
       )
     }
-
-  @Suppress("UNCHECKED_CAST")
-  private val Stage.regions: List<String>
-    get() = context["regions"] as List<String>? ?: emptyList()
 }
--- a/orca-dry-run/src/main/kotlin/com/netflix/spinnaker/orca/dryrun/stub/BakeOutputStub.kt
+++ b/orca-dry-run/src/main/kotlin/com/netflix/spinnaker/orca/dryrun/stub/BakeOutputStub.kt
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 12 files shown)


orca-dry-run/src/main/kotlin/com/netflix/spinnaker/orca/dryrun/stub/BakeOutputStub.kt
```diff
@@ -0,0 +1,50 @@
...
+@Component
+class BakeOutputStub : OutputStub {
+
+  override fun supports(stageType: String) = stageType == "bake"
+
+  override fun outputs(stage: Stage) =
+    if (stage.parent?.type == "bake") {
+      emptyMap()
+    } else {
+      mapOf(
```
robfletcher: We'd need to add a Titus variant here.

robzienert: Were you thinking of putting the cloudprovider specific logic in this single class or break them up into specific classes (my preference)?

orca-dry-run/src/main/kotlin/com/netflix/spinnaker/orca/dryrun/stub/FindImageFromTagsOutputStub.kt
```diff
@@ -0,0 +1,44 @@
...
+      "deploymentDetails" to stage.regions.map { region ->
+        Pair(randomHex(8), randomNumeric(4)).let { (id, build) ->
+          mapOf(
+            "imageName" to "${stage.execution.application}-${randomNumeric(1)}.${randomNumeric(4)}.0-h${build}.${randomHex(7)}-x86_64-${timestamp()}-trusty-hvm-sriov-ebs",
+            "imageId" to "ami-$id",
+            "jenkins" to mapOf(
+              "number" to build,
+              "host" to "http://jenkins/",
+              "name" to "ami-$id"
+            ),
```
robzienert: Why does this stub include jenkins things?

orca-dry-run/src/main/kotlin/com/netflix/spinnaker/orca/dryrun/stub/random.kt
```diff
@@ -0,0 +1,40 @@
...
+  .absoluteValue
+  .toString(16)
+  .substring(0 until minOf(length, 16))
+
+internal fun randomNumeric(length: Int) = UUID
+  .randomUUID()
+  .leastSignificantBits
+  .absoluteValue
+  .toString(10)
+  .substring(0 until minOf(length, 16))
```
robzienert: Any significance to the `16` max length?

PR #1925, opened 2018-01-31
fix(queue): Execution w/o initial stages should go TERMINAL immediately

ajordens:


ajordens:
This will immediately fail the pipeline vs leaving it in a "fake" `RUNNING` state that prevents future executions of the pipeline from starting (when concurrent execution is disabled).

asher:
LGTM

PR #1926, opened 2018-01-31
chore(queue): Increase work queue poll interval

robzienert:
Orca doesn't do nearly enough work to warrant such a low default, even at high load.


ajordens:
mega  🌮 👍

robfletcher:
Will this have a noticeable cooling effect on Amazon's datacenters?

robzienert:
I'm fairly certain we can call global warming a solved problem now.

PR #1932, opened 2018-02-02
fix(artifacts): fix bind behavior

lwander:
Also added some tests to protect against future refactors

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
@@ -16,9 +16,7 @@
 
 package com.netflix.spinnaker.orca.pipeline.util;
 
-import java.util.*;
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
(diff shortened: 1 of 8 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -52,66 +63,76 @@ public ArtifactResolver(ObjectMapper objectMapper, ExecutionRepository execution
...
     if (isEmpty(id)) {
       return null;
     }
 
     List<ExpectedArtifact> expectedArtifacts;
     if (stage.getContext() instanceof StageContext) {
-      expectedArtifacts = ((StageContext) stage.getContext()).getAll("resolvedExpectedArtifacts");
+      expectedArtifacts = (List<ExpectedArtifact>) Optional.ofNullable((List) ((StageContext) stage.getContext()).getAll("resolvedExpectedArtifacts"))
+          .map(list -> list.stream()
+              .map(it -> objectMapper.convertValue(it, new TypeReference<List<ExpectedArtifact>>() {}))
```
robfletcher: Feels like we should be able to avoid this by declaring the property on the trigger correctly. Will take a proper look when I get to work.

lwander:
No worries @robfletcher - I could have prevented this had I put those tests in place earlier ;)

robfletcher:
Yeah, I think when you're dealing with Groovy code tests play 2 roles. They're not just verifying correct functionality but they're also doing type checking. This kind of issue is really hard to foresee when making changes like this.

PR #1934, opened 2018-02-02
fix(trigger): Remove redundant parsing of artifacts

robfletcher:


lwander:
Let me try this out locally to make sure there isn't any parsing weirdness that the tests aren't catching...

lwander:
Works as expected

PR #1937, opened 2018-02-02
fix(pipelines): Improve pipeline error reporting

ezimanyi:
When a pipeline fails validation, due to its being disabled, having
an invalid template, or being missing an artifact, log a failed execution
to the execution history instead of only generating an error in the
logs.

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -77,15 +77,45 @@ class OperationsController {
 
   @RequestMapping(value = "/orchestrate", method = RequestMethod.POST)
   Map<String, Object> orchestrate(@RequestBody Map pipeline, HttpServletResponse response) {
-    parsePipelineTrigger(executionRepository, buildService, pipeline)
-    Trigger trigger = objectMapper.convertValue(pipeline.trigger, Trigger)
-
-    boolean plan = pipeline.plan ?: false
-    for (PipelinePreprocessor preprocessor : (pipelinePreprocessors ?: [])) {
-      pipeline = preprocessor.process(pipeline)
+    if (pipeline.plan) {
+      parseAndValidatePipeline(pipeline)
+      log.info('Not starting pipeline (plan: true): {}', value("pipelineId", pipeline.id))
+      return pipeline
     }
 
-    pipeline.trigger = trigger
+    Throwable pipelineError = null
+    try {
+      parseAndValidatePipeline(pipeline)
+    } catch (Throwable t) {
+      pipelineError = t
+    }
+
+    def augmentedContext = [trigger: pipeline.trigger]
+    def processedPipeline = contextParameterProcessor.process(pipeline, augmentedContext, false)
+
+    if (pipelineError == null) {
+      startPipeline(processedPipeline)
+    } else {
+      markPipelineFailed(processedPipeline, pipelineError)
+      throw pipelineError
+    }
+  }
+
+  private void parseAndValidatePipeline(Map pipeline) {
+    try {
+      parsePipelineTrigger(executionRepository, buildService, pipeline)
+
+      for (PipelinePreprocessor preprocessor : (pipelinePreprocessors ?: [])) {
+        pipeline = preprocessor.process(pipeline)
+      }
+    } catch (Throwable t) {
+      throw t
+    } finally {
+      // Even if an exception is thrown above, we need to convert pipeline.trigger to
+      // a Trigger object, as the code to log the error to the execution history is
+      // expecting this.
+      pipeline.trigger = objectMapper.convertValue(pipeline.trigger, Trigger)
+    }
 
     def json = objectMapper.writeValueAsString(pipeline)
     log.info('received pipeline {}:{}', value("pipelineId", pipeline.id), json)
@@ -99,22 +129,9 @@ class OperationsController {
       convertLinearToParallel(pipeline)
     }
 
-    if (plan) {
-      log.info('not starting pipeline (plan: true): {}', value("pipelineId", pipeline.id))
-      if (pipeline.errors != null) {
-        throw new ValidationException("Pipeline template is invalid", pipeline.errors as List<Map<String, Object>>)
-      }
-      return pipeline
-    }
-
-    def augmentedContext = [trigger: pipeline.trigger]
-    def processedPipeline = contextParameterProcessor.process(pipeline, augmentedContext, false)
-
     if (pipeline.errors != null) {
       throw new ValidationException("Pipeline template is invalid", pipeline.errors as List<Map<String, Object>>)
     }
-
-    startPipeline(processedPipeline)
   }
 
   private void parsePipelineTrigger(ExecutionRepository executionRepository, BuildService buildService, Map pipeline) {
@@ -247,6 +264,16 @@ class OperationsController {
     log.info('requested pipeline: {}', json)
 
     def pipeline = executionLauncher.start(PIPELINE, json)
+
+    [ref: "/pipelines/${pipeline.id}".toString()]
+  }
+
+  private Map<String, String> markPipelineFailed(Map config, Throwable t) {
+    injectPipelineOrigin(config)
+    def json = objectMapper.writeValueAsString(config)
+    log.info('requested pipeline: {}', json)
+
+    def pipeline = executionLauncher.logFailedExecution(PIPELINE, json, t)
 
     [ref: "/pipelines/${pipeline.id}".toString()]
   }
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -77,17 +77,19 @@ class OperationsController {
 
   @RequestMapping(value = "/orchestrate", method = RequestMethod.POST)
   Map<String, Object> orchestrate(@RequestBody Map pipeline, HttpServletResponse response) {
-    if (pipeline.plan) {
-      parseAndValidatePipeline(pipeline)
-      log.info('Not starting pipeline (plan: true): {}', value("pipelineId", pipeline.id))
-      return pipeline
-    }
-
-    Throwable pipelineError = null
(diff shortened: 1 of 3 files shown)


ezimanyi:
Here's the error handling pull request intended to address spinnaker/spinnaker#2255

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -77,15 +77,45 @@ class OperationsController {
 
   @RequestMapping(value = "/orchestrate", method = RequestMethod.POST)
   Map<String, Object> orchestrate(@RequestBody Map pipeline, HttpServletResponse response) {
-    parsePipelineTrigger(executionRepository, buildService, pipeline)
-    Trigger trigger = objectMapper.convertValue(pipeline.trigger, Trigger)
+    if (pipeline.plan ?: false) {
```
lwander: In groovy you don't need `false` here, `pipeline.plan` being null is "falsey"

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -77,15 +77,45 @@ class OperationsController {
...
+      parseAndValidatePipeline(pipeline)
+      log.info('not starting pipeline (plan: true): {}', value("pipelineId", pipeline.id))
```
lwander: I think we generally capitalize the start of these log message

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -77,15 +77,45 @@ class OperationsController {
...
+      return pipeline
+    }
 
-    boolean plan = pipeline.plan ?: false
-    for (PipelinePreprocessor preprocessor : (pipelinePreprocessors ?: [])) {
-      pipeline = preprocessor.process(pipeline)
+    Throwable pipelineError
```
lwander: I would initialize to `null`

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ezimanyi: Fixed.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ezimanyi: Fixed.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ezimanyi: Fixed.

ezimanyi:
Here's a screen shot of the executions view showing an execution that failed due to an unresolved artifact:
artifact-error

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -77,15 +77,45 @@ class OperationsController {
...
+  private void parseAndValidatePipeline(Map pipeline) {
+    try {
+      parsePipelineTrigger(executionRepository, buildService, pipeline)
 
-    pipeline.trigger = trigger
+      for (PipelinePreprocessor preprocessor : (pipelinePreprocessors ?: [])) {
+        pipeline = preprocessor.process(pipeline)
+      }
+    } catch (Throwable t) {
+      throw t
```
ajordens: this catch & throw the same thing seems unnecessary?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -251,6 +268,16 @@ class OperationsController {
     [ref: "/pipelines/${pipeline.id}".toString()]
   }
 
+  private Map<String, String> markPipelineFailed(Map config, Throwable t) {
+    injectPipelineOrigin(config)
+    def json = objectMapper.writeValueAsString(config)
+    log.info('requested pipeline: {}', json)
+
+    def pipeline = executionLauncher.logFailedExecution(PIPELINE, json, t)
```
ajordens: This reads slightly off ... `logFailedExecution` but it returns a new execution? This would imply that the method is doing something other than logging.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -77,15 +77,45 @@ class OperationsController {
 
   @RequestMapping(value = "/orchestrate", method = RequestMethod.POST)
   Map<String, Object> orchestrate(@RequestBody Map pipeline, HttpServletResponse response) {
-    parsePipelineTrigger(executionRepository, buildService, pipeline)
-    Trigger trigger = objectMapper.convertValue(pipeline.trigger, Trigger)
+    if (pipeline.plan) {
+      parseAndValidatePipeline(pipeline)
+      log.info('Not starting pipeline (plan: true): {}', value("pipelineId", pipeline.id))
+      return pipeline
```
ajordens: I realize we have `Map<String, Object>` here but the long expected result from this endpoint is either:
(a) an exception
or
(b) ["ref": "/.../{id}"]

It seems a little strange to echo the pipeline json back

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -77,15 +77,45 @@ class OperationsController {
...
+    }
+
+    Throwable pipelineError = null
```
ajordens: Out of curiosity, why are we dealing with `Throwable` instead of plain ol' `Exception`?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -77,15 +77,45 @@ class OperationsController {
...
+    try {
+      parseAndValidatePipeline(pipeline)
```
ajordens: Seems like the logic could be cleaned up here such that we only need to call `parseAndValidatePipeline(pipeline)` once and then take some follow-up action depending `pipeline.plan` ?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ezimanyi: Echoing back the pipeline when `pipeline.plan` is true was the behavior prior to this change; if we think it makes sense to change, I can do so, but didn't want to change that behavior as part of this change.  I agree that it seems a bit strange.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ezimanyi: Good point; no good reason.  I'll change this to `Exception`

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ajordens: Oops ... yes I just checked on that behavior and was coming back in here to remove this comment.

'tis a bit confusing but we can leave as is for now.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ezimanyi: Yeah, that's reasonable, I'll clean this up

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ezimanyi: This is really just so we can have a `finally` block that inflates the trigger JSON into a `Trigger` object regardless of whether `parsePipelineTrigger` succeeds or fails.  (As even if it fails, we need the trigger to exist so we can properly log the error in `markPipelineFailed`.)  I'm open to a better way of achieving this, though, if you have any suggestions.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ajordens: Perhaps easiest addressed by `s/logFailedExecution()/fail()` or something along those lines?

ezimanyi: Maybe this is something I can clarify by changing the name of `logFailedExecution`?  We are creating an execution (so that the error surfaces in the execution history), but rather than trying to start it we're just putting it in `TERMINAL` status.  Would it be more clear to call this `createFailedExecution` or something similar?

ajordens:
Couple minor things that are probably worth addressing pre-merge but otherwise 👍 -- good work!

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ezimanyi: Yup, I'll change the name to `fail()`, I agree that's clearer

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ajordens: You can still have the `finally` block ... I don't believe the behavior would be any different if you removed the `catch`?

ezimanyi: Right, I'd been mistaken in thinking that would silence all errors, but you're correct.

PR #1938, opened 2018-02-02
fix(triggers): derived properties of Jenkins triggers

robfletcher:
- buildInfo and properties are looked up from the Jenkins master not sent with the trigger
- fixed tests broken in IntelliJ when non-null annotation checks compiled in
- any `@Nonnull` list or map accepts null on the constructor and defaults the value


ezimanyi:
Verified that the tests that were failing earlier in IntelliJ (`OperationsControllerSpec.groovy`) now pass after this change.

PR #1941, opened 2018-02-03
fix(triggers): evaluate incoming triggers (& artifacts)

lwander:
In parsing the `Trigger` objects, we wind up unable to evaluate SPEL (which only works on maps lists & strings). To get around this, I defer parsing the trigger until after SPEL has been evaluated.

This should keep the behavior of https://github.com/spinnaker/orca/pull/1937 the same, since it requires the trigger to be parsed before the error message is reported (please confirm @ezimanyi).

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -16,7 +16,6 @@
 
 package com.netflix.spinnaker.orca.controllers
 
-import javax.servlet.http.HttpServletResponse
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.kork.web.exceptions.InvalidRequestException
 import com.netflix.spinnaker.kork.web.exceptions.ValidationException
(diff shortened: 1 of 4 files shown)


lwander:
Closes https://github.com/spinnaker/spinnaker/issues/2296

robfletcher:
I was wondering last night if we might be able to get Jackson to automatically resolve expressions where possible.

lwander:
We could also use reflection to try and evaluate all getters/setters on objects - but that will start evaluating things that weren't touched before

lwander:
I think that's OK if you just keep recursing. The evaluator already has the requirement that the input and output type of an expression must be the same, so as long as you recurse until you hit a string, you'll be able to at least do some transformation on any object.

In any case, I think I'll merge this for now.

robfletcher:
Yeah, go ahead & merge.

I think the problem is Jackson can never bind an incoming expression to an `int` field for example

lwander:
Ah good point :(

robfletcher:
Unless it can be resolved right away which is probably the case for triggers but will be a challenge if we start trying to strongly typed stage contexts, etc.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -94,27 +96,23 @@ class OperationsController {
 
     def augmentedContext = [trigger: pipeline.trigger]
     def processedPipeline = contextParameterProcessor.process(pipeline, augmentedContext, false)
+    processedPipeline.trigger = objectMapper.convertValue(processedPipeline.trigger, Trigger)
 
     if (pipelineError == null) {
       startPipeline(processedPipeline)
     } else {
+      throw pipelineError
```
ezimanyi: Why are we now throwing the error before calling `markPipelineFailed`?  This will effectively revert the change made in #1937 as we'll no longer be logging errors parsing pipelines and will just throw a server-side error.  (Also, everything after this line is unreachable.)

lwander: Whoops, that was left around from debugging a test, I thought that was backed out already. Will remove

PR #1942, opened 2018-02-05
fix(pipeline/validate): throw pipeline error _after_ recording failure

lwander:
@ezimanyi


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -101,8 +101,6 @@ class OperationsController {
     if (pipelineError == null) {
       startPipeline(processedPipeline)
     } else {
-      throw pipelineError
-      log.error("Failed to run pipeline:", pipelineError)
```
lwander: fyi I dropped this since the below "throw" will show the exception anyhow

ezimanyi:
LGTM

PR #1943, opened 2018-02-05
test(pipelines): Improve failed pipeline tests

ezimanyi:
Update the tests to enforce that failure to parse a pipeline
leads to an entry being added to the execution history.


ezimanyi:
@lwander: Updated my unit tests so that they now catch regressions such as that in #1941.

PR #1947, opened 2018-02-06
fix(spel): Fix assumptions made that were broke by #1892.

jtk54:
Fix assumptions made that were broke by #1892.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
@@ -111,7 +111,10 @@ public class ContextParameterProcessor {
       context.put("parameters", trigger.getParameters());
     }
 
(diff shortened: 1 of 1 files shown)


jtk54:
Don't merge yet, need to add unit test.

jtk54:
I don't know if that's the right fix yet, feels off that the webhook resp lands in `buildInfo`. Talking with Gard and Jorgen.

ethanfrogers:
would moving the response out of `buildInfo` constitute a breaking change?

jtk54:
I'd imagine yes, we're still breaking the contract no matter how weird it was.

lwander:
We can always put it into another field alongside `buildInfo`, and then figure out what to do with that field later. In the end a user writing SPEL shouldn't have to reference `buildInfo` to get a webhook output.

ethanfrogers:
yeah, that seems like a good idea. i'm not familiar with how the accessor methods work, but maybe that would be use case for one? then we can deprecate `buildInfo` once people have a change to move to the new field or the accessor method.

jtk54:
Merging this, will address relocating the response body in a later PR.

jervi:
I've made a shot at the first stage of the relocations in #1959. Please take a look.

PR #1951, opened 2018-02-06
chore(echo): Stop storing redundant information with each stage

robfletcher:
We've been storing this `stageDetails` object in the context of every stage for a long time but we don't actually need it for anything. We don't send that data to Echo and it's all completely redundant.


cfieber:
in favour of not storing redundant data - how certain are you that this isn't used when rendering the execution in the UI?

LGTM if we aren't using it there

robfletcher:
@cfieber `rg stageDetails .` in deck returns nothing. Looks like it's not being used.

PR #1953, opened 2018-02-07
fix(trigger): Fix bug with trying to read trigger during initialization

robfletcher:


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -61,11 +61,19 @@ class DependentPipelineStarter implements ApplicationContextAware {
     pipelineConfig.trigger = new PipelineTrigger(
       parentPipeline,
       parentPipelineStageId,
-      principal?.username ?: user ?: "[anonymous]",
-      buildParameters(pipelineConfig, suppliedParameters),
```
robfletcher: This was the problem. It was passing `pipelineConfig` to this method which will read `trigger` from it, but it's doing so while assigning `trigger`. Never going to work.

ajordens:
LGTM ... this is basically a restoration of the order in which we were building `trigger.parameters` right?

robfletcher:
Right.

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/pipeline/EnabledPipelineValidatorSpec.groovy
```diff
@@ -121,7 +121,7 @@ class EnabledPipelineValidatorSpec extends Specification {
     execution = pipeline {
       application = "whatever"
       pipelineConfigId = "1337"
-      trigger = new PipelineTrigger(null, [strategy: true])
+      trigger = new PipelineTrigger(pipeline {}, [strategy: true])
```
robfletcher: Running these tests in IntelliJ could fail if compiler `@Nonnull` checks are enforced.

ezimanyi:
LGTM

PR #1959, opened 2018-02-07
fix(webhook): Deprecate use of buildInfo

jervi:
The webhook stage puts a lot of arbitrary information into buildInfo, that broke assumptions done other places in Orca (#1892).
This commit deprecates the use of buildInfo, and adds information about the deprecation and how to migrate, to the stage output. The webhook output is now added beneath the key `webhook`.
The deprecated stuff should be removed after some time.

Note: Most error messages have been fully migrated, meaning I did not leave a copy at the root of `context`. I did the assumption that only successful results are being used in later stages.

Discussion of this issue: #1947.


jtk54:
Needs a rebase and we can merge this.

jervi:
Rebased 👍

jervi:
Hehe, that rebase didn't last long :p

jtk54:
Got it this time ;)

PR #1960, opened 2018-02-07
fix(persistence): fix bugs when running w/previousRedis

asher:
- fix bugs that could break pipelines running via previousRedis during a migration
- add related test coverage, fix test issues if running with compile time @Nonnull checks (intellij)

### DIFF
--- a/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepositoryTck.groovy
+++ b/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepositoryTck.groovy
@@ -32,6 +32,7 @@ import spock.lang.*
 import static com.netflix.spinnaker.orca.ExecutionStatus.*
 import static com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder.newStage
 import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType
+import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.ORCHESTRATION
 import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.PIPELINE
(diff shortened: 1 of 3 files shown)


orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepositoryTck.groovy
```diff
@@ -94,10 +95,12 @@ abstract class ExecutionRepositoryTck<T extends ExecutionRepository> extends Spe
     def runningExecution = orchestration {
       status = RUNNING
       buildTime = 0
+      trigger = new ManualTrigger(null, "fnord", [:], null, null)
```
robfletcher: Hail Eris!

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepositoryTck.groovy
```diff
@@ -566,16 +574,103 @@ class JedisExecutionRepositorySpec extends ExecutionRepositoryTck<JedisExecution
     100   || 4
   }
 
+  def "can store an orchestration already in previousRedis back to previousRedis"() {
```
robfletcher: Is that actually what we want to do? Don't we want to migrate it to the new Redis?

asher: Migration is still the work of `MultiRedisOrchestrationMigrationNotificationAgent` and `MultiRedisPipelineMigrationNotificationAgent`, which only operate over completed executions. We'll want to rewrite the migration process once `DynomiteExecutionRepository` is merged in with schema changes and I'd like to tackle migrating running executions then too.

robfletcher: Ok, cool

PR #1961, opened 2018-02-07
fix(triggers): verify a trigger is a Jenkins trigger before trying to pull buildInfo

robfletcher:


robfletcher:
Fixes spinnaker/spinnaker#2319

PR #1963, opened 2018-02-08
feat(queue): Keiko to 5.x message migrators

robzienert:
This is how I should've written the migrators for 6.x 🤦‍♂️


orca-queue-redis/src/test/kotlin/com/netflix/spinnaker/orca/q/redis/migration/KeikoMessageDeserializerTest.kt
```diff
@@ -0,0 +1,73 @@
...
+
+object KeikoMessageDeserializerTest : SubjectSpek<KeikoMessageDeserializer>({
+
+  val jsonParser: JsonParser = mock()
+
+  subject(CachingMode.GROUP) {
+    KeikoMessageDeserializer()
+  }
+
+  describe("handling a keiko message") {
```
ajordens: Thoughts on a quick test to verify `return mapper.readValue(p.valueAsString, Message::class.java)` when a message is encountered w/o `kind`?

PR #1964, opened 2018-02-08
Remove old pipeline queuing in favor of handling with message queue

robfletcher:
Instead of maintaining a queue of pipelines waiting to be executed this change means that `StartExecutionHandler` will just ignore any pipeline if its `limitConcurrent` flag is `true` and there is already a running instance. At the other end `CompleteExecutionHandler` emits a new message type `StartWaitingExecutions` that will check for not started executions and start the next one, cancelling any remaining not started instances if `keepWaitingPipelines` is `false`.


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
```diff
@@ -63,6 +57,9 @@ class CompleteExecutionHandler(
           publisher.publishEvent(
             ExecutionComplete(this, message.executionType, message.executionId, status)
           )
+          execution.pipelineConfigId?.let {
+            queue.push(StartWaitingExecutions(it, purgeQueue = !execution.isKeepWaitingPipelines))
+          }
```
robfletcher: Here's where it fires the message to go check for any waiting executions.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartExecutionHandler.kt
```diff
@@ -40,36 +42,57 @@ class StartExecutionHandler(
...
+      publisher.publishEvent(ExecutionComplete(this, execution.type, execution.id, execution.status))
+    } else {
+      log.warn("Execution (type: ${execution.type}, id: {}, status: ${execution.status}, application: {})" +
+        " cannot be started unless state is NOT_STARTED. Ignoring StartExecution message.",
+        value("executionId", execution.id),
+        value("application", execution.application))
+    }
+  }
+
+  private fun Execution.shouldQueue(): Boolean {
```
robfletcher: Queuing logic moved into `StartExecutionHandler`. We can simply no-op if there's already an instance of this pipeline running.

PR #1966, opened 2018-02-08
feat(artifacts) Artifact passing in pipelines

ezimanyi:
When a pipeline is triggered by another pipeline, allow the
second pipeline to see any artifacts that were emitted by
the first pipeline.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
@@ -70,6 +70,30 @@ public class ArtifactResolver {
     }
   }
 
+  public @Nonnull
+  List<Artifact> getArtifacts(@Nonnull Execution execution) {
+    // Get all artifacts emitted by the execution's stages; we'll sort the stages topologically,
+    // then reverse the result so that artifacts from later stages will appear
+    // earlier in the results.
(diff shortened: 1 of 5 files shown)


ezimanyi:
@lwander

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -70,6 +70,30 @@ public ArtifactResolver(ObjectMapper objectMapper, ExecutionRepository execution
     }
   }
 
+  public @Nonnull
+  List<Artifact> getArtifacts(@Nonnull Execution execution) {
```
lwander: Could we call this "getAllArtifacts" for clarity?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -80,21 +89,42 @@ class DependentPipelineStarter implements ApplicationContextAware {
     }
     pipelineConfig.trigger = trigger
 
+    def artifactError = null
+    try {
+      artifactResolver?.resolveArtifacts(pipelineConfig)
+    }
```
lwander: Usually we put `catch` on the same line as this closing `}`

PR #1970, opened 2018-02-09
fix(artifacts): fix passing of pipeline artifacts

lwander:
Pipelines that don't declare "expected artifacts" still need to be able to pass artifacts into their triggers

@ezimanyi PTAL


orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarterSpec.groovy
```diff
@@ -308,4 +308,70 @@ class DependentPipelineStarterSpec extends Specification {
...
+        [:],
+        null
+    )
+
+    then:
+    result.trigger.getArtifacts().size() == 2
+    result.trigger.getArtifacts()*.name.contains(testArtifact1.name)
+    result.trigger.getArtifacts()*.name.contains(testArtifact2.name)
+    result.trigger.resolvedExpectedArtifacts.size() == 1
+    result.trigger.resolvedExpectedArtifacts*.boundArtifact.name == [testArtifact1.namet ad]
```
ezimanyi: Looks like a typo at the end of this line

PR #1974, opened 2018-02-10
fix(trigger): fix NPE caused by changing PipelineTrigger to a map

robfletcher:


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -60,10 +58,11 @@ class DependentPipelineStarter implements ApplicationContextAware {
...
     pipelineConfig.trigger = [
-      type                  : "pipeline",
-      user                  : principal?.username ?: user ?: "[anonymous]",
-      parentExecution       : parentPipeline,
-      parentPipelineStageId : parentPipelineStageId
+      type                 : "pipeline",
+      user                 : principal?.username ?: user ?: "[anonymous]",
+      parentExecution      : parentPipeline,
+      parentPipelineStageId: parentPipelineStageId,
+      parameters           : [:]
```
robfletcher: without this subsequent lines NPE when they try to reference `pipelineConfig.trigger.parameters`

PR #1976, opened 2018-02-11
Delete old listeners from pre v3 orca

robfletcher:


robfletcher:
I thought you’d already merged a change that killed these about a week ago!

PR #1944, opened 2018-02-05
feat(rollback): Support for automatic rollback when a deployment fails

ajordens:
This PR injects a `RollbackCluster` stage if a particular cluster
deployment has automatic rollback enabled.

Automatic rollback can be enabled by adding the following to a deploy
stage:

```
      "rollback": {
        "onFailure": true
      }
```

PR #1978, opened 2018-02-12
feat(core): Allow a synthetic stage to override `stageTimeoutMs`

ajordens:
```
def pipeline = pipeline {
  stage {
    id = "1"

    stage {
      id = "2"

      stage {
        id = "3"
      }
    }
  }
}
```

This PR allows a synthetic stage (#2 in ^^ example) to override the
`stageTimeoutMs` that will be used by a child synthetic stage (#3).

The current behavior has a synthetic stage only looking at it's root
top-level stage for `stageTimeoutMs`.

Will be used by spinnaker/orca#1944

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
@@ -486,12 +486,22 @@ public class Stage implements Serializable {
     return topLevelStage;
   }
 
-  /**
-   * Returns the top-most stage timeout value if present.
-   */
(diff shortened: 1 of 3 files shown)


ajordens:
@emjburns  -- Not sure if there is/was a good reason to only look at the top-level stage for a `stageTimeoutMs` override? 

In the case of automatic rollbacks, I'd like to support rollbacks that occurred as a result of a timeout. As such, I need to be able to inject synthetic stages with their own `stageTimeoutMs`.

robfletcher:
I think the reason was that we thought it could only be configured on the top level. Maybe that was an incorrect assumption.

PR #1977, opened 2018-02-12
Trigger payloads

robfletcher:
The mistake I made with the previous implementation of trigger types was thinking that the `type` field specified the kind of payload the trigger contains. This is wrong, as a manual trigger can have Jenkins build info, for example.

This implementation instead identifies the type based on the fields that exist in the JSON. This means any type of trigger can contain any type of payload.

PR #1983, opened 2018-02-13
fix(core): use correct object mapper for context processing

lwander:
The default object mapper was not ignoring unknown props, causing various trigger types to crash pipelines.


lwander:
Closing in favor of #1977 (but will still cherry pick this  into release-1.6.x)

PR #1987, opened 2018-02-13
fix(trigger): Add trigger to processed pipeline

jervi:
Further fixes for #1937. To also make it work with #1941, I couldn't just simply copy the trigger as was done before.


jervi:
Opened #1988 instead (targeting `master` instead)

PR #1991, opened 2018-02-14
fix(pipelines): improve error msg if running disabled pipeline

asher:


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/StartPipelineTask.groovy
```diff
@@ -53,6 +53,10 @@ class StartPipelineTask implements Task {
       throw new IllegalArgumentException("The referenced ${isStrategy ? 'custom strategy' : 'pipeline'} cannot be located (${pipelineId}")
     }
 
+    if (pipelineConfig.get('disabled') != null && pipelineConfig.get('disabled') == true) {
```
robfletcher: ```if (pipelineConfig.getOrDefault('disabled', false))```

asher: danke!

PR #1992, opened 2018-02-14
fix(queue): Only remove synthetic stages if there are replacements

ajordens:
Currently any non-started synthetic stage is removed from the graph
even if there are no `onFailure` stages to replace them.


ajordens:
@christopherthielen -- I think it's good to remove the expectation on particular stages existing in `deck` but this should avoid the issue in the short-term.

PR #1994, opened 2018-02-15
feat(artifacts): allow stages to 'produce artifacts'

lwander:
https://github.com/spinnaker/spinnaker/issues/2339

@ezimanyi PTAL

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
@@ -58,13 +58,12 @@ public class FindArtifactFromExecutionTask implements Task {
 
     List<Artifact> priorArtifacts = artifactResolver.getArtifactsForPipelineId(pipeline, executionOptions.toCriteria());
 
-    Artifact match = artifactResolver.resolveSingleArtifact(expectedArtifact, priorArtifacts);
+    Artifact match = artifactResolver.resolveSingleArtifact(expectedArtifact, priorArtifacts, false);
 
     if (match == null) {
       return new TaskResult(ExecutionStatus.TERMINAL);
     }
 
-    expectedArtifact.setBoundArtifact(match);
     outputs.put("resolvedExpectedArtifacts", Collections.singletonList(expectedArtifact));
     outputs.put("artifacts", Collections.singletonList(match));
 
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
@@ -16,11 +16,6 @@
 
 package com.netflix.spinnaker.orca.pipeline.util;
 
-import java.io.IOException;
-import java.util.*;
-import java.util.stream.Collectors;
-import javax.annotation.Nonnull;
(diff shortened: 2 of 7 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
```diff
@@ -64,7 +64,6 @@ public TaskResult execute(@Nonnull Stage stage) {
       return new TaskResult(ExecutionStatus.TERMINAL);
     }
 
-    expectedArtifact.setBoundArtifact(match);
```
lwander: I removed this because the "artifactResolver" handles this logic.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -196,33 +187,55 @@ public void resolveArtifacts(@Nonnull Map pipeline) {
...
+      }
+
+      if (resolved == null && expectedArtifact.isUseDefaultArtifact() && expectedArtifact.getDefaultArtifact() != null) {
+        resolved = expectedArtifact.getDefaultArtifact();
+      }
+
+      if (resolved == null) {
+        throw new IllegalStateException(format("Unmatched expected artifact %s could not be resolved.", expectedArtifact));
+      } else {
+        expectedArtifact.setBoundArtifact(resolved);
```
ezimanyi: This is really minor, bit it feels to me that `expectedArtifact.setBoundArtifact` and `resolvedArtifacts.add(resolved)` would more logically live a few lines above after `resolved = expectedArtifact.getDefaultArtifact();`.  It also makes it more clear that we are explicitly calling `setBoundArtifact` iff we don't call `resolveSingleArtifact`.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -196,33 +187,55 @@ public void resolveArtifacts(@Nonnull Map pipeline) {
...
+        .stream()
+        .filter(expectedArtifact::matches)
+        .collect(toList());
     switch (matches.size()) {
       case 0:
         return null;
-      case 1:
-        return matches.get(0);
       default:
-        throw new IllegalStateException(format("Expected artifact %s matches multiple artifacts %s", expectedArtifact, matches));
```
ezimanyi: Curious why we are now allowing multiple artifacts to match and picking the first, rather than expecting exactly one to match as before

ezimanyi:
@lwander : One questions, one small comment; otherwise LGTM

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
lwander: Agreed

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
lwander: This code was originally written when expected artifacts only lived in triggers - and we generally treat artifacts at the start of execution as unordered. But with this change, allowing expected artifact in the middle of an execution it's desirable to always bind the most recent artifact.

Thinking about it I'll add a flag to make sure still fail triggers with ambiguous matches

PR #1997, opened 2018-02-16
feat(aws): Support for pinning min=desired capacity on source server group

ajordens:
This PR attempts to avoid unnecessary and problematic autoscale events
that occur as the total # of instances increase when a deploy is
happening.

As instance count increases, there is natural downward pressure
on the cluster resulting in the source server group being scaled down.

It is particularly bad for rolling red/black where the old server group
is not immediately disabled.

The expectation is that the source group will be returned to its
original unpinned capacity when the deploy completes (or fails!).

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy
@@ -17,41 +17,141 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.providers.aws
 
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerGroupStage
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.AbstractDeployStrategyStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.DeployStagePreProcessor
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroupResolver
+import com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategy
 import com.netflix.spinnaker.orca.kato.pipeline.support.StageData
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
+
+import static com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategySupport.getSource
 
 @Component
 class AwsDeployStagePreProcessor implements DeployStagePreProcessor {
   @Autowired
   ApplySourceServerGroupCapacityStage applySourceServerGroupSnapshotStage
 
+  @Autowired
+  ResizeServerGroupStage resizeServerGroupStage
+
+  @Autowired
+  TargetServerGroupResolver targetServerGroupResolver
+
   @Override
-  List<DeployStagePreProcessor.StepDefinition> additionalSteps() {
+  List<StepDefinition> additionalSteps(Stage stage) {
+    def stageData = stage.mapTo(StageData)
+    if (stageData.strategy == "rollingredblack") {
+      // rolling red/black has no need to snapshot capacities
+      return []
+    }
+
     return [
-        new DeployStagePreProcessor.StepDefinition(
-          name: "snapshotSourceServerGroup",
-          taskClass: CaptureSourceServerGroupCapacityTask
-        )
+      new StepDefinition(
+        name: "snapshotSourceServerGroup",
+        taskClass: CaptureSourceServerGroupCapacityTask
+      )
     ]
   }
 
   @Override
-  List<DeployStagePreProcessor.StageDefinition> afterStageDefinitions() {
-    return [
-        new DeployStagePreProcessor.StageDefinition(
-          name: "restoreMinCapacityFromSnapshot",
-          stageDefinitionBuilder: applySourceServerGroupSnapshotStage,
-          context: [:]
+  List<StageDefinition> beforeStageDefinitions(Stage stage) {
+    def stageData = stage.mapTo(StageData)
+    if (shouldPinSourceServerGroup(stageData.strategy)) {
+      def resizeContext = getResizeContext(stageData)
+      resizeContext.pinMinimumCapacity = true
+
+      return [
+        new StageDefinition(
+          name: "Pin ${resizeContext.serverGroupName}".toString(),
+          stageDefinitionBuilder: resizeServerGroupStage,
+          context: resizeContext
         )
-    ]
+      ]
+    }
+
+    return []
+  }
+
+  @Override
+  List<StageDefinition> afterStageDefinitions(Stage stage) {
+    def stageData = stage.mapTo(StageData)
+    def stageDefinitions = []
+    if (stageData.strategy != "rollingredblack") {
+      // rolling red/black has no need to apply a snapshotted capacity (on the newly created server group)
+      stageDefinitions << new StageDefinition(
+        name: "restoreMinCapacityFromSnapshot",
+        stageDefinitionBuilder: applySourceServerGroupSnapshotStage,
+        context: [:]
+      )
+    }
+
+    def unpinServerGroupStage = buildUnpinServerGroupStage(stageData)
+    if (unpinServerGroupStage) {
+      stageDefinitions << unpinServerGroupStage
+    }
+
+    return stageDefinitions
+  }
+
+  @Override
+  List<StageDefinition> onFailureStageDefinitions(Stage stage) {
+    def stageData = stage.mapTo(StageData)
+    def stageDefinitions = []
+
+    def unpinServerGroupStage = buildUnpinServerGroupStage(stageData)
+    if (unpinServerGroupStage) {
+      stageDefinitions << unpinServerGroupStage
+    }
+
+    return stageDefinitions
   }
 
   @Override
   boolean supports(Stage stage) {
     def stageData = stage.mapTo(StageData)
-    return stageData.useSourceCapacity && stageData.cloudProvider == "aws" && stageData.strategy != "rollingredblack"
+    return stageData.cloudProvider == "aws" // && stageData.useSourceCapacity
+  }
+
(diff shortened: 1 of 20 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy
```diff
@@ -17,41 +17,118 @@
...
+
+    return []
+  }
+
+  @Override
+  List<StageDefinition> afterStageDefinitions(Stage stage) {
+    def stageData = stage.mapTo(StageData)
+    def stageDefinitions = []
+    if (stageData.strategy != "rollingredblack") {
+      // rolling red/black has no need to apply a snapshotted capacity (on the newly created server group)
```
ajordens: this preprocessor was previously excluding rolling redblack completely ... specifically because we don't need to manipulate the capacity of the newly generated server group (the rolling red/black strategy already does it!)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy
```diff
@@ -17,41 +17,118 @@
...
   @Override
   boolean supports(Stage stage) {
     def stageData = stage.mapTo(StageData)
-    return stageData.useSourceCapacity && stageData.cloudProvider == "aws" && stageData.strategy != "rollingredblack"
+    return stageData.cloudProvider == "aws" && stageData.useSourceCapacity
+  }
+
+  private static boolean shouldPinSourceServerGroup(String strategy) {
+    // TODO-AJ consciously only enabling for rolling red/black -- will add support for other strategies after it's working
+    return strategy == "rollingredblack"
```
ajordens: Rolling red/black is still flagged as `experimental` and I'd like to soft roll this out to a couple internal groups before adding to red/black. 

I also need to make sure we've got a suitable story for ensuring the capacity gets unpinned in the event of failure.

pinning min=desired on the source server group is something that has been asked (by internal users @ netflix) for years!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy
```diff
@@ -17,41 +17,118 @@
...
+  }
+
+  private Map<String, Object> getResizeContext(StageData stageData) {
```
ajordens: TBH this came from `RollingRedBlackStrategy`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/AbstractDeployStrategyStage.groovy
```diff
@@ -97,15 +97,17 @@ abstract class AbstractDeployStrategyStage extends AbstractCloudProviderAwareSta
     Strategy strategy = (Strategy) strategies.findResult(noStrategy, {
       it.name.equalsIgnoreCase(stage.context.strategy) ? it : null
     })
+
+    def preProcessors = deployStagePreProcessors.findAll { it.supports(stage) }
     def stages = strategy.composeFlow(stage)
 
     def stageData = stage.mapTo(StageData)
-    deployStagePreProcessors.findAll { it.supports(stage) }.each {
```
ajordens: `strategy.composeFlow()` manipulates the stage context and I'd like to determine pre-processors based on the original unmodified stage

specifically the `RollingRedBlackStrategy` sets `useSourceCapacity` to `false`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -89,7 +91,6 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
 
     def targetPercentages = stageData.getTargetPercentages()
     if (targetPercentages.size() == 0 || targetPercentages[-1] != 100) {
-      log.info("Inserting implicit 100% final target percentage...")
```
ajordens: Rather useless log message!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -107,17 +108,19 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
       SyntheticStageOwner.STAGE_AFTER
     )
 
+    def source = getSource(targetServerGroupResolver, stageData, baseContext)
```
ajordens: This could potentially make a call to clouddriver ... only needs to happen once

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ResizeServerGroupTask.groovy
```diff
@@ -40,10 +39,23 @@ class ResizeServerGroupTask extends AbstractServerGroupTask {
...
+  Map getAdditionalContext(Stage stage, Map operation) {
     [capacity: operation.capacity]
   }
 
+  /**
+   * Track the _original_ capacity of the server group being resized in case it needs to be subsequently restored.
+   */
+  @Override
+  Map<String, Object> getAdditionalOutputs(Stage stage, Map operation) {
+    def originalCapacityKey = "originalCapacity.${operation.serverGroupName}".toString()
```
ajordens: Supports a similar pattern to that used in `AbstractAwsScalingProcessTask`

```
def processes = filterProcesses(
      targetServerGroup,
      (stage.context.processes ?: stage.context["scalingProcesses.${asgName}" as String]) as List<String>
    )
```

A `ResizeServerGroupStage` can optionally look for original capacity and use it to unpin the `min`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/ResizeStrategySupport.groovy
```diff
@@ -0,0 +1,53 @@
...
+
+package com.netflix.spinnaker.orca.kato.pipeline.support
+
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.RollingRedBlackStageData
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroupResolver
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+
+public class ResizeStrategySupport {
+  static ResizeStrategy.Source getSource(TargetServerGroupResolver targetServerGroupResolver,
```
ajordens: Lifted from `RollingRedBlackStrategy`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/ScaleToClusterResizeStrategy.groovy
```diff
@@ -73,6 +75,6 @@ class ScaleToClusterResizeStrategy implements ResizeStrategy{
     capacity.desired = Math.max(capacity.min, capacity.desired)
     capacity.desired = Math.min(capacity.desired + increment, capacity.max)
 
-    return capacity
+    return new CapacitySet(null, capacity)
```
ajordens: This strategy doesn't take the capacity from a single server group (rather it looks for the max across every server group in a cluster) ... as such there is no suitable original capacity to capture.

ajordens:
Wouldn't mind some eyes @robfletcher / @lwander.

@lwander -- this largely involves some AWS-specific bits we've always had in-place (pinning min=desired capacity while server groups are deploying) ... it's quite possible other cloud providers may benefit from a similar treatment.

I still need to circle around in an attempt to restore the altered capacities if a task or pipeline fails but figured it's worth getting an early review anyhow.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy
```diff
@@ -17,41 +17,118 @@
...
+import com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategy
 import com.netflix.spinnaker.orca.kato.pipeline.support.StageData
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
+import static com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategySupport.getSource
+
 @Component
 class AwsDeployStagePreProcessor implements DeployStagePreProcessor {
```
ajordens: In case you haven't seen these before, a `DeployStagePreProcessor` allows you to inject stages before/after a deploy.

`aws` has been using this to pin min=desired capacity while a new server group is being deployed.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
lwander: sorry :'|

ajordens: :)

amanya:
I found that this PR is causing us errors when resizing server groups. I have a pipeline with a deploy and then a resize server group stage with this configuration:

image

I get this error:

image

ajordens:
Doh, fixed in spinnaker/orca#2010 and will be merged shortly @amanya.

PR #1998, opened 2018-02-16
fix(spel): Support new return types in SpEL expressions

jervi:
Make SpEL support the new strongly typed `Trigger`s and some other models. Gets rid of errors like this: `Failed to evaluate [expression] found getter for requested trigger but rejected due to return type interface com.netflix.spinnaker.orca.pipeline.model.Trigger`


robfletcher:
Good catch.

PR #2003, opened 2018-02-17
chore(bakery): remove baseOs and baseLabel enums

asher:


ajordens:
This seems reasonable?

PR #2005, opened 2018-02-17
fix(artifacts): fix parsing buildInfo when artifacts missing

asher:


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.java
```diff
@@ -269,7 +269,9 @@ private String extractPackageVersion(Map<String, Object> artifact, String filePr
       .stream()
       .filter(it -> {
         Map<String, Object> buildInfo = (Map<String, Object>) it.getOutputs().get("buildInfo");
-        return buildInfo != null && artifactMatch((List<Map<String, String>>) buildInfo.get("artifacts"), packageFilePattern);
+        return buildInfo != null &&
+          buildInfo.containsKey("artifacts") &&
+          artifactMatch((List<Map<String, String>>) buildInfo.get("artifacts"), packageFilePattern);
```
ajordens: IMO I'd add a null check to `artifactMatch` instead ... it would also protect against a buildInfo that looked like `[artifacts: null]`.

asher: sounds good, didn't think of that

PR #2006, opened 2018-02-19
feat(persistence): Dynomite ExecutionRepository

asher:
- this PR excludes dynomite configurate classes (to use classes recently added to kork after RedisClientDelegate use is refactored)
- DynomiteExecutionRepositorySpec is pending an upcoming DynomiteClient release that is compatible with EmbeddedRedis.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/dynomite/DynomiteExecutionRepository.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/dynomite/DynomiteExecutionRepository.java
@@ -14,47 +14,55 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.orca.pipeline.persistence.jedis;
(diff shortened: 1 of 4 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/dynomite/DynomiteExecutionRepository.java
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  @Override
+  protected String executionKey(Stage stage) {
+    return format("{%s:%s}", stage.getExecution().getType(), stage.getExecution().getId());
+  }
+
+  @Override
+  protected String pipelineKey(String id) {
+    return format("{pipeline:%s}", id);
```
robzienert: I think this and `orchestrationKey` should use the `ExecutionType` enum and `toString` instead.

asher: ➕

PR #2011, opened 2018-02-21
fix(clouddriver): Getting task from clouddriver master when timed out attempts to replicas

jeyrschabu:
- keeping track if we have tried master
- added fx to look up task from master directly (KatoRestService)

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
@@ -72,21 +72,29 @@ class MonitorKatoTask implements RetryableTask {
     }
 
     Task katoTask
+    def skipReplica = stage.context."kato.task.skipReplica" ?: false
     try {
-      katoTask = kato.lookupTask(taskId.id).toBlocking().first()
+      katoTask = kato.lookupTask(taskId.id, skipReplica as Boolean).toBlocking().first()
     } catch (RetrofitError re) {
       //handle a 404 if a task update has not successfully replicated to a read replica
       if (re.kind == RetrofitError.Kind.HTTP && re.response.status == HttpURLConnection.HTTP_NOT_FOUND) {
         def firstNotFoundRetry = stage.context."kato.task.firstNotFoundRetry" as Long
+
         def now = clock.millis()
         def ctx = [:]
         if (firstNotFoundRetry == null || firstNotFoundRetry == -1) {
           ctx['kato.task.firstNotFoundRetry'] = now
           firstNotFoundRetry = now
         }
+
         if (now - firstNotFoundRetry > TASK_NOT_FOUND_TIMEOUT) {
+          if (skipReplica) {
+            // immediately fail the first time it gets a 404 directly from the master
+            throw re
+          }
+
           registry.counter("monitorKatoTask.taskNotFound.timeout").increment()
-          throw re
+          ctx['kato.task.skipReplica'] = true
         }
 
         registry.counter("monitorKatoTask.taskNotFound.retry").increment()
--- a/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTaskSpec.groovy
+++ b/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTaskSpec.groovy
@@ -34,6 +34,8 @@ import spock.lang.Specification
 import spock.lang.Subject
 import spock.lang.Unroll
 
+import static com.netflix.spinnaker.orca.test.model.ExecutionBuilder.stage
+
 class MonitorKatoTaskSpec extends Specification {
 
   def now = Instant.now()
@@ -43,7 +45,7 @@ class MonitorKatoTaskSpec extends Specification {
   def "result depends on Kato task status"() {
     given:
     task.kato = Stub(KatoService) {
-      lookupTask(taskId) >> Observable.from(new Task(taskId, new Task.Status(completed: completed, failed: failed), [], []))
+      lookupTask(taskId, false) >> Observable.from(new Task(taskId, new Task.Status(completed: completed, failed: failed), [], []))
     }
 
     and:
@@ -68,7 +70,7 @@ class MonitorKatoTaskSpec extends Specification {
   def "result depends on Kato task status and result object size for create/upsert operations"() {
     given:
     task.kato = Stub(KatoService) {
-      lookupTask(taskId) >> Observable.from(new Task(taskId, new Task.Status(completed: true), resultObjects, []))
+      lookupTask(taskId, false) >> Observable.from(new Task(taskId, new Task.Status(completed: true), resultObjects, []))
     }
 
     and:
@@ -127,7 +129,7 @@ class MonitorKatoTaskSpec extends Specification {
     }
     def stage = new Stage(Execution.newPipeline("orca"), "whatever", ctx)
     task.kato = Stub(KatoService) {
-      lookupTask(taskId) >> { retrofit404() }
+      lookupTask(taskId, false) >> { retrofit404() }
     }
 
     when:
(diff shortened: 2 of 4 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -72,21 +72,29 @@ class MonitorKatoTask implements RetryableTask {
     }
 
     Task katoTask
+    def firstNotFoundRetrySkippingReplica = stage.context."kato.task.firstNotFoundRetry.skipReplica" as Long
```
ajordens: Including the `firstNotFoundRetry` in this key seems strange ... 

`def skipReplica = stage.context."kato.task.skipReplica" ?: false` instead?

I don't really think it needs to be a timestamp?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -72,21 +72,29 @@ class MonitorKatoTask implements RetryableTask {
...
     try {
-      katoTask = kato.lookupTask(taskId.id).toBlocking().first()
+      katoTask = kato.lookupTask(taskId.id, firstNotFoundRetrySkippingReplica != null).toBlocking().first()
     } catch (RetrofitError re) {
       //handle a 404 if a task update has not successfully replicated to a read replica
       if (re.kind == RetrofitError.Kind.HTTP && re.response.status == HttpURLConnection.HTTP_NOT_FOUND) {
-        def firstNotFoundRetry = stage.context."kato.task.firstNotFoundRetry" as Long
+        def firstNotFoundRetry = !firstNotFoundRetrySkippingReplica ? stage.context."kato.task.firstNotFoundRetry" as Long
```
ajordens: I don't think this is necessary

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -72,21 +72,29 @@ class MonitorKatoTask implements RetryableTask {
...
+
         def now = clock.millis()
         def ctx = [:]
         if (firstNotFoundRetry == null || firstNotFoundRetry == -1) {
           ctx['kato.task.firstNotFoundRetry'] = now
           firstNotFoundRetry = now
         }
+
         if (now - firstNotFoundRetry > TASK_NOT_FOUND_TIMEOUT) {
+          if (firstNotFoundRetrySkippingReplica) {
```
ajordens: Is this just an optimization? Are you thinking you want it to immediately fail the first time it gets a 404 directly from the master?

If so maybe a comment along those lines. I also think the variable being called `skipReplica` is a little clearer than `firstNotFoundRetrySkippingReplica`.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTaskSpec.groovy
```diff
@@ -146,15 +146,58 @@ class MonitorKatoTaskSpec extends Specification {
...
+    }
+
+    when:
+    def result = task.execute(stage)
+
+    then:
+    result.status == ExecutionStatus.RUNNING
+    result.context['kato.task.notFoundRetryCount'] == 1
+    if (!skipReplicaFlag) {
+      result.context['kato.task.firstNotFoundRetry.skipReplica'] == null
```
ajordens: I would avoid the if/else in your `then` block and just make this equal some concrete variable you specify in your `where` block down below.

ie.

```
blah | blarg || shouldSkipReplica
x    | y     || true
```

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTaskSpec.groovy
```diff
@@ -146,15 +146,58 @@ class MonitorKatoTaskSpec extends Specification {
...
+    } else {
+      result.context['kato.task.firstNotFoundRetry.skipReplica'] != null
+    }
+
+    where:
+    taskId = "katoTaskId"
```
ajordens: making a constant for `taskId` down here doesn't really add a lot of value ... you could just define that in your `given` block

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTaskSpec.groovy
```diff
@@ -146,15 +146,58 @@ class MonitorKatoTaskSpec extends Specification {
...
+    if (previousRetry) {
+      ctx.put('kato.task.firstNotFoundRetry', now.minusMillis(elapsed).toEpochMilli())
+    } else {
+      if (previousRetryFlag != null) {
+        ctx.put('kato.task.firstNotFoundRetry', previousRetryFlag)
+      }
+    }
+
+    def stage = new Stage(Execution.newPipeline("orca"), "whatever", ctx)
+    task.kato = Stub(KatoService) {
```
ajordens: This sort of thing might make more sense as a `Mock` that you then configure and assert calls against in your `then` block

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
jeyrschabu: I had something else in mind using the stamp, but since we are only retrying the master once, I like the idea of using a boolean

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTaskSpec.groovy
```diff
@@ -146,15 +146,65 @@ class MonitorKatoTaskSpec extends Specification {
...
+    def taskId = "katoTaskId"
+    def elapsed = MonitorKatoTask.TASK_NOT_FOUND_TIMEOUT + 1
+    task.kato = Mock(KatoService)
+
+    def stage = new Stage(Execution.newPipeline("orca"), "whatever", [
+      "kato.last.task.id": new TaskId(taskId),
+      "kato.task.firstNotFoundRetry": now.minusMillis(elapsed).toEpochMilli()
+    ])
+
+    assert !(stage.context['kato.task.skipReplica'] as Boolean)
```
ajordens: this seems unnecessary

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTaskSpec.groovy
```diff
@@ -146,15 +146,65 @@ class MonitorKatoTaskSpec extends Specification {
...
+
+    when:
+    def result = task.execute(stage)
+
+    then:
+    result.context['kato.task.skipReplica'] as Boolean
```
ajordens: this is maybe a little more readable as `result.context['kato.task.skipReplica'] == true`

IntelliJ might complain but I still find it more readable

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTaskSpec.groovy
```diff
@@ -146,15 +146,65 @@ class MonitorKatoTaskSpec extends Specification {
     "about to fail"      | MonitorKatoTask.TASK_NOT_FOUND_TIMEOUT     |  true         | 'N/A'
   }
 
+  def "should set kato.task.skipReplica=true when getting a task from clouddriver times out"() {
+    given:
+    def taskId = "katoTaskId"
+    def elapsed = MonitorKatoTask.TASK_NOT_FOUND_TIMEOUT + 1
+    task.kato = Mock(KatoService)
+
+    def stage = new Stage(Execution.newPipeline("orca"), "whatever", [
```
ajordens: There is a shorthand way of doing I think ..

```
def stage = stage {
  context = [
    "kato.last.task.id": ...
  ]
}
```

`stage {` comes from ExecutionBuilder

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTaskSpec.groovy
```diff
@@ -146,15 +146,65 @@ class MonitorKatoTaskSpec extends Specification {
...
+    def taskId = "katoTaskId"
     def ctx = [
       "kato.last.task.id": new TaskId(taskId)
     ]
     ctx.put('kato.task.firstNotFoundRetry', now.minusMillis(elapsed).toEpochMilli())
+    ctx.put('kato.task.skipReplica', true)
     def stage = new Stage(Execution.newPipeline("orca"), "whatever", ctx)
     task.kato = Stub(KatoService) {
-      lookupTask(taskId) >> { retrofit404() }
+      lookupTask(taskId, false) >> { retrofit404() }
```
ajordens: If we're setting `skipReplica == true` should we ever even expect a call to `lookupTask(taskId, false)`

PR #2014, opened 2018-02-23
feat(clouddriver): Improved support for partially disabling a server group

ajordens:
If `clouddriver` returns a specific set of instance ids that have been
disabled as part of a percentage-based disable (it will for aws!),
`orca` only needs to wait until these instances have been disabled.

This short circuits a recalculation that may be out of date if
autoscaling has occurred in the background.


robzienert:
Do like.

ajordens:
FYI @lwander -- this should be `aws`-only for the time being so I wouldn't expect it to be overly problematic.

PR #2016, opened 2018-02-24
fix(core): Correctly handle artifact matching with multiple packages

ajordens:
This appears to have been broken when the global context was removed.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.java
@@ -41,7 +41,7 @@ public class PackageInfo {
   private final boolean extractBuildDetails;
   private final boolean extractVersion;
   private final BuildDetailExtractor buildDetailExtractor;
-  private final Pattern packageFilePattern;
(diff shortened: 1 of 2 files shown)


ajordens:
FYI @bensnet1

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.java
```diff
@@ -52,7 +52,15 @@ public PackageInfo(Stage stage, String packageType, String versionDelimiter, boo
     this.mapper = mapper;
     this.buildDetailExtractor = new BuildDetailExtractor();
 
-    packageFilePattern = Pattern.compile(format("%s.*\\.%s", stage.getContext().get("package"), packageType));
+    // can be a space separated set of packages
+    if (stage.getContext().containsKey("package")) {
+      String packages = stage.getContext().get("package").toString();
```
ajordens: `package` can actually be a space-separated set of packages ... in such cases a regex of:
`package1 package2.*\.deb` does not make a lot of sense.

This worked in the old world because the Jenkins stage emitted `buildInfo` to the global context which happened to short circuit the call to `artifactMatch()`.

ajordens:
Mind giving this a quick glance @robfletcher when you have a moment

PR #2017, opened 2018-02-26
fix(docs) Correct headers parameter in orca.yml

avram:
Example had incorrect parameter for headers in preconfigured webhook configuration

Corrected to match actual field name in the corresponding Groovy: https://github.com/spinnaker/orca/blob/master/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/config/PreconfiguredWebhookProperties.groovy#L45


ethanfrogers:
👍

PR #2019, opened 2018-02-26
fix(spel): Revert "fix(spel): Operate on evaluated stage in handlers.…

jtk54:
… (#1972)"

This reverts commit d9e35a5a8986f7f7a3d3c13ce54a5d68bc4377f9. @imosquera FYI.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
@@ -72,42 +72,41 @@ class StartStageHandler(
 
   override fun handle(message: StartStage) {
     message.withStage { stage ->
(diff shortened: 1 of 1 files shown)


imosquera:
Thanks @jtk54 - This created issues because it was processing SpEL for top level stages.  For example in multi-region bakes, if you used `my-var-{region}.json` for packer template var file it would get replaced with an empty string: `my-var-.json` and then the substages that were created per region would not be able to replace the `{region}` in the downstream stage since they were already replaced.  

The way I solved this locally, but do not fully understand the effects, is to set `ignoreUnknownProperties` to true in the `ExpressionEvaluator` so that the variable isn't replaced with an empty string

https://github.com/spinnaker/orca/blob/fc87a8e99ec59a4e129646026987aa5adbde2d31/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionEvaluator.java#L22

PR #2020, opened 2018-02-27
chore(build): add debug flag to orca build

ezimanyi:
Starting orca with ./gradlew -DDEBUG=true now causes the JVM to
listen on port 8183 for a remote debugger.


lwander:
Can you add an entry to the README explaining how to take advantage of this?

ezimanyi:
Good point @lwander; I've documented this in the README.

PR #2021, opened 2018-02-27
refactor(artifacts): Move artifact binding to orca-core

ezimanyi:
In order to use the artifact binding functionality in the image bakery,
we'll need this functionality to be available in orca-core so it can
be called from orca-bakery.


ezimanyi:
Probably didn't need to pull this out into its own commit, but will help avoid conflicts to merge the file move sooner and will make the diff on the actual change cleaner.

PR #2022, opened 2018-02-27
Updating Kayenta code in preparation for new work

robfletcher:


robfletcher:
I've broken `orca-test` out into:

- `orca-test-groovy` (contains Groovy-specific test helpers)
- `orca-test-kotlin` (Kotlin-specific test helpers)
- `orca-test` (general test helpers written in Java to avoid dependency bloat, provides dummy Eureka, embedded Redis, etc.)

There's also now an `orca-kotlin` module for any extension methods we want to define and use across other modules. Take a look in there for the kind of thing I mean.

duftler:
@robfletcher Anything we should know here? Didn't know this PR was coming, it's a couple of thousand lines, and pretty difficult to tell if the contract between orca and kayenta has changed at all. Also not sure if the test coverage is the same. Would have been easier to read if the huge refactoring had been done first, or separately, before rewriting everything in kotlin. Or if we had talked about the proposed changes first.

I'm a bit concerned as we were planning on making kayenta public March 15, and that's just a couple of weeks away. Best as I can tell, every line of code in orca related to kayenta was just replaced?

If everything works exactly the same, and it's just a refactoring, then we won't spend the time trying to read through it. But please let us know.

Also, this is kind of ominous: "...in preparation for new work".

Is there more we should know?

robfletcher:
Sorry for throwing this in suddenly, I was out all last week and then wanted to get it merged so that @dreynaud could join in working on this. I really should have broken this up into 2 commits, one of which was the restructuring of `orca-test` and one the refactoring of the Kayenta work. We've been asked to pick up work on Kayenta support in Orca (& Spinnaker generally) so people can start using it by the end of Q1. I was told you guys didn't have the resources to commit to that timeframe. My first task was to understand what had happened so far and see how it aligned with our intentions. Basically we want to build on what's here to (optionally) deploy baseline and canary (control & experiment) server group with the intention that we provide a unified canary experience instead of the existing choice between "Canary" and "ACA task" stages which users have reported confusion over.

The changes I've made are basically to add classes representing the canary stage context and associated sub-types so we can take advantage of `Stage.mapTo` which saves a lot of decoding and wrangling of individual fields. There is a slight change (suggested by Michael Graff) in that the start and end time is now not individually specified per scope as the code was assuming they had to be the same anyway. Other than that it should be the same. Hopefully the `KayentaCanaryContext` class is clear in terms of how it expects the data to be structured.

The new work is around integrating an optional automated deployment, cleanup and cancellation and optional scaling operations in between canary intervals.

duftler:
Thanks for the additional details. We're excited to see work starting on the additional orchestration pieces, and are completely in favor of a unified canary experience.

Regarding this change:

> "the start and end time is now not individually specified per scope"

I haven't read through the pr in detail, but curious whether this change applies to the configuration of the stage in the ui, or to the canary execution request sent from orca to kayenta?

Asking since if one of those contracts changed, I'd think there needs to be a companion PR in one of those other repo's.

Also relevant here is that we've started adding additional folks to the repo's (one at a time), and shared with them an overview doc describing how to configure and bring up the new components, perform end-to-end experiments, and add new metrics service integrations (have just shared that doc with you). It includes snippets of stage configuration and full canary execution requests. Does that doc need to be updated as well?

PR #2035, opened 2018-03-01
refactor(metrics): Use an Atlas timer instead of gauges to track lag

robfletcher:


gradle.properties
```diff
@@ -18,4 +18,4 @@ junitVersion=1.0.2
 jupiterVersion=5.0.2
 junitLegacyVersion=4.12.0
 spekVersion=1.1.5
-keikoVersion=2.5.0
+keikoVersion=2.5.1
```
robfletcher: fixes lag being in the wrong unit

PR #2037, opened 2018-03-02
fix(clouddriver): Fix NPE while searching for ancestor stage

robzienert:


lwander:
should we cherry pick this into 1.6?

ajordens:
Can you shed more light on what the pipeline looked like where this happened?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ApplySourceServerGroupCapacityTask.groovy
```diff
@@ -158,11 +158,17 @@ class ApplySourceServerGroupCapacityTask extends AbstractServerGroupTask {
       ancestorStage.context.containsKey("sourceServerGroupCapacitySnapshot")
     }
 
-    return ancestors ? ancestors[0] : stage.execution.stages.find {
+    Stage ancestor = (ancestors != null && !ancestors.isEmpty()) ? ancestors[0] : stage.execution.stages.find {
       // find a synthetic sibling w/ 'sourceServerGroupCapacitySnapshot' in the event of there being no suitable
```
ajordens: The return `ancestors ?` should have been doing the same as the revised not null and not empty check?

robzienert:
@ajordens this isn't making pipelines fail, it appears that's the NPE is being caught and logged, but each of the executions I checked all completed successfully. It does happen super often though.

I changed the null/empty check because groovy truth isnt terribly obvious to me: it's not the cause.

ajordens:
Did the pipeline not have the source capacity snapshot? Should it have?

If things are failing in a way that is leaving the newly created server groups with min = desired pinned, that seems like a larger issue worth tackling.

PR #2042, opened 2018-03-07
fix(artifacts): Don't replace artifacts in trigger after resolution

wjoel:
Artifacts that have been received in a trigger should be kept in the pipeline context, not replaced by received or resolved pipeline artifacts.

While echo trigger pipelines, and can include `receivedArtifacts` as in https://github.com/spinnaker/echo/blob/dfe3a4e3ede09039754a421d469c67172a6bc472/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/DockerEventMonitor.java#L100, triggers from deck only include the trigger itself.

This pull request changes orca to not just overwrite artifacts received from the trigger itself, and instead consider them as equally important as the (pipeline concept of) `receivedArtifacts`.


wjoel:
Ping @spinnaker/reviewers

ezimanyi:
Looks good to me too!

PR #2043, opened 2018-03-08
fix(rollback): Better support for automatic rollbacks on RRB deploys

ajordens:
This PR will correctly unpin the source server group when an RRB
deployment fails and is automatically rolled back.

The previous server group is guaranteed to exist so the automatic
rollback _only_ needs to disable the new server group.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.orca.clouddriver.pipeline.providers.aws.CaptureSour
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.DisableServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.EnableServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerGroupStage
+import com.netflix.spinnaker.orca.kato.pipeline.strategy.Strategy
 import com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategy
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner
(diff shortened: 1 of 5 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
```diff
@@ -54,9 +55,24 @@ class ExplicitRollback implements Rollback {
...
+    Map disableServerGroupContext = new HashMap(parentStage.context)
+    disableServerGroupContext.serverGroupName = rollbackServerGroupName
+    def disableServerGroupStage = newStage(
+      parentStage.execution, disableServerGroupStage.type, "disable", disableServerGroupContext, parentStage, SyntheticStageOwner.STAGE_AFTER
+    )
 
+    def parentDeployStage = getParentDeployStage(parentStage)
+    def parentDeployStageStrategy = parentDeployStage?.context?.strategy as String
+    if (Strategy.fromStrategy(parentDeployStageStrategy) == Strategy.ROLLING_RED_BLACK) {
+      // no need to do anything but disable the newly deployed (and failing!) server group when dealing with a
```
ajordens: this _only_ kicks in when we're automatically rolling back something that was created as part of a rolling red/black ... they'll be a check box on the deploy configuration that enables automatic rollback

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -165,8 +165,20 @@ class CompleteStageHandler(
...
       builder.buildAfterStages(this, onFailureStages) { it: Stage ->
+        if (it.requisiteStageRefIds.any { notStartedSyntheticRefIds.contains(it) }) {
+          /*
+           * Avoid having any NOT_STARTED synthetic stage as a requisiteStageRefId as they will be subsequently removed
+           * from the execution.
+           *
+           * Opting for an emptySet() as the StartStageHandler will reject any attempt to start a stage with an
+           * upstream failure.
+           */
+          it.requisiteStageRefIds = emptySet()
```
ajordens: Alternatively I suppose `StartStage` message could be altered to include a flag indicating that upstream failures should be ignored.

This _seemed_ simpler and fit with the way that `Stage.ancestors()` is working. 

It fixes a problem wherein an upstream stage output was not being found because it was attempting to navigate using a `requisiteStageRefId` that did not exist versus using `parentStageId`. 

Thoughts? I'm not against modifying the `StartStageHandler` or equivalent if we think that's cleaner.

ajordens: Another alternative would have been to have `Stage.ancestors()` filter out any `requisiteStageRefId` that reference something that no longer exists.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
```diff
@@ -54,9 +55,24 @@ class ExplicitRollback implements Rollback {
   ApplySourceServerGroupCapacityStage applySourceServerGroupCapacityStage
 
   @JsonIgnore
-  def List<Stage> buildStages(Stage parentStage) {
-    def stages = []
+  List<Stage> buildStages(Stage parentStage) {
```
robfletcher: Is this method part of an interface?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
robfletcher: It seems like filtering it would make sense. What if there are multiple requisite stage ids some of which are SUCCEEDED? You'd still remove all of them? I guess it makes no practical difference but it seems like we lose information.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
ajordens: It is ... not sure why it didn't have an `@Override`

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
ajordens: That's fair ... I did consider it initially and will flip back to filtering. Right now we don't have synthetic stages with multiple requisites (afaik) but that might not always be the case.

PR #2045, opened 2018-03-09
debug(scale down): add some logging to try to identify why scale down…

robfletcher:
… doesn't always happen


ajordens:
Can we get the `executionId` included in these log messages

PR #2047, opened 2018-03-09
feat(core) expose stage to getDynamicBackoffPeriod

asher:
- refactor WaitTask to no longer busy poll during the wait period

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/WaitTask.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/WaitTask.java
@@ -25,6 +25,7 @@ import org.springframework.stereotype.Component;
 
 import javax.annotation.Nonnull;
 import java.time.Clock;
+import java.time.Duration;
 import java.time.Instant;
 
 import static com.netflix.spinnaker.orca.ExecutionStatus.RUNNING;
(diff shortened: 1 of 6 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/DynamicRetryableTask.java
```diff
@@ -0,0 +1,16 @@
+package com.netflix.spinnaker.orca;
+
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import com.netflix.spinnaker.orca.pipeline.model.Task;
+
+import java.time.Duration;
+
+public interface DynamicRetryableTask extends com.netflix.spinnaker.orca.Task {
```
robzienert: Looks like your imports are wonky.

ajordens:
TBH I'm not a fan of both `RetryableTask` and `DynamicRetryableTask`. 

IMO It seems this could just be folded into `RetryableTask.getDynamicBackoffPeriod()`.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/WaitTask.java
```diff
@@ -60,11 +66,32 @@
...
-  @Override public long getTimeout() {
+  @Override
+  public long getDynamicBackoffPeriod(Stage stage, Duration taskDuration) {
+    WaitStage.WaitStageContext context = stage.mapTo(WaitStage.WaitStageContext.class);
+
+    if (context.isSkipRemainingWait()) {
+      return 0L;
+    }
+
+    if (context.getStartTime() != null && context.getWaitDuration() != null) {
```
ajordens: Probably worth some inline commenting on what you're looking to achieve here

asher:
@ajordens re: folding into `RetryableTask.getDynamicBackoffPeriod()`, was concerned about changing the signature there and potentially breaking third party stages if any implement that method. If you think we're in the clear there, will do.

PR #2049, opened 2018-03-12
feat(artifacts): Use artifacts for Bake->Deploy GCE stages

ezimanyi:
Update the Bake stage to emit the baked image as an artifact, and the Deploy stage to optionally deploy an specified artifact instead of searching upstream for the most recent Bake or Find Artifact stage.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CompletedBakeTask.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CompletedBakeTask.groovy
@@ -40,7 +40,11 @@ class CompletedBakeTask implements Task {
     def bakeStatus = stage.context.status as BakeStatus
     def bake = bakery.lookupBake(region, bakeStatus.resourceId).toBlocking().first()
     // This treatment of ami allows both the aws and gce bake results to be propagated.
-    def results = [ami: bake.ami ?: bake.imageName, imageId: bake.ami ?: bake.imageName, artifact: bake.artifact ?: new Artifact()]
+    def results = [
+      ami: bake.ami ?: bake.imageName,
+      imageId: bake.ami ?: bake.imageName,
+      artifacts: bake.artifact ? [bake.artifact] : []
+    ]
     /**
      * TODO:
      * It would be good to standardize on the key here. "imageId" works for all providers.
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
@@ -62,8 +62,11 @@ class GoogleServerGroupCreator implements ServerGroupCreator, DeploymentDetailsA
     def stageContext = stage.getContext()
     String image
 
-    if (stageContext.containsKey("imageSource") && stageContext.get("imageSource") == "artifact") {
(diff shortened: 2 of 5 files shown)


ezimanyi:
The following change to deck https://github.com/spinnaker/deck/pull/4986 allows users to configure pipelines to deploy artifacts; until that change merges, this PR will cause no user-visible change.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CompletedBakeTask.groovy
```diff
@@ -40,7 +40,11 @@ class CompletedBakeTask implements Task {
     def bakeStatus = stage.context.status as BakeStatus
     def bake = bakery.lookupBake(region, bakeStatus.resourceId).toBlocking().first()
     // This treatment of ami allows both the aws and gce bake results to be propagated.
-    def results = [ami: bake.ami ?: bake.imageName, imageId: bake.ami ?: bake.imageName, artifact: bake.artifact ?: new Artifact()]
+    def results = [
+      ami: bake.ami ?: bake.imageName,
+      imageId: bake.ami ?: bake.imageName,
+      artifacts: bake.artifact ? [bake.artifact] : []
```
lwander: It's a little worrying that the bake itself can only emit one artifact... While this works here, we may want to address this in Rosco going forward.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
```diff
@@ -59,6 +58,27 @@ class GoogleServerGroupCreator implements ServerGroupCreator, DeploymentDetailsA
     return [[(ServerGroupCreator.OPERATION): operation]]
   }
 
+  private String getImage(Stage stage) {
+    def stageContext = stage.getContext()
+    String image
+
+    if (stageContext.containsKey("imageSource") && stageContext.get("imageSource") == "artifact") {
```
lwander: I think the `groovy` way to do this is `if (stageContext.imageSource == "artifact") {`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
```diff
@@ -59,6 +58,27 @@ class GoogleServerGroupCreator implements ServerGroupCreator, DeploymentDetailsA
...
+      def artifactId = stageContext.get("imageArtifactId")
```
lwander: I think you just want `stageContext.imageArtifactId` -- might also be worth checking that this isn't `null` to return a friendly failed execution message.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CompletedBakeTask.groovy
ezimanyi: Yes, I noticed that as well; I'll look into allowing a bake to produce more than one artifact when I dig more into Rosco.

avram:
In order for this to be extended to AWS, it would mainly entail adding support on the EC2 ASG side, correct? The Rosco / bake component is universal?

ezimanyi:
@avram : Yes, that's correct...Rosco should be able to correctly emit the baked image as an artifact regardless of the cloud provider used.  So to extend this to AWS would involve passing that artifact to the `AmazonServerGroupController` and having it use that image rather than the strings it currently uses to find the image to deploy.

PR #2050, opened 2018-03-13
feat(provider/kubernetes): retry manifest force cache refresh on failure

danielpeach:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
```diff
@@ -44,21 +56,25 @@ public TaskResult execute(Stage stage) {
     String account = getCredentials(stage);
     Map<String, List<String>> deployedManifests = (Map<String, List<String>>) stage.getContext().get("outputs.manifestNamesByNamespace");
 
+    ExecutionStatus status = ExecutionStatus.SUCCEEDED;
     for (Map.Entry<String, List<String>> entry : deployedManifests.entrySet()) {
       String location = entry.getKey();
-      entry.getValue().forEach(name -> {
```
danielpeach: Can't set `status` from inside lambda.

lwander:
Retry utils example here: https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java#L94

PR #2051, opened 2018-03-13
fix(artifacts): Fix deploying of specified image for gce

ezimanyi:


ezimanyi:
Added a spock test to catch this issue

sbwsg:
Thanks!

PR #2060, opened 2018-03-16
fix(queue): requeue running tasks from CancelStage

asher:
- ensures stages and tasks with long backoff times are promptly cancelled
after CancelExecution is called.

### DIFF
--- a/orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueIntegrationTest.kt
+++ b/orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueIntegrationTest.kt
@@ -422,6 +422,47 @@ class QueueIntegrationTest {
   }
 
   @Test
+  fun `terminal pipeline immediately cancels stages in other branches where tasks have long backoff times`() {
+    val pipeline = pipeline {
+      application = "spinnaker"
+      stage {
+        refId = "1"
+        type = "wait"
+        context = mapOf("waitTime" to 60)
+      }
+      stage {
+        refId = "2"
+        type = "dummy"
+        requisiteStageRefIds = emptyList()
+      }
+    }
+    repository.store(pipeline)
+
+    whenever(dummyTask.execute(argThat { refId == "2" })) doReturn TaskResult(TERMINAL)
+
+    context.runToCompletion(pipeline, runner::start, repository)
+
+    repository.retrieve(PIPELINE, pipeline.id).apply {
+      assertThat(status).isEqualTo(TERMINAL)
+      assertThat(stageByRef("2").status).isEqualTo(TERMINAL)
+      assertThat(stageByRef("1").status).isEqualTo(CANCELED)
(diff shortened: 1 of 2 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CancelStageHandler.kt
```diff
@@ -37,6 +40,21 @@ class CancelStageHandler(
 
   override fun handle(message: CancelStage) {
     message.withStage { stage ->
+      if (stage.status == RUNNING) {
```
ajordens: Lets add a javadoc comment as to why this is necessary (and particularly why it's safe to re-run -- something something we check for cancel immediately)

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CancelStageHandler.kt
```diff
@@ -37,6 +40,21 @@ class CancelStageHandler(
...
+          .filter { it.status == RUNNING }
+          .forEach {
+            queue.reschedule(
+              RunTask(
+                stage.execution.type,
+                stage.execution.id,
+                stage.execution.application,
+                stage.id,
+                it.id,
+                it.type)
```
ajordens: `)` should be on newline

orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueIntegrationTest.kt
```diff
@@ -421,6 +422,47 @@ class QueueIntegrationTest {
...
+        refId = "2"
+        type = "dummy"
+        requisiteStageRefIds = emptyList()
+      }
+    }
+    repository.store(pipeline)
+
+    whenever(dummyTask.execute(argThat { refId == "2" })) doReturn TaskResult(TERMINAL)
+
+    context.runToCompletion(pipeline, runner::start, repository)
```
asher: This actually takes at least 60 seconds to return (and the test fails) without the `CancelStageHandler` portion of this PR. With it, this test doesn't appreciably add to `orca-queue-redis` test runtime.

robfletcher: Sweet.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CancelStageHandler.kt
```diff
@@ -51,4 +87,8 @@ class CancelStageHandler(
       }
     }
   }
+
+  @Suppress("UNCHECKED_CAST")
+  private val com.netflix.spinnaker.orca.pipeline.model.Task.type
+    get() = Class.forName(implementingClass) as Class<out Task>
```
robfletcher: Idiomatic Kotlin 😎

PR #2062, opened 2018-03-16
Fix failure when re-entering aroundStages method of TargetServerGroupLinearStageSupport

robfletcher:
Spotted by @duftler.

This was caused by the new before / after stage mechanism defaulting to `aroundStages` twice. After the action has taken place it may not be able to resolve target server groups, though.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupLinearStageSupport.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupLinearStageSupport.groovy
@@ -31,9 +31,6 @@ abstract class TargetServerGroupLinearStageSupport implements StageDefinitionBui
 
   @Autowired TargetServerGroupResolver resolver
 
-  // TODO: get rid
-  @Autowired DetermineTargetServerGroupStage determineTargetServerGroupStage
-
   @Override
   String getName() {
     return type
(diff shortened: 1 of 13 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupLinearStageSupport.groovy
```diff
@@ -16,77 +16,128 @@
...
-    if (TargetServerGroup.isDynamicallyBound(stage)) {
-      return composeDynamicTargets(stage, params)
+  protected void taskGraphInternal(Stage stage, TaskNode.Builder builder) {}
+
+  @Override
+  final void beforeStages(
+    @Nonnull Stage parent,
+    @Nonnull StageGraphBuilder graph
+  ) {
+    if (isTopLevel(parent)) {
```
ajordens: worth some inline comments on what's going on here?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupLinearStageSupport.groovy
```diff
@@ -16,77 +16,173 @@
...
+      if (isDynamicallyBound(parent)) {
+        preDynamic(parent.context, graph)
+      } else {
+        preStatic(parent.context, graph)
+      }
+    }
   }
 
-  private List<Stage> composeStaticTargets(Stage stage, TargetServerGroup.Params params) {
-    if (stage.parentStageId) {
```
robfletcher: I've replicated this logic with my `isTopLevel` method but I'm concerned it's not quite correct. Shouldn't the check be whether the parent is the same type as this stage?

ajordens:
I say we give this a whirl in test ... I don't have much more context to add.

duftler:
I didn't read through the PR, but I did pull it down and the local/manual test operations that were previously failing now pass.

robfletcher:
The problem previously was that we'd end up calling the `aroundStages` when planning before stages and then again when planning after stages (following recent changes to do that just in time). That meant the target resolution stuff happened twice and the second time it might not be valid (because of what the stage did to its target). I've changed it so that everything happens in before stages. Before it was transforming its own context to deal with a single target then spawning copies to deal with each other target. It also had to do some complex shenanigoats to weave in before and after stages of _each_ target stage. Now the top level stage is a no op and just creates child before stages to deal with each target. Those child stages can worry about their own before and after stages so the whole thing is just a lot simpler.

PR #2063, opened 2018-03-18
chore(core): old pipeline cleanup support for dynomite

asher:


orca-core/src/main/java/com/netflix/spinnaker/orca/notifications/scheduling/OldPipelineCleanupPollingNotificationAgent.java
```diff
@@ -130,21 +137,43 @@ private void startPolling() {
...
+
+          // pipeline:app:foo -> foo
+          applications.addAll(result.getResult().stream().map(k -> k.split(":")[2]).collect(Collectors.toList()));
+          if ("0".equals(cursor)) {
+            break;
+          }
+        } else {
+          // Dynomite
+          final CursorBasedResult<String> lastResult = objectMapper.readValue(
+            objectMapper.writeValueAsBytes(dynoResult), new TypeReference<CursorBasedResult<String>>() {
```
asher: There has to be a better way to do this without running afoul of the rule against non-final local variables in lambdas.

asher:
Closing, will re-do after https://github.com/spinnaker/kork/pull/140 is merged

PR #2068, opened 2018-03-19
chore(*): redis delegate for activeExecutionMonitor

asher:


orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/RedisActiveExecutionsMonitor.kt
```diff
@@ -51,7 +52,7 @@ import java.util.concurrent.atomic.AtomicLong
 @Component
 class RedisActiveExecutionsMonitor(
   private val executionRepository: ExecutionRepository,
-  @Qualifier("jedisPool") private val pool: Pool<Jedis>,
+  private val redisClientDelegate: RedisClientDelegate,
```
asher: @robzienert should RCD be named here?

PR #2070, opened 2018-03-20
refactor(persistence): Unify Jedis/Dynomite Repos

asher:
Having separate `Jedis` vs. `Dynomite` execution repositories posed a number of problems around live migrations in a dual repository configuration.

This was partially due to the Dynomite repository using a different scheme for key names, in order to utilize Dynomite hash sets, I.e. `[pipeline:abc-123, pipeline:abc-123:stageIndex]` turns into `[{pipeline:abc-123}, {pipeline:abc-123}:stageIndex]`.  To support live migrations, we need to be able to find executions in either the current or previous store regardless of type or key names. This implements that logic, as well as the ability to generate key names based on the class type of each configured `RedisClientDelegate`. This should enable `Redis <---> Dynomite` migrations in either direction.


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
```diff
@@ -708,11 +779,39 @@ private void deleteInternal(JedisCommands c, ExecutionType type, String id) {
       .subscribeOn(scheduler);
   }
 
-  protected static String alljobsKey(ExecutionType type) {
+  protected ImmutableMap<String, RedisClientDelegate> fetchKey(String id) {
```
robzienert: I don't typically recommend using Apache Commons, but it's already on the classpath, so I'd definitely refactor to use `ImmutablePair` instead. That will simplify the usage of this method quite dramatically:

```java
import org.apache.commons.lang3.tuple.ImmutablePair;

ImmutablePair pair = ImmutablePair.of(pipelineKey(redisClientDelegate.id), redisClientDelegate);
// pair.getLeft(), pair.getRight()
ImmutablePair pair = ImmutablePair.nullPair();
```

robzienert: Creating a stream just to fish out the key / value from a map would be quite a lot of overhead.

asher: 👍

robzienert:
We should probably wrap these (read, minimum) calls to the delegates with a retry (as seen how I [impl'd clouddriver](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/jedis/RedisTaskRepository.java#L98-L100) and [here](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/jedis/RedisTaskRepository.java#L274-L300)).

PR #2072, opened 2018-03-21
fix(rollback): Ensure that rollback stages are injected correctly

ajordens:
This PR handles generating a `refId` for a synthetic stage that has
pre-existing siblings.

In such cases, the `refId` should be unique and offset by the number of
siblings.

All `onFailureStages()` should be run linearly otherwise you run the
risk of a parent stage being notified prematurely.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
@@ -106,7 +106,6 @@ class CreateServerGroupStage extends AbstractDeployStrategyStage {
     }
 
     if (strategySupportsRollback) {
-      // TODO: should this append after any existing stages in the graph?
       graph.add {
         it.type = rollbackClusterStage.type
         it.name = "Rollback ${stageData.cluster}"
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/graph/StageGraphBuilder.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/graph/StageGraphBuilder.java
@@ -148,11 +148,18 @@ public class StageGraphBuilder {
   }
 
   private String generateRefId() {
+    long offset = parent
+      .getExecution()
+      .getStages()
+      .stream()
+      .filter(i -> parent.getId().equals(i.getParentStageId()) && type == i.getSyntheticStageOwner())
+      .count();
+
     return format(
       "%s%s%d",
       parent.getRefId(),
       type == STAGE_BEFORE ? "<" : ">",
-      graph.nodes().size()
+      offset + graph.nodes().size()
     );
   }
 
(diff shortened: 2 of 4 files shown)


ajordens:
image

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
```diff
@@ -106,7 +106,6 @@ class CreateServerGroupStage extends AbstractDeployStrategyStage {
     }
 
     if (strategySupportsRollback) {
-      // TODO: should this append after any existing stages in the graph?
```
ajordens: The answer was yes ... I didn't see anything on `StageGraphBuilder` that would let you `connect()` to the last added node.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/graph/StageGraphBuilder.java
```diff
@@ -148,11 +148,18 @@ public void connect(@Nonnull Stage previous, @Nonnull Stage next) {
   }
 
   private String generateRefId() {
+    long offset = parent
```
ajordens: If we're adding stages to an already built execution, it's important to not re-use a `refId`.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/graph/StageGraphBuilder.java
```diff
@@ -148,11 +148,18 @@ public void connect(@Nonnull Stage previous, @Nonnull Stage next) {
...
+      .getExecution()
+      .getStages()
+      .stream()
+      .filter(i -> parent.getId().equals(i.getParentStageId()) && type == i.getSyntheticStageOwner())
```
ajordens: not necessary the most elegant, alternatives?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -165,7 +165,14 @@ class CompleteStageHandler(
 
     val graph = StageGraphBuilder.afterStages(this)
     builder().onFailureStages(this, graph)
+
     val onFailureStages = graph.build().toList()
+    onFailureStages.forEachIndexed { index, stage ->
+      if (index > 0) {
+        // all on failure stages should be run linearly
+        graph.connect(onFailureStages.get(index - 1), stage)
```
ajordens: Without this, we run the risk of one of the stages completing and notifying it's parent ... and completing the execution with another stage still running in parallel.

My expectation is that > 1 onFailure stages will run linearly (they did previously).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
robfletcher: Yeah, I think `StageGraphBuilder` could be a little easier to use for certain common scenarios.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/graph/StageGraphBuilder.java
robfletcher: Seems reasonable. I can't think of a neater way off the top of my head.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
robfletcher: I'd like to address the root issue as well (not in this PR). The execution should not complete if anything is still in a `RUNNING` state and if it does that's a bug.

PR #2073, opened 2018-03-21
chore(tests): added an extra verification to on failure stage tests

robfletcher:


robfletcher:
Meant to commit this yesterday with the other changes.

PR #2076, opened 2018-03-21
chore(tests): fix test that was slow and making real network connecti…

robfletcher:
…ons to a host we don't own

### DIFF
--- a/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/kato/tasks/quip/InstanceHealthCheckTaskSpec.groovy
+++ b/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/kato/tasks/quip/InstanceHealthCheckTaskSpec.groovy
@@ -72,13 +72,13 @@ class InstanceHealthCheckTaskSpec extends Specification {
     result.status == executionStatus
 
     where:
-    instances | responseCode | executionStatus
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/kato/tasks/quip/InstanceHealthCheckTaskSpec.groovy
```diff
@@ -90,6 +90,10 @@ class InstanceHealthCheckTaskSpec extends Specification {
     def stage = new Stage(pipe, 'instanceHealthCheck', [:])
     stage.context.instances = instances
 
+    and:
+    task.createInstanceService(_) >> instanceService
+    instanceService.healthCheck("healthCheck") >> new Response('http://foo.com', 200, 'OK', [], new TypedString("Good"))
```
robfletcher: Without this check we were making real connections to `http://foo.com:7001/healthCheck` which takes 10-20 seconds to fail. Also, that's pretty obnoxious.

ajordens: lol ok

PR #2078, opened 2018-03-22
feat(execution): Adding start time TTLs to executions and stages

robzienert:
Allows users to define a specific execution-level and/or stage-level start time TTL. This will allow teams to guarantee an upper bounds SLA for pipeline / task executions. Especially handy for the likes of Chaos Monkey.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
@@ -192,6 +192,20 @@ public class Execution implements Serializable {
     this.endTime = endTime;
   }
 
+  /**
+   * Gets the start ttl timestamp for this execution. If the execution has not
+   * started before this timestamp, the execution will immediately terminate.
+   */
+  private Long startTimeTtl;
+
+  public @Nullable Long getStartTimeTtl() {
+    return startTimeTtl;
+  }
+
+  public void setStartTimeTtl(@Nullable Long startTimeTtl) {
+    this.startTimeTtl = startTimeTtl;
+  }
+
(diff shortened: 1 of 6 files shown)


jrsquared:
v.nice

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
```diff
@@ -192,6 +192,20 @@ public void setEndTime(@Nullable Long endTime) {
     this.endTime = endTime;
   }
 
+  /**
+   * Gets the start ttl timestamp for this execution. If the execution has not
+   * started before this timestamp, the execution will immediately terminate.
+   */
+  private Long startTimeTtl;
+
+  public @Nullable Long getStartTimeTtl() {
```
robfletcher: What unit is this in? Comment suggests it's a timestamp but is it actually a duration? Could we use `java.time.Instant` or `java.time.Duration` as appropriate?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
```diff
@@ -71,6 +71,10 @@ class StartStageHandler(
           log.warn("Ignoring $message as stage is already ${stage.status}")
         } else if (stage.shouldSkip()) {
           queue.push(SkipStage(message))
+        } else if (stage.afterStartTimeTtl()) {
+          log.warn("Stage (${stage.id}) is being canceled because its start time is after TTL " +
+            "(executionId: ${message.executionId}")
```
robfletcher: Does this message end up somewhere deck can surface it to the user?

PR #2080, opened 2018-03-22
feat(metric): execution deserialization errors

asher:
This counter metric will be incremented whenever a stored execution
cannot be deserialized. This is an indicator of backwards incompatible
execution schema changes.


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
```diff
@@ -509,6 +517,11 @@ Execution retrieveOrchestrationForCorrelationId(
   }
 
   protected Execution buildExecution(@Nonnull Execution execution, @Nonnull Map<String, String> map, List<String> stageIds) {
+    Id serializationErrorId = registry
+      .createId("executions.deserialization.error")
+      .withTag("executionType", execution.getType().toString().toLowerCase())
+      .withTag("application", execution.getApplication().toLowerCase());
```
robzienert: We don't normally do the lowercasing anywhere else. I'm not sure we should deviate.

asher: Ok, consistency wins out over my love of all lowercase metrics.

PR #2083, opened 2018-03-22
feat(expressions): Adding ability to skip evaluation of specific keys

jeyrschabu:
- added an eval strategy interface that will allow for fine grained manipulation
- refactored some comments and var naming to match the rest of the class
- refactored to remove some logic selecting spel evaluator version


jeyrschabu:
Allows skipping evaluation of keys.
```
spelEvaluation:  [
   excludes : [
          [
            "key": "comments",
            "path": "."
          ]
      ]
]
```
`path`:  Currently not used but I am envisioning a way to be specific on which key needs to be skipped if more than one exists. By default all matched keys will be skipped by the evaluator

PR #2084, opened 2018-03-22
feat(orca-kayenta): move canaryExecutionRequest and application to top-level

skandragon:
This moves the canaryExecutionRequest, application, and config  to be top-level objects, to match a soon to be fixed Kayenta version.

BREAKING CHANGE:  This will require an upgrade of Kayenta after a pull request is merged there.


duftler:
Relates to: https://github.com/spinnaker/kayenta/pull/256

PR #2087, opened 2018-03-23
fix(after stages): Allow for just-in-time planning of after stages

robfletcher:
This is a bridge to the 6.19.x branch of Orca where after stages are _always_ planned just in time. During deployment this can be an issue if a 6.19.x instance starts a stage (and does not plan after stages) and a 6.18.x instance tries to complete the stage (finding no before stages or tasks and no planned after stages).

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
@@ -48,8 +48,27 @@ class CompleteStageHandler(
   override fun handle(message: CompleteStage) {
     message.withStage { stage ->
       if (stage.status in setOf(RUNNING, NOT_STARTED)) {
-        val status = stage.determineStatus()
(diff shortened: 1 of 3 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -48,8 +48,27 @@ class CompleteStageHandler(
...
-        if (status.isComplete && !status.isHalt) {
+        var status = stage.determineStatus()
+        if (status == NOT_STARTED) {
+          // Stage had no before stages or tasks so we'll see if it had any
+          // un-planned after stages
+          var afterStages = stage.firstAfterStages()
+          if (afterStages.isEmpty()) {
+            stage.planAfterStages()
+          }
+          afterStages = stage.firstAfterStages()
```
jtk54: Belongs in the `if`?

robfletcher: Yep, updated it.

PR #2088, opened 2018-03-23
fix(queue): Correctly use execution type enum for ExecutionContext

robzienert:
After refactoring Orca to use a singular `Execution` class, we never updated places we call `ExecutionContext` to correctly use the ExecutionType enum. This breaks `ByExecutionType` clouddriver routing rules, but unsure what else.


ajordens:
LGTM

It would be super sweet to get this into a release branch

robzienert:
I tried to update the base branch on this PR and it went totally crazy. Closed and will reopen targeting release branch.

PR #2091, opened 2018-03-26
refactor(core): Change type to Instant

robzienert:
Refactors the type of the just-merged startTimeTtl PR. @robfletcher did it this way so I can get more contributor points.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
@@ -28,6 +28,7 @@ import com.netflix.spinnaker.security.User;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.Serializable;
+import java.time.Instant;
 import java.util.*;
 
 import static com.netflix.spinnaker.orca.ExecutionStatus.NOT_STARTED;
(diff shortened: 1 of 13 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
```diff
@@ -196,13 +197,14 @@ public void setEndTime(@Nullable Long endTime) {
    * Gets the start ttl timestamp for this execution. If the execution has not
    * started before this timestamp, the execution will immediately terminate.
    */
-  private Long startTimeTtl;
+  private Instant startTimeTtl;
 
-  public @Nullable Long getStartTimeTtl() {
+  public @Nullable
+  Instant getStartTimeTtl() {
```
robfletcher: Not to be picky but is this really a TTL?

robzienert: What would you call it instead? It's time-to-live as far as it being valid. It's not a cache, but it becomes invalid after that timestamp.

robfletcher: I’m not bringing you solutions, I’m just bringing you problems.

PR #2093, opened 2018-03-28
fix(provider/kubernetes): cancel timed-out run job pods

lwander:
@robfletcher let me know if this is appropriate. This allows tasks to handle being timed out, which is particularly useful for hung "run job" stages.


lwander:
depends on https://github.com/spinnaker/clouddriver/pull/2462

tomaslin:
We've seen cases where the polling times out due to network connectivity issues or misconfiguration in a proxy, if I'm understanding what this code is doing correctly, it would prematurely try to stop jobs that are fine. Can we feature flag it?

tomaslin:
also on the same vein, if you're canceling the job when the job times out, there should be an option to also cancel the underlying job when the pipeline is stopped or the stage is canceled

lwander:
> We've seen cases where the polling times out due to network connectivity issues or misconfiguration in a proxy,

I'm pretty sure `TimeoutException` can only be thrown when the actual timeout is met. Do you see situations where stages are failing due to a task/stage timeout, even though the actual execution succeeded? See:

https://github.com/spinnaker/orca/pull/2093/files#diff-03be772de3ad7155b258785b71d92dbcR195 and https://github.com/spinnaker/orca/pull/2093/files#diff-03be772de3ad7155b258785b71d92dbcR217


> also on the same vein, if you're canceling the job when the job times out, there should be an option to also cancel the underlying job when the pipeline is stopped or the stage is canceled

Good idea -- I'll look into that

lwander:
Hey Tomas, can you confirm if the timeout exception is reachable for anything but an actual timeout? 

I'll defer the onCancel change to a later PR since it's a little more invovled.

katrielt:
> also on the same vein, if you're canceling the job when the job times out, there should be an option to also cancel the underlying job when the pipeline is stopped or the stage is canceled

+1 for that. IMHO, anything else will be inconsistent behavior

lwander:
Updated to handle the cancel case as well.

It's possible to hit a race condition where if you hit cancel immediately after starting the pipeline, it's possible the "delete job" request will run in clouddriver before the job is submitted, even though the execution is already killed. It seems the only way to avoid that is to block, but that's a bad pattern in the task handler.

lwander:
@tomaslin PTAL

PR #2095, opened 2018-03-28
Fix after stage logic

robfletcher:


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -54,40 +54,23 @@ class CompleteStageHandler(
       if (stage.status in setOf(RUNNING, NOT_STARTED)) {
         var status = stage.determineStatus()
         try {
-          if (status == NOT_STARTED) {
-            // Stage had no before stages or tasks so we'll see if it had any
-            // un-planned after stages
+          if (status == NOT_STARTED || (status.isComplete && !status.isHalt)) {
```
robfletcher: I merged these two branches that were doing almost exactly the same thing.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -54,40 +54,23 @@ class CompleteStageHandler(
...
-                queue.push(StartStage(message, it.id))
-              }
-
+            if (afterStages.isNotEmpty() && afterStages.any { it.status == NOT_STARTED }) {
+              afterStages
+                .filter { it.status == NOT_STARTED }
+                .forEach { queue.push(StartStage(message, it.id)) }
               return@withStage
-            } else {
+            } else if (status == NOT_STARTED) {
```
robfletcher: This is the only bit that was not in common between the two branches

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -54,40 +54,23 @@ class CompleteStageHandler(
...
             if (afterStages.isEmpty()) {
               stage.planAfterStages()
               afterStages = stage.firstAfterStages()
             }
-            if (afterStages.isNotEmpty()) {
-              afterStages.forEach {
-                queue.push(StartStage(message, it.id))
-              }
-
+            if (afterStages.isNotEmpty() && afterStages.any { it.status == NOT_STARTED }) {
```
robfletcher: I feel like it was bad that we'd run these again regardless.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -54,40 +54,23 @@ class CompleteStageHandler(
...
-            // check to see if this stage has any unplanned synthetic after stages
-            var afterStages = stage.firstAfterStages()
-            if (afterStages.isEmpty()) {
-              stage.planAfterStages()
-
-              afterStages = stage.firstAfterStages()
-              if (afterStages.isNotEmpty()) {
-                afterStages.forEach {
-                  queue.push(StartStage(message, it.id))
-                }
```
robfletcher: The fact that this `queue.push` call was inside the if statement on line 79 was what's been breaking Orca deploys. Old instance pre-planned after stages, new instance therefore didn't run them and just no-opped.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandlerTest.kt
```diff
@@ -721,6 +721,42 @@ object CompleteStageHandlerTest : SubjectSpek<CompleteStageHandler>({
...
           })
         }
+
+        it("does not do anything silly like running the after stage again") {
+          verify(queue, never()).push(isA<StartStage>())
+        }
+      }
+    }
+
+    describe("when all after stages have completed successfully") {
```
robfletcher: Test to ensure we don't re-run stages.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -103,13 +86,12 @@ class CompleteStageHandler(
...
           stage.status = TERMINAL
           stage.endTime = clock.millis()
-          repository.storeStage(stage)
         }
 
         stage.includeExpressionEvaluationSummary()
         repository.storeStage(stage)
 
-        if (stage.status in listOf(SUCCEEDED, FAILED_CONTINUE)) {
+        if (stage.status in listOf(SUCCEEDED, FAILED_CONTINUE, SKIPPED)) {
```
robfletcher: This only affects stages that have no synthetics _or_ tasks (of which there are zero, so it's theoretical). Still, it's not right they'd zombie.

PR #2096, opened 2018-03-29
feat(webhooks): support webhook artifacts

ethanfrogers:
as mentioned in spinnaker/spinnaker#2330, webhook stages now support
producing artifacts. this expects that there is a top-level `artifacts`
key in the response. these artifacts are then bound and available to
downstream stages.

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/pipeline/WebhookStage.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/pipeline/WebhookStage.groovy
@@ -24,6 +24,7 @@ import com.netflix.spinnaker.orca.pipeline.TaskNode
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import com.netflix.spinnaker.orca.webhook.tasks.CreateWebhookTask
 import com.netflix.spinnaker.orca.webhook.tasks.MonitorWebhookTask
+import com.netflix.spinnaker.orca.pipeline.tasks.artifacts.BindProducedArtifactsTask
 import groovy.util.logging.Slf4j
 import org.springframework.stereotype.Component
 
@@ -39,6 +40,10 @@ class WebhookStage implements StageDefinitionBuilder, RestartableStage, Cancella
     if (waitForCompletion?.toBoolean()) {
       builder
         .withTask("monitorWebhook", MonitorWebhookTask)
+    }
+    if (stage.context.containsKey("expectedArtifacts")) {
+      builder
+        .withTask(BindProducedArtifactsTask.TASK_NAME, BindProducedArtifactsTask.class);
     }
   }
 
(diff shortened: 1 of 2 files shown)


ethanfrogers:
@lwander PTAL

some thoughts, considerations:
1. right now this only works for non-monitored webhooks. i'm not entirely sure where this should be added in `MonitorWebhookTask`. perhaps @jervi could provide some insight.
2. this assumes a top-level `artifacts` key will be returned in the response. i wonder if that's a reasonable assumption to make of is we should make it a bit more flexible and allow a `artifactsJsonPath` option like `statusUrlJsonPath`. thoughts?

wjoel:
@ethanfrogers for 2, support for Jinja templates to do a transformation might be a good solution as we already use them for pubsub, https://www.spinnaker.io/reference/artifacts/in-pipelines/#artifacts-in-trigger-payloads (and, maybe, also for incoming webhooks, since we have `TemplateBasedArtifactExtractor` in echo, but I don't know if there's any documentation for that and I haven't used it myself yet).

ethanfrogers:
@wjoel oh wow, that's interesting. it'd be cool to see an example of that in action and how it's configured. i agree though, some way to transform a payload into the correct shape would make this ultimately flexible.

however, since the use of a top-level `artifacts` key is documented i feel better about Jinja being an extension of this rather than a replacement.

lwander:
Jinja would definitely be a extension of this -- see how it's configured/used here: https://github.com/spinnaker/echo/pull/197

You don't have to add that as a part of this PR, but I would keep it in mind

ethanfrogers:
thanks @lwander, that was helpful. 

i'll open an issue so that we can track that addition and have discussion around implementation.

ethanfrogers:
spinnaker/spinnaker#2594

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/pipeline/WebhookStage.groovy
```diff
@@ -40,6 +41,10 @@ class WebhookStage implements StageDefinitionBuilder, RestartableStage, Cancella
       builder
         .withTask("monitorWebhook", MonitorWebhookTask)
     }
+    if (stage.context.containsKey("expectedArtifacts")) {
+      builder
+        .withTask(BindProducedArtifactsTask.TASK_NAME, BindProducedArtifactsTask.class);
```
sbwsg: LGTM!

Java noob question: why does `BindProducedArtifactsTask.class` need to be passed here instead whereas `MonitorWebhookTask` is passed above without `.class`?

ethanfrogers: Java (all flavors) noob answer: because copypasta. 😓 i was referencing the usage in RunJob for this.

ethanfrogers: i can remove it to keep it consistent if need be.

ethanfrogers:
also, i can't merge to this repo so someone is gonna have to do it for me, heh.

sbwsg:
thanks, done!

PR #2105, opened 2018-03-30
fix(after stages): handle stage with pre-planned after stages

robfletcher:


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -54,7 +54,7 @@ class CompleteStageHandler(
       if (stage.status in setOf(RUNNING, NOT_STARTED)) {
         var status = stage.determineStatus()
         try {
-          if (status == NOT_STARTED || (status.isComplete && !status.isHalt)) {
+          if (status in setOf(RUNNING, NOT_STARTED) || (status.isComplete && !status.isHalt)) {
```
robzienert: I don't understand why this is changing, help.

robzienert: Nevermind.

robfletcher: `determineStatus` didn't return `RUNNING` before.

PR #2106, opened 2018-03-30
chore(canary-v2): Replace region with location.

duftler:


duftler:
Relates to https://github.com/spinnaker/kayenta/pull/270 and https://github.com/spinnaker/deck-kayenta/pull/275

PR #2108, opened 2018-04-02
feat(rolling push): respect specified order of terminations

emjburns:
If the order of instances is specified in rolling push (using `"order" == "given"`), then let's terminate instances in that order.

Let's also test to make sure that the order we terminate in is the order we expect.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/DetermineTerminationCandidatesTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/DetermineTerminationCandidatesTask.groovy
@@ -46,8 +46,8 @@ class DetermineTerminationCandidatesTask implements Task {
     def terminationInstancePool = knownInstanceIds
     if (stage.context.termination?.instances) {
       terminationInstancePool = knownInstanceIds.intersect(stage.context.termination?.instances)
-      if (stage.context.termination?.order == 'given') {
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/DetermineTerminationCandidatesTask.groovy
```diff
@@ -46,6 +46,9 @@ class DetermineTerminationCandidatesTask implements Task {
     def terminationInstancePool = knownInstanceIds
     if (stage.context.termination?.instances) {
       terminationInstancePool = knownInstanceIds.intersect(stage.context.termination?.instances)
+      if (stage.context.termination?.order == 'given') {
```
ajordens: minor but we don't need the `termination?.order` ... we've already guaranteed `termination` is non-null in this block.

same in the `.sort {}` just below

ajordens: minor but we don't need the `termination?.order` ... we've already guaranteed `termination` is non-null in this block.

same in the `.sort {}` below

emjburns: ah yes fixing both!

PR #2112, opened 2018-04-03
fix(queue): `FAILED_CONTINUE` tasks should not halt stage execution

ajordens:
The stage should continue executing even if a task completes with
the `FAILED_CONTINUE` status.


orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteTaskHandlerTest.kt
```diff
@@ -50,137 +50,83 @@ object CompleteTaskHandlerTest : SubjectSpek<CompleteTaskHandler>({
...
-      action("the handler receives a message") {
-        subject.handle(message)
-      }
-
-      it("updates the task state in the stage") {
-        verify(repository).storeStage(check {
-          it.tasks.first().apply {
-            assertThat(status).isEqualTo(SUCCEEDED)
-            assertThat(endTime).isEqualTo(clock.millis())
+  setOf(SUCCEEDED, FAILED_CONTINUE).forEach { successfulStatus ->
```
ajordens: TBH this is the only change -- everything else is fallout from indenting this block

ajordens:
Noticed when doing some local testing of the `AddServerGroupEntityTagsTask` that subsequent tasks were not being executed after `FAILED_CONTINUE` was returned.

I'm not sure we return that status very often (`AddServerGroupEntityTagsTask` and `CompleteCanaryTask` could do it).

robfletcher:
I think historically we never returned that status directly, it was only applied after the fact when dealing with the end of the stage.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteTaskHandler.kt
```diff
@@ -49,7 +49,7 @@ class CompleteTaskHandler(
       } else {
         repository.storeStage(mergedContextStage)
 
-        if (message.status != SUCCEEDED) {
+        if (!setOf(SUCCEEDED, FAILED_CONTINUE).contains(message.status)) {
```
robfletcher: `message.status !in setOf(SUCCEEDED, FAILED_CONTINUE)`

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteTaskHandlerTest.kt
robfletcher: Wish GitHub could ignore whitespace changes like IntelliJ can.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteTaskHandler.kt
ajordens: done

PR #2121, opened 2018-04-04
feat(artifacts): Orca sends an an artifact to clouddriver

ezimanyi:
When a deploy description sets the image to deploy to an artifact, Orca should send the entire artifact to clouddriver, rather than just the name of the image.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
```diff
@@ -49,34 +50,39 @@ class GoogleServerGroupCreator implements ServerGroupCreator, DeploymentDetailsA
       operation.credentials = operation.account
     }
 
-    operation.image = operation.image ?: getImage(stage)
-
-    if (!operation.image) {
-      throw new IllegalStateException("No image could be found in ${stage.context.region}.")
+    if (stage.context.imageSource == "artifact") {
+      operation.imageSource = "ARTIFACT"
```
lwander: This upper/lowercase conversion is a little odd

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
```diff
@@ -49,34 +50,39 @@ class GoogleServerGroupCreator implements ServerGroupCreator, DeploymentDetailsA
...
+      operation.imageArtifact = getImageArtifact(stage)
+    } else {
+      operation.imageSource = "STRING"
```
lwander: Is string the right noun? What about "legacy"?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
```diff
@@ -49,34 +50,39 @@ class GoogleServerGroupCreator implements ServerGroupCreator, DeploymentDetailsA
...
     return [[(ServerGroupCreator.OPERATION): operation]]
   }
 
-  private String getImage(Stage stage) {
+  private Artifact getImageArtifact(Stage stage) {
     def stageContext = stage.getContext()
+
+    def artifactId = stageContext.imageArtifactId as String
+    if (artifactId == null) {
+      throw new IllegalStateException("Image source was set to artifact but no artifact was specified.")
```
lwander: I think `IllegalArgumentException` is a little more 'correct'

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
ezimanyi: I chose 'STRING' because we're effectively giving clouddriver the image to deploy either as (1) a string that it needs to resolve in the current project or (2) an artifact.  See https://github.com/spinnaker/clouddriver/blob/935e4d23c22f7832488ad9404b076baad6501560/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/description/BaseGoogleInstanceDescription.groovy#L46

It's also not completely clear this is legacy, as we'll go into the `else` block when manually starting a cluster and passing an image from the drop-down.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
ezimanyi: Makes sense, I'll change this

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
ezimanyi: Makes sense, I'll change this

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/GoogleServerGroupCreator.groovy
ezimanyi: Agreed---ideally we'd set upper-case constants in the pipeline definition in deck, but unfortunately deck sets these as lower-case.  Given that we can't change that to upper-case without breaking existing pipelines or writing a pipeline migrator, for now leaving this as is.

PR #2122, opened 2018-04-04
fix(clouddriver): Refactor TargetReferenceLinearStageSupport to use before/afterStages

robzienert:
With the recent refactor of splitting `aroundStages` to `[before|after]Stages`, the `destroyAsg` stage started failing because `TargetReferenceLinearStageSupport` uses `aroundStages`, which would try to find the target server group once before the stage, and again after the stage. In the case of a destroy ASG operation, the stage would delete the ASG successfully, but fail to continue because it would try to re-resolve the target server group, but it wouldn't be there anymore.

This refactor updates the functionality to correctly use before/afterStages. Furthermore, since all of this code is still deprecated, I added a log statement so we can track down who is using this and start moving people off of it in hopes of a big PR full of red.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/TargetReferenceLinearStageSupport.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/TargetReferenceLinearStageSupport.groovy
@@ -19,13 +19,17 @@ package com.netflix.spinnaker.orca.kato.pipeline.support
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.orca.kato.pipeline.DetermineTargetReferenceStage
 import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder
+import com.netflix.spinnaker.orca.pipeline.graph.StageGraphBuilder
 import com.netflix.spinnaker.orca.pipeline.model.Stage
-import com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner
+import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
-import static com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder.newStage
 
+import javax.annotation.Nonnull
+
+@Slf4j
 @Deprecated
 abstract class TargetReferenceLinearStageSupport implements StageDefinitionBuilder {
+
   @Autowired
   ObjectMapper objectMapper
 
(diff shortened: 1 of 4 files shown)


robfletcher:
I think the problem with hoping to get rid of these stages is that OSS-land uses them.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/TargetReferenceLinearStageSupport.groovy
```diff
@@ -36,33 +40,55 @@ abstract class TargetReferenceLinearStageSupport implements StageDefinitionBuild
...
+    }
+  }
+
+  @Override
+  void afterStages(@Nonnull Stage parent, @Nonnull StageGraphBuilder graph) {
+    parent.resolveStrategyParams()
+
+    List<Stage> stages = composeTargets(parent, graph)
+    for (int i = 1; i < stages.size(); i++) {
+      graph.connect(stages.get(i - 1), stages.get(i))
```
robfletcher: where does it add `stages[0]`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/TargetReferenceLinearStageSupport.groovy
```diff
@@ -36,33 +40,55 @@ abstract class TargetReferenceLinearStageSupport implements StageDefinitionBuild
...
     if (descriptionList.empty) {
       throw new TargetReferenceNotFoundException("Could not find any server groups for specified target")
     }
     def first = descriptionList.remove(0)
     stage.context.putAll(first)
 
-    if (descriptionList.size()) {
-      return descriptionList.collect {
-        newStage(stage.execution, this.type, this.type, it, stage, SyntheticStageOwner.STAGE_AFTER)
+    return descriptionList.collect { description ->
```
robfletcher: do we actually still need to return anything here or can the method be `void`?

robzienert: We do, for the reason of being able to connect up the stages correctly, unless we don't really care about connects (was just keeping with backwards compat).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/TargetReferenceLinearStageSupport.groovy
robzienert: Updated to check for `size() > 1` before hitting this loop.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/TargetReferenceLinearStageSupport.groovy
robfletcher: Shouldn't it connect them inside of this method?

robzienert:
Do we actually know that OSS land uses them? The only people that would've been using them at this point are people who have written clients against our API, which I suspect is a very minimal case. I'd think we could get away with first disabling `DestroyAsgStage`, `DisableAsgStage`, `EnableAsgStage` and `ModifyScalingProcessStage`. If anyone complains, we can advise them on how to move off, but in say... 6 months actually delete the classes from Orca.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/TargetReferenceLinearStageSupport.groovy
robzienert: Mm... yeah. Probably should. Will change.

PR #2125, opened 2018-04-05
chore(kayenta): use new API for synthetic stages

robfletcher:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/cluster/RollbackClusterStage.java
```diff
@@ -89,21 +91,19 @@ public void afterStages(@Nonnull Stage parent, @Nonnull StageGraphBuilder graph)
       // propagate any attributes of the parent stage that are relevant to this rollback
       context.putAll(propagateParentStageContext(parent.getParent()));
 
-      Stage rollbackStage = graph.add((it) -> {
+      graph.append((it) -> {
```
robfletcher: This was incorrect. Using `add` would create a stage with no `requisiteStageRefIds` when what's intended here is to run each region in serial. Spotted it when updating stages to use the new `append` method.

ajordens: 👍

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/cluster/RollbackClusterStage.java
```diff
@@ -49,13 +49,15 @@ public RollbackClusterStage(RollbackServerGroupStage rollbackServerGroupStage,
   }
 
   @Override
-  public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
+  public void taskGraph(
```
ajordens: mr line breaky over here!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/graph/StageGraphBuilder.java
```diff
@@ -132,6 +134,28 @@ public void connect(@Nonnull Stage previous, @Nonnull Stage next) {
...
+   */
+  public @Nonnull Stage append(
+    @Nonnull Consumer<Stage> init
+  ) {
+    if (lastAdded == null) {
+      return add(init);
+    } else {
+      return connect(lastAdded, init);
+    }
+  }
```
robzienert: `append` is a great addition. Will update my other PR to use this instead.

robfletcher: Yeah, I found myself doing exactly that logic in various stage builders. Now you only need to keep variables of stages if you want to do something more complex like a fork/join.

PR #2127, opened 2018-04-06
fix(queue): `FAILED_CONTINUE` tasks should not halt stage execution

ajordens:
This PR improves upon spinnaker/orca#2112 by only executing subsequent
tasks when a task _explicitly_ completes with the `FAILED_CONTINUE`
status.

An execution being flagged `"continuePipeline": true` is not enough to
cause subsequent tasks to execute if an unsuccessful status is
merely overridden to `FAILED_CONTINUE`.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
@@ -99,7 +99,7 @@ class RunTaskHandler(
                 }
                 TERMINAL                             -> {
                   val status = stage.failureStatus(default = result.status)
-                  queue.push(CompleteTask(message, status))
+                  queue.push(CompleteTask(message, status, result.status))
                   trackResult(stage, taskModel, status)
                 }
                 else                                 ->
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/messages.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/messages.kt
@@ -73,10 +73,22 @@ data class CompleteTask(
   override val application: String,
   override val stageId: String,
   override val taskId: String,
-  val status: ExecutionStatus
+  val status: ExecutionStatus,
+  val originalStatus: ExecutionStatus?
 ) : Message(), TaskLevel {
   constructor(source: TaskLevel, status: ExecutionStatus) :
-    this(source.executionType, source.executionId, source.application, source.stageId, source.taskId, status)
+    this(source, status, status)
+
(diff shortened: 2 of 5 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/messages.kt
```diff
@@ -73,10 +73,22 @@ data class CompleteTask(
   override val application: String,
   override val stageId: String,
   override val taskId: String,
-  val status: ExecutionStatus
+  val status: ExecutionStatus,
+  val originalStatus: ExecutionStatus = status
```
ajordens: Tracking the original / unmodified task status

I don't particularly like this _but_ I'm not sure it's avoidable given the calculation happening in `RunTaskHandler`. It feels like we could migrate to a point where the logic to override a task status is in the `CompleteTaskHandler` instead.

Can you think of a more elegant way to do this @robfletcher?

ajordens:
```

File: orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/messages.kt (hunk 1 above)
Comment (robfletcher) [reply orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/messages.kt]: I like the idea of having `CompleteTaskHandler` decide. Could we then just use the `isComplete` on `ExecutionStatus` to decide whether to run the next task?

File: orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt [hunk 2]
```diff
@@ -85,7 +85,7 @@ class RunTaskHandler(
                 }
                 TERMINAL, CANCELED                   -> {
                   val status = stage.failureStatus(default = result.status)
-                  queue.push(CompleteTask(message, status))
+                  queue.push(CompleteTask(message, status, result.status))
```
robfletcher: Yeah, it feels like the `failureStatus` calculation would make more sense in `CompleteTaskHandler`.

ajordens: Agreed ... I think we still need to do a pass (like this) where we're supplying the explicit value separately to remain backwards compatible. 

Then we can change `CompleteTaskHandler` to only look at the original status ... then we can set status == originalStatus ... then we can remove originalStatus.

A pain but I think this will migrate us without significant issue.

ajordens: Or maybe we can just take a shortcut and do it all in one swoop.

robfletcher: Oh, I see. Didn't realize that's what you intended. Given the infrequency of stages returning that is it really worth the effort?

PR #2128, opened 2018-04-06
feat(bake/manifest): bake manfest stage

lwander:
image


lwander:
https://github.com/spinnaker/spinnaker/issues/2636

PR #2129, opened 2018-04-06
fix(queue): Override RunTask message ack timeout

robzienert:
Depends on https://github.com/spinnaker/keiko/pull/31


ajordens:
Probably a _tad_ high but I'm not sure it'll break anything (if it does than the broken bits are worth a better look as I'm reluctant to have anything rely on a `RunTask` being re-queued immediately).

robzienert:
May as well adjust it now if we think it's high; 10 min? 15?

ajordens:
I'd be happy with 10min

PR #2130, opened 2018-04-06
fix(queue): Pass original task status along for `CANCELED`

ajordens:
Similar to `TERMINAL`, we should pass the original task completion
status in the `CompleteTask` message whenever it has been overridden.


ajordens:
The `CANCELED` check just went in today and was missed in my earlier PR.

PR #2133, opened 2018-04-09
fix(queue): reschedule child pipeline when parent canceled

asher:


emjburns:
LGTM, unless there's some problem I'm not thinking of with a stage having direct access to the queue.

robzienert:
I'm not particularly excited to sprinkle the queue around the codebase. Why not include this logic inside of `CancelStageHandler`? https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CancelStageHandler.kt#L83

In doing so, we'll keep all queue things localized and all cancelable stages will benefit from this prompt cancelation logic.

emjburns:
It seems that the  `pipeline stage` is a little different than other stages, because it needs to tell the pipeline that it's running to reschedule all of its messages. Those are tied to a different execution id. Is there some way to do that in the 'cancelStageHandler` that I'm missing?

asher:
Could do this in CancelStageHandler, but would have to special case based on i.e. `stage.type.equals("pipeline")` as PipelineStage isn't generic. So either spillover of queue logic into orca-front50, or of orca-front50 into orca-queue. I'm into whichever is least offensive :)

robzienert:
I would lean more towards putting the logic in the queue module. Keeping all of the logic about how the queue works, including its edge cases, inside of the same module strikes me as much more maintainable long-term.

asher:
Sounds good, going to close this PR and open a new one with the logic moved.

PR #2135, opened 2018-04-09
fix(expressions): Better handling for composite spel

jeyrschabu:
- broke out the main transform method into type specific ones
- small cosmetic changes: corrected inconsitency var names including allowUnknownKeys
- added a way to stringify certain types before returning


jeyrschabu:
- This also fixes the issue where `${stage('stageName')['status']}` required a `toString()`.  Also fixes an issue where a composite expression like `a.${ENV}.b.${ENV}.c` returned `a..b..c` if `${ENV}` doesn't resolve to a value. 
- This change will ensure `a.${ENV}.b.${ENV}.c` returns `a.${ENV}.b.${ENV}.c`.
- It also made sense to break transform methods into type specific ones, the generic method had some smells making it hard to reason about.

PR #2139, opened 2018-04-11
fix(queue): reschedule child pipeline when parent canceled

asher:


orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueIntegrationTest.kt
```diff
@@ -424,6 +426,47 @@ class QueueIntegrationTest {
...
+    }
+    repository.retrieve(PIPELINE, childPipeline.id).apply {
+      assertThat(stageByRef("wait").status == RUNNING)
+    }
+
+    context.runToCompletion(childPipeline, runner::start, repository)
+
+    repository.retrieve(PIPELINE, childPipeline.id).apply {
+      assertThat(isCanceled).isTrue()
+      assertThat(stageByRef("wait").wasShorterThan(10000L)).isTrue()
```
asher: This assert fails if `queue.push(RescheduleExecution(child))` is removed from `CancelStageHandler`.

emjburns:
LGTM. Nice tests

PR #2140, opened 2018-04-11
feat(artifacts): Parse Jenkins job properties for artifacts

avram:
Jenkins job stages support producing artifacts; artifacts will be parsed out from the existing properties file output that Jenkins jobs can declare. A key `artifacts` will be parsed for artifacts that match the declared artifacts.

This is the Orca work for https://github.com/spinnaker/spinnaker/issues/2643

Please don't merge yet-- this is my first Spinnaker PR and I need to validate this more; this current code is just a PoC and seems too easy.

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/JenkinsStage.groovy
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/JenkinsStage.groovy
@@ -25,6 +25,7 @@ import com.netflix.spinnaker.orca.igor.tasks.StopJenkinsJobTask
 import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder
 import com.netflix.spinnaker.orca.pipeline.TaskNode
 import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.pipeline.tasks.artifacts.BindProducedArtifactsTask
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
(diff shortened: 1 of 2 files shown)


ezimanyi:
A couple of quick pointers based on what you have here (which you noted of course is just a POC):
* The artifacts emitted by Jenkins aren't necessarily in the same format as artifacts Spinnaker expects, so there will likely need to be some translation for that to happen
* The `BindProducedArtifactsTask` is expecting any artifacts it binds to be in `stage.context.artifacts`, so you'll need to somehow read those from the Jenkins job result and put them there.  I'd suggest taking a look at `BakeStage.groovy` for a good example (in that case in `CompletedBakeTask` we handle the artifacts, before calling `BindProducedArtifacts`)

Thanks a lot for contributing to Spinnaker---this PR looks good so far!

avram:
The terminology can be confusing, but I'm not intending for this to be picking up Jenkins artifacts, but rather I'm relying on the user to declare that the Jenkins job will emit a properties file, and that the properties file will include the key `artifacts`, containing an array of legal Artifact objects.

That assumes, I think reasonably, that the Jenkins job author can construct legal Spinnaker-style Artifact descriptions in their job and put them in a property file.

I also assume that the property file will be in one of the supported formats that could support complex objects-- but both the YML and JSON formats should work fine; Java `.properties` files less so.

I thought about leaning on the actual Jenkins job artifacts, but those aren't usually that valuable unless they were actually published somewhere, in which case we'd need to know where they were published and not just pull them from the Jenkins instance.

avram:
I also really wanted to be able to write a test for the _stage_ -- I ended up augmenting a test for one of the _tasks_, which means I didn't get to properly test the way that the stage generates the task graph and the way that task context is fed into stage context. If you can point me to some tests that have working patterns of testing total stages, I'll make stage-level tests which should better cover the feature.

ezimanyi:
I agree it makes sense to have the author of the Jenkins job construct a legal spinnaker artifact (as you say, Jenkins artifacts are in general not really useful unless published somewhere).

Regarding tests for the stage, there are a few stages that have tests of varying degree of thoroughness...looks like `RollbackServerGroupStageSpec.groovy` does some validation of the generated task graph.  `ManualJudgementStageSpec.groovy` is also fairly well tested, so might be a good example.

ezimanyi:
Looks great to me.  @avram, if you're set with this I can merge it...just let me know!  Thanks!

avram:
I've asked @imosquera to help bang on this some too because I don't feel like I've tested this enough.

avram:
@imosquera verified this as well. Let's get this merged!

PR #2141, opened 2018-04-11
chore(highlander): Instrument cluster-wide tasks.

jtk54:


ajordens:
At a minimum you'll want to include the `executionId` ... also the logged payloads for these is going to large if it's the entire server group map?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractClusterWideClouddriverTask.groovy
```diff
@@ -102,6 +102,7 @@ abstract class AbstractClusterWideClouddriverTask extends AbstractCloudProviderA
     }
 
     List<Map> serverGroups = cluster.get().serverGroups
+    log.debug("Server groups fetched from cluster: ${serverGroups}")
```
duftler: Since you're trying to troubleshoot a potential filtering problem, why not just pull out the names of the server groups here (and below) instead of serializing their full representations?

PR #2145, opened 2018-04-12
fix(front50): Wait for pipeline to be updated before completing stage

ajordens:
This offers a very small bandaid around situations wherein a Front50
backed by S3 does not always provide read-after-write consistency.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorFront50Task.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorFront50Task.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


ajordens:
This is of questionable value but some local test identified at least one save pipeline took ~46s to reconcile. 

Subsequent saves were of course immediate.

If you have multiple `front50` instances this is not going to guarantee that all instances reflect the newly saved pipeline. 

I'm inclined to merge as it provides minor _incremental_ value.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorFront50Task.java
```diff
@@ -0,0 +1,89 @@
...
+          .findFirst();
+
+        if (!pipeline.isPresent()) {
+          return TaskResult.SUCCEEDED;
+        }
+
+        Long lastModifiedTime = Long.valueOf(pipeline.get().get("updateTs").toString());
+        return (lastModifiedTime > stage.getStartTime()) ? TaskResult.SUCCEEDED : TaskResult.RUNNING;
+      } catch (Exception e) {
+        return TaskResult.RUNNING;
```
dreynaud: might help to log the exception here

PR #2149, opened 2018-04-13
feat(core): Support for both sync/async application event listeners

robzienert:
Some plumbing work in preparation for implementing the QoS management system, which will be an `extensionpoint` to Orca's core functionality. Orca is currently written so that all Spring events are asynchronous, but we'll be adding event extensions that will need to be synchronous. Spring's normal behavior is to opt into async listeners, with sync being the default; in our case we need the inverse which requires switching up some Spring plumbing. 

Supports both class-based event listeners, as well as method-level event listeners. You can opt a listener of either kind into synchronous execution via the new `@Sync` annotation.

I've also addressed a series of TODOs that move the event task executor pool stuff into `orca-core` from `orca-queue`, since I was already touching this system.

At the point of this PR, there's no synchronous listeners.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/annotations/Sync.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/annotations/Sync.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/annotations/Sync.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Used for flagging application event listeners as synchronous. By default, orca's event listeners will be setup
+ * as asynchronous unless explicitly flagged with this annotation.
+ */
+@Target({ElementType.METHOD, ElementType.TYPE})
```
robfletcher: Why `METHOD`?

robfletcher: Oh, ignore me. Annotated listener methods. Got it.

PR #2150, opened 2018-04-15
feat(provider/kubernetes): v2 Add stable/failed manifests to output

wjoel:
This makes it possible for downstream stages to determine which manifests failed.


wjoel:
Great question, I'll have to check that and get back to you. Probably the former?

wjoel:
Yep, `stableManifests` is non-empty and `failedManifests` is empty (in this case waiting for just one other manifest, it would contain other actual failed manifests if they had reached a known failed state) when I set `stageTimeoutMs` to a low value.

I'm not sure if we can intercept that and set `failedManifests` to `allManifests.removeAll(stableManifests)` or if we even want to, since we don't _know_ if those manifests are actually failed. I can see arguments for both options.

wjoel:
Rebased, I like it the way it is.

PR #2152, opened 2018-04-16
feat(artifacts) Expose child pipeline outputs as outputs of a pipeline stage

avram:
Per Slack discussion with @ezimanyi https://spinnakerteam.slack.com/archives/C091CCWRJ/p1523425204000140; my intended use case is to access artifacts emitted by the child pipeline, but there may be other utility as well.

This appears to be the easiest way to do this, but it relies on the deprecated `getContext()` on `Execution`. In order to avoid that, I could recreate the outputs by walking the stages in the child pipeline, or just create a non-deprecated `getOutputs()` on `Execution`. If someone can comment on _why_ the method is deprecated, that might help. It was deprecated in a89c57b7d35ac8370f1955b465750ef0748c7f7b by @robfletcher.

I did _not_ add support for declaring Expected Artifacts and emitting only expected artifacts in the pipeline stage, but I can do that as well. Perhaps @lwander can comment on which behavior would be more in line with the underlying design of artifact resolution?

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
@@ -151,7 +151,6 @@ public class Execution implements Serializable {
     this.keepWaitingPipelines = keepWaitingPipelines;
   }
 
-  @Deprecated
   @JsonIgnore
   public @Nonnull Map<String, Object> getContext() {
     return Stage.topologicalSort(stages)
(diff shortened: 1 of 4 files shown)


lwander:
I think if you add a "bind produced artifacts" task after the child pipeline is called, and enable "produces artifacts" in deck (I see you've done this before) it should be able to bind any child pipeline's produced artifacts automatically.

avram:
Ok. I'll wire in the rest of the tasks for artifact handling then!

avram:
Ok. I'll wire in the rest of the tasks for artifact handling then!

ezimanyi:
@avram : This looks good; I'm actually not sure why that method is deprecated, as it seems useful and it's not clear to me what alternatives exist beyond just implementing the method again in the calling code.  I'd suggest using the method despite the deprecation (and potentially even removing the deprecation), though curious if @lwander has thoughts about why it's deprecated.

lwander:
I don't know unfortunately, let's try pinging @robfletcher to see if he knows

ezimanyi:
Chatted about this with @robfletcher.  The `@Deprecated` annotation was added back when `getContext` function was just an accessor for the now-removed `Execution.context` member variable.  Since then, `getContext` was refactored to construct the context by looking at the stages and accumulating the outputs of the stages.  Given that the function has now been refactored to no longer depend on the deprecated (and now removed) variable, we can just remove the `@Deprecated` annotation.

PR #2155, opened 2018-04-17
config(metrics): tag task duration metric with task type

dreynaud:
With this, we should be able to see a breakdown of time taken for different task types (looking at you, `ServerGroupCacheForceRefreshTask`)

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
@@ -85,22 +85,22 @@ class RunTaskHandler(
               when (result.status) {
                 RUNNING                              -> {
                   queue.push(message, task.backoffPeriod(taskModel, stage))
-                  trackResult(stage, taskModel, result.status)
+                  trackResult(stage, task, taskModel, result.status)
                 }
                 SUCCEEDED, REDIRECT, FAILED_CONTINUE -> {
                   queue.push(CompleteTask(message, result.status))
-                  trackResult(stage, taskModel, result.status)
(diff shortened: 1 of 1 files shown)


ajordens:
This may (or may not) blow out the tag space and prevent us from looking at multiple hours of this metric?

robzienert:
I said this, but Daniel looked at me like I was being ridiculous.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -113,24 +113,25 @@ class RunTaskHandler(
...
       }
     }
   }
 
-  private fun trackResult(stage: Stage, taskModel: com.netflix.spinnaker.orca.pipeline.model.Task, status: ExecutionStatus) {
+  private fun trackResult(stage: Stage, task: Task, taskModel: com.netflix.spinnaker.orca.pipeline.model.Task, status: ExecutionStatus) {
     val id = registry.createId("task.invocations")
       .withTag("status", status.toString())
       .withTag("executionType", stage.execution.type.name.capitalize())
+      .withTag("taskType", task::class.qualifiedName)
```
robfletcher: You already have access to `taskModel` which has an `implementingClass` property that gives you exactly this value. No need to introduce another parameter.

dreynaud: sweet, thanks!

dreynaud:
@ajordens I count <300 task classes. This is moderately high, but I don't think it is prohibitively expensive. There are lots of metrics with a lot more cardinality, so I'm not overly worried about this one.

PR #2157, opened 2018-04-17
More tags

dreynaud:
Keep the current metrics with the application tag:
* task.invocations
* task.invocations.duration

and introduce new ones without the application tag but with the task type,
account and region if applicable:
* task.invocations.withType
* task.invocations.duration.withType

This way, we should be able to dig into performance issues without blowing
up the cardinality of our metrics (#applications * #taskTypes may get high)

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
@@ -17,10 +17,7 @@
 package com.netflix.spinnaker.orca.q.handler
 
 import com.netflix.spectator.api.BasicTag
-import com.netflix.spectator.api.Id
 import com.netflix.spectator.api.Registry
-import com.netflix.spectator.api.Tag
-import com.netflix.spectator.api.histogram.BucketCounter
 import com.netflix.spinnaker.orca.*
 import com.netflix.spinnaker.orca.ExecutionStatus.*
 import com.netflix.spinnaker.orca.exceptions.ExceptionHandler
@@ -36,6 +33,7 @@ import com.netflix.spinnaker.orca.q.CompleteTask
 import com.netflix.spinnaker.orca.q.InvalidTaskType
 import com.netflix.spinnaker.orca.q.PauseTask
 import com.netflix.spinnaker.orca.q.RunTask
+import com.netflix.spinnaker.orca.q.metrics.MetricsTagHelper
 import com.netflix.spinnaker.orca.time.toDuration
 import com.netflix.spinnaker.orca.time.toInstant
 import com.netflix.spinnaker.q.Message
@@ -65,6 +63,8 @@ class RunTaskHandler(
   override fun handle(message: RunTask) {
     message.withTask { stage, taskModel, task ->
       val execution = stage.execution
+      val thisInvocationStartTimeMs = clock.millis()
+
       try {
         if (execution.isCanceled) {
           task.onCancel(stage)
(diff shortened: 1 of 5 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -128,36 +131,34 @@ class RunTaskHandler(
...
+    val taskInvocationsWithTypeId = registry.createId("task.invocations.withType")
+      .withTags(commonTags)
+      .withTag("taskType", taskModel.implementingClass)
+      .withTag("account", stage.context["account"].toString()?: "n/a")
+      .withTag("region", stage.context["region"].toString()?: "n/a")
+
+    registry.counter(taskInvocationsId).increment()
+    registry.counter(taskInvocationsWithTypeId).increment()
+
+    val elapsedMillis = System.currentTimeMillis() - (taskModel.startTime ?: 0)
```
robfletcher: `RunTaskHandler` has a `clock` property we could use instead of `System.currentTimeMillis()`.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -128,36 +131,34 @@ class RunTaskHandler(
...
-      "gt5m"
-    } else {
-      "lt5m"
+    hashMapOf(
+      "task.invocations.duration" to taskInvocationsId.tags(),
+      "task.invocations.duration.withType" to taskInvocationsWithTypeId.tags()
+    ).forEach {
+      name, tags ->
+        val id = registry.createId(name).withTags(tags)
+        registry.timer(id).record(elapsedMillis, TimeUnit.MILLISECONDS)
```
robfletcher: Is this valid use of a timer given we'll repeatedly generate a value for the same id every time this handler is invoked?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
dreynaud: indeed, changed!

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
brharrington: It will just update the stats for the timer with that id, so the usage of the timer should be fine. However, based on the duration strings, I wonder if a long task timer would be more appropriate for this use-case, see:

http://netflix.github.io/spectator/en/latest/intro/timer/

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -128,36 +131,34 @@ class RunTaskHandler(
...
 
-    val distributionId = registry.createId("task.invocations.duration").withTags(id.tags())
-    BucketCounter
-      .get(registry, distributionId, { v -> bucketDuration(v) })
-      .record(System.currentTimeMillis() - (taskModel.startTime ?: 0))
-  }
+    val taskInvocationsWithTypeId = registry.createId("task.invocations.withType")
+      .withTags(commonTags)
+      .withTag("taskType", taskModel.implementingClass)
+      .withTag("account", stage.context["account"].toString()?: "n/a")
```
copperlight: The `/` character is invalid and will be rewritten to an underscore `_`.

See http://insight-docs.prod.netflix.net/atlas/java-client/#allowed-characters for more details.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
dreynaud: Thanks for the suggestion @brharrington, looks like this is not going to work in this context because tasks can get started and completed on different instances

dreynaud:
Updated, thanks *@copperlight*

2018-04-17 15:51 GMT-07:00 Matthew Johnson <notifications@github.com>:

dreynaud:
@robfletcher I made the change we discussed this morning in 8b1fbc4c, which records separately the task invocation and completion times

@ajordens 8b1fbc4c is a significant change to the metrics we publish, what do you think?

dreynaud:
@ajordens this is correct, the time since task start is now reported at task completion in `CompleteTaskHandler`. I think `task.completions.duration` is a good starting point to monitor the perception of Spinnaker slowness from a user point of view, and `task.invocations.duration` _may_ be useful for us to debug slowness issues. We can remove it if it's not useful.

ajordens:
wcgw ... 👍

PR #2160, opened 2018-04-18
fix(core) Populate trigger.artifacts when it's a manual execution with the context of a previous build trigger

PerGon:
When a pipeline is initiated via manual execution and using the context of a previous build, the field `trigger.artifacts` is not populated. 
The artifacts information is in `trigger.buildinfo`, but when the execution happens via the build trigger, Echo puts the artifacts into `pipeline.receivedArtifacts` which will then end in `trigger.artifacts`. Such doesn't happen when it's a manual execution.

~~This PR will make the `ArtifactResolver` look into `trigger.buildinfo` to check if there are artifacts to be processed.~~

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -195,6 +195,9 @@ class OperationsController {
           log.warn("Igor is not enabled, unable to lookup build artifacts. Fix this by setting igor.enabled: true")
         } else {
           buildInfo.artifacts = buildArtifactFilter.filterArtifacts(buildInfo.artifacts)
+          if (trigger.type == "manual") {
+            trigger.artifacts = buildInfo.artifacts
+          }
         }
 
       }


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -163,7 +163,17 @@ public void resolveArtifacts(@Nonnull Map pipeline) {
     List<Artifact> artifactsFromTrigger = (List<Artifact>) Optional.ofNullable((List) trigger.get("artifacts"))
       .map(list -> list.stream().map(it -> objectMapper.convertValue(it, Artifact.class)).collect(toList()))
       .orElse(emptyList());
-    List<Artifact> receivedArtifacts = Stream.concat(receivedArtifactsFromPipeline.stream(), artifactsFromTrigger.stream()).collect(toList());
+
+    List<Artifact> artifactsFromTriggerBuildInfo = emptyList();
+    if (trigger.get("type").equals("manual") && trigger.get("buildInfo") != null) {
```
lwander: When are we populating the artifacts from build info when it's not a manual execution?

ezimanyi:
I wonder if we should try to come up with a more general solution to this---it seems like it will cause maintenance issues if we have duplicated code in Echo and Orca to translate build properties into artifacts, and which one runs depends on whether it's a triggered or a manual execution.  This also makes me think of how we should handle in general re-running pipelines that received artifacts (basically every instance of `withReceivedArtifacts` in Echo); ideally we wouldn't need to push all of those into Orca to handle manual executions.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
PerGon: It happens in Echo [in here](https://github.com/spinnaker/echo/blob/0718fddd458b14b752acaeaf2795775656f6b60c/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/BuildEventMonitor.java#L88)

PerGon:
I've committed a change that changes the initial approach I was having for a much simpler one.

@ezimanyi I agree. I'm not entirely sure how we could do it. Currently, the flow goes something like this:
Build trigger:
Igor --> Echo --> Orca
Manual Trigger:
Gate --> Orca

Maybe a way to solve this would be to change the Manual Trigger for a flow like this:
Gate --> Echo --> Orca

I haven't dug very much in what other cases there are, but a manual execution can either have context from a build, parent pipeline or neither of those.

How do you suggest this could be changed?
Thanks

ezimanyi:
@PerGon : This change looks good for the specific case of build artifacts, so I'd be fine merging this and then thinking more generally about the broader question.

I do think your idea of making manual triggers go through Echo might be a good idea; it feels like if we want manual runs to be as similar as possible to regular runs, we should try to trigger them in the same way (which is via a message in Echo).  Maybe something we can think about for any broader work we do on this...though that's obviously a larger scope than the change here.

PR #2164, opened 2018-04-19
feat(qos): Adding initial naive Quality of Service module

robzienert:
We've been experiencing production pain in the form of higher bursty execution
load than what Orca's downstream services are able to take if offered in a
firehose fashion. This PR offers a framework (albeit simplistic first impl)
for enforcing QoS of executions.

The basic premise is that a new extension hook is added when executions are
initially saved into the system, they'll potentially be set with a BUFFERED
status, which will prevent StartExecutionHandler from seeing it, and thus
beginning any work.

A BufferStateSupplier is implemented (currently based off system-wide active
executions) that can flip the system from INACTIVE to ACTIVE. In the case of
this basic implementation, if a threshold of active execution is surpassed the
QoS system is set to ACTIVE.

Once ACTIVE, an ExecutionBufferActuator listening to the new
BeforeInitialExecutionPersist event will start passing the execution through
a pluggable & configurable set of ordered BufferPolicies. A BufferPolicy is
responsible for determining an action on the execution, either to ENQUEUE,
BUFFER or ABSTAIN from a decision. If ENQUEUE, the execution will be dropped
onto the queue immediately, BUFFER will set the execution's state to BUFFERED
and ABSTAIN will no-op.

On the other end, an ExecutionPromoter is always running on an interval,
regardless of BufferState, to promote BUFFERED executions into NOT_STARTED
(and thus enqueued). The ExecutionPromoter, similar to the
ExecutionBufferActuator, will pass all candidate executions through a list
of ordered PromotionPolicies. These Policies will apply heuristics for
filtering and ordering what executions will be promoted.

In both forms of Policy, a Policy can force an action if it decides that it
knows best. If multiple Policies perform a force, the highest precedence
Policy will win.

---

Currently, the implementation is very naive and basically on/off. Future work
will be done to implement application & execution criticality to inform
policies. I'd also like to profile individual clouddriver clients as another
form of BufferStateSupplier, so we can back off on work we're sending to
clouddriver if orca starts getting higher latencies or increased error rates.

There will be additional further work in attaching system notifications to
executions so that Orca can propagate QoS actions up to end users via the API
and UI so people aren't left wondering why their execution isn't immediately
starting.

This concept of execution buffering was chosen over something implemented
on the queue so that we can ensure in-flight executions are finished as
quickly as possible, rather than slowing all executions down. It's possible
this QoS system may grow more sophisticated in the future to tune QoS of
in-flight executions based on criticality, but that's a lot more work and a
little more dicey.

By default, everything is disabled.

### DIFF
--- a/orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/BufferPolicy.kt
+++ b/orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/BufferPolicy.kt
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.qos
+
+import com.netflix.spinnaker.orca.pipeline.model.Execution
+import org.springframework.core.Ordered
+
+/**
+ * A policy responsible for determining whether an [Execution] should be buffered.
+ */
+interface BufferPolicy : Ordered {
+
+  fun apply(execution: Execution): BufferResult
+
+  override fun getOrder() = 0
+}
+
+enum class BufferAction {
+  /**
+   * Marks the execution for buffering.
+   */
+  BUFFER,
+
+  /**
+   * Marks the execution for moving from buffered to queued.
+   */
+  ENQUEUE
+}
+
+/**
+ * @param action The action to take
+ * @param force If true, the chain of policies will be short-circuited and this result's action will be used.
+ * @param reason A human-friendly reason for the action.
+ */
+data class BufferResult(
+  val action: BufferAction,
+  val force: Boolean,
+  val reason: String
+)
--- a/orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
+++ b/orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.qos
+
+import com.netflix.spinnaker.orca.ExecutionStatus.BUFFERED
+import com.netflix.spinnaker.orca.annotations.Sync
+import com.netflix.spinnaker.orca.events.BeforeInitialExecutionPersist
+import com.netflix.spinnaker.orca.pipeline.model.Execution
+import com.netflix.spinnaker.orca.qos.BufferAction.BUFFER
+import com.netflix.spinnaker.orca.qos.BufferAction.ENQUEUE
+import com.netflix.spinnaker.orca.qos.BufferState.ACTIVE
+import net.logstash.logback.argument.StructuredArguments.value
+import org.slf4j.LoggerFactory
+import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
+
+/**
+ * Determines if an execution should be buffered.
+ */
+@Component
+class ExecutionBufferActuator(
+  private val bufferStateSupplier: BufferStateSupplier,
+  policies: List<BufferPolicy>
+) {
+
+  private val log = LoggerFactory.getLogger(ExecutionBufferActuator::class.java)
+
+  private val orderedPolicies = policies.sortedByDescending { it.order }.toList()
+
+  @Sync
+  @EventListener(BeforeInitialExecutionPersist::class)
+  fun beforeInitialPersist(event: BeforeInitialExecutionPersist) {
+    if (bufferStateSupplier.get() == ACTIVE) {
+      val execution = event.execution
+      withActionDecision(execution) {
+        when (it.action) {
+          BUFFER -> {
+            log.warn("Buffering execution: {}, reason: ${it.reason}", value("executionId", execution.id))
+            execution.status = BUFFERED
+          }
+          ENQUEUE -> {
+            log.debug("Enqueuing execution: {}, reason: ${it.reason}", value("executionId", execution.id))
+          }
+        }
+      }
+    }
+  }
+
+  fun withActionDecision(execution: Execution, fn: (BufferResult) -> Unit) {
+    orderedPolicies
+      .map { it.apply(execution) }
+      .let { bufferResults ->
+        if (bufferResults.isEmpty()) {
+          return@let null
+        }
+
+        val forcedDecision = bufferResults.firstOrNull { it.force }
+        if (forcedDecision != null) {
+          return@let forcedDecision
+        }
+
+        // Require all results to call for enqueuing the execution, otherwise buffer.
+        val enqueue = bufferResults.all { it.action == ENQUEUE }
+        val reasons = bufferResults.joinToString(",") { it.reason }
+
+        return@let BufferResult(
+          action = if (enqueue) ENQUEUE else BUFFER,
+          force = false,
+          reason = reasons
+        )
+      }
+      ?.run { fn.invoke(this) }
+  }
+}
--- a/orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
+++ b/orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.qos
+
+import com.netflix.spinnaker.orca.ExecutionStatus.NOT_STARTED
+import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
+import net.logstash.logback.argument.StructuredArguments.value
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.BeanInitializationException
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+
+/**
+ * Marker interface if someone wants to override the default promoter.
+ */
+interface ExecutionPromoter
+
+@Component
+@ConditionalOnMissingBean(ExecutionPromoter::class)
+class DefaultExecutionPromoter(
+  private val executionRepository: ExecutionRepository,
+  private val policies: List<PromotionPolicy>
+) : ExecutionPromoter {
+
+  private val log = LoggerFactory.getLogger(ExecutionPromoter::class.java)
+
+  init {
+    if (policies.isEmpty()) {
+      throw NoPromotionPolicies("At least one PromotionPolicy must be defined")
+    }
+  }
+
+  @Scheduled(fixedDelayString = "\${qos.executionPromoter.intervalMs:5000}")
+  fun promote() {
+    executionRepository.retrieveBufferedExecutions()
+      .sortedByDescending { it.buildTime }
+      .let {
+        // TODO rz - This is all temporary mess and isn't meant to live long-term. I'd like to calculate until
+        // result.finalized==true or the end of the list, then be able to pass all contributing source & reason pairs
+        // into a log, with a zipped summary that would be saved into an execution's system notifications.
+        var lastResult: PromotionResult? = null
+        var candidates = it
+        policies.forEach { policy ->
+          val result = policy.apply(candidates)
+          if (result.finalized) {
+            return@let result
+          }
+          candidates = result.candidates
+          lastResult = result
+        }
+        lastResult ?: PromotionResult(
+          source =  javaClass.simpleName,
+          candidates = candidates,
+          finalized = true,
+          reason = "No promotion policy resulted in an action"
+        )
+      }
+      .also { result ->
+        result.candidates.forEach {
+          log.info("Promoting execution {} for work: {}", value("executionId", it.id), result.reason)
+          executionRepository.updateStatus(it.id, NOT_STARTED)
+        }
+      }
+  }
+
+  private class NoPromotionPolicies(message: String) : BeanInitializationException(message)
+}
--- a/orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/PromotionPolicy.kt
+++ b/orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/PromotionPolicy.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.qos
+
+import com.netflix.spinnaker.orca.pipeline.model.Execution
+import org.springframework.core.Ordered
+
+/**
+ * A policy responsible for determining what execution(s) should be promoted.
+ *
+ * PromotionPolicies can either reduce the list of candidates or re-order them, as well as short-circuit lower
+ * precedence PromotionPolicies.
+ */
+interface PromotionPolicy : Ordered {
+
+  fun apply(candidates: List<Execution>): PromotionResult
+
+  override fun getOrder() = 0
+}
+
+/**
+ * @param candidates A potentially reduced or re-ordered list of promotion candidates.
+ * @param finalized Flag controlling whether or not the [candidates] are the final list of executions to promote.
+ * Setting this this value to true will short-circuit lower-precedence [PromotionPolicy]s.
+ * @param reason A human-friendly reason for the promotion result.
+ */
+data class PromotionResult(
+  val source: String,
+  val candidates: List<Execution>,
+  val finalized: Boolean,
+  val reason: String
+)
--- a/orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/bufferstate/ActiveExecutionsBufferStateSupplier.kt
+++ b/orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/bufferstate/ActiveExecutionsBufferStateSupplier.kt
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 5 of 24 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
```diff
@@ -501,6 +502,18 @@ Execution retrieveOrchestrationForCorrelationId(
     });
   }
 
+  @Override
+  public @Nonnull List<Execution> retrieveBufferedExecutions() {
+    // TODO rz - This is definitely not a healthy way to do this.
```
asher: Already todo'd but we should definitely have an index of buffered executions before this goes live.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
```diff
@@ -0,0 +1,81 @@
...
+  fun withActionDecision(execution: Execution, fn: (BufferResult) -> Unit) {
+    orderedPolicies
+      .map { it.apply(execution) }
+      .let { policies ->
+        // If any policies have flagged themselves to force an action, the first force is used. Otherwise, the last
+        // policy's action will be used.
+        // TODO rz - In the future, better heuristics around all policies informing the decision may need to be
+        // implemented.
+        try {
+          return@let policies.firstOrNull { it.force } ?: policies.last()
```
asher: Why does the last policy win and how is ordering determined? It's not clear to me what behavior should be expected if i.e. there are 3 active policies where 1 returns BUFFER and 2 ENQUEUE, but none set force==true. Still wrapping my head around the full PR, but I have an initial naive preference for considering BUFFER=false, ENQUEUE=true, ABSTAIN=true where an execution will be enqueued `if (policy1 && policy2 ... && policyN) == true`.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
```diff
@@ -0,0 +1,81 @@
...
+          source =  javaClass.simpleName,
+          candidates = candidates,
+          finalized = true,
+          reason = "No promotion policy resulted in an action"
+        )
+      }
+      .also { result ->
+        result.candidates.forEach {
+          log.info("Promoting execution {} for work: {}", value("executionId", it.id), result.reason)
+          executionRepository.updateStatus(it.id, NOT_STARTED)
```
asher: should we enqueue a StartExecution message here?

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/BufferPolicy.kt
```diff
@@ -0,0 +1,57 @@
...
+   */
+  BUFFER,
+
+  /**
+   * Marks the execution for moving from buffered to queued.
+   */
+  ENQUEUE,
+
+  /**
+   * Abstain from performing a policy action. If all policies abstain, the the execution will be enqueued.
```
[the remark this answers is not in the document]
robzienert: That's right. Only one enqueue is required. I hadn't really thought about conflicts, I'd been going for first `force` flag, or last policy. It's probable that we'll want to adjust this in the future.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
```diff
@@ -0,0 +1,81 @@
...
+        } catch (e: NoSuchElementException) {
```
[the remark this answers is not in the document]
robzienert: I could wire up a no-op policy by default so there's always a policy present.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
robzienert: +1

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
```diff
@@ -0,0 +1,81 @@
...
+        when (it.action) {
+          BUFFER -> {
+            log.warn("Buffering execution: {}, reason: ${it.reason}", value("executionId", execution.id))
+            execution.status = BUFFERED
+          }
+          ENQUEUE -> {
+            log.debug("Enqueuing execution: {}, reason: ${it.reason}", value("executionId", execution.id))
+          }
+          ABSTAIN -> {
+            log.debug("Abstaining from quality of service, enqueuing execution: {}, reason: ${it.reason}", value("executionId", execution.id))
```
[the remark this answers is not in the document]
robzienert: We probably don't need it. I had hopes for it in a half-asleep late-night coding session, but don't think it's necessary.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
robzienert: This decision logic is by no means ideal (I even have a TODO right above saying that this needs to be done better). I think requiring all to ENQUEUE is probably the easiest strategy to reason about. Do like.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
robzienert: I don't want to directly influence the behavior of the queue from this code. If I added a start execution message here, then `StartExecutionHandler`'s logic around enforcing limitConcurrent, etc would subverted and likely introduce bugs.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/bufferstate/ActiveExecutionsBufferStateSupplier.kt
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Supplies the buffering state based on a threshold of system-wide active executions.
+ *
+ * As an immediate implementation, this supplier depends on [RedisActiveExecutionMonitor] to update the registry with
+ * system-wide active executions. The polling interval of that monitor will determine the resolution of the buffering
+ * state. By default, this is 60 seconds.
+ */
+@Component
+class ActiveExecutionsBufferStateSupplier(
```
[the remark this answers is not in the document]
robzienert: That's exactly the behavior of this PR. I don't want to have the logic of determining whether executions should potentially be buffered (whatever it may be in the future) in the actual persist execution path, since it could potentially be an expensive operation.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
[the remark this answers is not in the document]
robfletcher: Or, y'know, just account for the possibility of `policies` being empty.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/bufferstate/ActiveExecutionsBufferStateSupplier.kt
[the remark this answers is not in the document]
robfletcher: Oh, so _all_ executions get buffered when buffering is happening? Then the decider will gate some of them thru quickly?

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
robfletcher: I feel like `ENQUEUE` winning is probably right. Err on the side of caution.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
robfletcher: I still feel like we should potentially unify that logic. It seems like 2 different ways of doing very similar things.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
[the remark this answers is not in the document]
robfletcher: Can we maybe lose it and add it back if it ever turns out we need it?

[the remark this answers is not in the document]
robzienert: 1000% yes. I'm happy to trim things down.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/bufferstate/ActiveExecutionsBufferStateSupplier.kt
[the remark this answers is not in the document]
robzienert: That's right.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
robzienert: Although you could also say erring on the side of caution is to `BUFFER` the execution. If the system is under load, I would want the system to BUFFER by default. But yes, we should simplify.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
robzienert: In this implementation, it's keeping all of the logic in a single place.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
```diff
@@ -0,0 +1,81 @@
...
+
+  private val log = LoggerFactory.getLogger(ExecutionPromoter::class.java)
+
+  init {
+    if (policies.isEmpty()) {
+      throw NoPromotionPolicies("At least one PromotionPolicy must be defined")
+    }
+  }
+
+  @Scheduled(fixedDelayString = "\${qos.executionPromoter.intervalMs:5000}")
```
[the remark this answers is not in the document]
robzienert: What benefit would we have in a jitter?

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/bufferstate/ActiveExecutionsBufferStateSupplier.kt
```diff
@@ -0,0 +1,71 @@
...
+  private var state: BufferState = INACTIVE
+
+  private val bufferingId = registry.createId("qos.buffering")
+
+  @Scheduled(fixedDelayString = "\${pollers.qosUpdateStateIntervalMs:5000}")
+  private fun updateCurrentState() {
+    val activeExecutions = registry.gauges()
+      .filter { it.id().name() == "executions.active" }
+      .map { it.value() }
+      .reduce { p: Double, o: Double -> o + p }
```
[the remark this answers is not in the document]
robzienert: This is basically handling that the interface of `map` returns a collection. This just reduces the collection to a single value, even though the `filter` should only have the collection be a size of 1.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionBufferActuator.kt
robfletcher: D'oh. That's what I meant. Sorry.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/bufferstate/ActiveExecutionsBufferStateSupplier.kt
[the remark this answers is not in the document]
robfletcher: Ok, that seems legit. I like the simplicity.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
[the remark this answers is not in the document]
emjburns: Mostly so that all instances aren't doing the work at the same time. Discussed offline, seems like there's no need to right now.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/BufferPolicy.kt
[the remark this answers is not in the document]
robzienert: What if a policy wants to force the alternative behavior? Regardless of which default we choose, I could see a world where we would want a policy to force its decision and overrule the others. 

For example, what if we buffer by default but a team is responding to an incident and _needs_ the executions to go through; it'd be nice to have a break-the-glass policy that people could interact with that forces a particular action. Alternatively, if we choose a default enqueue by default, what if there's a high-value customer that needs more capacity, we might force buffering other executions to ensure the high-value customer gets the guaranteed capacity they need to fulfill their executions within an SLA.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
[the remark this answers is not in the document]
ajordens: TBH there's enough jitter in instance launch time that they wouldn't be running at the same time anyhow. 

What happens though if an instance is disabled ... it probably shouldn't be running this stuff?

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
```diff
@@ -0,0 +1,81 @@
...
+  fun promote() {
+    executionRepository.retrieveBufferedExecutions()
```
ajordens: Imagine a thousand buffered executions -- is this slow?

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
ajordens: Thoughts around what would happen for tightly cron'd pipelines? (ie. the ones we have that are */5)

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
```diff
@@ -0,0 +1,81 @@
...
+      }
+      .also { result ->
+        result.candidates.forEach {
+          log.info("Promoting execution {} for work: {}", value("executionId", it.id), result.reason)
+          executionRepository.updateStatus(it.id, NOT_STARTED)
+        }
+      }
+  }
+
+  private class NoPromotionPolicies(message: String) : BeanInitializationException(message)
```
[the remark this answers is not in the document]
ajordens: Why you gotta be different!

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/PromotionPolicy.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+interface PromotionPolicy : Ordered {
+
+  fun apply(candidates: List<Execution>): PromotionResult
+
+  override fun getOrder() = 0
+}
+
+/**
+ * @param candidates A potentially reduced or re-ordered list of promotion candidates.
```
ajordens: missing a `@param source`

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/bufferstate/ActiveExecutionsBufferStateSupplier.kt
```diff
@@ -0,0 +1,71 @@
...
+  private val registry: Registry
+) : BufferStateSupplier {
+
+  private val log = LoggerFactory.getLogger(ActiveExecutionsBufferStateSupplier::class.java)
+
+  private var state: BufferState = INACTIVE
+
+  private val bufferingId = registry.createId("qos.buffering")
+
+  @Scheduled(fixedDelayString = "\${pollers.qosUpdateStateIntervalMs:5000}")
```
ajordens: Not sure this `pollers.qosUpdateStateIntervalMs` pattern fits the existing structure of `pollers`

Perhaps

```
pollers:
  qos:
    updateStateIntervalMs: 5000
```

We have enough spare threads in our `orca` threadpool for these new `@Scheduled` right? I think we do but it's worth being conscious of what that max number of threads is.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/strategy/naive/NaivePromotionPolicy.kt
```diff
@@ -0,0 +1,35 @@
...
+@Component
+@ConditionalOnProperty("qos.strategy.naive.enabled")
+class NaivePromotionPolicy : PromotionPolicy {
+
+  override fun apply(candidates: List<Execution>) =
+   PromotionResult(
+     source = javaClass.simpleName,
+     candidates = candidates.subList(0, 1),
+     finalized = false,
+     reason = "Naive policy promotes 1 execution every cycle"
```
ajordens: For this particular case, `candidates` is currently (initial default) sorted list by `buildTime` right? 

So we're just naively promoting the most recently built execution (or maybe it's the oldest execution).

Also this is just the naive implementation, I wonder if it's worth being explicit about the sort order rather than just `subList(0, 1)` .. or maybe it just doesn't matter at all!

orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
```diff
@@ -503,6 +504,18 @@ Execution retrieveOrchestrationForCorrelationId(
     });
   }
 
+  @Override
+  public @Nonnull List<Execution> retrieveBufferedExecutions() {
+    // TODO rz - This is definitely not a healthy way to do this.
+    return Observable.concat(
```
ajordens: Lulz. Answers my earlier question re: how slow this is.

ajordens: We may want to think through what happens to an `ORCHESTRATION` that is initiated through the UI ... it's possible that _should not_ ever be buffered?

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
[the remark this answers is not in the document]
robzienert: Good call, I'll get these pollers discovery activated.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
robzienert: I haven't tested this, but it shouldn't be any different than if there were tightly cron'd pipelines that had longer execution times than their cron schedule: they'll just stack up.

orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
robzienert: Definitely, good thought! This would be pretty straight forward; create an `OriginBufferPolicy` that returns a `force=true` to `ENQUEUE` the execution.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/strategy/naive/NaivePromotionPolicy.kt
robzienert: It would promote the oldest execution in its current form (since all candidates are fist sorted, as you note). I figured pre-sorting the candidates by build time would likely save us some cycles if / when we get to a chain of promotion policies.

I think for a naive implementation (and likely one that won't live very long), this should suffice for now?

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/bufferstate/ActiveExecutionsBufferStateSupplier.kt
ajordens: By default the `taskScheduler` was a single threaded pool but it's been increased -- from this metric it looks like we're 10 per instance.

These pollers are running per-instance (right?) so we just need to be conscious that we don't have 10 `@Scheduled` as we run the risk that something goes sideways and execution gets all out of wack (it's happened with the default single threaded pool).

orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
ajordens: 👍

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/strategy/naive/NaivePromotionPolicy.kt
ajordens: I agree.

I'm mentally thinking through what it might look like to precalculate a score up-front so that you could store in a sorted set and avoid having to fetch the world.

orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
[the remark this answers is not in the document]
robfletcher: Why you gotta use pseudo Hungarian Notation but only for exception types?

PR #2165, opened 2018-04-19
fix(expressions): Include eval summary for unevaluated composite expressions

jeyrschabu:
- including error msg if composite spel is not evaluated or is partially evaluated


jeyrschabu:

PR #2166, opened 2018-04-19
fix(q): actually using dead letter handler

emjburns:
Upgrading version of keiko.
Actually using our dead letter handler so that tasks/stages/executions are successfully canceled on error.


robzienert:
Haaa. This would explain a few things. We'll definitely want this pulled into the release candidate of 1.7.x as well. cc @lwander

PR #2175, opened 2018-04-24
refactor(*): Split redis code to its own module

robzienert:
Step one of v2 persistence. Functionally, nothing in this PR has changed. This PR breaks out all Redis code into its own `orca-redis` module:

* Introduced `NotificationClusterLock` interface so our `NotificationPollingAgents` can have a locking implementation that may not necessarily be Redis. Also added `RedisNotificationClusterLock` which is functionally the same as what was there previously.
* Moved all Redis-related code to `orca-redis`. This includes the migrators that were in `orca-front50` (strong doubts that they belonged in that module to begin with). 
* Removed Redis dependency on in `orca-queue`.
* Moved all Redis-related config into single package.
* Added `orca-core-tck`, which includes `AbstractExecutionRepositoryTck`.

Things to cleanup after this:

* `OldPipelineCleanup...Agent` and `TopApplicationExecutionCleanup...Agent` were moved into `orca-redis`. Non-ideal. Will refactor these to be storage impl-agnostic in a separate PR.

Other things of note:

* I couldn't remove Rx from `orca-core` because `ExecutionRepository` includes it as part of its API. Womp womp. From the looks, that's the only spot, though.
* Next steps will be a PR to get `orca-sql` in place & baseline Spring config to run Redis vs SQL (and also Redis & SQL for red/black).


robfletcher:
WRT the agents, it doesn’t necessarily feel wrong to include them in `orca-redis` as who knows if another storage implementation would even need them.

robzienert:
Ah, yes. I just mean I want to move them back to core without redis (if it makes sense to do so). I think they'll be valuable regardless of the persistence

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/kato/pipeline/ResizeAsgStageSpec.groovy
```diff
@@ -37,20 +33,6 @@ class ResizeAsgStageSpec extends Specification {
...
-  def setupSpec() {
-    embeddedRedis = EmbeddedRedis.embed()
-  }
-
-  def cleanup() {
-    embeddedRedis.jedis.withCloseable { it.flushDB() }
-  }
-
-  Pool<Jedis> jedisPool = embeddedRedis.pool
-
```
robfletcher: Holy… why was this…?

robzienert: No idea. Haha.

PR #2176, opened 2018-04-24
fix(kubernetes) - fixes WaitForClusterDisable timeout for kubernetes …

imosquera:
addresses the issue discussed here
https://github.com/spinnaker/spinnaker/issues/2690


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/WaitForClusterDisableTask.groovy
```diff
@@ -79,9 +79,14 @@ class WaitForClusterDisableTask extends AbstractWaitForClusterWideClouddriverTas
     // Even if the server group is disabled, we want to make sure instances are down
     // to prevent downstream stages (e.g. scaleDownCluster) from having to deal with disabled-but-instances-up server groups
     def targetServerGroup = serverGroup.get()
-    if (targetServerGroup.isDisabled() || stage.context.desiredPercentage) {
+    if (targetServerGroup.isDisabled() && stage.context.desiredPercentage) {
       return !waitForRequiredInstancesDownTask.hasSucceeded(stage, targetServerGroup as Map, targetServerGroup.getInstances(), interestingHealthProviderNames)
     }
+    
+    if(targetServerGroup.isDisabled()) {
```
sbwsg: missing space before (

ajordens:
I suspect this will regress a recent issue fixed by @dreynaud.

Just because the server group comes back as disabled doesn't mean that all instances in it have been disabled (yet).

The preference is that this task should wait until the server group and all its instances have been disabled.

lwander:
@ajordens the core issue here is that this check is run during a deployment, whose health providers check for instance (pod/container) health rather that load balancer health. When the check is made to see if all instances are disabled, it's still using the deployment's notion of health. Is there a way around this by injecting different health providers for this task alone?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/WaitForClusterDisableTask.groovy
dreynaud: this is going to fail tests in `WaitForClusterDisableTaskSpec.groovy`

imosquera: Yup, based on @ajordens feedback looks like we're going to take a different approach for Kubernetes.

lwander:
@ajordens or @dreynaud, any thoughts on the health providers?

ajordens:
What would it look like to allow special casing per-cloudprovider -- `kubernetes` in this particular case?

imosquera:
@ajordens - can  you see the updated PR, that injects a health supplier specifically for our use-case.  I followed a similar pattern to the Titus supplier - https://github.com/armory-io/orca/blob/a5c1254ce9b0442b6b72961bc4b5cdd921abb844/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesSupplier.java

ajordens:
@imosquera The `Supplier` approach looks good to me

lwander:
@sbwsg I'm going to patch this into 1.7 once merged here

PR #2180, opened 2018-04-25
feat(redis): Conditional prop for RedisExecutionRepository

robzienert:
We'll want this guy for Project SuperQualifiedLobotomy. No one will ever figure out that code name.


orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
```diff
@@ -55,6 +56,7 @@
 import static redis.clients.jedis.BinaryClient.LIST_POSITION.BEFORE;
 
 @Component
+@ConditionalOnProperty(value = "executionRepository.redis.enabled", matchIfMissing = true)
```
robfletcher: So that's on by default?

robzienert: Yes. This is backwards compatible with existing configurations.

PR #2185, opened 2018-04-26
chore(*): Upgrade to Boot 1.5.10

robzienert:
Necessary for integrating FPs.


robfletcher:
How does this work with the Spring Platform declared in Spinnaker-dependencies?

PR #2189, opened 2018-04-27
feat(core): Change Execution & Stage IDs to use ULIDs

robzienert:
Switching from `java.util.UUID` to [ulid](https://github.com/ulid/spec) (via impl [sulky-ulid](https://github.com/huxi/sulky/tree/master/sulky-ulid)) to get us a time-sortable ID in executions and stages. This is internals work for SQL.


robzienert:
@emjburns Good question. I only converted models that would have their own tables in SQL. For our initial implementation, tasks will just be serialized as part of the stage JSON blob, and won't have a table for themselves.

robfletcher:
Also Tasks are not currently assigned UUIDs. They're done like this ```task.id = (stage.tasks.size + 1).toString()```

PR #2193, opened 2018-04-30
fix(pipeline_template): Don't share snakeyaml instance across threads

ezimanyi:
The snakeyaml processor is not thread-safe, and we are currently sharing
it across threads, leading to intermittent NPEs when concurrent uses put
it in an inconsistent state.


ezimanyi:
Also closes https://github.com/spinnaker/spinnaker/issues/2243

PR #2200, opened 2018-05-01
config(metrics): best effort at pulling account, region, cloud from c…

dreynaud:
…ontext

This is closer to what actual tasks are doing, in particular ServerGroupCacheForceRefreshTask.


robzienert:
Why not just use `orca-core`? What goes in `-common` that wouldn't have a place in `-core`?

dreynaud:
@robzienert it's definitely possible. Since it's more of a utility thing and not particularly a "core" thing, I thought maybe we'd want to avoid polluting core. What says @robfletcher ?

ajordens:
tbh I'd probably stash it in `core` personally.

robfletcher:
I kinda like the idea of keeping it out of core but it's not a huge deal, I don't think

dreynaud:
Deleted orca-common and moved `CloudProviderAware` into core

PR #2201, opened 2018-05-02
fix(core): don't immediately restart pipelines if another is running

robfletcher:
If a pipeline has `limitConcurrent = true` and someone tries to restart a failed pipeline while another is already running, that restart will be queued just like any other executions.


robfletcher:
Renamed and moved everything around.

robzienert:
Bester.

PR #2205, opened 2018-05-03
refactor(core): Remove RxJava from ExecutionRepository interface

robzienert:
No functional changes. Not really prerequisite work for Squirrelly Quagmire Lounge, but will be a nice thing to have anyway. I didn't go about changing any internals of `RedisExecutionRepository` - I think using it inside is totally fine and probably makes a good deal of sense still.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
@@ -58,7 +58,7 @@ public interface ExecutionRepository {
 
   @Nonnull Iterable<Execution> retrieve(ExecutionType type);
 
-  @Nonnull Observable<Execution> retrievePipelinesForApplication(@Nonnull String application);
+  @Nonnull Iterable<Execution> retrievePipelinesForApplication(@Nonnull String application);
 
   @Nonnull Observable<Execution> retrievePipelinesForPipelineConfigId(
     @Nonnull String pipelineConfigId, @Nonnull ExecutionCriteria criteria);
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
@@ -24,6 +24,8 @@ import rx.Observable;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.*;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import static java.util.stream.Collectors.toList;
 
@@ -60,7 +62,7 @@ public interface ExecutionRepository {
 
   @Nonnull Iterable<Execution> retrievePipelinesForApplication(@Nonnull String application);
 
-  @Nonnull Observable<Execution> retrievePipelinesForPipelineConfigId(
+  @Nonnull Iterable<Execution> retrievePipelinesForPipelineConfigId(
     @Nonnull String pipelineConfigId, @Nonnull ExecutionCriteria criteria);
 
   @Nonnull Observable<Execution> retrieveOrchestrationsForApplication(
@@ -114,6 +116,13 @@ public interface ExecutionRepository {
 
     @Override public int hashCode() {
       return Objects.hash(limit, statuses);
+    }
+  }
+
+  final class IterableUtil {
+
+    public static <T> Stream<T> toStream(Iterable<T> iterable) {
+      return StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterable.iterator(), Spliterator.ORDERED), false);
     }
   }
 }
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
@@ -56,7 +56,7 @@ public interface ExecutionRepository {
 
   void delete(@Nonnull ExecutionType type, @Nonnull String id);
 
-  @Nonnull Observable<Execution> retrieve(ExecutionType type);
+  @Nonnull Iterable<Execution> retrieve(ExecutionType type);
 
   @Nonnull Observable<Execution> retrievePipelinesForApplication(@Nonnull String application);
 
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
@@ -19,7 +19,6 @@ import com.netflix.spinnaker.orca.ExecutionStatus;
 import com.netflix.spinnaker.orca.pipeline.model.Execution;
 import com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
-import rx.Observable;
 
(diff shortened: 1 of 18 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplateService.java
```diff
@@ -87,9 +89,8 @@ public Execution retrievePipelineOrNewestExecution(@Nullable String executionId,
       ExecutionRepository.ExecutionCriteria criteria = new ExecutionRepository.ExecutionCriteria().setLimit(1);
       try {
         return executionRepository.retrievePipelinesForPipelineConfigId(pipelineConfigId, criteria)
```
ajordens: any issue if the underlying result was empty ... ?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/ProjectController.groovy
```diff
@@ -63,9 +66,9 @@ class ProjectController {
       statuses: (statuses.split(",") as Collection)
     )
 
-    def allPipelines = rx.Observable.merge(pipelineConfigIds.collect {
+    def allPipelines = pipelineConfigIds.collect {
```
ajordens: is this a situation where we previously would have been doing multiple ops in parallel (on Schedulers.io()) but now are doing it linearly?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -89,22 +94,20 @@ class TaskController {
 
     def orchestrations = executionRepository
```
ajordens: another possible situation where we were fetching tasks in parallel?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -89,22 +94,20 @@ class TaskController {
...
+      .collect({ Execution orchestration -> convert(orchestration) })
+      .subList(0, Math.min(orchestrations.size(), limit))
   }
 
   @PreAuthorize("@fiatPermissionEvaluator.storeWholePermission()")
   @PostFilter("hasPermission(filterObject.application, 'APPLICATION', 'READ')")
   @RequestMapping(value = "/tasks", method = RequestMethod.GET)
   List<OrchestrationViewModel> list() {
-    executionRepository.retrieve(ORCHESTRATION).toBlocking().iterator.collect {
+    executionRepository.retrieve(ORCHESTRATION).collect {
```
ajordens: probably a super duper dangerous endpoint -- maybe we should just remove it?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -169,9 +172,10 @@ class TaskController {
 
     def ids = pipelineConfigIds.split(',')
 
-    def allPipelines = rx.Observable.merge(ids.collect {
+    def allPipelines = ids.collect {
```
ajordens: one more that looks possibly like parallel -> linear?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplateService.java
robzienert: The code here was already designed around the fact that it'll handle a `NoSuchElementException`. I could've further refactored this to not behave in this way, but didn't want unrelated changes to blur things up.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/ProjectController.groovy
robzienert: Hmm, yes. I got a little carried away here. I'll revert this change.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
robzienert: Also another case of getting carried away.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
robzienert: Agreed. I had a mental todo to investigate what (if anything) uses this endpoint and remove if possible.

robzienert:
Updated feedbacks, please take another gander, plz and thank you. Will merge tomorrow if no new words.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -69,11 +77,9 @@ class TaskController {
 
   @PreAuthorize("hasPermission(#application, 'APPLICATION', 'READ')")
   @RequestMapping(value = "/applications/{application}/tasks", method = RequestMethod.GET)
-  List<Execution> list(@PathVariable String application,
-                       @RequestParam(value = "limit", defaultValue = "3500")
-                         int limit,
-                       @RequestParam(value = "statuses", required = false)
-                         String statuses) {
+  List<OrchestrationViewModel> list(@PathVariable String application,
```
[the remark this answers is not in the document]
robfletcher: IIRC it was already returning `OrchestrationViewModel` but someone had declared the method signature wrong and Groovy just doesn't care.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
```diff
@@ -56,14 +57,14 @@
 
   void delete(@Nonnull ExecutionType type, @Nonnull String id);
 
-  @Nonnull Observable<Execution> retrieve(ExecutionType type);
+  @Nonnull Iterable<Execution> retrieve(ExecutionType type);
```
cfieber: would it make sense to have these return Collection instead of Iterable? I don't think we actually are streaming results out of the implementation, and maybe simplifies some of the transformations to streams?

cfieber: or even Stream<Execution> for that matter

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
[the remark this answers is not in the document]
robzienert: Testing the endpoint against prestaging Redis, there's no functional difference in this change, it's just making the method signature correctly reflect what the method is actually returning.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
robzienert: I wanted to keep `Iterable<Execution>` because I want to introduce cursor-based pagination into the SQL implementation. What I was thinking would be to have a `class ExecutionIterable : Iterable<Execution>` that would automatically handle going back out to the database to fetch the next page of results without us, the end users, having to build in that functionality.

I'll take a look at what it'd take to return a Stream interface and still be able to provide the above functionality.

robfletcher: Iterable is more cross-language friendly too. Stream is really only useful in Java.

PR #2206, opened 2018-05-03
feat(artifacts): Orca passes artifacts to bake to rosco

ezimanyi:
With this change, bake stages now support a field packageArtifactIds
that contains an array artifacts representing packages to bake into
the produced image.  This field will not be visible in the UI until
the corresponding change is made to deck.


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -136,6 +140,10 @@ class CreateBakeTask implements RetryableTask {
 
     Map requestMap = packageInfo.findTargetPackage(allowMissingPackageInstallation)
 
+    requestMap.packageArtifacts = stage.context.packageArtifactIds.collect { String artifactId ->
```
sbwsg: is there any risk that `packageArtifactIds` wouldn't be present in the stage's context?  if so would suggest adding a guard here

PR #2208, opened 2018-05-03
feat(core): fast property based kill-switch for pipeline buffering

robfletcher:


orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/KillSwitchBufferStateSupplier.kt
```diff
@@ -0,0 +1,33 @@
...
+
+package com.netflix.spinnaker.orca.qos
+
+import com.netflix.spinnaker.kork.transientconfig.TransientConfigService
+import com.netflix.spinnaker.orca.qos.BufferState.ACTIVE
+import com.netflix.spinnaker.orca.qos.BufferState.INACTIVE
+import org.springframework.stereotype.Component
+
+@Component
+class KillSwitchBufferStateSupplier(
```
robzienert: Did you test that this actually gets wired up correctly? I think at this point we'll have multiple BufferStateSuppliers, but the interfaces today expect only one.

PR #2217, opened 2018-05-07
fix(core): prevent app permissions from being lost during an app upse…

danielpeach:
…rt operation

### DIFF
--- a/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
+++ b/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.applications.tasks
 
+import com.netflix.spinnaker.fiat.model.resources.Permissions
 import com.netflix.spinnaker.orca.front50.model.Application
 import com.netflix.spinnaker.orca.front50.tasks.AbstractFront50Task
 import groovy.transform.CompileStatic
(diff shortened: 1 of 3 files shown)


orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
```diff
@@ -44,10 +45,15 @@ class UpsertApplicationTask extends AbstractFront50Task {
     } else {
       log.info("Creating application (name: ${application.name})")
       front50Service.create(application)
+      if (application?.permission?.permissions == null) {
```
lwander: You can probably groovify this with `if (!application?.permission?.permissions)`

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
```diff
@@ -44,10 +45,15 @@ class UpsertApplicationTask extends AbstractFront50Task {
...
+        application.setPermissions(Permissions.EMPTY)
+      }
     }
 
     try {
-      front50Service.updatePermission(application.name, application.permission)
+      if (application?.permission?.permissions != null) {
```
lwander: Similar here to above

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
danielpeach: I want this condition to be true for an empty map - `null` should mean no change to the permissions, whereas `[:]` should mean an empty set of permissions.

lwander: It'll also be true if `application` or `permission` are `null` when written like this -- is that intended?

danielpeach: Yes for `permission`, no for `application`, so I'll take out the null check for application.

ttomsu:
LGTM

PR #2218, opened 2018-05-07
config(okhttp3): pick up okhttp3 as retrofit client

dreynaud:


orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/RetrofitConfiguration.groovy
```diff
@@ -34,40 +36,41 @@ import org.springframework.context.annotation.Scope
...
+   @Value('${okHttp3Client.connectionPool.maxIdleConnections:5}')
    int maxIdleConnections
 
-   @Value('${okHttpClient.connectionPool.keepAliveDurationMs:300000}')
+   @Value('${okHttp3Client.connectionPool.keepAliveDurationMs:300000}')
    int keepAliveDurationMs
 
-   @Value('${okHttpClient.retryOnConnectionFailure:true}')
+   @Value('${okHttp3Client.retryOnConnectionFailure:true}')
    boolean retryOnConnectionFailure
```
robzienert: I'm not so sure we want to change the namespacing of these configs, I feel like it'd be really hard for our OSS users to track down potential regressions in config.

dreynaud: what happens if we have two `@ConfigurationProperties` classes that use the same `prefix=okHttpClient`?

PR #2220, opened 2018-05-09
feat(monitoring): Allow for `executions.running` to be tracked for a specific set of applications

ajordens:
Calculating this metric for orchestrations can be a somewhat heavy
operation and has been left as an explicit opt-in.

```
pollers:
  enhancedMonitoring:
    enabled: true
    intervalMs: 60000
    applications:
      - my_application_foo
      - my_application_bar
```

In such a configuration, the # of `RUNNING` orchestrations will be
determined every 60s and emitted as the `executions.running` metric.

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/config/EnhancedMonitoringConfiguration.java
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/config/EnhancedMonitoringConfiguration.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -50,6 +52,7 @@ import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.
 import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.PIPELINE
 import static java.time.ZoneOffset.UTC
 
+@Slf4j
```
ajordens: We were using `log` in here that would have eventually NPE'd

orca-web/src/main/groovy/com/netflix/spinnaker/config/EnhancedMonitoringConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+      Id applicationSpecificId = runningOrchestrationsId.withTag("application", application);
+      orchestrationCountPerApplication.put(
+        application,
+        registry.gauge(applicationSpecificId, new AtomicLong(0))
+      );
+    }
+  }
+
+  @Scheduled(fixedDelayString = "${pollers.enhancedMonitoring.intervalMs:60000}")
+  void refresh() {
```
ajordens: Runs on every instance (because every instance can publish the metrics) ... seems less than ideal to me.

Alternatively could lock and write something to `redis` that is than read when emitting the metric. 

Not really sure it's going to be problematic so I didn't go that far initially ... should I?

robzienert: Yeah I don't think it's a problem as-is.

PR #2225, opened 2018-05-09
fix(queue): Allow restart of a failed branch while pipeline is running

robfletcher:
AKA: the Cadmium exception

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
@@ -16,7 +16,7 @@
 
 package com.netflix.spinnaker.orca.q.handler
 
-import com.netflix.spinnaker.orca.ExecutionStatus
+import com.netflix.spinnaker.orca.ExecutionStatus.RUNNING
 import com.netflix.spinnaker.orca.exceptions.ExceptionHandler
 import com.netflix.spinnaker.orca.ext.parent
 import com.netflix.spinnaker.orca.pipeline.model.Execution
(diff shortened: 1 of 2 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
```diff
@@ -89,15 +90,16 @@ internal interface OrcaMessageHandler<M : Message> : MessageHandler<M> {
...
+    val configId = pipelineConfigId
+    return when {
+      !isLimitConcurrent -> false
+      configId == null   -> false
+      else               -> {
+        val criteria = ExecutionCriteria().setLimit(2).setStatuses(RUNNING)
+        repository
+          .retrievePipelinesForPipelineConfigId(configId, criteria)
+          .any { it.id != id }
+      }
```
robfletcher: I tried to make this code a little clearer but basically the logic is the same as before _except that_ it will return `false` if the only pipeline with the same config id that's running is the one you're asking about.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/StartExecutionHandlerTest.kt
```diff
@@ -275,7 +274,10 @@ object StartExecutionHandlerTest : SubjectSpek<StartExecutionHandler>({
           pipeline.isLimitConcurrent = true
           runningPipeline.isLimitConcurrent = true
 
-          whenever(pendingExecutionService.depth(configId)) doReturn 1
+          whenever(
+            repository.retrievePipelinesForPipelineConfigId(eq(configId), any())
+          ) doReturn listOf(runningPipeline)
+
```
robfletcher: I think this was a merge goof, that `depth` method is irrelevant here.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/StartExecutionHandlerTest.kt
```diff
@@ -306,7 +308,36 @@ object StartExecutionHandlerTest : SubjectSpek<StartExecutionHandler>({
...
+          subject.handle(message)
+        }
+
+        it("starts the new pipeline") {
+          verify(repository).updateStatus(PIPELINE, message.executionId, RUNNING)
+          verify(queue).push(isA<StartStage>())
+        }
+      }
+
+      and("the pipeline is not allowed to run concurrently but the only pipeline already running is the same one") {
```
robfletcher: test for the new behavior

PR #2232, opened 2018-05-10
feat(redis): Add index for buffered executions

robzienert:


orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
```diff
@@ -510,12 +519,16 @@ Execution retrieveOrchestrationForCorrelationId(
...
+    List<Observable<Execution>> observables = allRedisDelegates().stream()
+      .map(d -> Arrays.asList(
+        retrieveObservable(PIPELINE, allBufferedExecutionsKey(PIPELINE), queryAllScheduler, d),
+        retrieveObservable(ORCHESTRATION, allBufferedExecutionsKey(ORCHESTRATION), queryAllScheduler, d)
+      ))
+      .flatMap(List::stream)
+      .collect(Collectors.toList());
+
+    return Observable.merge(observables)
+      .filter(e -> e.getStatus() == BUFFERED)
```
robzienert: I kept the filter here in case this method is called while some other executions get their statuses updated out of band. Maybe not needed.

PR #2233, opened 2018-05-10
fix(expressions): improved evaluateExpression api

cfieber:
* include error details on failure
* enable helper functions (allowUnknownKeys=true)

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
@@ -294,9 +294,9 @@ class TaskController {
     def evaluated = contextParameterProcessor.process(
       [expression: expression],
       [execution: execution],
-      false
+      true
     )
-    return [result: evaluated?.expression]
+    return [result: evaluated?.expression, detail: evaluated?.expressionEvaluationSummary]
   }
 
   @PreAuthorize("hasPermission(#application, 'APPLICATION', 'READ')")


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -294,9 +294,9 @@ class TaskController {
     def evaluated = contextParameterProcessor.process(
       [expression: expression],
       [execution: execution],
-      false
+      true
```
cfieber: `true` here to enable helper functions (`stage`, `deployedServerGroups`, etc)

christopherthielen: wonder why this was false to begin with?

robzienert: @jeyrschabu Probably knows

tomaslin:
LGTM

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
cfieber: @danielpeach Any concern with this? (you added it w/ false in #1336 - not sure what the exact use case was there)

danielpeach: nope, no memory of why though

PR #2234, opened 2018-05-11
Okhttp3 oui oui

dreynaud:
config(rpc): add an OkClient bean (in addition to Ok3Client)
    
    This allows FiatAuthenticationConfig to find a client it expects, while
    connections to echo, front50 and clouddriver go through the okhttp3 client
    (as confirmed with breakpoints).
    
    Other notable change: Ok3Client now has singleton scope, because that's
    what the okhttp3 documentation recommends (because each client now has
    its own connection pool instead of sharing a global singleton pool).

### DIFF
--- a/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/RetrofitConfiguration.groovy
+++ b/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/RetrofitConfiguration.groovy
@@ -47,18 +47,18 @@ import java.util.concurrent.TimeUnit
 @EnableConfigurationProperties
 class RetrofitConfiguration {
 
-   @Value('${okHttp3Client.connectionPool.maxIdleConnections:5}')
+   @Value('${okHttpClient.connectionPool.maxIdleConnections:5}')
    int maxIdleConnections
 
-   @Value('${okHttp3Client.connectionPool.keepAliveDurationMs:300000}')
+   @Value('${okHttpClient.connectionPool.keepAliveDurationMs:300000}')
    int keepAliveDurationMs
 
-   @Value('${okHttp3Client.retryOnConnectionFailure:true}')
+   @Value('${okHttpClient.retryOnConnectionFailure:true}')
    boolean retryOnConnectionFailure
 
    @Bean(name = ["retrofitClient"])
    @Scope(ConfigurableBeanFactory.SCOPE_SINGLETON)
-   Ok3Client retrofitClient(OkHttp3ClientConfiguration okHttpClientConfig) {
+   Ok3Client ok3Client(OkHttp3ClientConfiguration okHttpClientConfig) {
      final String userAgent = "Spinnaker-${System.getProperty('spring.application.name', 'unknown')}/${getClass().getPackage().implementationVersion ?: '1.0'}"
      OkHttpClient.Builder builder = okHttpClientConfig.create()
      builder.addNetworkInterceptor(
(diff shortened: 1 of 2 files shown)


orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/RetrofitConfiguration.groovy
```diff
@@ -34,40 +37,60 @@ import org.springframework.context.annotation.Scope
...
+
 @Configuration
 @CompileStatic
-@Import(OkHttpClientConfiguration)
+@Import(OkHttp3ClientConfiguration)
 @EnableConfigurationProperties
 class RetrofitConfiguration {
 
-   @Value('${okHttpClient.connectionPool.maxIdleConnections:5}')
+   @Value('${okHttp3Client.connectionPool.maxIdleConnections:5}')
```
ajordens: Not sure these need to change (it's a breaking config change)

orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/RetrofitConfiguration.groovy
```diff
@@ -34,40 +37,60 @@ import org.springframework.context.annotation.Scope
...
-   @Value('${okHttpClient.retryOnConnectionFailure:true}')
+   @Value('${okHttp3Client.retryOnConnectionFailure:true}')
    boolean retryOnConnectionFailure
 
-   @Bean(name = ["retrofitClient", "okClient"])
-   @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
-   OkClient retrofitClient(@Qualifier("okHttpClientConfiguration") OkHttpClientConfiguration okHttpClientConfig) {
+   @Bean(name = ["retrofitClient"])
+   @Scope(ConfigurableBeanFactory.SCOPE_SINGLETON)
+   Ok3Client retrofitClient(OkHttp3ClientConfiguration okHttpClientConfig) {
```
ajordens: since we're naming beans explicitly ... perhaps more consistent to just call this `ok3Client(...)` and the other `okClient(...)` (rather than `retrofitClient` and `oldRetrofitClient`)?

orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/RetrofitConfiguration.groovy
```diff
@@ -34,40 +37,60 @@ import org.springframework.context.annotation.Scope
...
 import org.springframework.core.annotation.Order
 import retrofit.RestAdapter.LogLevel
-import retrofit.client.Client
 import retrofit.client.OkClient
 
+import java.util.concurrent.TimeUnit
+
 @Configuration
 @CompileStatic
-@Import(OkHttpClientConfiguration)
```
ajordens: we don't need to import the old one at all?

orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/RetrofitConfiguration.groovy
```diff
@@ -52,22 +56,41 @@ class RetrofitConfiguration {
    @Value('${okHttpClient.retryOnConnectionFailure:true}')
    boolean retryOnConnectionFailure
 
-   @Bean(name = ["retrofitClient", "okClient"])
-   @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
-   OkClient retrofitClient(@Qualifier("okHttpClientConfiguration") OkHttpClientConfiguration okHttpClientConfig) {
+   @Bean(name = ["retrofitClient"])
+   @Scope(ConfigurableBeanFactory.SCOPE_SINGLETON)
```
robzienert: Was SINGLETON intentional instead of PROTOTYPE? I don't totally understand why we'd have the old OkClient as a PROTOTYPE to begin with, but switching to a SINGLETON is a pretty dramatic departure from prior functionality.

dreynaud: I know, seems a bit scary, eh? But I guess that's the right thing to do if you believe the okhttp3 docs:

```
There is no longer a global singleton connection pool. In OkHttp 2.x, all OkHttpClient instances shared a common connection pool by default. In OkHttp 3.x, each new OkHttpClient gets its own private connection pool. Applications should avoid creating many connection pools as doing so prevents connection reuse. Each connection pool holds its own set of connections alive so applications that have many pools also risk exhausting memory!

The best practice in OkHttp 3 is to create a single OkHttpClient instance and share it throughout the application. Requests that needs a customized client should call OkHttpClient.newBuilder() on that shared instance. This allows customization without the drawbacks of separate connection pools.
```

robzienert: Wonderful. That's exactly the sort of behavior I would want out of a connection pool anyway! Thanks.

PR #2242, opened 2018-05-16
Revert "feat(core): revert support for parallel after stages (#2194)"

danielpeach:
Do not merge.

This reverts commit f99de659b08ae9afe3b0e0b66c88a2d578e075e8.

Was reverted because the parent stage never completed, but I haven't been able to reproduce this. Would appreciate a look over to see if there is an obvious situation that I'm not thinking of that would produce the bug.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ContinueParentStageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ContinueParentStageHandler.kt
@@ -17,15 +17,12 @@
 package com.netflix.spinnaker.orca.q.handler
 
 import com.netflix.spinnaker.orca.ExecutionStatus.NOT_STARTED
-import com.netflix.spinnaker.orca.ext.allBeforeStagesComplete
-import com.netflix.spinnaker.orca.ext.anyBeforeStagesFailed
-import com.netflix.spinnaker.orca.ext.firstAfterStages
-import com.netflix.spinnaker.orca.ext.hasTasks
+import com.netflix.spinnaker.orca.ext.*
 import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner.STAGE_BEFORE
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
 import com.netflix.spinnaker.orca.q.CompleteStage
 import com.netflix.spinnaker.orca.q.ContinueParentStage
-import com.netflix.spinnaker.orca.q.StartStage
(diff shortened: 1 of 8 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ContinueParentStageHandler.kt
```diff
@@ -46,14 +43,23 @@ class ContinueParentStageHandler(
...
-        when {
-          stage.hasTasks() -> stage.runFirstTask()
-          else -> queue.push(CompleteStage(stage))
+      if (message.phase == STAGE_BEFORE) {
+        if (stage.allBeforeStagesComplete()) {
+          when {
+            stage.hasTasks() -> stage.runFirstTask()
+            else             -> queue.push(CompleteStage(stage))
+          }
+        } else if (!stage.anyBeforeStagesFailed()) {
```
lwander: What if this is true? Do we have to handle the failure here?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ContinueParentStageHandler.kt
```diff
@@ -63,18 +69,7 @@ class ContinueParentStageHandler(
     if (firstTask.status == NOT_STARTED) {
       queue.push(StartTask(this, firstTask))
     } else {
-      log.warn("Ignoring $messageType for ${id} as tasks are already running")
-    }
-  }
-
-  private fun Stage.runAfterStages() {
```
lwander: Why was this deleted?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ContinueParentStageHandler.kt
danielpeach: I think the `ContinueParentStage` message is only sent in the case where a synthetic stage doesn't fail (the failing synthetic stage would end up completing the execution).

I think this handles the case where a `ContinueParentStage` message is sent -> a sibling synthetic stage fails -> this handler picks up the `ContinueParentStage` message and should do nothing because the execution is going to be completed by a different handler.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ContinueParentStageHandler.kt
danielpeach: Don't know, this was in the original revert. It's not used.

duftler:
@ajordens @robfletcher Any objections to merging this one?

ajordens:
Merge it and we'll see what happens.

PR #2251, opened 2018-05-17
feat(alerting): detect zombie executions

robfletcher:


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
```diff
@@ -176,4 +206,13 @@ class AtlasQueueMonitor
    */
   private val MessageNotFound.counter: Counter
     get() = registry.counter("queue.message.notfound")
+
+  private fun hasBeenAroundAWhile(execution: Execution): Boolean =
+    Instant.ofEpochMilli(execution.buildTime!!)
+      .isBefore(clock.instant().minus(10, MINUTES))
```
robzienert: Would like to see this threshold configurable. I imagine we'll want to tweak this as we go along.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
```diff
@@ -71,6 +84,23 @@ class AtlasQueueMonitor
...
+      .mergeWith(repository.retrieve(ORCHESTRATION))
+      .subscribeOn(zombieCheckScheduler.orElseGet(Schedulers::io))
+      .filter { it.status == RUNNING }
+      .filter(this::hasBeenAroundAWhile)
+      .filter(this::queueHasNoMessages)
+      .doOnCompleted {
+        log.info("Completed zombie check in ${Duration.between(startedAt, clock.instant())}")
+      }
+      .subscribe {
+        registry.counter("zombie.execution", "id", it.id).increment()
```
robzienert: We definitely don't want do add the execution ID as a tag. Would much prefer a simple counter, and then log more specific information.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
robfletcher: Yep

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
robfletcher: It would be nice if there was enough info in the alert to at least point you at something useful. Application? Pipeline name?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
```diff
@@ -71,6 +84,23 @@ class AtlasQueueMonitor
     _lastState.set(queue.readState())
   }
 
+  @Scheduled(fixedDelayString = "\${queue.zombie.check.frequency:3600000}")
+  fun checkForZombies() {
+    val startedAt = clock.instant()
+    repository.retrieve(PIPELINE)
+      .mergeWith(repository.retrieve(ORCHESTRATION))
```
robfletcher: I'd really like to make methods to specifically fetch running pipelines. Wouldn't make much difference for Redis since we'd still likely filter in memory but it would be much zippier with SQL.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
robfletcher: Well, no. Orchestrations do have a name though, but if you're worried about tag bloat it's probably not a good idea to use that.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
robzienert: Yes, 💯 . Just a `retrieve(ExecutionType, ExecutionCriteria)` method would have a lot of utility.

robfletcher: Ok, gimme a minute & I'll make that happen.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
```diff
@@ -71,6 +88,27 @@ class AtlasQueueMonitor
...
+      .filter(this::queueHasNoMessages)
+      .doOnCompleted {
+        log.info("Completed zombie check in ${Duration.between(startedAt, clock.instant())}")
+      }
+      .subscribe {
+        val tags = mutableListOf<Tag>(BasicTag("application", it.application))
+        if (it.type == PIPELINE && it.pipelineConfigId != null) {
+          tags.add(BasicTag("name", it.name))
+        }
+        registry.counter("zombie.execution", tags).increment()
```
robfletcher: Tag with application and pipeline name if it's a pipeline that exists in fiddy

ajordens: I'd suggest not including the pipeline name ... that's potentially going to blow up big

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
```diff
@@ -71,6 +88,27 @@ class AtlasQueueMonitor
...
+      .filter(this::queueHasNoMessages)
+      .doOnCompleted {
+        log.info("Completed zombie check in ${Duration.between(startedAt, clock.instant())}")
+      }
+      .subscribe {
+        val tags = mutableListOf<Tag>(
+          BasicTag("application", it.application),
+          BasicTag("type", it.type.name)
+        )
+        registry.counter("zombie.execution", tags).increment()
```
robfletcher: Ok, just application and tag.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
```diff
@@ -42,7 +56,11 @@ class AtlasQueueMonitor
 @Autowired constructor(
   private val queue: MonitorableQueue,
   private val registry: Registry,
-  private val clock: Clock
+  private val repository: ExecutionRepository,
+  private val clock: Clock,
+  @Value("\${queue.zombie.check.enabled:false}")private val zombieCheckEnabled: Boolean,
```
ajordens:
```
queue:
  zombie:
    check:
      enabled: true
```

Is that what we'd like?

```
queue:
  zombieDetection:
    enabled: true 
    frequency: ...
```

Not sure if ^^ is better or not?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
```diff
@@ -71,6 +89,29 @@ class AtlasQueueMonitor
     _lastState.set(queue.readState())
   }
 
+  @Scheduled(fixedDelayString = "\${queue.zombie.check.frequency:3600000}")
```
ajordens: `frequency` vs `frequencyMs`

ajordens: We also have other examples in `orca.yml` where we're calling this `intervalMs` 🤷‍♂️

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitorTest.kt
```diff
@@ -55,14 +65,15 @@ object AtlasQueueMonitorTest : SubjectSpek<AtlasQueueMonitor>({
     on { counter(eq("queue.duplicate.messages"), anyVararg<String>()) } doReturn duplicateCounter
     on { counter("queue.lock.failed") } doReturn lockFailedCounter
     on { timer("queue.message.lag") } doReturn messageLagTimer
+    on { counter(eq("zombie.execution"), any<Iterable<Tag>>()) } doReturn zombieCounter
```
ajordens: `queue.zombies` ? 

`queue.zombie.executions`?

Figure keeping the tag names somewhat consistent for queues might prove beneficial

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/AtlasQueueMonitor.kt
robfletcher: ok, mr picky

ajordens: it's not like I suggested putting a space before `private val`

PR #2253, opened 2018-05-18
fix(pipeline/expressions): reference trigger as map, not object

lwander:
`buildExecutionContext` is only called by stages trying to evaluate SPeL inside a stage on some input text. The issue is, at this point in the execution, the trigger has already been cast to something other than the generic `Trigger.class`, which some expressions depend on. This explains why typical SPeL evaluation doesn't run into this problem, and only the "ContitionalOnExpression" and "DeployManifest" stages see this.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.pipeline.util;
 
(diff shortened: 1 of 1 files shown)


jtk54:
LGTM, needs a rebase.

lwander:
Going to wait to let @robfletcher chime in

avram:
Does this fix https://github.com/spinnaker/spinnaker/issues/2700?

lwander:
Yes assuming that you're seeing that behavior only in the deploy manifest stage when referencing an artifact to deploy.

lwander:
paging @robfletcher or @robzienert

robzienert:
This seems right to me.

PR #2259, opened 2018-05-22
chore(queue): Record metrics on task timeouts

robzienert:


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -220,6 +218,11 @@ class RunTaskHandler(
     }
   }
 
+  private fun Registry.timeoutCounter(stageType: String, taskType: String) =
+    counter(
+      createId("queue.task.timeouts").withTags("stageType", stageType, "taskType", taskType)
```
ajordens: I wonder if there's value in `application` or `executionType`? (it'd be fair to say we don't care about applications in the metric and can just look at logs instead). 

Otherwise LGTM.

robzienert: Good thought! Will add both, I can see them both being useful in metrics.

PR #2221, opened 2018-05-09
feat(provider/ecs): added support for non-ecr docker repositories

BrunoCarrier:
feat(provider/ecs): added support for non-ecr docker repositories

@robzienert @ajordens

PR #2268, opened 2018-05-25
fix(caching): Fix bug resetting stageData after successful cache refresh

ezimanyi:
If an on-demand cache update returns a 200 response, we exit immediately
from the ServerGroupCacheForceRefreshTask without needing to poll. In
that code path, we currently don't reset the stage data on server groups
to refresh, so a later ServerGroupCacheForceRefreshTask in the same
stage uses the prior (now stale) data.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
@@ -112,19 +112,29 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
       return Optional.empty()
     }
 
-    def status = RUNNING
+    boolean allUpdatesApplied = false
     refreshableServerGroups.each { Map<String, String> model ->
       try {
         def response = cacheService.forceCacheUpdate(cloudProvider, REFRESH_TYPE, model)
         if (response.status == HttpURLConnection.HTTP_OK) {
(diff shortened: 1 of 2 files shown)


ezimanyi:
The primary fix here is the failure to call `stageData.reset()` when returning `SUCCESS` directly from `performForceCacheRefresh`.  (This change already happens when we return after at least one polling cycle since the change to `execute` in https://github.com/spinnaker/orca/commit/cf82ebab2dab0936b6f6d831e334775493ade479#diff-1a970c91247ddfba97c681a020e91b2a.)

This was resulting in cases where a later call to `ServerGroupCacheForceRefereshTask` in the same stage failed to actually refresh the cache (due to its relying on the results from the prior stage).

I also updated it so that `performForceCacheRefresh` only returns `SUCCEEDED` if *all* cache refreshes were applied immediately.  I wasn't hitting this particular case (and am not really sure if there would ever be a case where some server groups update synchronously and others asynchronously in the same stage), but it looked like the current behavior was a bug.

ezimanyi:
The Travis failure here looks unrelated to the change; the same failure is seen on https://github.com/spinnaker/orca/pull/2264 and https://github.com/spinnaker/orca/pull/2221.

I'll try to figure out what's going on here

ezimanyi:
Looks like the test failure is being fixed in https://github.com/spinnaker/orca/pull/2269; should pass here once that merges and I rebase.

robzienert:
Sorry. :'(

ezimanyi:
I accidentally committed part of my debugging/trying to fix the broken unit test on the last commit; reverted that now as the unit test was fixed in master.  This PR no longer touches the `ServerGroupCacheForceRefreshTaskSpec.groovy` tests, which I had never intended it to anyway.

PR #2271, opened 2018-05-29
refactor(retrofit): PoC Retrofit 2 in orca-bakery

robfletcher:
Not necessarily for merge but more an RFC. This is kinda what things might/could look like if we adopted Retrofit 2.


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/BakeryApi.java
```diff
@@ -0,0 +1,50 @@
...
+  Call<BakeStatus> lookupStatus(
+    @Path("region") String region, @Path("statusId") String statusId);
+
+  @GET("/api/v1/{region}/bake/{bakeId}")
+  Call<Bake> lookupBake(
+    @Path("region") String region, @Path("bakeId") String bakeId);
+
+  //
+  // Methods below this line are not supported by the Netflix Bakery, and are only available
+  // iff bakery.roscoApisEnabled is true.
```
robzienert: iff

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/BakeryService.java
```diff
@@ -0,0 +1,25 @@
...
+
+  BakeStatus createBake(String region, BakeRequest bake, String rebake);
+
+  BakeStatus lookupStatus(String region, String statusId);
+
+  Bake lookupBake(String region, String bakeId);
+
+  //
+  // Methods below this line are not supported by the Netflix Bakery, and are only available
+  // iff bakery.roscoApisEnabled is true.
```
robzienert: iff

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/RetrofitBakeryService.java
```diff
@@ -0,0 +1,66 @@
...
+  @Override
+  public BakeStatus createBake(String region, BakeRequest bake, String rebake) {
+    return retrofit(() -> api.createBake(region, bake, rebake));
+  }
+
+  @Override public BakeStatus lookupStatus(String region, String statusId) {
+    return retrofit(() -> api.lookupStatus(region, statusId));
+  }
+
+  @Override public Bake lookupBake(String region, String bakeId) {
```
robzienert: Your editor does some really, really weird stuff with annotations.

orca-retrofit2/orca-retrofit2.gradle
```diff
@@ -0,0 +1,7 @@
+apply from: "$rootDir/gradle/kotlin.gradle"
+
+dependencies {
+  compile project(":orca-core")
+  compile "com.squareup.retrofit2:retrofit:2.4.0"
```
robzienert: Since we already bring in okhttp3, we'll want to exclude retrofit's transitive dep on it so we don't have two versions floating around.

robfletcher:
Where do we leak okhttp3? Other than the fact we have a dependency on it.

robzienert:
We aren't, I tricked myself into thinking `HttpException` was okhttp3, but it's not. Sorry for the confusion.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/RetrofitBakeryService.java
```diff
@@ -0,0 +1,66 @@
...
+    return retrofit(() -> api.getBaseImage(cloudProvider, imageId));
+  }
+
+  private <T> T retrofit(Supplier<Call<T>> call) {
+    try {
+      Response<T> response = call.get().execute();
+      if (response.isSuccessful()) {
+        return response.body();
+      } else {
+        throw new HttpException(response);
```
ajordens: does retrofit2 break the contract of throwing a `RetrofitError` on non-200?

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/RetrofitBakeryService.java
```diff
@@ -0,0 +1,66 @@
...
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
ajordens: This probably warrants something a little cleaner ... brief look at docs show an `IOException` being somewhat equivalent to a `NETWORK` error.

Not sure if trying to replicate a similar exception is worthwhile ... https://gist.github.com/koesie10/bc6c62520401cc7c858f#file-retrofitexception-java

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/MonitorBakeTask.groovy
```diff
@@ -48,16 +48,16 @@ class MonitorBakeTask implements OverridableTimeoutRetryableTask {
...
       if (isCanceled(newStatus.state) && previousStatus.state == BakeStatus.State.PENDING) {
         log.info("Original bake was 'canceled', re-baking (executionId: ${stage.execution.id}, previousStatus: ${previousStatus.state})")
         def rebakeResult = createBakeTask.execute(stage)
         return new TaskResult(ExecutionStatus.RUNNING, rebakeResult.context, rebakeResult.outputs)
       }
 
       new TaskResult(mapStatus(newStatus), [status: newStatus])
-    } catch (RetrofitError e) {
-      if (e.response?.status == 404) {
+    } catch (HttpException e) {
```
ajordens: We'll something similar to the `RetrofitExceptionHandler` (https://github.com/spinnaker/orca/blob/master/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/RetrofitExceptionHandler.groovy) that speaks `HttpException` and whatever we wrap network errors with.

ajordens:
Looks reasonable to me once we come to some consensus on propagating different types of exceptions.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/RetrofitBakeryService.java
robzienert: I hadn't thought about the migration pain that comes with having to support both `RetrofitError` and `HttpException`. From a shared lib perspective, we've got... `RetrySupport` -- anything else? Too bad Java doesn't support implicit interfaces. :(

I'm not attached to the exact `RetrofitError` impl, but I do think wrapping `HttpException` / `IOException` in a common exception would be valuable.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/RetrofitBakeryService.java
robfletcher: Yep

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/RetrofitBakeryService.java
robfletcher: Yeah, I deliberately didn't sink too much time into this. We need to have a good exception strategy here. I really just wanted to see what you guys thought of this approach of wrapping a Retrofit interface with a (very similar) service interface.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/BakeryApi.java
robfletcher: Comment was just copied from existing `BakeryService`

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/RetrofitBakeryService.java
ajordens: Thus far `RetrySupport` doesn't really factor into anything retrofit-related. 

At least as far as `orca` goes, it should be reasonably easy to track down everywhere that RetrofitError is currently being used.

In 4 years of hindsight, it would have been nice to have wrapped RetrofitError with something of our own.

orca-retrofit2/src/main/kotlin/com/netflix/spinnaker/orca/retrofit2/NetworkingException.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package com.netflix.spinnaker.orca.retrofit2
+
+import okhttp3.Request
+import java.io.IOException
+
+class NetworkingException(
+  val request: Request,
+  cause: IOException
+) : RuntimeException(cause)
```
robzienert: Do we want this to extend `HttpException` instead? Left as-is, we'll regress in error handling behavior since `RetrofitError` encompassed both HTTP- & network-level errors.

robfletcher: I was wondering about that but I didn't want to no-op things like `code()` on the sub-class. Possibly we need an overall exception that can contain a cause that is either `HttpException` or `IOException`.

robzienert: I like that. I just don't want us to have to convert from a single `catch (RetrofitError)` to now catching two different exceptions, I feel a refactor like that will have a bunch of unintended consequences.

robzienert: The more I think about it, the more I suspect the converter that Adam linked is probably the right way forward.

robfletcher: You mean the exception handler?

robzienert: Yes.

robzienert:
What're we thinking about this PR?

PR #2272, opened 2018-06-04
feat(webhooks): Protected preconfigured webhook stages

amanya:
This PR allows protecting preconfigured webhook stages using Fiat roles.

So if we add a `permissions` section in `orca.yml` that way:

```
webhook:
  preconfigured:
  - label: Deploy PaaS
    description: Deploys using PaaS
    method: POST
    url: https://example.com/deploy/
    type: deployWithPaaS
    permissions:
      READ: 
      - some-role
      WRITE:
      - some-role
```

Will cause that when creating a new pipeline, the preconfigured webhook `Deploy PaaS` will only be shown to users that have any of the roles listed in the section `permissions.READ`

At the same time, when executing the pipeline, the `Deploy PaaS` stage will only succeed if the user has any of the roles listed in the section `permissions.WRITE`

This allows us to create preconfigured webhooks tailored to distinct teams with their own settings.


orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/pipeline/PreconfiguredWebhookStage.groovy
```diff
@@ -43,6 +55,24 @@ class PreconfiguredWebhookStage extends WebhookStage {
...
+        throw new PreconfiguredWebhookUnauthorizedException((String) user, (String) stage.type)
+      }
+    }
+
+    def user = new User()
+    if (stage.lastModified) {
+      user.setAllowedAccounts(stage.lastModified.allowedAccounts)
+      user.setUsername(stage.lastModified.user)
+
+      UserPermission.View userPermission = fiatService.getUserPermission(user.toString())
```
stewchen: This doesn't seem to be used to check the preconfigured webhook

amanya: Sorry, I left it here from an old test, removed now

PR #2278, opened 2018-06-07
feat(destroy): Destroy latest server group after failed Deploy

jasonkshar:
Similar to ed90df775b9e5870c2c6779818293c90a34edb7d but with a Destroy stage instead
Can be enabled by adding the following to a deploy stage:
```
  "destroy": {
    "onFailure": true
  },
```


lwander:
Any reason why this being merged into 1.6? Also, we don't accept features into release branches, only fixes. Could you reopen against master?

@ajordens what do you think of this idea?

ajordens:
Feature is reasonable but I'd think of it more along the lines of, I want to rollback on failure _and_ destroy the latest server group.

To that end, I'd add a `destroyLatest: true` (or something along those lines) to the existing `rollback: {}` block.

I'll take a better look at the actual code change in this PR later this morning.

jasonkshar:
I don't mind _rolling_ it into the `rollback` feature. I'll leave this up for a bit, let me know what you think. I'll reopen it against the master branch after the weekend.

willgorman:
To provide some more context on this, we are using this with DC/OS where rollback isn't supported through Spinnaker (since we don't have control over a load balancer to let us do things like disable).  If a deploy stage fails, the only thing that can be done is destroy that latest server group that was just deployed.  We could stub out some stuff for rollbacks to make them a no-op in DC/OS so if that's still the preferred way to get this in then we'll do that.

PR #2282, opened 2018-06-13
feat(kayenta): support analysis lifetimes shorter than 1 hour

sbwsg:
lifetimeHours continues to be supported for backwards compatibility.

lifetimeDuration can be passed as a string matching this format: https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-

https://github.com/spinnaker/kayenta/issues/283
Supporting UI: https://github.com/spinnaker/deck-kayenta/pull/311


sbwsg:
Updated tests to use `lifetimeDuration`

PR #2283, opened 2018-06-13
feat(kayenta): generate scopes from canary deployments

danielpeach:


orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/PropagateDeployedClusterScopeTask.kt
```diff
@@ -16,46 +16,64 @@
...
+      deployedCanaryClusters.putAll(mapOf(
+        "experimentLocation" to location,
+        "experimentServerGroup" to serverGroup
       ))
-    } else {
-      TaskResult.SUCCEEDED
     }
+
+    return TaskResult(SUCCEEDED, emptyMap<String, Any>(), mapOf(
+      "deployedCanaryClusters" to deployedCanaryClusters
```
danielpeach: For canary configs using Stackdriver, we'd like to be able to say right here that the resource type we just deployed is a `gce_instance` or `ec2_instance`. But we'd have to look up the canary config and determine that it's using Stackdriver, which is very ugly. I imagine other metric stores are going to want to do something similar.

duftler: Could the type of metric service in use be provided in advance via the stage config? Doesn't the pipeline editor already have this info when the canary stage is first configured?

orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/RunKayentaCanaryTask.kt
```diff
@@ -20,33 +20,66 @@ import com.netflix.spinnaker.orca.ExecutionStatus.SUCCEEDED
...
   private val kayentaService: KayentaService
 ) : Task {
-
-  private val log = LoggerFactory.getLogger(javaClass)
-
   override fun execute(stage: Stage): TaskResult {
     val context = stage.mapTo<RunCanaryContext>()
+
+    val scopes = stage.deployedCanaryClusters?.let { deployedClusters ->
+      // Assume for now that if we deployed canary clusters that
```
duftler: This makes sense if the plan is to hide the `locations/scopes` controls in the stage config when the stage is also handling the provisioning.

PR #2285, opened 2018-06-13
feat(destroy): Destroy latest server group after failed Deploy

jasonkshar:
Similar to ed90df775b9e5870c2c6779818293c90a34edb7d but can include a Destroy stage
Can be enabled by adding the following to a deploy stage:
```
  "rollback": {
    "onFailure": true,
    "destroyLatest": true
  }
```

Previous PR attempt: https://github.com/spinnaker/orca/pull/2278


lwander:
fyi @ajordens

jasonkshar:
so whatcha think about this? yay? nay?

robzienert:
@jasonkshar Will be merging. Didn't notice the thumbs.

PR #2294, opened 2018-06-18
feat(core): Adding custom trigger deserializer capability

robzienert:
Woowww.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/CustomTriggerDeserializerSupplier.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/CustomTriggerDeserializerSupplier.kt
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/CustomTriggerDeserializerSupplier.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.netflix.spinnaker.orca.pipeline.model.Trigger
+
+/**
+ * Provides a [predicate] & [deserializer] pair for custom trigger types.
+ *
+ * The [predicate] will return true if the [deserializer] should be used
+ * for the provided JsonNode. If more than one [predicate] returns true,
+ * the first supplier will be chosen.
+ */
+data class CustomTriggerDeserializerSupplier(
```
robfletcher: Why a data class with 2 lambdas instead of just an interface with 2 methods?

robzienert: Why an interface with 2 methods and not a data class?

robzienert:
@emjburns Ping on this guy - do we still want this behavior?

emjburns:
@robzienert yes I believe we do.

PR #2295, opened 2018-06-19
fix(auth): Avoid propagating auth to triggered pipelines

ajordens:
BREAKING CHANGE

This addresses a potential security vulnerability wherein a user could
configure a pipeline to be triggered by the completion of another
privileged pipeline.

Currently, the authentication details of the privileged pipeline would
be propagated to the triggered pipeline.

To enable service accounts for implicitly triggered pipelines, you must
enable `services.fiat.enabled`.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
@@ -52,7 +52,12 @@ class DependentPipelineStarter implements ApplicationContextAware {
   @Autowired(required = false)
   ArtifactResolver artifactResolver
 
-  Execution trigger(Map pipelineConfig, String user, Execution parentPipeline, Map suppliedParameters, String parentPipelineStageId) {
+  Execution trigger(Map pipelineConfig,
+                    String user,
+                    Execution parentPipeline,
+                    Map suppliedParameters,
+                    String parentPipelineStageId,
+                    User principal) {
     def json = objectMapper.writeValueAsString(pipelineConfig)
 
     if (pipelineConfig.disabled) {
@@ -60,8 +65,6 @@ class DependentPipelineStarter implements ApplicationContextAware {
     }
 
     log.info('triggering dependent pipeline {}:{}', pipelineConfig.id, json)
-
-    User principal = getUser(parentPipeline)
 
     pipelineConfig.trigger = [
       type                 : "pipeline",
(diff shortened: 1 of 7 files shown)


ajordens:
@spinnaker/google-reviewers -- Is there a historical reason to propagate authentication details when a pipeline is triggered on success/failure?

Strikes me as a potential vulnerability as I don't think there are any controls over whom can trigger off of whom?

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarterSpec.groovy
```diff
@@ -81,7 +81,8 @@ class DependentPipelineStarterSpec extends Specification {
       triggeredPipelineConfig,
       null /*user*/,
       parentPipeline, [:],
-      null
+      null,
+      true
```
ajordens: Just blanket updated these tests.

If there's some agreement that we should fix this, I'll add a set of tests to validate the `false` path.

lwander:
Dang @ttomsu is OOO for a while and knows most about this.... Maybe @stewchen can help?

danielpeach:
@dibyom I guess this is the reason why pipeline triggers don't have a `Run As User` field. (Agree with @ajordens that this is wrong).

ajordens:
Any concern in seeing this fixed? 

By fixed I mean not propagating authentication to a triggered pipeline (vs one that is started via a `Start Pipeline` stage).

dibyom:
This sounds like the right thing to do. But wouldn't it potentially break any current users using dependent pipelines if the dependent pipeline is touching an account/application that is protected?

duftler:
We could mark the commit as a breaking change and also mention it on slack. If the change is set to be released in 1.9, that would give anybody depending on the wrong behavior time to adapt.

danielpeach:
Should also add the `Run As User` field to pipeline triggers in Deck. Otherwise users won't have any way to fix their triggers.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -61,7 +66,7 @@ class DependentPipelineStarter implements ApplicationContextAware {
 
     log.info('triggering dependent pipeline {}:{}', pipelineConfig.id, json)
 
-    User principal = getUser(parentPipeline)
+    User principal = propagateAuthentication ? getUser(parentPipeline) : null
```
stewchen: What about the case where both pipelines have the same permissions? This will break those users

ajordens:
@danielpeach https://github.com/spinnaker/deck/pull/5473 should add the `Run As User` field to pipeline triggers ... basically c&p of what was in the other triggers.

ajordens:
Rejig'd a bit to introduce a check against `services.fiat.enabled` in the `DependentPipelineExecutionListener`. 

If `fiat` is enabled and the matching trigger a `runAsUser`, we'll propagate that.

PR #2303, opened 2018-06-28
feat(expressions): Adding support for failing stage on failed evaluations

jeyrschabu:
- fail stage on `failOnFailedExpressions==true` and failed expressions
- `failOnFailedExpressions` must be in the context and has to be `true`
- sample stage:
```
 {
      "name": "Wait",
      "refId": "1",
      "requisiteStageRefIds": [],
      "type": "wait",
      "waitTime": 120,
     "failOnFailedExpressions": true,
     "comments": "${failing spel missing closing parens"
}
```


duftler:
This is a great idea.

PR #2304, opened 2018-06-28
feat(orca-web): Support parallel jobs in tasks submitted to /ops

christopherthielen:


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -299,7 +299,10 @@ class OperationsController {
   }
 
   private Map<String, String> startTask(Map config) {
-    convertLinearToParallel(config)
+    def linear = config.stages.every { it.refId == null }
```
ajordens: Should it be `every` or `any`?

christopherthielen: I stole this code from the pipeline handler.  happy to go either way.

https://github.com/spinnaker/orca/blob/f56e9f24992a1c4d5ea1640304f56080ee46a616/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy#L133-L136

PR #2307, opened 2018-07-03
fix(bake): strip baseOs from titus bake reqs

asher:


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -144,6 +144,12 @@ class CreateBakeTask implements RetryableTask {
       artifactResolver.getBoundArtifactForId(stage, artifactId)
     }
 
+    // Workaround for deck/titusBakeStage.js historically injecting baseOs=trusty into stage definitions;
+    // baseOs is unnecessary for titus container image bakes
+    if ("titus" == requestMap.cloudProviderType) {
```
cfieber: could theoretically get away with keying this off storeType == "docker" too ?

PR #2308, opened 2018-07-05
fix(front50): Correctly extract the pipeline identifier

ajordens:
The `SavePipelineTask` will now correctly capture the identifier for a
newly created pipeline.

The `MonitorFront50Task` requires a pipeline identifier in order to
verify that a pipeline has been updated, previously was no-op'ing for
new pipelines.

relates to spinnaker/front50#333

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorFront50Task.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorFront50Task.java
@@ -40,12 +40,18 @@ public class MonitorFront50Task implements RetryableTask {
   private final Front50Service front50Service;
 
   private final int successThreshold;
+  private final int gracePeriodMs;
 
   @Autowired
   public MonitorFront50Task(Optional<Front50Service> front50Service,
-                            @Value("${tasks.monitorFront50Task.successThreshold:0}") int successThreshold) {
+                            @Value("${tasks.monitorFront50Task.successThreshold:0}") int successThreshold,
+                            @Value("${tasks.monitorFront50Task.gracePeriodMs:5000}") int gracePeriodMs) {
     this.front50Service = front50Service.orElse(null);
     this.successThreshold = successThreshold;
+
+    // some storage providers round the last modified time to the nearest second, this allows for a configurable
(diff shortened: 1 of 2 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
```diff
@@ -92,9 +92,17 @@ public TaskResult execute(Stage stage) {
...
-    if (pipeline.containsKey("id")) {
-      // a newly created pipeline will not yet have an `id`
-      outputs.put("pipeline.id", pipeline.get("id"));
+    try {
+      Map<String, Object> savedPipeline = (Map<String, Object>) objectMapper.readValue(
+        response.getBody().in(), Map.class
+      );
+      outputs.put("pipeline.id", savedPipeline.get("id"));
+    } catch (Exception e) {
+      log.error("Unable to deserialize saved pipeline, reason: ", e.getMessage());
```
ajordens: really just here to handle a situation where front50 hasn't been deployed and we need to fallback to the previous behavior of trying to fetch it out of the existing pipeline.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorFront50Task.java
```diff
@@ -88,9 +94,14 @@ public TaskResult execute(@Nonnull Stage stage) {
...
 
           Long lastModifiedTime = Long.valueOf(pipeline.get().get("updateTs").toString());
-          if (lastModifiedTime < stage.getStartTime()) {
+          if (lastModifiedTime < (stage.getStartTime() - gracePeriodMs)) {
             return TaskResult.RUNNING;
           }
+
+          try {
+            // small delay between verification attempts
+            Thread.sleep(1000);
```
christopherthielen: It retrieves the pipeline from front50 multiple times (assume it will hit various front 50 instances).  If it gets the updated pipeline 5 times in a row (with a second in between each fetch), then we're happy.   If it gets a stale copy even once, reschedule the whole the stage.

That right?

ajordens: Yar.

PR #2309, opened 2018-07-06
fix(pipelines): Remove test endpoint that was accidentally added.

joonlim:


jtk54:
Thanks @robzienert for pointing this out.

PR #2310, opened 2018-07-09
feat(redis): Adding instrumented redis execution repo

robzienert:
Pulling forward some interfaces from the -nflx SQL work and implementing a Redis-mirror of the SQL instrumented execution repository delegate.

This will allow us to get some better apples-to-apples comparisons of redis vs sql perf.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DelegatingExecutionRepository.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DelegatingExecutionRepository.kt
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DelegatingExecutionRepository.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+package com.netflix.spinnaker.orca.pipeline.persistence
+
+/**
+ * Allows writing delegating execution repository extensions, while continuing
+ * to expose the concrete internal implementation.
+ */
+interface DelegatingExecutionRepository<T : ExecutionRepository> : ExecutionRepository {
+  fun getDelegate(): T
+}
```
robzienert: This isn't used by OSS yet, but is necessary for `-nflx`'s `DualExecutionRepository` implementation. That will be pulled forward later.

PR #2312, opened 2018-07-09
fix(titus): Include `application` in context of a job run

ajordens:
This avoids doing a `Names.parse()` on the titus job name which happens
to be a UUID.

Also correct the spelling of `application`.


tomaslin:
LGTM

PR #2314, opened 2018-07-10
feat(authz):  Generate spinnaker managed service accounts based on pipeline roles.

dibyom:
Part of https://github.com/spinnaker/spinnaker/issues/3037

This adds an Orca task to generate/update service accounts if the pipeline spec contains a `roles` field.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.front50.tasks;
+
+import com.google.common.collect.ImmutableMap;
+import com.netflix.spinnaker.fiat.model.UserPermission;
+import com.netflix.spinnaker.fiat.model.resources.Role;
+import com.netflix.spinnaker.fiat.model.resources.ServiceAccount;
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator;
+import com.netflix.spinnaker.fiat.shared.FiatStatus;
+import com.netflix.spinnaker.orca.ExecutionStatus;
+import com.netflix.spinnaker.orca.RetryableTask;
+import com.netflix.spinnaker.orca.TaskResult;
+import com.netflix.spinnaker.orca.front50.Front50Service;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import javax.annotation.Nonnull;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Component;
+import retrofit.client.Response;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+/**
+ * Save a pipeline-scoped Fiat Service Account. The roles from
+ * this service account are used for authorization decisions when
+ * the pipeline is executed from an automated trigger.
+ */
+
+@Component
+@Slf4j
+public class SaveServiceAccountTask implements RetryableTask {
+
+  private static final String SERVICE_ACCOUNT_SUFFIX = "@managed-service-account";
+
+  @Autowired(required = false)
+  private FiatStatus fiatStatus;
+
+  @Autowired(required = false)
+  private Front50Service front50Service;
+
+  @Autowired(required = false)
+  private FiatPermissionEvaluator fiatPermissionEvaluator;
+
+  @Override
+  public long getBackoffPeriod() {
+    return TimeUnit.SECONDS.toMillis(1);
+  }
+
+  @Override
+  public long getTimeout() {
+    return TimeUnit.SECONDS.toMillis(30);
+  }
+
+  @Nonnull
+  @SuppressWarnings("unchecked")
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    if (!fiatStatus.isEnabled()) {
+      throw new UnsupportedOperationException("Fiat is not enabled, cannot save roles.");
+    }
+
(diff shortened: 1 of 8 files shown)


orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTaskSpec.groovy
```diff
@@ -126,4 +126,67 @@ class SavePipelineTaskSpec extends Specification {
...
+      name: 'my pipeline',
+      stages: [],
+      triggers: [
+        [
+          type: 'cron',
+          enabled: true
+        ]
+      ]
+    ]
+    def stage = new Stage(Execution.newPipeline("orca"), "whatever", [
```
danielpeach: Should be able to use the `stage` builder for tests here:

```
def stage = stage {
  context = [
    pipeline: Base64.encoder.encodeToString(...),
   "pipeline.serviceAccount": expectedRunAsUser
  ]
}
```
Stage builder: `import static com.netflix.spinnaker.orca.test.model.ExecutionBuilder.stage`.

danielpeach: Hmm I guess other tests in this spec don't use that builder

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTaskSpec.groovy
```diff
@@ -126,4 +126,67 @@ class SavePipelineTaskSpec extends Specification {
...
+      pipeline: Base64.encoder.encodeToString(objectMapper.writeValueAsString(pipeline).bytes)
+    ])
+
+    when:
+    stage.getContext().put('pipeline.serviceAccount', expectedRunAsUser)
+    front50Service.savePipeline(_) >> { Map<String, Object> newPipeline ->
+      runAsUser = newPipeline.get("triggers")?.get(0)?.get('runAsUser')
```
danielpeach: In groovy you might as well just walk down the map without `.get`: `newPipeline.triggers[0].runAsUser`.

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTaskSpec.groovy
```diff
@@ -0,0 +1,124 @@
...
+
+  def "should do nothing if no pipeline roles present"() {
+    given:
+    def pipeline = [
+      application: 'orca',
+      name: 'my pipeline',
+      stages: []
+    ]
+    def stage = new Stage(Execution.newPipeline("orca"), "whatever", [
+            pipeline: Base64.encoder.encodeToString(objectMapper.writeValueAsString(pipeline).bytes)
```
danielpeach: Indentation is off here

danielpeach: Should also be able to use the stage builder here, but up to you.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    Map<String, Object> pipeline;
+    try {
+      pipeline = (Map<String, Object>) stage.decodeBase64("/pipeline", Map.class);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("pipeline must be encoded as base64", e);
+    }
+
+    // Since no permissions are specified, the pipeline is either not using triggers
```
danielpeach: You're leaving me hanging after "either not using triggers" (something's missing)

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTaskSpec.groovy
dibyom: Good to know since this method seems to be deprecated. Changed it in the other file but leaving it as is here for consistency.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/model/FiatServiceAccount.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.List;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class FiatServiceAccount {
```
ttomsu: These domain model objects should be available in the `fiat-api` module - you probably shouldn't recreate them.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java
```diff
@@ -0,0 +1,164 @@
...
+
+  @Nonnull
+  @SuppressWarnings("unchecked")
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    if (!fiatStatus.isEnabled()) {
+      throw new UnsupportedOperationException("Fiat is not enabled, cannot save permissions.");
+    }
+
+    if (front50Service == null) {
```
ttomsu: My best estimation is that there should never be a time when `front50Service` is null, so you can probably remove this check & the `required=false` annotation above.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java
```diff
@@ -0,0 +1,164 @@
...
+      log.warn("User {} is not authorized with all roles for pipeline", user);
+      return new TaskResult(ExecutionStatus.TERMINAL);
+    }
+
+    FiatServiceAccount svcAcct = new FiatServiceAccount(generateSvcAcctName(pipeline), permissions);
+
+    // Creating a service account with an existing name will overwrite it
+    // i.e. perform an update for our use case
+    // TODO(dibyom): If permissions are unmodified, skip the create/update below.
+    Response response = front50Service.saveServiceAccount(svcAcct);
```
ttomsu: Haven't looked at Orca code in a long time, so I don't remember if there is a proper Retrofit handler when this RPC returns a 4xx/5xx status code. You may want to have a look around at other `Task`s and see if their calls to other subsystems are wrapped with try/catch blocks.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java
dibyom: The only place I see `front50.enabled` set to false is in `halconfig/orca-bootstrap.yml` 
Would requiring `front50Service` always be a problem for halyard bootstrap? @lwander

lwander: Yes it would break the bootstrapping process. Orca doesn't need front50 when only running executions.

PR #2316, opened 2018-07-11
fix(cache/ondemand): re-request missed server group cache refreshes 

emptywee:
Partially addresses https://github.com/spinnaker/spinnaker/issues/3030 and treats symptoms when clouddriver doesn't return servergroups from ondemand cache refresh which orca expects. Previously, the logic was a bit broken and orca wouldn't issue a new request in a timely manner if clouddriver lost it.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
@@ -16,8 +16,6 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.servergroup
 
-import org.codehaus.groovy.runtime.StackTraceUtils
-
 import java.time.Clock
 import java.util.concurrent.TimeUnit
 import com.fasterxml.jackson.annotation.JsonProperty
@@ -57,7 +55,7 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
 
   @Override
   TaskResult execute(Stage stage) {
-    StackTraceUtils.sanitize(new Exception()).printStackTrace()
+    
     if ((clock.millis() - stage.startTime) > autoSucceedAfterMs) {
       /*
        * If an issue arises performing a refresh, wait at least 10 minutes (the default ttl of a cache record) before
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.servergroup
 
+import org.codehaus.groovy.runtime.StackTraceUtils
+
 import java.time.Clock
 import java.util.concurrent.TimeUnit
 import com.fasterxml.jackson.annotation.JsonProperty
@@ -55,6 +57,7 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
 
   @Override
   TaskResult execute(Stage stage) {
+    StackTraceUtils.sanitize(new Exception()).printStackTrace()
     if ((clock.millis() - stage.startTime) > autoSucceedAfterMs) {
       /*
        * If an issue arises performing a refresh, wait at least 10 minutes (the default ttl of a cache record) before
@@ -62,6 +65,7 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
        *
        * Under normal operations, this refresh task should complete sub-minute.
        */
+      log.warn("Force cache refresh issue arose. Shorting the circuit.")
       return new TaskResult(SUCCEEDED, ["shortCircuit": true])
     }
 
@@ -74,10 +78,10 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
       return optionalTaskResult.get()
     }
 
-    boolean allAreComplete = processPendingForceCacheUpdates(
-      stage.execution.id, account, cloudProvider, stageData, stage.startTime
-    )
+    boolean allAreComplete = processPendingForceCacheUpdates(stage.execution.id, account, cloudProvider, stageData, stage.startTime)
+
     if (allAreComplete) {
+      log.debug("Force cache refresh is complete. Stage reset.")
       // ensure clean stage data such that a subsequent ServerGroupCacheForceRefresh (in this stage) starts fresh
       stageData.reset()
     }
@@ -115,6 +119,7 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
     boolean allUpdatesApplied = false
     refreshableServerGroups.each { Map<String, String> model ->
       try {
+        log.info("Force cache refresh POST a cache refresh request to clouddriver with model: ${model}")
         def response = cacheService.forceCacheUpdate(cloudProvider, REFRESH_TYPE, model)
         if (response.status == HttpURLConnection.HTTP_OK) {
           // cache update was applied immediately, no need to poll for completion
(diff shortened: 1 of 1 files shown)


emptywee:
Requesting @lwander @cfieber or @ajordens to review and accept/reject :)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -55,13 +55,15 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
...
   TaskResult execute(Stage stage) {
+    
     if ((clock.millis() - stage.startTime) > autoSucceedAfterMs) {
       /*
        * If an issue arises performing a refresh, wait at least 10 minutes (the default ttl of a cache record) before
        * short-circuiting and succeeding.
        *
        * Under normal operations, this refresh task should complete sub-minute.
        */
+      log.warn("Force cache refresh issue arose. Shorting the circuit.")
```
lwander: Could you log the time it took to short circuit as well?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -154,26 +157,46 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
                                                   String cloudProvider,
                                                   StageData stageData,
                                                   Long startTime) {
+    // log.debug("Force cache refresh calling clouddriver with refresh type ${REFRESH_TYPE}")
```
lwander: Please remove

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -154,26 +157,46 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
...
     def pendingForceCacheUpdates = cacheStatusService.pendingForceCacheUpdates(cloudProvider, REFRESH_TYPE)
-    boolean isRecent = (startTime != null) ? pendingForceCacheUpdates.find { it.cacheTime >= startTime } : false
+    log.debug("Force cache refresh clouddriver response was: ${pendingForceCacheUpdates}.")
+    boolean pendingFound = pendingForceCacheUpdates.find { it.cacheTime >= startTime }
+    boolean isRecent = (startTime != null) ? (pendingFound != null ) ? pendingFound : false : false
```
lwander: This is effectively doing the same as before, right? It's arguably a little harder to read due to the nested ternary and `null` check of a `boolean` type.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -154,26 +157,46 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
...
+          log.debug("Force cache refresh the model ${model} has been already processed, returning true.")
           return true
         }
 
+        if (!forceCacheUpdate) {
+          // there is no pending cache update, force it again in the event that it was missed
+          stageData.removeRefreshedServerGroup(model.serverGroup, model.region, model.account)
+          log.warn("Unable to find pending cache refresh request (model: ${model}), forcing a new cache refresh.")
+
+          return false
```
lwander: I'm confused what the change is here -- the `if (!forceCacheUpdate) {` and `def forceCacheUpdate = ...` chunks of code were just copied from below. The only thing that seems to happen differently is that we return `false` in this case earlier than we otherwise would have anyway

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
emptywee: Sure.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
emptywee: Ok.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
emptywee: Yes, this can be reverted to what it was.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
emptywee: Earlier, isRecent would be null and stageData.removeRefreshedServerGroup wouldn't be called, thus orca wouldn't POST another ondemand request to clouddriver. It would keep GET-ing cache refresh data from clouddriver, setting isRecent to null and go over and over. Other option would be adding to the `if (!isRecent)` condition, but I don't like duplicating code.

emptywee: In other words, checking for isRecent or not should be only done when we get a valid response from clouddriver (a response with the servergroup we are interested in). So I am fixing the logic here. If response did not have data with the server group we need, we assume that clouddriver lost it somewhere and we put a request for it in the next cycle.

Agreed?

lwander: Ah got it - yes that sounds better. Can @ajordens or @cfieber double check?

emptywee:
@lwander @cfieber @ajordens Also, added some performance improvement to this. Instead of waiting for another cycle to re-issue a cache refresh request to clouddriver, do it right after we have realized that. Potential savings may go up to 20s of time spent on a force cache refresh task.

emptywee:
Any feedback, guys?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -115,6 +119,7 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
     boolean allUpdatesApplied = false
     refreshableServerGroups.each { Map<String, String> model ->
       try {
+        log.info("Force cache refresh POST a cache refresh request to clouddriver with model: ${model}")
```
robzienert: I don't really understand this log message. What's the intent here? I'm happy to help massage it.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -154,26 +159,57 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
...
+          } catch (e) {
+            stageData.errors << e.message
+          }
+          return false
+        }
+
         if (!isRecent) {
           // replication lag -- there are no pending force cache refreshes newer than this particular stage ... retry in 10s
-          log.debug(
+          log.warn(
```
robzienert: I'm not sure I agree with this escalation of log level, nor the one below at the end of the function. Can you explain why you believe these should be louder?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
emptywee: This is mostly to indicate that orca posted a new FCR to clouddriver. Helpful to see what's being sent to clouddriver.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
emptywee: Probably because it should not really happen? /shrug If you feel like it is regular and typical thing, then yeah, we should leave it at the debug level.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -55,13 +55,17 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
...
   TaskResult execute(Stage stage) {
+    
     if ((clock.millis() - stage.startTime) > autoSucceedAfterMs) {
       /*
        * If an issue arises performing a refresh, wait at least 10 minutes (the default ttl of a cache record) before
        * short-circuiting and succeeding.
        *
        * Under normal operations, this refresh task should complete sub-minute.
        */
+      log.warn("After {}ms no progress has been made with the force cache refresh. Shorting the circuit.",
```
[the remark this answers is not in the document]
emptywee: Don't know what to say :)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy
```diff
@@ -74,10 +78,10 @@ class ServerGroupCacheForceRefreshTask extends AbstractCloudProviderAwareTask im
       return optionalTaskResult.get()
     }
 
-    boolean allAreComplete = processPendingForceCacheUpdates(
-      stage.execution.id, account, cloudProvider, stageData, stage.startTime
-    )
+    boolean allAreComplete = processPendingForceCacheUpdates(stage.execution.id, account, cloudProvider, stageData, stage.startTime)
+
     if (allAreComplete) {
+      log.debug("Force cache refresh is complete. Stage reset.")
```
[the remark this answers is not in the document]
emptywee: Well, personally, I find Spinnaker sources lacking of debug logging, so in order to debug it you have to clone it, add it, compile it, build an image, push it, deploy it on the cluster and start actually debug it. All that could be avoided if it had enough logging from day 1. But it's up to you guys. I am just trying to make it better.

emptywee:
Other than logging concerns, anything else to look at? My colleague also pointed out that `boolean allUpdatesApplied = false` logic seemed off. It should be set to true, and then set to false in the loop if anything is not updated. I sorta agreed, but haven't looked into it for real yet.

```
    boolean allUpdatesApplied = false
    refreshableServerGroups.each { Map<String, String> model ->
      try {
        log.info("Force cache refresh POST a cache refresh request to clouddriver with model: ${model}")
        def response = cacheService.forceCacheUpdate(cloudProvider, REFRESH_TYPE, model)
        if (response.status == HttpURLConnection.HTTP_OK) {
          // cache update was applied immediately, no need to poll for completion
          allUpdatesApplied = true
        }

        stageData.refreshedServerGroups << model
      } catch (e) {
        stageData.errors << e.message
      }
    }
```

This code (it's from the current orca source code) suggests that if any response in the loop returned HTTP_OK, all updates are applied.

ajordens:
```
log.warn(
            "No recent pending force cache refresh updates found, retrying in 10s (lag: {}ms, model: {}, executionId: {})",
            System.currentTimeMillis() - startTime,
            model,
            executionId
          )
```

^^ is an example of including the `executionId` which will help tie it all together.

I believe your fork is slightly out of date --> https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy#L115 is the change I'm referring to.

I pulled this PR locally, rebased it against master and gave it a quick sanity test with a couple pipelines. 

Everything looked as I would expect _except_ for the `Processed force cache refresh request in` log message. There's a bit of a fundamental problem with any of the calculations we're doing off of the `start` value ... turns out that is actually the stage start time and not the task start time. 

The value being reported in that log output is going to greatly exceed the actual time taken to do the force cache refresh. 

It looks to have been this way for quite some time and I wouldn't suggest trying to tackle that as part of this PR. Tidy up the other bits of logging and we can get this merged. 

A fix for determining the proper `start` would probably be along the lines of looking at `stage.tasks` and picking the `RUNNING` one with an `implementationClass` of `ServerGroupCacheForceRefreshTask`.

emptywee:
Yes, my bad. You're right. Since so many days have passed since the original PR was submitted, I will have to rebase and tidy it up according to the recommendations. Thanks.

emptywee:
@ajordens does it look okay now? hopefully gets merged and released in 1.9.x

ajordens:
Not sure what was up with travis --- I ended up pulling your commits into a new PR and merging that down (https://github.com/spinnaker/orca/pull/2367).

emptywee:
Thanks!

PR #2318, opened 2018-07-12
feat(kubernetes): Resolve docker containers passed as artifacts

ezimanyi:
Kubernetes V1 stages now allow docker containers to be specified as artifacts. Orca will resolve the artifacts and pass the resulting containers to Clouddriver.


ezimanyi:
https://github.com/spinnaker/spinnaker/issues/2962

PR #2323, opened 2018-07-13
fix(kayenta): be ok with iso-8601 non-instant format

emjburns:
We definitely specify iso-instant format, but some people (like me) definitely don't understand the difference between iso-8601 and iso-8601 instant. As a first step at making data parsing better, can we just do this simple string replace?

In the future, it would definitely be better to have a date picker/guidance/validation in the UI.


skandragon:
I think the fix is to do whatever is needed to use the ISO_OFFSET_DATE_TIME parser, which will handle both the +HH:MM format, and the Z shortcut for +00:00

PR #2329, opened 2018-07-17
fix(pipeline_template): Strings with boolean shall remain strings

xoonz:
Hello,

A template containing a deploy manifest stage with an annotation containing true gets converted to a boolean when generated for a pipeline which fails the deployment since annotations may only be strings.

Exemple:
For a template containing
```yaml
- apiVersion: extensions/v1beta1
  kind: Deployment
  metadata:
    annotations:
        prometheus.io/scrape: "true"
```

We get in the pipeline template on spinnaker
```json
"annotations": {
    "prometheus.io/scrape": "true"
}
```
And the in the generated pipeline we got
```
"annotations": {
    "prometheus.io/scrape": true
}
```
Which during the deployment got us this error
```
Annotations: ReadString: expects \" or n, but found t, error found in #10 byte of ...|/scrape\":true}
```

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/JsonRenderedValueConverter.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/JsonRenderedValueConverter.java
@@ -43,8 +43,6 @@ public class JsonRenderedValueConverter implements RenderedValueConverter {
       } catch (NumberFormatException ignored) {
         return NumberUtils.createLong(rendered);
       }
-    } else if (rendered.equals("true") || rendered.equals("false")) {
-      return Boolean.parseBoolean(rendered);
     } else if (rendered.startsWith("{{") || (!rendered.startsWith("{") && !rendered.startsWith("["))) {
       return rendered;
     }
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/YamlRenderedValueConverter.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/render/YamlRenderedValueConverter.java
@@ -27,7 +27,7 @@ import java.util.List;
 
 public class YamlRenderedValueConverter implements RenderedValueConverter {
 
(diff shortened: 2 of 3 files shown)


robzienert:
Hmm, I'm somewhat concerned what the repercussions of such a change could be. That said, there's some other use cases that shouldn't type-cast. Thanks for the submission - going to mull this over a bit.

xoonz:
From my understanding the behavior was introduced to compensate for the inability in templates to use jinja without quoting as braces conflicts with yaml syntax.

So anyone with templates and boolean variables are impacted.

Another possibility would be to translate a quoted boolean e.g. when having 
```
"'true'"
```
to translate to
```
"true"
```

robzienert:
That's correct, the behavior was introduced for Jinja's syntax, that said it was introduced at the very beginning. Making the change to remove type-casting as it exists today would be a backwards-incompatible change and I need to figure out how that would affect current users.

xoonz:
As a follow-up for this problem we stumbled upon another case of content conversion when trying to work with ambassador, a service-mesh solution that we wanted to use to split traffic between services in a canary deployment scenario. To configure it, you could either pass a yaml or json document in an annotation on a service.

We managed to make it work in a pipeline but when trying to transcript it in a template we came across the same problem with booleans here:
* Passing a quoted yaml document, the string gets converted and an error is returned about another yaml document being defined within the template
* Passing a quoted json document, the string gets converted into a object and kubectl complained about the annotation requiring a string

We managed to use yes instead of true for the original issue but I couldn't find a solution for this problem here. We stopped working with ambassador so it is not a blocking matter for us but I do think this behaviour is a real problem as well as a limitation that should be tackled.

lwander:
Hey @xoonz thank you for submitting the PR! As @robzienert said this is a tricky case where the fact that we chose Jinja is biting us in a nasty way. I don't see a good way to fix this for v1 of MPT.

For what it's worth, we're (@jtk54) starting to write a spec for MPT v2, which doesn't make use of any additional server-side rendering, so values like booleans and integers are interpreted unambiguously. If you want to track the progress I would get in touch with us on the #dev channel in slack.

xoonz:
Hello, 

Thanks for the feedback. I agree it's a breaking change so I wouldn't expect to see that fixed in a minor version. Looking forward for the new version, any broad idea of when we can expect it?

For information on how to bypass the problem (we need annotations containing integers) we are working with init containers that add the annotations.

lwander:
We should have a specification out later this quarter, and that will contain more detailed timelines/dates.

lwander:
I'm going to close this for now.

PR #2330, opened 2018-07-17
fix(caching): Poll for completion if any cache update is not immediate

ezimanyi:
Currently the ServerGroupForceRefreshTask will return SUCCEEDED if any cache refresh is applied immediately; it should only do this if all cache refreshes are applied immediately.


ezimanyi:
I claimed to have fixed this bug in https://github.com/spinnaker/orca/pull/2268#issuecomment-392154888, but turns out that PR didn't actually fix it as pointed out by @kcbaltz.  As noted in the prior comment, I'm not sure how common it would be for some server groups to return 200 and others 202; but either way it makes sense to fix this bug.

ajordens:
LGTM --- I'm not actually sure it's possible to have a situation where you would have both 200 vs 202.

PR #2334, opened 2018-07-20
fix(retrofit): Do not retry requests that cannot succeed

robzienert:
While fumbling around in our logs, I noticed a bunch of logs like this:

`retrofit.RetrofitError: Path parameter "imageId" value must not be null.`

Added a condition in the Retrofit ExceptionHandler to never retry these sort of requests since they'll never succeed anyway.

While digging, I also noticed the `GetCommitsTask` makes a call to front50 that is never used, so I tore that out.

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
@@ -74,7 +74,7 @@ class GetCommitsTask implements DiffTask {
 
     try {
       // get app config to see if it has repo info
-      repoInfo = getRepoInfo(stage.context.account, stage.context.application)
+      repoInfo = getRepoInfo(stage.context.application)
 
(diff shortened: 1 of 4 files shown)


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
```diff
@@ -123,7 +123,8 @@ class GetCommitsTask implements DiffTask {
 
       return new TaskResult(ExecutionStatus.SUCCEEDED, outputs)
     } catch (RetrofitError e) {
-      if (e.response?.status == 404) { // just give up on 404
+      if (e.response?.status == 404 || e.kind == RetrofitError.Kind.UNEXPECTED) {
+        // just give up on 404 or internal errors
         log.error("got a 404 from igor for : [repoType: ${repoInfo?.repoType} projectKey:${repoInfo?.projectKey} repositorySlug:${repoInfo?.repositorySlug} sourceCommit:${sourceInfo} targetCommit: ${targetInfo}]")
```
ajordens: error message might be confusing if it's one of the `UNEXPECTED` and not a `404`

PR #2335, opened 2018-07-20
chore(trafficguard): Add metrics to traffic guards

robzienert:
Also added some markers for things to do to Traffic Guards to improve reliability of the feature itself.

My current plan to improve Traffic Guards is to add a new purpose-specific endpoint clouddriver for the data that guards need:

1. `GET /trafficGuards/{application}/clusters/{account}/{name}` - Get a cluster.
2. `GET /trafficGuards/{application}/instanceServerGroups?instanceIds=[]` - Get a `Map<InstanceId, ServerGroup>` response.

These two endpoints will allow us to cut down the number of network calls `orca -> clouddriver` in a traffic guarded codepath by quite a lot and also allow us to get the freshest data. Within clouddriver we'll end up going direct to the cloud provider for data if we don't detect a traffic guard condition. Using this PR to collect opinions (although will probably merge once we're good on the metric).

Before going down this path at all, I've added a new metric so we know how many cases we're actually catching with traffic guards by application/account, and also updated the log message to not imply an actual failure.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
@@ -16,19 +16,23 @@
 
 package com.netflix.spinnaker.orca.clouddriver.utils;
 
-import java.util.*;
-import java.util.stream.Collectors;
+import com.google.common.collect.ImmutableMap;
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.moniker.Moniker;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.Location;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup;
 import com.netflix.spinnaker.orca.front50.Front50Service;
 import com.netflix.spinnaker.orca.front50.model.Application;
-import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -53,6 +61,7 @@ public void verifyInstanceTermination(String serverGroupNameFromStage,
                                         Location location,
                                         String cloudProvider,
                                         String operationDescriptor) {
+    // TODO rz - Expose a single `instance server groups` endpoint in clouddriver; returns a map<instanceid, servergroup>.
```
dreynaud: do you mean `map<servergroup, List<instanceId>>` like this seems to compute?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -135,13 +148,17 @@ private void verifyOtherServerGroupsAreTakingTraffic(String serverGroupName, Mon
       ).collect(Collectors.toList());
 
       log.debug(
-        "Traffic guard check has failed (cluster: {}, account: {}, context: {})",
+        "Detected traffic guard condition (cluster: {}, account: {}, context: {})",
```
dreynaud: 👌

dreynaud: alternatively, I like the text from the `TrafficGuardException` below. We could re-use that and tack on the extra context

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
robzienert: Ha yes.

PR #2338, opened 2018-07-26
feat(spel): add support for Instant.java in spel

lwander:
From jasonfillo on slack

> jasonfillo [3:25 PM]
> LocalDate is just for Date operations. Instant will do Date and Time and is tied directly to UTC which all your backend, database, business logic, whatever should probably be using


lwander:
Want to double check with someone from netflix first that this is OK @spinnaker/netflix-reviewers

PR #2340, opened 2018-07-30
refactor(webhooks): add webhook error information to the pipeline

zthroo:
Adds visibility for webhook stage failures.


orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -44,6 +44,11 @@ class CreateWebhookTask implements RetryableTask {
 
   @Override
   TaskResult execute(Stage stage) {
+    Map<String, ?> outputs = [webhook: [:]]
+    // TODO: The below parameter is deprecated and should be removed after some time
+    Map<String, ?> outputsDeprecated = [deprecationWarning: "All webhook information will be moved beneath the key 'webhook', " +
+      "and the keys 'statusCode', 'buildInfo', 'statusEndpoint' and 'error' will be removed. Please migrate today."]
```
robzienert: I think we should write a pipeline migrator inside of Front50 to do this for our users, rather than forcing manual toil on people -- a lot of users setup a pipeline and then never look at it again.

robzienert: Looks like this isn't your change, though, so won't let that prevent a merge.

PR #2341, opened 2018-07-30
fix(spel): Fix escaping of SpEL expressions

ezimanyi:
Currently the SpEL evaluator returns an error if the result contains any SpEL; this means that it's impossible to escape expressions of the form "${VARIABLE:default}" (which are common in Helm templates) so that they are left verbatim by the SpEL evaluator. Change the way that the evaluator detects partial evaluation so that the result is allowed to contain strings that look like SpEL.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.java
@@ -108,6 +108,7 @@ public class ExpressionTransform {
                                  EvaluationContext evaluationContext,
                                  ExpressionEvaluationSummary summary,
                                  Map<String, Object> additionalContext) {
+    boolean isPartiallyEvaluated = false;
     if (isExpression(source)) {
       String literalExpression = includeExecutionParameter(source.toString());
       Object result = null;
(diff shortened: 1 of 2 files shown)


ezimanyi:
@jeyrschabu : I was running into some issues escaping the string "${" in SpEL because the evaluator throws an error when the result contains "${".  I think this change shouldn't break https://github.com/spinnaker/orca/pull/2165; at the very least, all of the unit tests from that PR still pass.  Mind taking a look and letting me know if this looks good to you?  Thanks!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.java
```diff
@@ -123,6 +124,7 @@ private Object transformString(Object source,
             String value = e.getValue(evaluationContext, String.class);
             if (value == null) {
               value = String.format("${%s}", e.getExpressionString());
+              isPartiallyEvaluated = true;
```
lwander: Why does this mean it's partially evaluated?

ezimanyi: If we're in this block, we were not able to resolve a SpEL expression, so just dropped the expression verbatim back in the output---so we didn't completely evaluate the SpEL.  If there's a better term for it, I can change the variable name.  (I took the term "partially evaluated" from the unit tests in the referenced task.)

lwander: Ah got it, was caught up the canonical definition of [partial evaluation](https://en.wikipedia.org/wiki/Partial_evaluation). I'd probably pick something else, like `failedEvaluation`, `expressionUnresolved`, or something similar.

PR #2342, opened 2018-07-31
feat(traffic guards): perform check on disableInstancesInDiscovery

dreynaud:
Perform the same check as DisableInstancesTask and TerminateInstancesTask


dreynaud:
Looks good in prestaging, it blocked a disable in an ASG with only one instance, did not block it if the ASG had 2 instances

PR #2343, opened 2018-08-01
feat(webhooks): adds optional map usable in preconfigured webhooks...

zthroo:
…not visible in the UI

This adds the option to have a map that you can reference in your preconfigured webhooks that is not visible in the UI.

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -330,6 +330,7 @@ class OperationsController {
         preconfiguredProperties: it.preconfiguredProperties,
         noUserConfigurableFields: it.noUserConfigurableFields(),
         parameters: it.parameters,
+        parameterData: it.parameterData,
       ]
     }
   }
--- a/orca-web/src/test/groovy/com/netflix/spinnaker/orca/controllers/OperationsControllerSpec.groovy
+++ b/orca-web/src/test/groovy/com/netflix/spinnaker/orca/controllers/OperationsControllerSpec.groovy
@@ -644,8 +644,8 @@ class OperationsControllerSpec extends Specification {
       createPreconfiguredWebhook("Webhook #2", "Description #2", "webhook_2", null)
     ]
     preconfiguredWebhooks == [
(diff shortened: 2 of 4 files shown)


robzienert:
Let's rename this property, `parameterData` is not very descriptive of its intended functionality. Perhaps `hiddenParameters`?

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/config/PreconfiguredWebhookProperties.groovy
```diff
@@ -41,6 +41,7 @@ class PreconfiguredWebhookProperties {
     String description
     String type
     List<WebhookParameter> parameters
+    Map<String, Map<String, String>> parameterData
```
robzienert: Why isn't this `List<WebhookParameter>`? Based on the description of this PR, I would expect it to be the same as `parameters`.

zthroo: Apologies, I wasn't clear enough. This isn't adding hidden parameters, this is a map of data that can be used in expressions with the keys being parameters. An easy way to get data based on the provided parameters that you don't want to have to be entered as a parameter.

robzienert: OK, I see. Sorry if these are dumb questions, I'm not intimately knowledgable of the webhook stage.

Shouldn't there be more code involved in this change? How is this map actually used? Is the intention that this map is only used via Deck? Do we want to have this support via the API as well (my inclination would be yes).

zthroo: No problem.

No, there isn't any more code involved in this change. Our use case for this map is having it store API keys needed for posting. One of the parameters is a account name that is the key to the API key needed, and in the post request we resolve to the value. We may make this available from deck in the future to increase ease of use. Is that what you mean by having this support via the API? Having the information sent back to deck so it can use it?

lwander: Are these meant to be useable via SpEL as well? Or is this primarily a way to deal with secret configuration for the webhooks?

zthroo:
So sorry, I forgot about this. I'll get it updated.

zthroo:
All updated. :)

PR #2346, opened 2018-08-01
feat(canary): allow for multiple control/experiment deployments per c…

danielpeach:
…anary stage


orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/PropagateDeployedClusterScopeTask.kt
```diff
@@ -1,61 +0,0 @@
...
-import com.netflix.spinnaker.orca.TaskResult
-import com.netflix.spinnaker.orca.ext.mapTo
-import com.netflix.spinnaker.orca.kayenta.model.CanaryConfigScope
-import com.netflix.spinnaker.orca.kayenta.model.KayentaCanaryContext
-import com.netflix.spinnaker.orca.kayenta.model.deployments
-import com.netflix.spinnaker.orca.pipeline.model.Stage
-import org.springframework.stereotype.Component
-
-@Component
-class PropagateDeployedClusterScopeTask : Task {
```
danielpeach: This was never running, and will not compile now. I'm not sure how the later canary stages will pick up the deployed server group names yet.

orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/model/Deployments.kt
```diff
@@ -56,22 +62,64 @@ internal data class Baseline(
...
   val cloudProvider: String,
   val account: String,
-  val moniker: Moniker,
+  val moniker: Moniker?,
   val availabilityZones: Map<String, Set<String>>,
-  val region: String?
+  val region: String?,
+  val application: String?,
+  val stack: String?,
+  val freeFormDetails: String?
```
robzienert: Forgive my ignorance here, but why add these additional values if we've got moniker? Shouldn't we just try to converge around that instead of separate properties?

danielpeach: We should be converging - but looks like most of the providers don't support moniker. Deck and Clouddriver don't support monikers for either EC2 or GCE (for deploy ops, I mean).

orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/PropagateDeployedClusterScopes.kt
```diff
@@ -0,0 +1,70 @@
...
+      }
+      experiment.mapTo<DeployServerGroupContext>().deployServerGroups.entries.first().let { (location, serverGroups) ->
+        scope["experimentLocation"] = location
+        scope["experimentScope"] = serverGroups.first()
+      }
+      scope
+    }
+
+    return TaskResult(ExecutionStatus.SUCCEEDED, emptyMap<String, Any>(), mapOf(
+      "deployedClusters" to scopes
```
duftler: Can't quite wrap my head around the naming. Are we deploying clusters, or server groups?

orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/RunKayentaCanaryTask.kt
```diff
@@ -37,16 +40,45 @@ class RunKayentaCanaryTask(
...
+      ),
+      experimentScope = scope.experimentScope.copy(
+        scope = cluster.experimentScope,
+        location = cluster.experimentLocation
+      )
+    )
+  }
+}
+
+internal data class DeployedCluster(
```
duftler: Does this really describe a cluster? Looks like a pair of server groups that are likely in the same cluster I suppose.

orca-kayenta/src/test/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/PropagateDeployedClusterScopesTest.kt
```diff
@@ -0,0 +1,106 @@
...
+          }
+        }
+      }
+    }
+
+    subject.execute(pipeline.stageByRef("1")).outputs["deployedClusters"]?.let { clusters ->
+      it("summarizes deployments, joining experiment and control pairs") {
+        OrcaObjectMapper.newInstance().convertValue<List<DeployedCluster>>(clusters).let {
+          assertThat(it).containsExactlyInAnyOrder(
+            DeployedCluster(
```
duftler: Nice.

orca-kayenta/src/test/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/RunKayentaCanaryTaskTest.kt
```diff
@@ -0,0 +1,104 @@
...
+import java.time.Duration
+import java.time.Instant
+import java.util.*
+
+object RunKayentaCanaryTaskTest : Spek({
+
+  val kayenta: KayentaService = mock()
+  val subject = RunKayentaCanaryTask(kayenta)
+
+  given("an upstream stage that generated a summary of upstream control & experiment server groups") {
```
duftler: Reminds me of the naming question.

danielpeach:
@duftler renamed almost everything from cluster to server group. I kept the `CleanupCanaryClusters` stage, since it operates on clusters rather than individual server groups (it synthesizes `DisableCluster` and `DestroyCluster` stages).

PR #2349, opened 2018-08-03
fix(core): Avoid revisiting stages when traversing ancestors

mtweten:
See https://github.com/spinnaker/spinnaker/issues/2895 for the issue that introduced this change.

This change introduces a set to track visited nodes when traversing ancestors. Without this, it's possible to get into some bad scenarios depending on a pipeline's graph structure. 

Given a structure like this:
graph

The current ancestor code will end up traversing the same stages repeatedly and adding duplicate stages to the final ancestors list. As more "columns" are added to the graph, performance degrades further. 

In one of our pipelines with 6 columns of 10 stages each, we've seen the ancestors logic take nearly 30 minutes to complete when trying to view the pipeline in deck. Gate times out on the call to get pipeline executions due to hystrix timeouts, but the request continues in orca and ties up a thread for 30 minutes.

In this screenshot you can see an example execution of a pipeline with the same structure described above, in which I have the "column" stages not execute, and the single wait stages in between execute (all wait stages are set to wait for only 3 seconds). 

exampleexecution

You can see the further down the graph it gets, the longer each wait stage takes to actually execute and complete, taking far longer than the actual 3 seconds it's supposed to wait.


mtweten:
Looks like my change [here](https://github.com/spinnaker/orca/pull/2349/files#diff-b9a1511802a5c213abc5e0d0e977f8e6R366) is causing this [TargetGroupServer test](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroupSpec.groovy#L76) to fail, since that implicitly calls down into the ancestors code, and the test doesn't set up a stage with a non-null execution object.

From the Stage object, it looks like Execution is [expected to be Nonnull](https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java#L164).

Ok to just update that test to construct the Stage for the test with an Execution object to avoid the NPE?

mtweten:
Regarding Rob's comment here: https://github.com/cerner/orca/pull/3#issuecomment-410827508

> Is the time spent because ancestors gets called multiple times, or is it a single call being so inefficient? I'm wondering if we can restructure the way relationships are stored, memoize the call, or something like that in order to make things more efficient in the first place.

I did try some naive memoization code to just store the ancestors in the stage after they are calculated, but that seemed to blow up memory usage pretty bad. I didn't pursue it any further since I considered this a more elegant solution (only additional cost is storing the refIds in the visited set).

asher:
Looks good, thanks for submitting this

PR #2352, opened 2018-08-06
fix(redis): fix multi-redis configuration

dmart:
Fix 2 issues when Orca is configured with redis.connectionPrevious
  - Fix expression syntax
  - Change argument names to resolve the correct beans


jonsie:
@robzienert

PR #2353, opened 2018-08-07
feat(queue): Adding completed executions metric

robzienert:
How we didn't have this metric from the start, I'm not totally sure. Here it be, though.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.q.handler
 
+import com.netflix.spectator.api.Registry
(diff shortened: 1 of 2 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
```diff
@@ -57,6 +60,10 @@ class CompleteExecutionHandler(
           publisher.publishEvent(
             ExecutionComplete(this, message.executionType, message.executionId, status)
           )
+          registry.counter(completedId.withTags(
+            "status", status.name,
+            "executionType", execution.type.name
```
ajordens: Thoughts on adding `application` to this metric? (we have it on `executions.started`)

robzienert: I had a reason for not including it, but I can't remember, so it was probably rubbish.

PR #2355, opened 2018-08-08
feat(core): Genericize execution migrators

robzienert:
Squashes PollingAgentExecutionRepository methods into the
root ExecutionRepository interface. This separatino was more
pain that it was worth.

Pulled forward the DualExecutionRepository that was written
for Orca SQL into orca-core.

For the main show, I've refactored the two execution data
migrators so that they're persistence backend agnostic.
Combined with DualExecutionRepository, you can migrate from
one backend to another without writing new migration logic,
making the operational story a little simpler for an already
tough process.


orca-core/src/main/java/com/netflix/spinnaker/orca/notifications/scheduling/OldPipelineCleanupPollingNotificationAgent.java
```diff
@@ -169,7 +170,7 @@ private void cleanup(List<PipelineExecutionDetails> executions) {
     executions.subList(0, (executions.size() - minimumPipelineExecutions)).forEach(p -> {
       long startTime = p.startTime == null ? p.buildTime : p.startTime;
       long days = ChronoUnit.DAYS.between(Instant.ofEpochMilli(startTime), Instant.ofEpochMilli(clock.millis()));
-      if (days > thresholdDays && !executionRepository.hasEntityTags(PIPELINE, p.id)) {
```
robzienert: The `hasEntityTags` interface method no longer does anything. It was an interface added to support some Netflix functionality that never made it to OSS and has subsequently been removed from our extension. I removed it from OSS entirely as a result.

orca-migration/src/main/kotlin/com/netflix/spinnaker/orca/pipeline/persistence/migration/OrchestrationMigrationAgent.kt
```diff
@@ -0,0 +1,86 @@
...
+
+      if (pendingOrchestrations.isNotEmpty()) {
+        log.info("${pendingOrchestrations.size} orchestrations yet to complete $applicationName [$index/${allApplications.size}]")
+      }
+
+      if (migratableOrchestrations.isNotEmpty()) {
+        log.info("${migratableOrchestrations.size} orchestrations to migrate ($applicationName) [$index/${allApplications.size}]")
+
+        migratableOrchestrations.forEach {
+          val execution = dualExecutionRepository.previous.retrieve(it.type, it.id)
```
robzienert: This was an oddity of the prior migrators that I kept. I'm not totally certain why we're refreshing the execution even though we already have it, but I'm not eager to deviate from the existing logic. This choice is consistent for both migrators.

asher: This was me being stupid, safe to drop and avoid the duplicate read.

PR #2356, opened 2018-08-08
fix(clouddriver): Handle missing server groups when checking instances

ajordens:
This PR addresses a couple of situations where we were not handling a
non-existent server group.

We encountered some odd edge cases where `WaitForAllInstancesNotUpTask`
would end up getting stuck waiting on a server group that had previously
been destroyed. Easily reproduced by submitting slightly offset destroy
server group operations.

Should a server group with the same name happen to be created again, the
task would detect that and succeed (since a new server group has no UP
instances when first created!).

These tasks will now short-circuit if they're unable to find a server
group _AND_ they were expecting one to exist.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/OortHelper.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/OortHelper.groovy
@@ -58,7 +58,9 @@ class OortHelper {
                                                    String serverGroupName,
                                                    String location,
                                                    String cloudProvider) {
-    return new TargetServerGroup(convert(oortService.getServerGroup(account, location, serverGroupName) , Map))
+    return convertedResponse(Map) {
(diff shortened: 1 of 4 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/OortHelper.groovy
```diff
@@ -58,7 +58,9 @@ class OortHelper {
                                                    String serverGroupName,
                                                    String location,
                                                    String cloudProvider) {
-    return new TargetServerGroup(convert(oortService.getServerGroup(account, location, serverGroupName) , Map))
+    return convertedResponse(Map) {
```
ajordens: This method actually expects an `Optional` to be returned ... it used to be this way but Groovy somehow allowed the `return new TargetServerGroup()`.

emjburns:
LGTM

PR #2357, opened 2018-08-08
feat(bake): parse version from kork Artifacts

emjburns:
If there are `artifacts` present in the trigger, and those are the requested artifacts, parse the version out of them and then submit that to the bakery. Added some tests around this.

Refactor pass #1 at `PackageInfo` to make variable names make sense. My desire is to do a refactor #2 to make this class easier to reason about and modify. Added some comments to aid others in understanding this class.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.java
@@ -20,6 +20,7 @@ import java.util.*;
 import java.util.regex.Pattern;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.java
```diff
@@ -69,132 +87,139 @@ private boolean isUrl(String potentialUrl) {
...
-    if (buildArtifacts.isEmpty() && triggerArtifacts.isEmpty()) {
-      return request;
+    if (buildArtifacts.isEmpty() && triggerArtifacts.isEmpty() && artifacts.isEmpty()) {
+      return stageContext;
     }
 
     List<String> missingPrefixes = new ArrayList<>();
     String fileExtension = format(".%s", packageType);
 
-    // There might not be a request.package so we look for the package name from either the buildInfo or trigger
```
emjburns: Above we say ```if (stageContext.get("package") == null || stageContext.get("package").equals("") || isUrl(stageContext.get("package").toString())) {
      return stageContext;
    }```

That indicates to me that we will have already failed if there are no artifacts specified. Please let me know if I'm missing something and this actually cannot be removed!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.java
```diff
@@ -69,132 +87,151 @@ private boolean isUrl(String potentialUrl) {
...
+          + triggerArtifact.get("filename")
+          + ") and produced artifacts ("
+          + matchedArtifact.getVersion() + versionDelimiter + matchedArtifact.getVersion()
+          + ")");
+      }
+
+      if (!buildArtifact.isEmpty() && matchedArtifact != null && !extractPackageVersion(buildArtifact, prefix, fileExtension).equals(matchedArtifact.getVersion())) {
+        throw new IllegalStateException("Found build artifact in both Jenkins stage ("
+          + matchedArtifact.getVersion() + versionDelimiter + matchedArtifact.getVersion()
+          + ")and produced artifacts ("
```
cfieber: `) and produced artifacts (`

PR #2358, opened 2018-08-08
feat(locking): Adds basic locking support.

cfieber:
Introduces a lock manager and Acquire/Release lock stages.

Locking is disabled completely by default, but can be enabled
with the configuration property:
```yaml
locking:
  enabled: true
```

Locking can run in learning mode where locking
failures are non fatal and are just logged:
```yaml
locking:
  enabled: true
  learningMode: true
```

Note that this initial implementation does not wire the acquire and
release lock stages into any operations, so the only way to exercise
them is to explicitly add them into a pipeline.

An example minimal pipeline that exercises locking looks like:
```json
{
  "application": "lockingtest",
  "stages": [
    {
      "name": "Acquire Lock",
      "refId": "1",
      "requisiteStageRefIds": [],
      "type": "acquireLock",
      "lock": {
        "lockName": "testlock"
      }
    },
    {
      "name": "Wait",
      "refId": "2",
      "requisiteStageRefIds": ["1"],
      "type": "wait",
      "waitTime": 5
    },
    {
      "name": "Release Lock",
      "refId": "3",
      "requisiteStageRefIds": ["2"],
      "type": "releaseLock"
    }
  ]
}
```

This is a subset of what was initially proposed in #1799 - the core locking constructs without wiring into any existing stages. Synthetic stage construction and monikers for cluster identification changed significantly enough while #1799 was stagnating, so most of the wiring into operations needs a revisit.

closes #1799

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/TaskExecutionInterceptor.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/TaskExecutionInterceptor.java
@@ -1,0 +1,32 @@
+package com.netflix.spinnaker.orca;
+
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+
+/**
+ * TaskExecutionInterceptor is a hook point to customize the specific execution of a task.
+ *
+ * Before execute is called on a Task, all TaskExecutionInterceptors will be called. The resulting
+ * Stage object from beforeTaskExecution is passed to subsequent invocations of TaskExecutionInterceptor
+ * and then used for the invocation of execute on Task.
+ *
+ * After a Task completes with a TaskResult, all TaskExecutionInterceptors are called. The resulting
+ * TaskResult is passed to subsequent invocations ot TaskExecutionInterceptor and the final TaskResult
+ * is used as the output of the task.
+ *
+ * A TaskExecutionInterceptor can specify the maximum backoff that should be allowed. As an example, the
+ * LockExtendingTaskExecutionInterceptor needs to ensure that a task doesn't delay longer than the
+ * lock extension. The minimum maxTaskBackoff among all registered TaskExecutionInterceptors will be used
+ * to constrain the task backoff supplied by a RetryableTask.
+ */
+public interface TaskExecutionInterceptor {
+
+  default long maxTaskBackoff() { return Long.MAX_VALUE; }
+
+  default Stage beforeTaskExecution(Task task, Stage stage) {
+    return stage;
+  }
+
+  default TaskResult afterTaskExecution(Task task, Stage stage, TaskResult taskResult) {
+    return taskResult;
+  }
+}
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/locks/ExecutionCompleteLockReleasingListener.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/locks/ExecutionCompleteLockReleasingListener.java
@@ -1,0 +1,48 @@
+package com.netflix.spinnaker.orca.locks;
+
+import com.netflix.spinnaker.orca.events.ExecutionComplete;
+import com.netflix.spinnaker.orca.pipeline.AcquireLockStage;
+import com.netflix.spinnaker.orca.pipeline.model.Execution;
+import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationListener;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ExecutionCompleteLockReleasingListener implements ApplicationListener<ExecutionComplete> {
+  private final Logger logger = LoggerFactory.getLogger(getClass());
+  private final ExecutionRepository executionRepository;
+  private final LockManager lockManager;
+  private final LockingConfigurationProperties lockingConfigurationProperties;
+
+  @Autowired
+  public ExecutionCompleteLockReleasingListener(ExecutionRepository executionRepository,
+                                                LockManager lockManager,
+                                                LockingConfigurationProperties lockingConfigurationProperties) {
+    this.executionRepository = executionRepository;
+    this.lockManager = lockManager;
+    this.lockingConfigurationProperties = lockingConfigurationProperties;
+  }
+
+  @Override
+  public void onApplicationEvent(ExecutionComplete event) {
+    if (!lockingConfigurationProperties.isEnabled()) {
+      return;
+    }
+    if (event.getStatus().isHalt()) {
+      Execution execution = executionRepository.retrieve(event.getExecutionType(), event.getExecutionId());
+      execution.getStages().forEach(s -> {
+        if (AcquireLockStage.PIPELINE_TYPE.equals(s.getType())) {
+          try {
+            LockContext lc = s.mapTo("/lock", LockContext.LockContextBuilder.class).withStage(s).build();
+            lockManager.releaseLock(lc.getLockName(), lc.getLockValue(), lc.getLockHolder());
+          } catch (LockFailureException lfe) {
+            logger.info("Failure releasing lock in ExecutionCompleteLockReleasingListener - ignoring", lfe);
+          }
+        }
+      });
+    }
+  }
+}
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockContext.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockContext.java
@@ -1,0 +1,137 @@
+package com.netflix.spinnaker.orca.locks;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.netflix.spinnaker.orca.pipeline.model.Execution;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class LockContext {
+  public static class LockContextBuilder {
+
+    public static class LockValueBuilder {
+      private String application;
+      private String type;
+      private String id;
+      private Stage stage;
+
+      LockValueBuilder() {
+        this(null, null, null, null);
+      }
+
+      @JsonCreator
+      public LockValueBuilder(@JsonProperty("application") String application,
+                              @JsonProperty("type") String type,
+                              @JsonProperty("id") String id) {
+        this(application, type, id, null);
+      }
+
+      LockValueBuilder(String application, String type, String id, Stage stage) {
+        this.application = application;
+        this.type = type;
+        this.id = id;
+        this.stage = stage;
+      }
+
+      public LockValueBuilder withStage(Stage stage) {
+        this.stage = stage;
+        return this;
+      }
+
+      public LockManager.LockValue build() {
+        final Optional<Execution> execution = Optional.ofNullable(stage).map(Stage::getExecution);
+
+        final String application = Optional.ofNullable(this.application).orElseGet(() ->
+          execution
+            .map(Execution::getApplication)
+            .orElse(null));
+
+        final String type = Optional.ofNullable(this.type).orElseGet(() ->
+          execution
+            .map(Execution::getType)
+            .map(Execution.ExecutionType::toString)
+            .orElse(null));
+
+        //TODO(cfieber): lockValue.id:
+        // if we are a child pipeline / strategy we probably need to use the parent execution id
+        final String id = Optional.ofNullable(this.id).orElseGet(() ->
+          execution
+            .map(Execution::getId)
+            .orElse(null));
+
+        return new LockManager.LockValue(application, type, id);
+      }
+    }
+
+
+    private String lockName;
+    private LockValueBuilder lockValue;
+    private String lockHolder;
+    private Stage stage;
+
+    public LockContextBuilder() {
+      this(null, null, null, null);
+    }
+
+    @JsonCreator
+    public LockContextBuilder(
+      @JsonProperty("lockName") String lockName,
+      @JsonProperty("lockValue") LockValueBuilder lockValue,
+      @JsonProperty("lockHolder") String lockHolder) {
+      this(lockName, lockValue, lockHolder, null);
+    }
+
+    LockContextBuilder(String lockName, LockValueBuilder lockValue, String lockHolder, Stage stage) {
+      this.lockName = lockName;
+      this.lockValue = lockValue;
+      this.lockHolder = lockHolder;
+      this.stage = stage;
+    }
+
+    public LockContextBuilder withStage(Stage stage) {
+      this.stage = stage;
+      Optional.ofNullable(lockValue).ifPresent(lv -> lv.withStage(stage));
+      return this;
+    }
+
+    public LockContext build() {
+      final LockManager.LockValue lockValue =
+        Optional.ofNullable(this.lockValue)
+          .orElseGet(() -> new LockValueBuilder().withStage(stage))
+          .build();
+
+      final String lockHolder = Optional.ofNullable(this.lockHolder).orElseGet(() ->
+        Optional.ofNullable(stage)
+          .map(Stage::getId)
+          .orElse(null));
+
+      return new LockContext(lockName, lockValue, lockHolder);
+    }
+  }
+
+  private final String lockName;
+  private final LockManager.LockValue lockValue;
+  private final String lockHolder;
+
+  LockContext(String lockName,
+              LockManager.LockValue lockValue,
+              String lockHolder) {
+    this.lockName = Objects.requireNonNull(lockName);
+    this.lockValue = Objects.requireNonNull(lockValue);
+    this.lockHolder = Objects.requireNonNull(lockHolder);
+  }
+
+  public String getLockName() {
+    return lockName;
+  }
+
+  public LockManager.LockValue getLockValue() {
+    return lockValue;
+  }
+
+  public String getLockHolder() {
+    return lockHolder;
+  }
+}
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockExtendingTaskExecutionInterceptor.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockExtendingTaskExecutionInterceptor.java
@@ -1,0 +1,112 @@
+package com.netflix.spinnaker.orca.locks;
+
+import com.netflix.spinnaker.orca.ExecutionStatus;
+import com.netflix.spinnaker.orca.Task;
+import com.netflix.spinnaker.orca.TaskExecutionInterceptor;
+import com.netflix.spinnaker.orca.TaskResult;
+import com.netflix.spinnaker.orca.pipeline.AcquireLockStage;
+import com.netflix.spinnaker.orca.pipeline.ReleaseLockStage;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import com.netflix.spinnaker.orca.pipeline.tasks.AcquireLockTask;
+import com.netflix.spinnaker.orca.pipeline.tasks.ReleaseLockTask;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.time.Duration;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+
+@Component
+public class LockExtendingTaskExecutionInterceptor implements TaskExecutionInterceptor {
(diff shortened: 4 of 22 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockingConfigurationProperties.java
```diff
@@ -0,0 +1,53 @@
...
+  private int backoffBufferSeconds = 10;
+  private final DynamicConfigService dynamicConfigService;
+
+  @Autowired
+  public LockingConfigurationProperties(DynamicConfigService dynamicConfigService) {
+    this.dynamicConfigService = dynamicConfigService;
+  }
+
+  public boolean isLearningMode() {
+    return dynamicConfigService.getConfig(Boolean.class, "locking.learningMode", learningMode);
```
cfieber: I don't entirely hate this pattern of injecting the `DynamicConfigService` into the `@ConfigurationProperties` object

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CancelExecutionHandler.kt
```diff
@@ -51,8 +46,6 @@ class CancelExecutionHandler(
 
       // then, make sure those runTask messages get run right away
       queue.push(RescheduleExecution(execution))
-
-      publisher.publishEvent(ExecutionComplete(this, message.executionType, message.executionId, CANCELED))
```
cfieber: this was happening too early when published explicitly here - it ends up being fired from the CompleteExecutionHandler now

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -168,8 +179,8 @@ class RunTaskHandler(
   private fun Task.backoffPeriod(taskModel: com.netflix.spinnaker.orca.pipeline.model.Task, stage: Stage): TemporalAmount =
     when (this) {
       is RetryableTask -> Duration.ofMillis(
-        getDynamicBackoffPeriod(stage, Duration.ofMillis(System.currentTimeMillis() - (taskModel.startTime
-          ?: 0)))
+        Math.min(getDynamicBackoffPeriod(stage, Duration.ofMillis(System.currentTimeMillis() - (taskModel.startTime
```
cfieber: Constrains the maximum backoff so that a task won't get rescheduled too far in the future where its locks won't get extended. Essentially hijacks the WaitStage's attempt to backoff until it would be ready to complete and makes sure it gets re-run at least often enough to keep the locks alive

orca-redis/src/main/java/com/netflix/spinnaker/orca/locks/RedisLockManager.java
```diff
@@ -0,0 +1,244 @@
...
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import static java.util.Arrays.asList;
+
+@Component
+public class RedisLockManager implements LockManager {
```
cfieber: This could be refactored at some point to use the locking in kork

orca-core/src/main/java/com/netflix/spinnaker/orca/TaskExecutionInterceptor.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * After a Task completes with a TaskResult, all TaskExecutionInterceptors are called. The resulting
+ * TaskResult is passed to subsequent invocations ot TaskExecutionInterceptor and the final TaskResult
+ * is used as the output of the task.
+ *
+ * A TaskExecutionInterceptor can specify the maximum backoff that should be allowed. As an example, the
+ * LockExtendingTaskExecutionInterceptor needs to ensure that a task doesn't delay longer than the
+ * lock extension. The minimum maxTaskBackoff among all registered TaskExecutionInterceptors will be used
+ * to constrain the task backoff supplied by a RetryableTask.
+ */
```
robzienert: Big fan of docs! Love the docs.

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/ExecutionCompleteLockReleasingListener.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+    if (event.getStatus().isHalt()) {
+      Execution execution = executionRepository.retrieve(event.getExecutionType(), event.getExecutionId());
+      execution.getStages().forEach(s -> {
+        if (AcquireLockStage.PIPELINE_TYPE.equals(s.getType())) {
+          try {
+            LockContext lc = s.mapTo("/lock", LockContext.LockContextBuilder.class).withStage(s).build();
+            lockManager.releaseLock(lc.getLockName(), lc.getLockValue(), lc.getLockHolder());
+          } catch (LockFailureException lfe) {
+            logger.info("Failure releasing lock in ExecutionCompleteLockReleasingListener - ignoring", lfe);
```
robzienert: What's the effect of failing to release the lock? The log level & ignoring sounds like it's not a big deal, but not releasing sounds like a big deal.

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockContext.java
```diff
@@ -0,0 +1,139 @@
...
+    public LockContext build() {
+      final LockManager.LockValue lockValue =
+        Optional.ofNullable(this.lockValue)
+          .orElseGet(() -> new LockValueBuilder().withStage(stage))
+          .build();
+
+      //TODO(cfieber):
+      //lockHolder - if we are a synthetic stage then should we use parent stage's id?
+      // if we are a top level stage, lock with our id
+      // the corresponding release lock stage should have an explicit stage id in it..
```
robzienert: If we're a synthetic stage in say... a canary or something new and exciting, we'd want locks on individual synthetic deploy stages yes?

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockExtendingTaskExecutionInterceptor.java
```diff
@@ -0,0 +1,99 @@
...
+      LockContext lc = s.mapTo("/lock", LockContext.LockContextBuilder.class).withStage(s).build();
+      AtomicInteger count = heldLocks.computeIfAbsent(new HeldLock(lc.getLockName(), lc.getLockValue()), hl -> new AtomicInteger(0));
+      if (AcquireLockStage.PIPELINE_TYPE.equals(s.getType())) {
+        count.incrementAndGet();
+      } else {
+        count.decrementAndGet();
+      }
+    });
+
+    List<HeldLock> toExtend = heldLocks.entrySet().stream().filter(me -> me.getValue().get() > 0).map(Map.Entry::getKey).collect(Collectors.toList());
```
robzienert: Can we newline the different bits of this stream?

robzienert: Generally, I prefer things under 120 columns so I don't need to scroll.

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockFailureException.java
```diff
@@ -0,0 +1,34 @@
...
+
+  private static String buildMessage(String lockName, @Nullable LockManager.LockValue currentLockValue) {
+    Optional<LockManager.LockValue> lv = Optional.ofNullable(currentLockValue);
+    return "Failed to acquire lock " + lockName
+      + " currently held by " + lv.map(LockManager.LockValue::getApplication).orElse("UNKNOWN")
+      + "/" + lv.map(LockManager.LockValue::getType).orElse("UNKNOWN")
+      + "/" + lv.map(LockManager.LockValue::getId).orElse("UNKNOWN");
+
+
+  }
```
robzienert: Many lines.

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockingConfigurationProperties.java
robzienert: I like it.

orca-redis/orca-redis.gradle
```diff
@@ -2,6 +2,7 @@ apply from: "$rootDir/gradle/kotlin.gradle"
 apply from: "$rootDir/gradle/spock.gradle"
 
 dependencies {
+  compile project(":orca-core")
```
robzienert: What's the advantage of calling this out explicitly if it's a dependency of front50? Just explicitness?

orca-redis/src/main/java/com/netflix/spinnaker/orca/locks/RedisLockManager.java
```diff
@@ -0,0 +1,244 @@
...
+  private static final int LOCK_VALUE_APPLICATION_IDX = 0;
+  private static final int LOCK_VALUE_TYPE_IDX = 1;
+  private static final int LOCK_VALUE_IDX = 2;
+
+  private static final int JEDIS_POOL_MAX_ATTEMPTS = 3;
+  private static final int JEDIS_POOL_INITIAL_BACKOFF_MILLIS = 10;
+  private static final int JEDIS_POOL_BACKOFF_INCREMENT_MILLIS = 10;
+
+  private final Logger log = LoggerFactory.getLogger(getClass());
+  private final Pool<Jedis> jedisPool;
```
robzienert: Would like to see this use `RedisClientDelegate` instead.

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockContext.java
ajordens: ... was typing the same thing as @robzienert. 

The individual synthetic deploy stages have the execution windows attached (pretty sure?), so we'd want to lock at that level.

But there are probably more normal cases where we'd want to use the parent stage id.

orca-redis/src/main/java/com/netflix/spinnaker/orca/locks/RedisLockManager.java
```diff
@@ -0,0 +1,244 @@
...
+          kv("requestLockValue.type", lockOperation.lockValue.getType()),
+          kv("requestLockValue.id", lockOperation.lockValue.getId()),
+          kv("requestLockHolder", Optional.ofNullable(lockOperation.lockHolder).orElse("UNSPECIFIED")),
+          lfe);
+        if (lockingConfigurationProperties.isLearningMode()) {
+          return;
+        }
+        throw lfe;
+      } else {
+        log.info("Exception during {} for lock {} requested by {} {} {} {}",
```
ajordens: `info` the right level for this?

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/ExecutionCompleteLockReleasingListener.java
cfieber: Worst case of failing to release here would be that the lock still exists and is unavailable until the TTL expires (120s by default).

I feel like this is more likely to happen if the execution got in a bad way and stalled and did not retain the lock, and releasing failed because the lock no longer exists - in which case this shouldn't matter because the execution is done

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockContext.java
cfieber: in hindsight I think this comment is kind of unnecessary - the holder is just an id for one (of potentially many) holders of a lock - used to enable the reentrant locking bits. I think I need to move this comment into the LockValueBuilder where it is figuring out the lockValue (which will be the execution.id by default). There we need a story for locking in child pipelines or strategies.

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockingConfigurationProperties.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Component
+@ConfigurationProperties("locking")
+public class LockingConfigurationProperties {
+  private boolean learningMode = true;
```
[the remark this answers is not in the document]
cfieber: still does nothing with locking.enabled=false (the default) but if you do enable, it will only be in learning mode instead of full enforcing

orca-redis/orca-redis.gradle
cfieber: yeah pretty much - directly needs stuff in orca core (and nothing in orca-fiddy)

orca-redis/src/main/java/com/netflix/spinnaker/orca/locks/RedisLockManager.java
cfieber: +1
Going to look at / probably move the bit of retry logic I have around getting a resource from the pool into kork

orca-redis/src/main/java/com/netflix/spinnaker/orca/locks/RedisLockManager.java
cfieber: I'll switch to debug - if we are in learning mode then this is not a fatal thing, if we are not in learning mode then we'll propagate an exception out of here and fail elsewhere

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockExtendingTaskExecutionInterceptor.java
```diff
@@ -0,0 +1,112 @@
...
+  private void extendLocks(Task task, Stage stage) {
+    if (!lockingConfigurationProperties.isEnabled()) {
+      return;
+    }
+    if (task instanceof AcquireLockTask || task instanceof ReleaseLockTask) {
+      return;
+    }
+
+    Map<HeldLock, AtomicInteger> heldLocks = new HashMap<>();
+    Set<String> lockTypes = new HashSet<>(Arrays.asList(AcquireLockStage.PIPELINE_TYPE, ReleaseLockStage.PIPELINE_TYPE));
```
dreynaud: If we're not against guava, this would work well here:

       Set<String> lockTypes = ImmutableSet.of(AcquireLockStage.PIPELINE_TYPE, ReleaseLockStage.PIPELINE_TYPE))

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockExtendingTaskExecutionInterceptor.java
```diff
@@ -0,0 +1,112 @@
...
+        count.incrementAndGet();
+      } else {
+        count.decrementAndGet();
+      }
+    });
+
+    List<HeldLock> toExtend = heldLocks
+      .entrySet()
+      .stream()
+      .filter(me -> me.getValue().get() > 0)
```
dreynaud: is it possible to have a value > 0 for a lock that was acquired some time in the past but has expired?

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockManager.java
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * A stage that operates on a particular cluster should acquire the lock for that cluster. Any synthetic stages
+ * can also acquire the same lock. For example:
+ *
+ * As an example, a deploy with Red/Black would go through the following:
+ *
+ *
+ * lockValue = new LockValue(application, execution.type, execution.id)
+ * createServerGroup
+ *   acquireLock(clusterName, lockValue, createServerGroup.stage.id)
```
dreynaud: do we really need that top-level lock? I'd like to think we would only need it for destructive actions (i.e. the `disableCluster` and `scaleDown` cluster below)

orca-redis/orca-redis.gradle
dreynaud: explicitness is goodness

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockingConfigurationProperties.java
```diff
@@ -0,0 +1,53 @@
...
+  @Autowired
+  public LockingConfigurationProperties(DynamicConfigService dynamicConfigService) {
+    this.dynamicConfigService = dynamicConfigService;
+  }
+
+  public boolean isLearningMode() {
+    return dynamicConfigService.getConfig(Boolean.class, "locking.learningMode", learningMode);
+  }
+
+  public void setLearningMode(boolean learningMode) {
```
dreynaud: why the setters? For testing?

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockManager.java
```diff
@@ -0,0 +1,121 @@
...
+   * @throws LockFailureException if the lock is currently held with a different lockValueApplication/lockValueType/lockValue
+   */
+  void acquireLock(String lockName, LockValue lockValue, String lockHolder, int ttlSeconds) throws LockFailureException;
+
+  /**
+   * Extends a named lock ttl.
+   * @param lockName The name of the lock.
+   * @param lockValue The value of the lock - if the lock is already held with this value, the ttl extension is successful otherwise lock extension fails.
+   * @param ttlSeconds How long to extend the lock for.
+   * @throws LockFailureException if the lock is currently held with a different lockValueApplication/lockValueType/lockValue
```
dreynaud: what's the expected behavior if you attempt to extend a lock that does not exist (e.g. expired or it slipped on a banana on its way to redis)

orca-redis/src/main/java/com/netflix/spinnaker/orca/locks/RedisLockManager.java
```diff
@@ -0,0 +1,243 @@
...
+    if (!lockingConfigurationProperties.isEnabled()) {
+      return;
+    }
+    try {
+      lockManagementOperation.accept(lockOperation);
+    } catch (Throwable t) {
+      if (t instanceof LockFailureException) {
+        LockFailureException lfe = (LockFailureException) t;
+        Optional<LockValue> currentLockValue = Optional.ofNullable(lfe.getCurrentLockValue());
+        log.debug("LockFailureException during {} for lock {} currently held by {} {} {} requested by {} {} {} {}",
```
dreynaud: gimme metrics please

want

orca-redis/src/main/java/com/netflix/spinnaker/orca/locks/RedisLockManager.java
```diff
@@ -0,0 +1,243 @@
...
+          return;
+        }
+
+        if (t instanceof RuntimeException) {
+          throw (RuntimeException) t;
+        }
+        throw new RuntimeException("Exception in RedisLockManager", t);
+      }
+    }
+
```
dreynaud: sneaky newline

orca-redis/src/test/groovy/com/netflix/spinnaker/orca/locks/RedisLockManagerSpec.groovy
```diff
@@ -0,0 +1,202 @@
...
+
+    when:
+    redisLockManager.acquireLock('foo', lv('bazinga'), 'baz', 300)
+
+    then:
+    def ex = thrown(LockFailureException)
+    ex.currentLockValue.id == 'bar'
+  }
+
+  def "should be able to release non-existant lock"() {
```
dreynaud: existent

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/DetermineLockTask.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    Map<String, Object> lock = (Map<String, Object>) stage.getContext().get("lock");
+    if (lock != null && lock.containsKey("lockValue")) {
+      return TaskResult.SUCCEEDED;
+    }
+
+    Optional<StageNavigator.Result> lockStageResult = stageNavigator.ancestors(stage).stream().filter(r -> r.getStageBuilder() instanceof AcquireLockStage).findFirst();
```
dreynaud: can't we conceivably have multiple parent lock stages?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
dreynaud: ^ would be a great code comment, that

PR #2366, opened 2018-08-10
refactor(agents): All agents now use locking by default

robzienert:
Moves the `startPolling` behavior into a default method that uses locking to ensure a single instance per agent notification type.


orca-core/src/main/java/com/netflix/spinnaker/orca/notifications/AbstractPollingNotificationAgent.java
```diff
@@ -44,7 +47,21 @@ public AbstractPollingNotificationAgent(NotificationClusterLock clusterLock) {
 
   protected abstract String getNotificationType();
 
-  protected abstract void startPolling();
+  protected abstract void tick();
+
+  protected void startPolling() {
+    subscription = Observable
+      .timer(getPollingInterval(), TimeUnit.SECONDS, scheduler)
```
asher: This made sense in `RestorePinnedServerGroupsPoller` which was configured with an interval in seconds, but it looks like most other polling agents return a polling interval in ms, configured via a `pollers.*.intervalMs` property. Treating as seconds may prevent cleanup agents from running at all, depending on orca deployment frequency.

robzienert: Oh bloody hell.

PR #2371, opened 2018-08-14
fix(canary): When resolving final run score, filter on parent id and …

duftler:
…sort all child run canary stages.


duftler:
Fixes https://github.com/spinnaker/spinnaker/issues/3180.

PR #2378, opened 2018-08-15
feat(agents): store class names on MDC for sql query annotation

asher:
This change supplies scheduled agent class names to a jooq `renderEnd` execution listener so that resulting `SqlExecutionRepository` statements are annotated with the agent class, such as:

```select e.id, e.body, s.execution_id, s.body, e.status from (select `id`, `body`, `status` from orchestrations where (status IN ('BUFFERED')) order by id desc) as `e` left outer join (select `execution_id`, `body` from orchestration_stages) as `s` on (s.execution_id = e.id) -- agentClass: DefaultExecutionPromoter```

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/locks/ExecutionCompleteLockReleasingListener.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/locks/ExecutionCompleteLockReleasingListener.java
@@ -6,9 +6,12 @@ import com.netflix.spinnaker.orca.pipeline.model.Execution;
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: orca-core/src/main/java/com/netflix/spinnaker/orca/locks/ExecutionCompleteLockReleasingListener.java [hunk 1]
```diff
@@ -32,17 +35,23 @@ public void onApplicationEvent(ExecutionComplete event) {
...
+              lockManager.releaseLock(lc.getLockName(), lc.getLockValue(), lc.getLockHolder());
+            } catch (LockFailureException lfe) {
+              logger.info("Failure releasing lock in ExecutionCompleteLockReleasingListener - ignoring", lfe);
+            }
           }
-        }
-      });
+        });
+      } finally {
+        MDC.clear();
```
ajordens: `MDC.clear()` vs `MDC.remove()` ?

asher: Good catch, will fix in a follow-up PR.

PR #2382, opened 2018-08-17
feat(pipelines): add metrics for starting dependend pipelines

emjburns:
In `echo`, every time we start a pipeline we [emit a metric](https://github.com/spinnaker/echo/blob/master/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/monitor/BuildEventMonitor.java#L116). I'd like to emit a similar metric every time we start a dependent pipeline in orca to give a better picture of the number of pipelines triggered.


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -52,6 +54,9 @@ class DependentPipelineStarter implements ApplicationContextAware {
   @Autowired(required = false)
   ArtifactResolver artifactResolver
 
+  @Autowired
+  Registry registry
```
robzienert: Worth changing to an autowired constructor?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -130,6 +135,12 @@ class DependentPipelineStarter implements ApplicationContextAware {
       callable = AuthenticatedRequest.propagate({
         log.debug("Destination thread user: " + AuthenticatedRequest.getAuthenticationHeaders())
         pipeline = executionLauncher().start(PIPELINE, json)
+
+        String id = registry.createId("pipelines.triggered")
```
robzienert: This is actually `Id`, not a string. You'll probably get some unintended side effects by casting it as a string

emjburns: Oops, changing.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
emjburns: Can you point me to an example of someplace where we have autowired constructor with non required beans where we correctly fail to start up when things are missing that shouldn't be?

emjburns: Thx <3

PR #2384, opened 2018-08-21
feat(provider/kubernetes): v2 Include manifest when finding resource

wjoel:
Include the manifest found in addition to the artifacts. Rename the stage to `FindManifest`, similar to find image stages.

This makes it possible to, for example, retrieve the current number of replicas for a deployment and use this when updating that deployment, or to use that number when creating a new deployment (for example, when doing blue-green deployments).

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactsFromResourceTask.java
@@ -55,6 +55,7 @@ public class FindArtifactsFromResourceTask extends AbstractCloudProviderAwareTas
       ), 5, 1000, true);
 
     if (manifest != null) {
+      outputs.put("manifest", manifest.getManifest());
       outputs.put("artifacts", manifest.getArtifacts());
     } else {
       throw new IllegalArgumentException("Manifest " + stageData.manifestName + " not found in "


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/FindManifestStage.java
```diff
@@ -17,21 +17,21 @@
...
 import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder;
 import com.netflix.spinnaker.orca.pipeline.TaskNode;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
 import com.netflix.spinnaker.orca.pipeline.tasks.artifacts.BindProducedArtifactsTask;
 import org.springframework.stereotype.Component;
 
 @Component
-public class FindArtifactsFromResourceStage implements StageDefinitionBuilder {
-  public static final String PIPELINE_CONFIG_TYPE = "findArtifactFromResource";
+public class FindManifestStage implements StageDefinitionBuilder {
```
wjoel: @lwander are we ok with breaking backwards compatibility or do we need to keep the old stage and name around?

danielpeach:
Wouldn't `findArtifactFromManifest` be more consistent?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/FindManifestStage.java
lwander: I've once renamed a stage and it caused a lot of problems in the execution repository -- I think we need to keep the name around.

lwander:
Agreed w/ dan -- doesn't this also find artifacts in the manifest?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/FindManifestStage.java
wjoel: Alright, can just keep the old name then.

wjoel:
It does also find artifacts in the manifest/resource, the reason I called it "resource" is that (afaiui) the Kubernetes objects in the cluster are resources while the manifests describe which resources to create (for example, see the comments in https://kubernetes.io/docs/reference/kubectl/cheatsheet/#creating-objects). `describe` describes the resource while `get` retrieves the manifest of the resource.

However, many users (perhaps even most) seem to use "manifest" to refer to both the manifest (textual description in YAML or JSON files) as well as the resource (or Kubernetes object), so being precise about this might not be useful. I also didn't like the way "Find artifacts from manifest (Manifest)" looked.

But anyway, perhaps just keep the old name, then?

wjoel:
Updated to keep the same name of the stage and just include the manifest in the output.

PR #2385, opened 2018-08-22
fix(metrics): null value is string null, optional values actually optional

emjburns:
Daniel told me not to be so clever and just use `null`.

Also fixing orca startup when optional beans aren't actually present.


joonlim:
LGTM!

PR #2386, opened 2018-08-22
fix(pipelines): Check Optionals for value before calling get().

joonlim:


joonlim:
@emjburns @jtk54 Please review at your convenience.

jtk54:
@emjburns SGTM. Joon can you close this out?

PR #2394, opened 2018-08-25
feat(sql): Adding SQL execution persistence backend

robzienert:
Sorry it's gigantic. I can break this up if people want.

For Netflix reviewers, here's the sole difference:

1. MariaDB is not coming along for the OSS ride, but I showed how we do it here: https://github.com/robzienert/orca-mariadb-extension
1. `orca-sql-mysql` is a new module. We'll need to exclude the mysql connector in our deployments.

### DIFF
--- a/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
+++ b/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
@@ -1,0 +1,684 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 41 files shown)


orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -0,0 +1,653 @@
...
+    return ActiveExecutionsReport(
+      record.get(0, Int::class.java),
+      record.get(1, Int::class.java)
+    )
+  }
+
+  override fun retrievePipelinesForPipelineConfigIdsBetweenBuildTimeBoundary(pipelineConfigIds: MutableList<String>,
+                                                                             buildTimeStartBoundary: Long,
+                                                                             buildTimeEndBoundary: Long): Observable<Execution> {
+    TODO("not implemented")
```
robzienert: We don't use this... gotta impl before it gets merged.

PR #2398, opened 2018-08-30
fix(cancel): consistent CANCELED event

cfieber:
Routes cancelation into the CompleteExecutionHandler, allowing it to
determine the final pipeline status, and not updating the pipeline status
until other branches with running stages have been marked as canceled.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/ExecutionStatus.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/ExecutionStatus.java
@@ -19,7 +19,6 @@ package com.netflix.spinnaker.orca;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Optional;
 
 public enum ExecutionStatus {
   /**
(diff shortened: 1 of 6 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/ExecutionStatus.java
```diff
@@ -31,62 +30,62 @@
    * The task is still running and the {@code Task} may be re-executed in order
    * to continue.
    */
-    RUNNING(false, false),
```
cfieber: fixed some funky formatting in here only (was initially exploring adding a CANCELING state but backed that out)

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
```diff
@@ -56,20 +56,24 @@ class CompleteExecutionHandler(
...
-            "application", execution.application
-          )).increment()
+          var runningStages: Boolean = false
           if (status != SUCCEEDED) {
             execution.topLevelStages.filter { it.status == RUNNING }.forEach {
+              runningStages = true
               queue.push(CancelStage(it))
             }
           }
+          if (!runningStages) {
```
cfieber: CompleteExecution gets queued at the end of every branch, and also requeued as needed, so this allows the other stages to cancel, complete their branch, and queue a CompleteExection.

Only once we don't have any stages to cancel do we update the status to its final value and publish the ExecutionComplete event

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -358,7 +358,6 @@ class TaskController {
         reason
       )
     }
-    executionRepository.updateStatus(PIPELINE, id, ExecutionStatus.CANCELED)
```
cfieber: turns out this was the main cause of my melted brain - queueing a CancelExecution message but also monkeying with the underlying state directly is a bad combo

I suspect the side-effect of this is that it will take longer for the pipeline to reflect a CANCELED state in the UI as its not updated to that until it is actually done - maybe there is a use case here for creating that artificial state (essentially it is running with a different label) but we do already have a  cancelled boolean flag on the execution so we can I think get by without introducing another execution status.

robfletcher: I seem to recall trying to argue someone out of making that change but being brain fried at the time and not being able to articulate my concerns very well.

robfletcher: I just wonder if having a new state is a better thing than having state + boolean combo.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
robfletcher: Seems sane. Any danger of a race condition here? I don't think so.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandlerTest.kt
```diff
@@ -205,18 +205,6 @@ object CompleteExecutionHandlerTest : SubjectSpek<CompleteExecutionHandler>({
...
-      it("updates the pipeline status") {
-        verify(repository).updateStatus(PIPELINE, pipeline.id, stageStatus)
-      }
-
-      it("publishes an event") {
-        verify(publisher).publishEvent(check<ExecutionComplete> {
-          assertThat(it.executionType).isEqualTo(pipeline.type)
-          assertThat(it.executionId).isEqualTo(pipeline.id)
-          assertThat(it.status).isEqualTo(stageStatus)
-        })
```
robfletcher: Is there a corresponding test that it _doesn't_ update the status if there _are_ still other stages doing stuffs.

cfieber: There is a `verifyNoMoreInteractions(queue)` in here, I'll add

cfieber:
```kotlin
it("does not update execution status") {
  verifyNoMoreInteractions(repository)
}
```

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
cfieber: I also don't think so. So probably.

PR #2401, opened 2018-08-30
Revert "fix(cancel): consistent CANCELED event (#2398)"

cfieber:
This reverts commit 8a814a1d2e505a3bbe8a051667157ea2c157ab9b.


cfieber:
found a case where a TERMINAL branch in a pipeline is not properly triggering cancellation of other branches - investigating how right or wrong that behaviour is

PR #2404, opened 2018-09-05
feat(core): Pipeline/Orchestration cleanup should only run for redis

ajordens:
Optimized variants for sql will be forthcoming.


ajordens:
FYI @asher and @robzienert.

I'll have some sql-specific variants that'll be enabled when redis is disabled.

PR #2406, opened 2018-09-05
fix(artifacts): Populate parent artifacts when a templated pipeline is trigger by a parent pipeline

PerGon:
Bug: when a templated pipeline is triggered by a "parent" pipeline, the parent pipeline artifacts get lost.

This happens because of the `PipelinePreprocessor` behavior which mutates the `pipelineConfig` object and the `pipelineConfig.receivedArtifacts` end up "disappearing".
Setting up the `receivedArtifacts`in `pipelineConfig` after it has been massaged by the `PipelinePreprocessor` fixes this issue and the artifacts received by the parent pipeline show in the artifacts list.


PerGon:
@spinnaker/reviewers

PR #2408, opened 2018-09-05
fix(dryrun): Rollback cluster stages can run for real

robfletcher:
The stage runs a task that identifies the cluster to roll back. That info is useful and it's just a read operation. Also, if that has not happened construction of the after stages that actually perform the rollback operations fails.


erikmunson:
:tada: thanks rob

PR #2411, opened 2018-09-11
Stop leaking Redis processes

robfletcher:


orca-test-redis/src/main/java/com/netflix/spinnaker/orca/test/redis/EmbeddedRedisConfiguration.java
```diff
@@ -18,8 +18,8 @@
     return redis;
   }
 
-  @Bean public Pool<Jedis> jedisPool() {
-    return redisServer().getPool();
+  @Bean public Pool<Jedis> jedisPool(EmbeddedRedis redisServer) {
+    return redisServer.getPool();
```
robfletcher: Calling `redisServer()` directly is bad since the process never gets destroyed. Instead I've made it use the bean.

ajordens:
Great success!

PR #2412, opened 2018-09-11
fix(redis): Cleaning up kork redis client configuration

robzienert:
Depends on https://github.com/spinnaker/kork/pull/188.

Opening early to give some additional context to above PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,7 @@ allprojects {
   group = "com.netflix.spinnaker.orca"
 
   ext {
-    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '1.0.13'
+    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '1.0.15'
   }
 
   def checkLocalVersions = [spinnakerDependenciesVersion: spinnakerDependenciesVersion]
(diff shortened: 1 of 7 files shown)


ezimanyi:
@robzienert : I'm getting an NPE when I try to test this.  It looks like the issue is [here](https://github.com/spinnaker/kork/blob/00d5c31a0c48063f9c2adc6bb59e6318ef24a32c/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedisPool.java#L87)---as `initPool` gets called by the super's constructor so before `delegated` gets assigned in the constructor of `InstrumentedJedisPool`.

I'm not sure why I didn't get that NPE when just pulling the new Kork library alone into Orca (as the bug looks to be in Kork) but I do get it when trying to boot Orca with Kork 2.0.8 + this change.

robzienert:
Huh, ok. I didn't get that when I was running locally. Thanks for the heads up - I'll get a fix out as soon as possible.

ezimanyi:
Fix to that bug looks good, but I'm running into one more issue.  I think that the `redisClientDelegate` bean [here](https://github.com/spinnaker/orca/blob/669a37b963450ddd1738ed4cb4401b6de9955d45/orca-redis/src/main/java/com/netflix/spinnaker/orca/telemetry/RedisActiveExecutionsMonitor.kt#L56) needs to be changed to a `redisClientSelector` as you've done with the other beans.  (Currently getting a missing bean error on startup.)

robzienert:
Right you are. I guess I had that disabled in my local. Thanks!

ezimanyi:
Latest commit is working great :+1:

PR #2415, opened 2018-09-17
fix(spring): Disable JDBC autoconfigure

robzienert:
Fixes https://github.com/spinnaker/spinnaker/issues/3342

Totally forgot about this; had a todo in our config and everything, my bad!


ezimanyi:
Thanks!  Tested, orca starts up now; going to merge right away to unbreak things.

PR #2417, opened 2018-09-18
fix(front50): Allow pipeline triggers to refer to triggers via spel

Quintok:
using a pipeline that takes a default parameter with spEL has minimal context (seems to be only scminfo & artifacts) to execute against as the trigger is not passed in as context.

However, this is not true for [/orchestrate](https://github.com/spinnaker/orca/blob/768e9c4bb7c4f1ae075677f78e9e2484eef210bf/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy#L108) endpoint.  This means the spel is evaluated only upon manual execution and not on trigger by parent execution.

- Pipeline A has parameter with the default value of `${trigger.parentExecution.trigger.parameters.myParameter}`
- Pipeline B has a parameter with the name 'myParameter'
- Pipeline A is triggered by pipeline B


Quintok:
If my assumptions are correct about the context of the trigger being passed in for spEL evaluation I think a second pr refactoring this logic centrally would be wise. 

Both methods are quite complex and ordering of evaluation seems to be critical. See https://github.com/spinnaker/orca/pull/2406 which reorders the evaluation of the same method.

ezimanyi:
This looks correct to me, but wouldn't mind if someone with more SpEL experience took a look too.

PR #2418, opened 2018-09-18
feat(sql): Support multiple root liquibase changesets

robzienert:


robzienert:
Need to fight the great fight with Spring a bit more.

PR #2419, opened 2018-09-19
fix(queue): `ContinueParentStageHandler` should complete if all after stages are complete

ajordens:
This addresses a scenario where an execution would never complete if
the set of after stage statuses contained a mix of SUCCEEDED and
TERMINAL.

See `onFailureStages()` on `CreateServerGroupStage`.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
```diff
@@ -139,6 +141,9 @@ class CreateServerGroupStage extends AbstractDeployStrategyStage {
         ]
       }
     }
+
+    // any on-failure stages from the parent should be executed _after_ the rollback completes
+    super.onFailureStages(stage, graph)
```
ajordens: It's important that any stages emitted from `super.onFailureStages()` are injected after the rollback.

This avoids trying to unpin capacity on a server group that has a mix of enabled and disabled instances (the rollback will ensure that everything is enabled first!).

PR #2425, opened 2018-09-24
feat(webhooks): Add optional sort order to preconfigured webhook params

brantburnett:
This will allow the user to control the order these parameters are
displayed in the UI. It will default to 0 if not set.


orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/config/PreconfiguredWebhookProperties.groovy
```diff
@@ -88,6 +88,7 @@ class PreconfiguredWebhookProperties {
     String defaultValue
     String description
     ParameterType type = ParameterType.string
+    int order
```
ethanfrogers: Should this be `int order = 0` or will it default to 0?

brantburnett: I've tested and confirm that order returns 0 through Gate if not specified in orca-local.yml.

ethanfrogers: 👍

PR #2426, opened 2018-09-24
fix(metrics): Correctly handle potentially empty account tag

robzienert:
At long last, we know exactly where we had that bad metric.

An example of the invalid measurements (there's a ton of these):

```
invalid measurement: Measurement(task.invocations.duration.withType:account=:atlas.dstype=rate:cloudProvider=aws:executionType=PIPELINE:isComplete=true:region=n_a:statistic=totalOfSquares:status=SUCCEEDED:taskType=com.netflix.spinnaker.orca.igor.tasks.StartScriptTask,1537820460000,0.0)
```

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/MetricsTagHelper.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/MetricsTagHelper.kt
@@ -2,8 +2,8 @@ package com.netflix.spinnaker.orca.q.metrics
 
 import com.netflix.spectator.api.BasicTag
 import com.netflix.spinnaker.orca.ExecutionStatus
-import com.netflix.spinnaker.orca.pipeline.model.Stage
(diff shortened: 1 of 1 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/MetricsTagHelper.kt
```diff
@@ -14,17 +14,26 @@ class MetricsTagHelper : CloudProviderAware {
...
-        BasicTag("account", helper.getCredentials(stage) ?: "n_a"),
+        BasicTag("account", helper.getCredentials(stage).valueOrNa()),
 
         // sorting regions to reduce the metrics cardinality
         BasicTag("region", helper.getRegions(stage).let {
           if (it.isEmpty()) { "n_a" }
           else { it.sorted().joinToString(",") }
         }))
+
+    private fun String?.valueOrNa(): String {
```
robzienert: I refrained from `valueOrNah`, because I'm a model professional.

ajordens:
LGTM

PR #2420, opened 2018-09-20
feat(locking): adds locking to traffic guarded operations

cfieber:
Uses TrafficGuard as a signal that a lock should be held on a cluster
for an operation. Adds locking to
* deploy stages via AbstractDeployStrategyStage
* TargetServerGroup stages via TargetServerGroupLinearStageSupport
* Cluster-wide stages via AbstractClusterWideClouddriverOperationStage

PR #2431, opened 2018-09-27
fix(locking): fixes locking in strategies / child pipelines

cfieber:
fix for a couple items broken by #2420 :

* locking in strategies and child pipelines will now use the parent pipeline id 
* traffic guard getting a 403 on an application lookup is non fatal (treated as application not existing)
* also increases the poll interval for monitoring child pipelines from 1s -> 15s

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/ParallelDeployStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/ParallelDeployStage.groovy
@@ -61,12 +61,13 @@ class ParallelDeployStage implements StageDefinitionBuilder {
 
   @CompileDynamic
   protected Collection<Map<String, Object>> parallelContexts(Stage stage) {
+    def defaultStageContext = new HashMap(stage.context)
     if (stage.execution.type == PIPELINE) {
       Trigger trigger = stage.execution.trigger
       if (trigger.strategy && trigger instanceof PipelineTrigger) {
         // NOTE: this is NOT the actual parent stage, it's the grandparent which is the top level deploy stage
         Stage parentDeployStage = trigger.parentExecution.stageById(trigger.parameters.parentStageId)
-        Map cluster = parentDeployStage.context as Map
+        Map cluster = new HashMap(parentDeployStage.context as Map)
         cluster.strategy = 'none'
         if (!cluster.amiName && trigger.parameters.amiName) {
           cluster.amiName = trigger.parameters.amiName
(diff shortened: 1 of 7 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockContext.java
```diff
@@ -54,12 +55,21 @@ public LockValueBuilder withStage(Stage stage) {
...
-            .map(Execution::getId)
-            .orElse(null));
+        final String id = Optional.ofNullable(this.id).orElseGet(() -> {
+          if (!execution.isPresent()) {
+            return null;
+          }
+
+          final Execution exec = execution.get();
+          if (exec.getTrigger() instanceof PipelineTrigger) {
+            return Optional.ofNullable(((PipelineTrigger) exec.getTrigger()).getParentStage())
```
robzienert: Can a strategy have an embedded strategy?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy
```diff
@@ -35,7 +35,7 @@ class MonitorPipelineTask implements OverridableTimeoutRetryableTask {
   @Autowired
   ExecutionRepository executionRepository
 
-  long backoffPeriod = 1000
+  long backoffPeriod = TimeUnit.SECONDS.toMicros(15)
```
robzienert: I think you want `toMillis` here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/ParallelDeployStage.groovy
```diff
@@ -61,12 +61,13 @@ class ParallelDeployStage implements StageDefinitionBuilder {
 
   @CompileDynamic
   protected Collection<Map<String, Object>> parallelContexts(Stage stage) {
+    def defaultStageContext = new HashMap(stage.context)
```
cfieber: previously there was a bunch of nastiness in here that was mutating the original stage (including changing the type so that a different afterStages call would accidentally get invoked when the stage completed which was the thing breaking locking)

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy
cfieber: hah, yessir

orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockContext.java
cfieber: it can't - the deploy stage in the strategy is just a placeholder for where to inject the cluster context from the parent

the child pipeline however, that probably totally maybe can...

ajordens:
Not sure if I see anything in here that would have fixed up the expression bug in our flip-flop pipeline? 

Worth double checking that it is back to being happy, otherwise looks like an improvement to me.

cfieber:
Going to merge this independent of the expression business, will test that and follow up with another PR

PR #2434, opened 2018-10-01
fix(locking): determine lock should not fail when locking is disabled

cfieber:
Rolling in the locking code to an existing system running pipelines can cause a new node to plan a disable lock stage when the corresponding acquire lock stage was not planned. If locking is not enabled, the determine lock failure becomes non fatal and just generates a stubbed out lock context.


ajordens:
LGTM

PR #2435, opened 2018-10-02
fix(webhook): ignore empty expectedArtifacts list in webhook stage response

sbwsg:
Before: An empty `expectedArtifacts` array in a webhook stage definition results in a failed pipeline with no explanation provided, and an error in Orca related to missing artifacts.

After: The pipeline is allowed to continue when the webhook response doesn't contain any artifacts.

Closes https://github.com/spinnaker/spinnaker/issues/3354


dotdotdotpaul:
Cool beans.  😄

PR #2438, opened 2018-10-03
fix(core): Fix http client wrapper by honoring system properties if they are set

PerGon:
Orca-core is using a custom built http client which by default would not respect any system properties (in our case `http.proxyHost` and `http.proxyPort`) 

With this, any jvm property will be respected if no custom value is set!


ajordens:
I worry about unintentional pickup of any `javax.*` properties that someone may have set for another purpose.

I'd probably err on the side of explicit proxy configuration for expressions.

ajordens:
Anyone from @spinnaker/netflix-reviewers @spinnaker/google-reviewers feel differently?

lwander:
Yes this might cause surprising side effects for anyone relying on any of Spinnaker's defaults up until now.

PerGon:
Yeah, sounds reasonable. Thanks for the feedback!

I've given it a second try 😅

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -89,6 +92,13 @@ public long getRetryInterval() {
       return executionCount <= MAX_RETRIES;
     });
 
+    String proxyHostname = System.getProperty(JVM_HTTP_PROXY_HOST);
+    if (proxyHostname != null) {
+      int proxyPort = System.getProperty(JVM_HTTP_PROXY_PORT) != null ? Integer.parseInt(System.getProperty(JVM_HTTP_PROXY_PORT)) : 8080;
```
jervi: Are you sure you shouldn't use `System.getenv()`? The above code will only find jvm properties, not system environment variables...

PerGon: Yeah, it's only jvm properties I'm after!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -89,6 +92,13 @@ public long getRetryInterval() {
...
+      LOGGER.info("Found system properties for proxy configuration. Setting up http client to use proxy with hostname {} and port {}", proxyHostname, proxyPort );
```
robzienert: Can you break this log line up at 120 char? Aside from that, this looks great, thanks!

PerGon: Done! Thanks

PR #2440, opened 2018-10-04
fix(front50): Default handling for any retrofit exceptions

ajordens:
The current try/catch suppresses any validation failures that were
raised by `front50`.

Regressed in spinnaker/orca#2387

### DIFF
--- a/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
+++ b/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
@@ -40,25 +40,20 @@ class UpsertApplicationTask extends AbstractFront50Task {
      */
 
     def existingApplication = fetchApplication(application.name)
-    try {
-      if (existingApplication) {
-        outputs.previousState = existingApplication
-        log.info("Updating application (name: ${application.name})")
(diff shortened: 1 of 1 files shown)


orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
```diff
@@ -40,25 +40,20 @@ class UpsertApplicationTask extends AbstractFront50Task {
...
+    }
 
-      if (application.permission?.permissions != null) {
-        front50Service.updatePermission(application.name, application.permission)
-      }
-    } catch (RetrofitError re) {
-      log.error("Could not create or update application permission", re)
-      return new TaskResult(ExecutionStatus.TERMINAL, [:], outputs)
+    if (application.permission?.permissions != null) {
+      front50Service.updatePermission(application.name, application.permission)
```
jtk54: If one of the front50Service calls fail, does the TaskResult return a `TERMINAL` state? I think that's why I wrapped the rest of the front50 calls in the try in the first place. Not sure what the default handling will do here.

ajordens: Yup, the task/execution is being flagged `TERMINAL`.

PR #2442, opened 2018-10-04
chore(queue): Add kv logs for zombie executions

robzienert:
Noticed we've got some zombos, but the logs that we emit aren't very friendly to machine parsing.


ajordens:
lgtm

PR #2444, opened 2018-10-05
fix(locking): more guards around locking

cfieber:
Originally the locking.enabled flag would cause the locking tasks to
no-op. This change guards the wiring of locking synthetic stages
as well, as a means of ensuring there are no extra unexpected stages
present when locking is disabled, and to better allow vetting the
feature


ajordens:
LGTM

PR #2445, opened 2018-10-05
feat(core): manually skip a stage

danielpeach:

### DIFF
--- a/orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
+++ b/orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
@@ -83,6 +83,11 @@ fun Stage.anyUpstreamStagesFailed(): Boolean =
 fun Stage.syntheticStages(): List<Stage> =
   execution.stages.filter { it.parentStageId == id }
 
+fun Stage.recursiveSyntheticStages(): List<Stage> =
+  syntheticStages() + syntheticStages().flatMap {
+    it.recursiveSyntheticStages()
+  }
+
 fun Stage.beforeStages(): List<Stage> =
   syntheticStages().filter { it.syntheticStageOwner == STAGE_BEFORE }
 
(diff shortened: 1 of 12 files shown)


orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
```diff
@@ -83,6 +83,12 @@ fun Stage.anyUpstreamStagesFailed(): Boolean =
 fun Stage.syntheticStages(): List<Stage> =
   execution.stages.filter { it.parentStageId == id }
 
+fun Stage.recursiveSyntheticStages(): List<Stage> =
+  syntheticStages() + syntheticStages().flatMap {
+    if (it.syntheticStages().isEmpty()) emptyList()
```
lwander: Don't think you need this `if`, if the stages are empty, the recursive `it.recursiveSyntheticStages()` never makes it into the `flatmap`, since there are no synthetic stages.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/SkipStageHandler.kt
```diff
@@ -37,8 +37,16 @@ class SkipStageHandler(
...
   override fun handle(message: SkipStage) {
     message.withStage { stage ->
-      if (stage.status in setOf(RUNNING, NOT_STARTED)) {
+      if (stage.status in setOf(RUNNING, NOT_STARTED) || stage.isManuallySkipped()) {
         stage.status = SKIPPED
+        if (stage.isManuallySkipped()) {
+          stage.recursiveSyntheticStages().forEach {
+            if (it.status !in setOf(SUCCEEDED, TERMINAL, FAILED_CONTINUE)) {
+              it.status = SKIPPED
+              repository.storeStage(it)
```
robzienert: I suspect you'd want to send new messages onto the queue to skip these synthetic stages, rather than just marking them skipped?

danielpeach: I'm not totally seeing why - any synthetic stage's tasks will have been completed as `SKIPPED` at this point, so there's nothing else to do but mark the stages as `SKIPPED`. Would it be more natural to bubble the `SkipStage` messages up - each synthetic stage marks itself as `SKIPPED`, then pushes a `SkipStage` message onto the queue with its parent as the next stage to be skipped?

robzienert: Oh, you're right. nvm. :)

PR #2447, opened 2018-10-05
add a traffic guard check on deregisterInstancesFromLoadBalancer

dreynaud:


dreynaud:
So apparently I merged this in August already? Wow 😅

PR #2472, opened 2018-10-18
feat(MPTv2): Adds update op for MPT v2 templates.

jtk54:


jtk54:
Note that dependent pipelines aren't checked yet in this implementation. Will add that soon.

PR #2477, opened 2018-10-19
fix(queue): remove max attempts from CompleteExecution

asher:
No longer setting MaxAttemptsAttribute on CompleteExecution messages due to a potential race condition when a pipeline has many parallel end branches completing within a close window.


ajordens:
LGTM

PR #2478, opened 2018-10-19
fix(tasks): Enforce consistent application task ordering

robzienert:
There have been some odd reports that, in some cases while using the `SqlExecutionRepository`, some applications' tasks are unsorted. This is weird, because SQL itself orders all executions that are returned from the repository. It turns out that this ordering is lost and sort of mangled in the TaskController.

This change moves the sorting back into the ExecutionRepository and forces canned comparators which can be implemented appropriately by the implementing backend. For Redis, this will continue to use the old runtime sorting strategy, whereas SQL will use SQL.

Removed the `RedisPollingAgentExecutionRepositorySpec` class as it doesn't actually test anything `RedisExecutionRepositorySpec` doesn't.


orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
```diff
@@ -464,8 +464,8 @@ public void delete(@Nonnull ExecutionType type, @Nonnull String id) {
   }
 
   @Override
-  public @Nonnull
-  Observable<Execution> retrieveOrchestrationsForApplication(@Nonnull String application, @Nonnull ExecutionCriteria criteria) {
+  public @Nonnull Observable<Execution> retrieveOrchestrationsForApplication(@Nonnull String application,
+                                                                             @Nonnull ExecutionCriteria criteria) {
```
emjburns: Is this your preferred way to do multi line function declarations with long names?

orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
```diff
@@ -545,6 +545,32 @@ public void delete(@Nonnull ExecutionType type, @Nonnull String id) {
     return currentObservable;
   }
 
+  @Nonnull
+  @Override
+  public List<Execution> retrieveOrchestrationsForApplication(@Nonnull String application,
+                                                              @Nonnull ExecutionCriteria criteria,
+                                                              @Nullable ExecutionComparator sorter) {
+
```
emjburns: extra line

robzienert:
Now with passing tests!

orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
robzienert: This is consistent with most of the Java/Groovy code in our projects. Kotlin has a different standard, which I'm fine with (new lining the first argument and indenting one). The key thing, for me, is not going over 120 characters where feasible.

I really dislike side-scrolling. :)

PR #2481, opened 2018-10-23
feat(web): Force cancel admin command

robzienert:


emjburns:
Oh! You are awesome. LGTM

PR #2484, opened 2018-10-23
config(retrofit): enable retries on HTTP 503 responses

dreynaud:
That is, even for non-idempotent request methods like POST. The motivation
is that this status code most likely indicates an error from a load balancer
that failed to reach a backend server, so it seems safe to retry.

This should in particular enable retries for createBake tasks.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,15 +21,14 @@ buildscript {
     maven { url "https://plugins.gradle.org/m2/" }
   }
   dependencies {
-    classpath "com.netflix.spinnaker.gradle:spinnaker-gradle-project:4.3.0"
-    classpath "org.springframework.boot:spring-boot-gradle-plugin:1.5.10.RELEASE"
+    classpath "com.netflix.spinnaker.gradle:spinnaker-dev-plugin:5.1.4"
     classpath "com.netflix.nebula:nebula-kotlin-plugin:$kotlinVersion"
     classpath "org.jetbrains.kotlin:kotlin-allopen:$kotlinVersion"
   }
 }
 
 allprojects {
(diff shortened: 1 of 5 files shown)


orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/RetrofitExceptionHandler.groovy
```diff
@@ -66,14 +66,25 @@ class RetrofitExceptionHandler implements ExceptionHandler {
...
+    if (e.kind == HTTP && e.response.status == HTTP_UNAVAILABLE) {
+      return true
+    }
+
+    return isIdempotentRequest(e) && (isNetworkError(e) || isGatewayErrorCode(e) || isThrottle(e))
   }
 
-  boolean isGatewayTimeout(RetrofitError e) {
+  boolean isGatewayErrorCode(RetrofitError e) {
     e.kind == HTTP && e.response.status in [HTTP_BAD_GATEWAY, HTTP_UNAVAILABLE, HTTP_GATEWAY_TIMEOUT]
```
robfletcher: Wouldn't it make sense to remove `HTTP_UNAVAILABLE` from this list since you've handled it already?

dreynaud: We could do that, but on the other hand if we keep it like that the method does more what it says -- it does tell you if you're having a gateway error of any kind. That opens it to future reuse in a different context potentially

PR #2485, opened 2018-10-24
chore(builds): Introduces composite builds via gradle plugin. Removes…

ttomsu:
… spring boot plugin.

Ref: https://github.com/spinnaker/spinnaker/issues/2844


ttomsu:
@robfletcher FYI - no more needing to make OSS releases to test your internal custom layer.

robzienert:
I. Love. It.

PR #2489, opened 2018-10-29
feat(MPTv2): Adds resolution for v2 templates.

jtk54:


lwander:
Can you describe what you mean by resolution? Is this for loading templates from various sources?

jtk54:
This resolves a config's source template no matter the protocol e.g. 'spinnaker://myTemplate', 'file://<path>'. This will be helpful for supporting a more reasonable development flow. If you want to hold off on these two until there's a clearer picture on what that looks like that's OK with me.

PR #2491, opened 2018-10-30
fix(configurationAccountName): propagate

skandragon:


orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/RunKayentaCanaryTask.kt
```diff
@@ -52,7 +52,7 @@ class RunKayentaCanaryTask(
       stage.execution.application,
       stage.execution.id,
       context.metricsAccountName,
-      context.storageAccountName /* configurationAccountName */, // TODO(duftler): Propagate configurationAccountName properly.
```
robfletcher: are you intending to remove that? The new property replaces it?

skandragon: Yep.
`storageAccountName` was used as a placeholder for `configurationAccountName` and also used properly.

PR #2492, opened 2018-10-31
feat(windows): enable synthetic execution windows

asher:


asher:
Was planning to use this with openconnect firmware deployment stages, but decided to go a different route.

PR #2493, opened 2018-10-31
feat(MPTv2): V2 template loader handler in /orchestrate.

jtk54:


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/handler/v2/V2Handlers.kt
```diff
@@ -0,0 +1,45 @@
...
+  private val templateLoader: V2TemplateLoader,
+  private val objectMapper: ObjectMapper,
+  private val contextParameterProcessor: ContextParameterProcessor,
+  private val registry: Registry
+): HandlerGroup {
+
+  override fun getHandlers(): List<Handler>
+    = listOf(
+    V2TemplateLoaderHandler(templateLoader, contextParameterProcessor, objectMapper)
+//    V1ConfigurationValidationHandler(),
```
lwander: Ok to remove?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/handler/v2/V2TemplateLoaderHandler.java
```diff
@@ -0,0 +1,126 @@
...
+        })
+        .collect(Collectors.toList());
+      template.setVariables(variables);
+    } else {
+      template = templateLoader.load(config.getPipeline().getTemplate());
+    }
+
+    // Ensure that any expressions contained with template variables are rendered
+    // TODO(jacobkiefer): This nests application, trigger, and pipelineTemplate under a 'variables'
+    // namespace in the SpEL processing. Should we un-nest this?
```
lwander: Are these nested only during the first evaluation, or during the execution?

jtk54: So far this is contained in the first evaluation. I'm not planning to carry this over to the execution resolution. The nesting feels clunky to me since you'd refer to `trigger` attributes by doing `${ variables.trigger.field }` which doesn't feel right.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/handler/v2/V2Handlers.kt
jtk54: Yes, these are placeholders until we implement stuff.

PR #2494, opened 2018-11-01
fix(core): sets cloud provider type on ephemeral server group poller

tomaslin:


ajordens:
LGTM

Quite possible we haven't had a desire to TTL anything but `aws` server groups.

PR #2495, opened 2018-11-02
fix(expressions): expressions in top level stage planning.

cfieber:
Ensures that a top level stage's context has expressions processed during planning.

(not necessary for synthietic stages as they are explicitly added to the graph withMergedContext())


robfletcher:
I eagerly await a world where this is no longer necessary

PR #2501, opened 2018-11-07
fix(authz): Use pipeline ids for managed service account names.

dibyom:
This was causing service accounts to be overwritten when two pipelines in different applications had the same name in https://github.com/spinnaker/spinnaker/issues/3562


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java
```diff
@@ -135,11 +135,7 @@ private String generateSvcAcctName(Map<String, Object> pipeline) {
     if (pipeline.containsKey("serviceAccount")) {
       return (String) pipeline.get("serviceAccount");
     }
-
-    String pipelineName = (String) pipeline.get("name");
```
dibyom: Could also use application name + pipeline name

dibyom: id makes more sense since that way the service account is not orphaned when a pipeline is renamed.

PR #2503, opened 2018-11-08
refactor(artifacts): Remove build artifact filtering from orca

ezimanyi:
BREAKING CHANGE: This is a breaking change for users who have manually set BuildArtifactFilter.maxArtifacts or BuildArtifactFilter.preferredArtifacts as environment variables to Orca. As the filtering logic is being moved to Igor, these environment variables need to be set on the Igor process.

Now that igor filters build artifacts before returning them to the caller, we don't need to apply post-filtering anymore.


ezimanyi:
This is the companion change to https://github.com/spinnaker/igor/pull/321; now that igor does this filtering before returning artifacts we can remove this logic from orca.

PR #2509, opened 2018-11-13
 feat(traffic-guards): consider capacity ratio

dreynaud:
Instead of just enforcing capacity > 0, we want to compare the current
capacity and the future capacity. This is meant to prevent incidents where
a production cluster could accidentally go from 200 to 3 instances for
example.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
@@ -137,22 +137,20 @@ public class TrafficGuard {
     int serverGroupCapacity = capacityByServerGroupName.get(serverGroupName);
     int futureCapacity = totalCapacity - serverGroupCapacity;
 
-    if (futureCapacity == 0) {
-      List<Map> context = targetServerGroups.stream()
-        .map(tsg -> ImmutableMap.builder()
-          .put("name", tsg.getName())
-          .put("disabled", tsg.isDisabled())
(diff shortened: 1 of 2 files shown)


dreynaud:
The changes are relatively big, but I think each commit is conceptually self-consistent. So it might make more sense to look at each commit than everything squashed

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -164,6 +174,32 @@ private void verifyOtherServerGroupsAreTakingTraffic(String serverGroupName, Mon
...
+        .put("name", tsg.getName())
+        .put("disabled", tsg.isDisabled())
+        .put("instances", tsg.getInstances())
+        .build())
+      .collect(Collectors.toList());
+  }
+
+  private int getServerGroupCapacity(TargetServerGroup serverGroup) {
+    return (int) serverGroup.getInstances().stream()
+      .filter(instance -> "Up".equals(instance.get("healthState")))
```
asher: Did you consider using an asg's min or desired instead of service discovery health? During a production incident where many instances are unhealthy, min or desired may be preferable for calculation a capacity floor.

dreynaud: I did not change this behavior, but I was surprised to find that this part of the code looked at number of instances actually up and not just the desired size for instance. They each have pros and cons, I'm not sure one number is the clear winner that should be used

PR #2510, opened 2018-11-13
fix(MPT): Always resolve artifacts

jervi:
It is required to be able to use dynamic template sources (SPeL in the `source` field) in combination with cron triggers; for instance if the source template is an artifact from the last successful CI build, but the execution is triggered using a cron trigger.


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -191,9 +191,7 @@ class OperationsController {
       }
     }
 
-    if (!pipeline.plan) {
```
jtk54: Changing this behavior would break [Echo triggering](https://github.com/spinnaker/echo/blob/master/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/orca/PipelineInitiator.java#L96) and likely `roer plan` since there is no guarantee the artifacts would be present.

jervi: I fail to see why it will break? We have been running this fix in production for about a week, and triggers of all kinds are still working. There are so many null-checks inside the `resolveArtifacts()` method that I don't think you will would be able to create a NPE here even if you tried. The only way is if `pipeline.trigger` is null, but then the code would have crashed much earlier.

jervi: It was like this prior to https://github.com/spinnaker/orca/pull/1835 btw.

jervi: You might even be able to revert https://github.com/spinnaker/echo/pull/398 as I think maybe this fixes the root cause. But I might not understand the problem completely.

jtk54: Plan requests from Echo may be fine with this change, but it's not clear if `roer plan` requests would have a trigger and hence would likely break. I suspect you are correct that this change nullifies https://github.com/spinnaker/echo/pull/398, but is this change sufficient or does the echo change need reverted? Would always resolving artifacts break expected artifact inheritance? I believe that's why it was added in the first place.

The overall issue here is that /orchestrate is overloaded with multiple concerns that need identified and split out into separate interfaces.

jtk54:

PR #2515, opened 2018-11-16
feat(kubernetes): ManifestForceCacheRefreshTask Metrics

Rigdon:
This adds the spectator registry to ManifestForceCacheRefreshTask as an initial start to getting more detailed metrics from the Kubernetes manifest tasks. In particular this will give us detailed timing emitted directly from this task on completion and, more importantly, a specific metric when the Force Cache Refresh has been "autoSucceeded" based on the 12 minute timeout. I followed the pattern [here](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy#L52-L60) to make testing a bit easier.

Fixes https://github.com/spinnaker/spinnaker/issues/3605

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableMap;
+import com.netflix.spectator.api.Id;
 import com.netflix.spinnaker.orca.Task;
 import com.netflix.spinnaker.orca.TaskResult;
 import com.netflix.spinnaker.orca.clouddriver.CloudDriverCacheService;
@@ -36,6 +37,7 @@ import org.springframework.stereotype.Component;
 import retrofit.client.Response;
 
 import java.io.IOException;
+import java.time.Clock;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -45,6 +47,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+
+import com.netflix.spectator.api.Registry;
 
 import static com.netflix.spinnaker.orca.ExecutionStatus.RUNNING;
 import static com.netflix.spinnaker.orca.ExecutionStatus.SUCCEEDED;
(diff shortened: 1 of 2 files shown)


Rigdon:
@ewiseblatt Would you mind giving this a look?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTaskSpec.groovy
```diff
@@ -0,0 +1,43 @@
+package com.netflix.spinnaker.orca.clouddriver.tasks.manifest
```
lwander: Missing license header, see other files as an example. Should include `Datadog`, check w/ your legal for details.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
```diff
@@ -62,24 +66,41 @@
...
   @Override
   public TaskResult execute(Stage stage) {
     Long startTime = stage.getStartTime();
     if (startTime == null) {
       throw new IllegalStateException("Stage has no start time, cannot be executing.");
     }
-    if ((System.currentTimeMillis() - startTime) > autoSucceedAfterMs) {
+    if ((clock.millis() - startTime) > autoSucceedAfterMs) {
       log.info("{}: Force cache refresh never finished processing... assuming the cache is in sync and continuing...", stage.getExecution().getId());
+      registry.counter(String.format("%s.timeoutSuccess", METRIC_PREFIX)).increment();
```
ewiseblatt: I think it would be nice to have a single timer with a tag to distinguish the outcomes.
The test at #100 looks suspicious (even though it is analogous to before). I'm not familiar with the code but suggests that autoSucceed is there in case getStartTime took a long time. I assume that makes sense.
If you had a common timer, you would be able to know how much time we are talking about for these cases, plus would be able to relate that to overall attempts.

Also in the final else clause (incidentally the "else" is superfluous since the if already returned) you have no metric. So if you want to know how long it takes refesh to succeed, and the failures are still going through that refreshManifests call, I'd think you'd want to update that timer with that else case as well. Besides giving complete coverage over the refreshManifests time, it would tell you how often you are going through each of the code paths.

Perhaps the outcome could have string values indicating timeout success, refresh success, and whatever that third case is.

If it makes sense, you coudl add a "success" tag and also include that initial exception. If getStartTime is so complicated that you are worrying about timeouts, then you might also be interested in failures.

ewiseblatt: @robzienert has more specific expertise about instrumenting within orca. He might have additional comments.

Rigdon: Yeah I think a single timer for success and failure will be good. I'd like to leave a separate counter specifically for this case though, when the 12 minute time limit for the Force Cache Refresh is hit.

ewiseblatt: wouldnt that case be the same as "statistic=count" and "outcome=AUTO_SUCCESS" or whatever you called the outcome?

Rigdon:
Pushed those suggestions. I opted for `successful` as the tag to match [this one](https://github.com/spinnaker/orca/blob/fc87a8e99ec59a4e129646026987aa5adbde2d31/orca-core/src/main/java/com/netflix/spinnaker/orca/listeners/MetricsExecutionListener.java#L66)

Rigdon:
I'll add an instance variable. We'd prefer to keep the extra counter specifically for the timeout case. It doesn't seem to add value to the timer.

Rigdon:
Question for my own edification @ewiseblatt: is there a performance reason for the instance variable or style preference?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTaskSpec.groovy
```diff
@@ -52,6 +52,6 @@ class ManifestForceCacheRefreshTaskSpec extends Specification {
     expect:
     taskResult.getStatus().isSuccessful()
     registry.counter("manifestForceCacheRefreshTask.refreshTimeout").count() == 1
-    registry.timer("manifestForceCacheRefreshTask.refreshComplete", "result", "success").count() == 1
+    registry.timer("manifestForceCacheRefreshTask.refreshComplete", "successful", "true").count() == 1
```
ewiseblatt: The example you are baseing this on is an outlier within spinnaker. All the other microservices use "success".  Actually they use various things, but "success" is the most common and one I would like to standardize on (and am working on proposing).

Rigdon:
I like `.duration`. I can remove the counter and add a tag `outcome` with values of `complete`, `auto_succeed`, and `failure`.

Rigdon:
I removed the counter and switched to `.duration` with a `result` tag as used in other parts of the codebase. This seems quite a bit cleaner now.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
```diff
@@ -89,9 +113,21 @@ public TaskResult execute(Stage stage) {
     stageData.manifestNamesByNamespace = manifestNamesByNamespace(stage);
 
     if (refreshManifests(cloudProvider, account, stageData)) {
+      registry.timer(durationTimerId.withTags("success", "true", "outcome", "complete"))
+        .record(duration, TimeUnit.MILLISECONDS);
```
ewiseblatt: I thought this was a different outcome than line 124 below. Especially now that you removed the counter.

Rigdon: Line 105 is the different outcome

PR #2517, opened 2018-11-16
 feat(webhooks): Allow a custom trust store for webhooks

ezimanyi:
This change allows users to configure a custom trust store that will be used for webhooks; it will be used in addition to the default trust store.  (The use case is where a user has an additional CA they would like to trust in addition to the standard CAs.)


ezimanyi:
Quick update on the configuration; I've added an `enabled` key so you'll need:
```
webhook:
  trust:
    enabled: true
    trustStore: <path to trust store in jks format>
    trustStorePassword: <string password for trustStore>
```
Halyard configuration is being added in the PR linked above, and I'll add documentation to spinnaker.io.

PR #2520, opened 2018-11-21
feat(swabbie): do not monitor delete image

emjburns:
Removing the monitor task from delete image. 

If the image does not actually get deleted due to a transient error, Swabbie will try again in the future.


ajordens:
👍 

This might explain the _huge_ spike to `AmazonNamedImageLookupController`?

emjburns:
@ajordens this morning? yup it might :)

PR #2524, opened 2018-11-26
fix(artifacts): Use client error return code for missing artifacts

ezimanyi:
When a pipeline is missing artifacts on startup, we currently throw an IllegalStateException, which eventually returns a 500 code to the client.  As this is a problem with the request, and it should not be retried, let's be more clear by making it an IllegalArgumentException, which returns a 400 code.


ezimanyi:
@dbyron0 : Yes this (and https://github.com/spinnaker/echo/pull/402) do indeed address that issue.  I'll close the issue!

PR #2525, opened 2018-11-26
feat(bake): pass owner as application owner

emjburns:
We use the owner field in the bake stage to tag images with their owner.
Right now, we just pull this from the stage.

This PR adds a lookup to front50 to find the application, and if that application exists and the owner exists it updates the owner to be that value.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
@@ -24,6 +24,8 @@ import com.netflix.spinnaker.orca.TaskResult
 import com.netflix.spinnaker.orca.bakery.api.BakeRequest
 import com.netflix.spinnaker.orca.bakery.api.BakeStatus
 import com.netflix.spinnaker.orca.bakery.api.BakeryService
+import com.netflix.spinnaker.orca.front50.Front50Service
+import com.netflix.spinnaker.orca.front50.model.Application
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import com.netflix.spinnaker.orca.pipeline.util.ArtifactResolver
 import com.netflix.spinnaker.orca.pipeline.util.OperatingSystem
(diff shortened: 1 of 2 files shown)


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -59,13 +67,34 @@ class CreateBakeTask implements RetryableTask {
...
     if (!bakery) {
       throw new UnsupportedOperationException("You have not enabled baking for this orca instance. Set bakery.enabled: true")
     }
 
+    // If application exists, we should pass the owner of the application as the user to the bakery
+    try {
+      if (front50Service != null) {
+        String appName = stage.execution.application
+        Application application = front50Service.get(appName)
+        if (application) {
```
ajordens: you'll _never_ get a `null` application back here ... check isn't necessary

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -59,13 +67,34 @@ class CreateBakeTask implements RetryableTask {
...
     String region = stage.context.region
     if (!bakery) {
       throw new UnsupportedOperationException("You have not enabled baking for this orca instance. Set bakery.enabled: true")
     }
 
+    // If application exists, we should pass the owner of the application as the user to the bakery
+    try {
+      if (front50Service != null) {
+        String appName = stage.execution.application
+        Application application = front50Service.get(appName)
```
ajordens: probably want some retries around this ...

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -59,13 +67,34 @@ class CreateBakeTask implements RetryableTask {
...
+        if (application) {
+          String user = application.email
+          if (user != null && user != "") {
+            stage.context.user = user
+          }
+        }
+      }
+    } catch (Exception e) {
```
ajordens: It's a little more proper to catch a `RetrofitError` here

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -59,13 +67,34 @@ class CreateBakeTask implements RetryableTask {
...
+      // ignore exception, we will just use the owner passed to us
+      if (!e.message.contains("404")) {
+        log.warn("Error retrieving application from front50, ignoring.", e)
```
ajordens: would be helpful to know the application name that couldn't be retrieve

PR #2527, opened 2018-11-27
fix(bake): initialize retry support

emjburns:
HA


robzienert:
This is a real good PR.

PR #2529, opened 2018-11-27
fix(webhook): using okhttp configs instead of defaults

emjburns:
The webhook okhttp wasn't using the defaults we have in the config. We were seeing requests time out after 10 seconds (which is not consistent with previous behavior). This PR updates okhttp to use the config values for read and connect timeout.

@ajordens let me know if you'd do this a different way, happy to make changes.


emjburns:
@ezimanyi no worries!

ajordens:
LGTM

PR #2530, opened 2018-11-28
fix(provider/kubernetes): manifest refresh is retryable

lwander:
This should slow down how often a failed refresh attempt is made by forcing orca to listen to the backoff period for the ManifestForeCacheRefresh task


lwander:

PR #2533, opened 2018-11-28
fix(authz): Sync roles after creating managed service user

jervi:
If you create a pipeline with roles populated, the managed service user is created, but the pipeline creation task then fails because fiat is unaware of the new user.


dibyom:
The call to create a service account in Front50 already syncs the users: https://github.com/spinnaker/front50/blob/master/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/ServiceAccountsController.groovy#L54

I think what you might be running into is https://github.com/spinnaker/spinnaker/issues/3631#issuecomment-441793748

jervi:
Ah, yes, you are probably right. Closing this then 👍

PR #2543, opened 2018-12-06
feat(MPTv2): Adds separate v2 plan endpoint.

jtk54:
Also fixes SpEL rendering for /orchestrate - was broken due to model changes in a previous PR. Maintains backwards compatibility for v1 plan. @jervi @gardleopard may be interested in this as well.

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/V2PipelineTemplateController.java
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/V2PipelineTemplateController.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/V2PipelineTemplateController.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+@RestController
+@RequestMapping(value = "/v2/pipelineTemplates")
+@ConditionalOnExpression("${pipelineTemplates.enabled:true}")
+public class V2PipelineTemplateController {
+
+  private static final Logger log = LoggerFactory.getLogger(V2PipelineTemplateController.class);
```
lwander: If we already have lombok here, you could just annotate with `@Slf4j` for the same effect

jtk54: Switched to the annotation.

PR #2549, opened 2018-12-07
fix(core): Remove use of JsonManagedReference on getStages() and replace with JsonIgnoreProperties

jonsie:
Resolves https://github.com/spinnaker/spinnaker/issues/3676

This is currently being tested in our preprod environment, but so far so good.

Additionally, tests added here (https://github.com/spinnaker/orca/pull/1772/files) pass.  Seems to be a safe replacement for `JsonManagedReference` annotation.


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
```diff
@@ -168,7 +165,7 @@ public void setKeepWaitingPipelines(boolean keepWaitingPipelines) {
 
   private final List<Stage> stages = new ArrayList<>();
 
-  @JsonManagedReference
+  @JsonIgnoreProperties(value = "execution")
```
jtk54: Can you drop a comment here that this field is a `JsonBackReference`? Does this ignore `stage.execution`? Maybe clarify that in a comment as well.

robzienert:
ping @jonsie - what's the status on this?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
jonsie: Done.

PR #2555, opened 2018-12-10
feat(pipeline): Support conditional disabling of `ServerGroupCacheForceRefreshTask`

ajordens:
```
stages.cloneServerGroupStage.forceCacheRefresh.enabled: false
```

Current belief is that `ServerGroupCacheForceRefreshTask` is being used
in a few spots unnecessarily and causing pipelines/stages to run longer
than needed.

We will test in our environment, if the belief holds true we'll either:
(a) remove the `ServerGroupCacheForceRefreshTask` or
(b) make it easier to disable `ServerGroupCacheForceRefreshTask` across the board

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/BulkDestroyServerGroupStage.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/BulkDestroyServerGroupStage.java
@@ -16,18 +16,27 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup;
 
-import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup;
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService;
 import com.netflix.spinnaker.orca.clouddriver.tasks.MonitorKatoTask;
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.*;
 import com.netflix.spinnaker.orca.kato.pipeline.Nameable;
 import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder;
 import com.netflix.spinnaker.orca.pipeline.TaskNode;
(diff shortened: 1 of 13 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/BulkDestroyServerGroupStage.java
```diff
@@ -16,40 +16,49 @@
...
-      builder
-        .withTask("bulkDisableServerGroup", BulkDisableServerGroupTask.class)
-        .withTask("monitorServerGroups", MonitorKatoTask.class)
-        .withTask("waitForNotUpInstances", WaitForAllInstancesNotUpTask.class)
-        .withTask("forceCacheRefresh", ServerGroupCacheForceRefreshTask.class)
-        .withTask("bulkDestroyServerGroup", BulkDestroyServerGroupTask.class)
-        .withTask("monitorServerGroups", MonitorKatoTask.class)
-        .withTask("waitForDestroyedServerGroup", BulkWaitForDestroyedServerGroupTask.class);
-    } catch (TargetServerGroup.NotFoundException e) {
-      builder
```
ajordens: Believe we long removed the bit that would have actually thrown `TargetServerGroup.NotFoundException`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/BulkDestroyServerGroupStage.java
```diff
@@ -16,40 +16,49 @@
...
   private static final String PIPELINE_CONFIG_TYPE = "bulkDestroyServerGroup";
 
+  private final DynamicConfigService dynamicConfigService;
+
+  @Autowired
+  public BulkDestroyServerGroupStage(DynamicConfigService dynamicConfigService) {
+    this.dynamicConfigService = dynamicConfigService;
+  }
+
+
```
robzienert: xtra line!

PR #2557, opened 2018-12-12
chore(traffic-guards): more details in error message and log

dreynaud:
Would be helpful to know what _has_ been found when the error is "could
not find x".

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
@@ -156,7 +156,10 @@ public class TrafficGuard {
                                    Collection<TargetServerGroup> currentServerGroups,
                                    String account,
                                    String operationDescriptor) {
-    Preconditions.checkArg(!serverGroupsGoingAway.isEmpty(), "serverGroupsGoingAway must not be empty");
+    if (serverGroupsGoingAway == null || serverGroupsGoingAway.isEmpty()) {
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -125,15 +125,15 @@ public void verifyTrafficRemoval(String serverGroupName, Moniker serverGroupMoni
     List<TargetServerGroup> targetServerGroups = ((List<Map<String, Object>>) cluster.get().get("serverGroups"))
       .stream()
       .map(TargetServerGroup::new)
-      .filter(tsg -> location.equals(tsg.getLocation()))
+      .filter(tsg -> location.equals(tsg.getLocation(location.getType())))
```
dreynaud: @ewiseblatt this is the fix attempt we were discussing earlier

ewiseblatt:
This lets the kube_smoke_test pass once again. I'm not sure if you want the added log of the non-matching services. I'm not sure if that's a common case or not.

I dont really understand the fix. I see you took away the precondition assertion with a softer check and return, but it isnt clear what's going on in this case as to where the problem is and how to avoid reintroducing it with a future change somewhere.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
ewiseblatt: ok. thanks for pointing that out. I thought it was a different one.

ewiseblatt:
LGTM. I appreciate your help looking into this.

PR #2559, opened 2018-12-12
feat(pipeline): Support conditional disabling of `ServerGroupCacheFoceRefreshTask`

ajordens:
Condition added to `ModifyScalingProcessStage` and `ApplySourceServerGroupCapacityStage`.


ajordens:
FYI @lwander @ttomsu.

Incrementally doing this in a few spots and doing some testing. Digging through code to try and make sure there aren't cases where we're explicitly expecting some current state to be available in the cache.

If it looks good for us, thought would be to add a single flag that would disable across the board ... and if no problems, just remove the `ServerGroupCacheForceRefreshTask` completely.

It's heavily feature flagged right now.

PR #2562, opened 2018-12-17
fix(orchestrate): Fixes starting/triggering v2 templated pipelines.

jtk54:


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePreprocessor.kt
```diff
@@ -44,6 +45,18 @@ class PipelineTemplatePreprocessor
...
     }
 
+    // TODO(jacobkiefer): We push the 'toplevel' v2 config into a 'config' field to play nice
+    // with MPT v1's opinionated TemplatedPipelineRequest. When we cut over, the template configuration
+    // should be lifted to the top level like users will specify them.
+    //
+    // We also need to ensure that 'type' and 'schema' are set properly upstream when saving v2 template configs.
+    if (pipeline.getOrDefault(V2PipelineTemplate.SCHEMA, null) == V2PipelineTemplate.V2_SCHEMA_VERSION) {
+      val templateConfig = HashMap(pipeline)
+      templateConfig.remove("trigger") // prevent collisions
```
lwander: What collisions could happen here?

jtk54: Probably a misleading comment - the template config doesn't naturally have a `trigger` field, though it has a `triggers`, which are normal pipeline trigger definitions. I can clarify that in the comment.

lwander: If a user supplied "trigger", what happens? We just delete it?

PR #2564, opened 2018-12-20
fix(traffic-guards): bypass the traffic guard capacity check during RRB

dreynaud:
The capacity check in TrafficGuard assumes entire server groups are going
away, which is not necessarily true in the case of rolling red/blacks.

In the RRB case, AbstractClusterWideClouddriverTask generates katoOps
for operations like disableServerGroup with a desiredPercentage property
and clouddriver is responsible for selecting which instances to disable.


ajordens:
Seems alright to me.

PR #2568, opened 2018-12-21
feat(clouddriver): Support `scaleDown` for rolling red/black deployments

ajordens:


ajordens:
Parity with normal red/black deployments as far as scaling down the source server group goes.

PR #2576, opened 2019-01-04
fix(pipelines): Allow caller to specify whether to resolve artifacts

ezimanyi:
Create a new '/plan' endpoint that allows callers to request a plan for a pipeline template. Any calls to '/orchestrate' with pipeline.plan set to true can be migrated to this endpoint, but the '/orchestrate' endpoint will continue to work.

This new endpoint has a new feature, the flag 'resolveArtifacts' that allows the caller to toggle whether to resolve artifacts when making the plan.


ezimanyi:
Fixes the bug noted in https://github.com/spinnaker/spinnaker/issues/2654#issuecomment-449546544

Depends on https://github.com/spinnaker/echo/pull/427

ezimanyi:

PR #2577, opened 2019-01-04
fix(pipelines): Add expected artifacts to pipeline template spec

ezimanyi:
Expected artifacts are not being returned from template plan requests as they are not in the spec; this is breaking display of templated pipelines with expected artifacts.


ezimanyi:
Partially fixes https://github.com/spinnaker/spinnaker/issues/2654#issuecomment-449546544

The UI was not properly rendering artifacts for templated piplines, because requests to `plan` were never returning the pipeline's expected artifacts.

ezimanyi:

PR #2581, opened 2019-01-05
fix(pipelines): Fix rendering of templated pipelines

ezimanyi:
When a templated pipeline is first configured, we call the plan endpoint to generate the pipeline for display in the UI. This rendered pipeline doesn't have the inherited fields included; these are only inherited on a save operation. In order to properly show the rendered pipeline after a plan request, also process any inherited fields when requesting a plan.

As this code is run on any save operation, it is already required to be idempotent (so subsequent saves without a change in the UI are no-ops); it is thus safe to run it after configuring the pipeline as well.


ezimanyi:
Fixes the issue in https://github.com/spinnaker/spinnaker/issues/2654#issuecomment-451341960

@robzienert @jtk54 : This is the proposed fix I discussed with both of you earlier today, please take a look and let me know if this looks reasonable

PR #2588, opened 2019-01-08
fix(scaleDown): Fixes scaleDown task for autoscaled gce mig.

jtk54:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.groovy
```diff
@@ -90,7 +90,7 @@ class GceDeployStagePreProcessor implements DeployStagePreProcessor  {
       )
     }
 
-    def unpinServerGroupStage = buildUnpinServerGroupStage(stageData)
+    def unpinServerGroupStage = buildUnpinServerGroupStage(stageData, false)
```
lwander: Rather than add this boolean flag in a few places, why not just set this to `null`? In fact, it looks like we can simplify this chunk of code here to just say `return stageDefinitions`

jtk54:
@spinnakerbot cherry-pick 1.10 1.11

jtk54:
@spinnakerbot cherry-pick 1.10

PR #2591, opened 2019-01-10
feat(urlRestrictions): Add urlRestrictions.rejectedIps flag to blacklist IPs/IP ranges from webhook calls and SpEL jsonFromUrl, propertiesFromUrl, and fromUrl calls.

joonlim:


orca-core/orca-core.gradle
```diff
@@ -40,6 +40,8 @@ dependencies {
   compile "com.fasterxml.jackson.module:jackson-module-kotlin:${spinnaker.version('jackson')}"
   compile "org.apache.commons:commons-lang3:3.7"
   compile "de.huxhorn.sulky:de.huxhorn.sulky.ulid:8.1.1"
+  compile "org.springframework.security:spring-security-web:4.2.10.RELEASE"
+  compile "javax.servlet:javax.servlet-api:4.0.1"
```
jtk54: Is this dependency used?

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -81,21 +83,36 @@ public Builder withRejectLinkLocal(boolean rejectLinkLocal) {
       return this;
     }
 
+    public List<String> getRejectedIps() {
```
jtk54: Seems like a lot of boilerplate code can be removed by using lombok `@Data` etc here.

orca-core/orca-core.gradle
```diff
@@ -40,6 +40,8 @@ dependencies {
   compile "com.fasterxml.jackson.module:jackson-module-kotlin:${spinnaker.version('jackson')}"
   compile "org.apache.commons:commons-lang3:3.7"
   compile "de.huxhorn.sulky:de.huxhorn.sulky.ulid:8.1.1"
+  compile "org.springframework.security:spring-security-web:4.2.10.RELEASE"
```
ezimanyi: This can be `compile spinnaker.dependency("springSecurityWeb")` which will pin it to our Spinnaker-wide dependency.

orca-core/src/main/java/com/netflix/spinnaker/orca/config/OrcaConfiguration.java
```diff
@@ -115,8 +116,8 @@ public ComparableLooseVersion comparableLooseVersion() {
   }
 
   @Bean
-  UserConfiguredUrlRestrictions userConfiguredUrlRestrictions(UserConfiguredUrlRestrictions.Builder userConfiguredUrlRestrictionProperties) {
-    return userConfiguredUrlRestrictionProperties.build();
+  UserConfiguredUrlRestrictions userConfiguredUrlRestrictions(UserConfiguredUrlRestrictions.Builder userConfiguredUrlRestrictionProperties, @Value("${urlRestrictions.rejectedIps:}") String[] rejectedIps) {
+    return userConfiguredUrlRestrictionProperties.withRejectedIps(Arrays.asList(rejectedIps)).build();
```
ezimanyi: You shouldn't need to manually wire in the config `urlRestrictions.rejectedIps` here.  If you use the existing configuration hash `userConfiguredUrlRestrictions` and set `userConfiguredUrlRestrictions.rejectedIps`, this will get auto-wired for you above.

orca-core/orca-core.gradle
joonlim: Code would not compile without it:

> Task :orca-core:compileJava FAILED
/usr/local/google/home/joonlim/dev/spinnaker/orca/orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java:145: error: cannot access HttpServletRequest
        if (ipMatcher.matches(u.getHost())) {
                     ^
  class file for javax.servlet.http.HttpServletRequest not found
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -123,6 +140,13 @@ public URI validateURI(String url) throws IllegalArgumentException {
         }
       }
 
+      for (String ip : rejectedIps) {
+        IpAddressMatcher ipMatcher = new IpAddressMatcher(ip);
+        if (ipMatcher.matches(u.getHost())) {
```
ezimanyi: It looks like you're passing the host name to `matches` here; is this function expecting the hostname or the IP address?

orca-core/src/test/groovy/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictionsSpec.groovy
```diff
@@ -0,0 +1,45 @@
...
+class UserConfiguredUrlRestrictionsSpec extends Specification {
+
+  def "validateURI fails if URI is a rejected IP"() {
+    setup:
+    def urlRestrictions = new UserConfiguredUrlRestrictions.Builder()
+      .withRejectedIps(Arrays.asList("10.63.245.46"))
+      .build()
+
+    when:
+    urlRestrictions.validateURI("10.63.245.46")
```
ezimanyi: These tests are passing an IP address to `validateURI` but it is actually expecting a host name.

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -123,6 +140,13 @@ public URI validateURI(String url) throws IllegalArgumentException {
...
+          throw new IllegalArgumentException("address " + u.getHost() + " is within blacklist: " + ip);
```
ezimanyi: Maybe use the same term we're configuring with, 'is within rejectedIps`

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
joonlim: The field name says address, but it only works with the hostname. Within the ```matches``` call, it calls InetAddress.getByName(address);

orca-core/src/test/groovy/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictionsSpec.groovy
joonlim: I believe I will have to get rid of this spec altogether.

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
ezimanyi: Ah, I see---that's confusing, but it does look like the `parseAddress` actually converts a hostname to an address.  But now in the new commit you've reverted to passing an IP address...so is the new commit incorrect?

joonlim: Sorry I added that by accident. I must've tested wrong before because I can now confirm that the method works with either a hostname or an address. I will add a new commit to change it to the way it was before.

joonlim: Fixed

PR #2593, opened 2019-01-11
fix(search): execution results < limit == no more matches

emjburns:
A user discovered a situation where SQL behaves differently than Redis:
```
$ SIZE=100
$ curl -s "https://spinnaker/applications/my_app/executions/search?application=my_app&pipelineName=DEPLOY&triggerTimeStartBoundary=1539846060000&triggerTimeEndBoundary=1539964800000&reverse=true&size=$SIZE" | jq '. | length'
27

$ SIZE=1000
$ curl -s "https://spinnaker/applications/my_app/executions/search?application=my_app&pipelineName=DEPLOY&triggerTimeStartBoundary=1539846060000&triggerTimeEndBoundary=1539964800000&reverse=true&size=$SIZE" | jq '. | length'
98
```

This happens because with redis, we'd get all the executions and then filter down to the matching ones. In sql, we are only getting `limit` number of executions and then filtering them down to the matching ones. This is unexpected.

This PR does a couple of things:
 - removes `Observables` from this endpoint, since we immediately block.
 - creates a new execution repository endpoint which returns all executions in a time period, sorted how you specify. This allows each executionRepositoryImpl to handle pagination and sorting however it would like.
- fixed the problem stated above.

### DIFF
--- a/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
+++ b/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
@@ -406,7 +406,7 @@ class SqlExecutionRepository(
   }
 
   override fun retrievePipelinesForPipelineConfigIdsBetweenBuildTimeBoundary(
-    pipelineConfigIds: MutableList<String>,
+    pipelineConfigIds: List<String>,
     buildTimeStartBoundary: Long,
     buildTimeEndBoundary: Long,
     executionCriteria: ExecutionCriteria,
@@ -442,7 +442,7 @@ class SqlExecutionRepository(
   }
 
   override fun retrieveAllPipelinesForPipelineConfigIdsBetweenBuildTimeBoundary(
-    pipelineConfigIds: MutableList<String>,
+    pipelineConfigIds: List<String>,
     buildTimeStartBoundary: Long,
     buildTimeEndBoundary: Long,
     executionCriteria: ExecutionCriteria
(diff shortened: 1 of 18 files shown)


orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -405,24 +405,69 @@ class SqlExecutionRepository(
...
+    while (moreResults) {
+      val results = retrievePipelinesForPipelineConfigIdsBetweenBuildTimeBoundary(
+        pipelineConfigIds,
+        buildTimeStartBoundary,
+        buildTimeEndBoundary,
+        executionCriteria,
+        page,
+        limit
+      )
+      moreResults = results.size == limit
```
robzienert: I'm always partial to a nice `results.size >= limit` myself, because I don't trust computers or myself.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -312,52 +313,63 @@ class TaskController {
...
+          return true
         }
+      }).collect(Collectors.toList())
+      pipelineConfigIds = pipelines*.id as List<String>
+    }
+
+    ExecutionCriteria executionCriteria =  new ExecutionCriteria()
+      .setSortType( reverse ? REVERSE_BUILD_TIME : BUILD_TIME )
+    if (statuses != null && statuses != "") {
+      executionCriteria.setStatuses(statuses.split(",").toList())
```
robzienert: Hmm. I wonder if it'd be useful to have a `ExecutionCriteria setStatuses(String statuses)` that will use a splitter?

emjburns: It's easy enough to do at the call site, I don't think adding a method would provide a lot of value.

emjburns:
@robzienert previously sorting was done in the controller, and I've moved it do be done in the execution repository. That's why I changed the sort order. I _think_ it still works the same 🤞 based on my tests.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
emjburns: Changed!

PR #2602, opened 2019-01-17
feat(webhooks): Ability to specify HTTP status codes that cause the stage to fail fast

marchello2000:
Adds `failFastStatusCodes` property to the webhook stage. This is an array of status
codes that will short circuit the retry logic and instead fail the stage

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
@@ -17,6 +17,7 @@
 
 package com.netflix.spinnaker.orca.webhook.tasks
 
+import com.fasterxml.jackson.annotation.JsonFormat
 import com.jayway.jsonpath.JsonPath
 import com.jayway.jsonpath.PathNotFoundException
 import com.jayway.jsonpath.internal.JsonContext
@@ -24,6 +25,7 @@ import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.RetryableTask
 import com.netflix.spinnaker.orca.TaskResult
 import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.webhook.config.WebhookProperties
 import com.netflix.spinnaker.orca.webhook.service.WebhookService
 import groovy.util.logging.Slf4j
 import org.apache.http.HttpHeaders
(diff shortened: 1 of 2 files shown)


marchello2000:
Note: this is not exposed in the UI yet

orca-webhook/src/test/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTaskSpec.groovy
```diff
@@ -122,14 +124,15 @@ class CreateWebhookTaskSpec extends Specification {
       payload: [:],
```
marchello2000: Note: this test, as best as I can tell, doesn't actually test a realistic scenario, since, in the event of non-success HTTP status code, Retrofit will throw an exception. I decided to leave it as is, but use the exception throwing mock in the next test, which ensures there are retries on 500's and 429 which is the behavior today.

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -60,11 +61,34 @@ class CreateWebhookTask implements RetryableTask {
        response = webhookService.exchange(method, url, payload, customHeaders)
     } catch (HttpStatusCodeException e) {
       def statusCode = e.getStatusCode()
+
+      if (failFastStatusCodes != null) {
+        def isArray = [Collection, Object[]].any { it.isAssignableFrom(failFastStatusCodes.getClass()) }
```
emjburns: Any reason you didn't use `instanceOf`, as in `failFastStatusCodes instanceOf Collection`? That seems a little cleaner.

ajordens: There's a few cases where we're casing and whatnot ... alternative would be to `stage.mapTo()` a concrete type and just let jackson do the deserialization for you.

marchello2000: Thanks! @ajordens pointed me in the direction for Jackson goodness, updating PR

orca-webhook/src/test/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTaskSpec.groovy
```diff
@@ -140,16 +143,113 @@ class CreateWebhookTaskSpec extends Specification {
...
+    ])
+    def bodyString = "Retry later, please"
+
+    createWebhookTask.webhookService = Stub(WebhookService) {
+      exchange(
+        HttpMethod.DELETE,
+        "https://my-service.io/api/",
+        [:],
+        [:]
+      ) >> {throw new HttpServerErrorException(HttpStatus.TOO_MANY_REQUESTS, HttpStatus.TOO_MANY_REQUESTS.name(), bodyString.getBytes(Charset.defaultCharset()), Charset.defaultCharset())}
```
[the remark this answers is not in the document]
marchello2000: 👍

emjburns:
LGTM I'll rebase and merge

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -49,22 +51,34 @@ class CreateWebhookTask implements RetryableTask {
     Map<String, ?> outputsDeprecated = [deprecationWarning: "All webhook information will be moved beneath the key 'webhook', " +
       "and the keys 'statusCode', 'buildInfo', 'statusEndpoint' and 'error' will be removed. Please migrate today."]
 
-    String url = stage.context.url
-    def method = stage.context.method ? HttpMethod.valueOf(stage.context.method.toString().toUpperCase()) : HttpMethod.POST
-    def payload = stage.context.payload
-    def customHeaders = stage.context.customHeaders
-    boolean waitForCompletion = (stage.context.waitForCompletion as String)?.toBoolean()
+    StageData stageData = stage.mapTo(null, StageData.class)
```
ajordens: The `null` seems off, I don't think that's needed is it?

`StageData stageData = stage.mapTo(StageData)`

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -124,4 +138,17 @@ class CreateWebhookTask implements RetryableTask {
       return new TaskResult(ExecutionStatus.TERMINAL, outputsDeprecated + outputs)
     }
   }
+
+  private static class StageData {
+    public String url
+    public Object payload
+    public Object customHeaders
+    public int[] failFastStatusCodes
```
ajordens: This should probably be a List<Integer> ... could default to an empty list too

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
marchello2000: thanks!

PR #2604, opened 2019-01-17
fix(applications): Validate app names based on cloud provider.

jtk54:


orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
```diff
@@ -30,6 +30,8 @@ import retrofit.RetrofitError
 @Component
 @CompileStatic
 class UpsertApplicationTask extends AbstractFront50Task {
+  private static final String VALID_NAME_REGEX = '^[a-zA-Z0-9-]+$'
```
ajordens: This regex is too restrictive ... we have applications with underscores and periods. 

Certain AWS features might not be available depending on names _but_ we haven't prevented those names from being used historically.

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
```diff
@@ -64,4 +73,10 @@ class UpsertApplicationTask extends AbstractFront50Task {
   String getNotificationType() {
     return "upsertapplication"
   }
+
+  private boolean isValidName(Application application) {
+    def cloudProviders = application?.cloudProviders ? application.cloudProviders.split(',') : []
+    boolean dashesAreValid = application.name.contains('-') ? cloudProviders?.every { it == 'kubernetes' } : true
```
ajordens: Would love an alternative to adding cloud provider-specific conditions into common orca tasks, we've got a couple already but should avoid adding more if possible (alternative would be an interface).

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
jtk54: AWS also doesn't allow `-` in app names either, right?

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
jtk54: I can refactor this out into another utils class so it's not baked into the task, sound OK?

ajordens: Should be an interface IMO rather than a Utils class that is called directly.

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
ajordens: Frigga doesn't allow it, but I _don't think_ AWS particularly cares about dashes for ASGs (at least).

jtk54:
Mirrored the UI app validation per cloud provider and dropped it in a trait, PTAL.

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy
```diff
@@ -39,6 +40,12 @@ class UpsertApplicationTask extends AbstractFront50Task {
      * Upsert application to all global registries.
      */
 
+    def validationErrors = validate(application)
+    if (validationErrors) {
+      log.error("Invalid application name, errors: ${validationErrors}")
```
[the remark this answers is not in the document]
jtk54: Nope.

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/utils/ApplicationNameValidator.groovy
```diff
@@ -0,0 +1,62 @@
...
+  List<String> validate(Application application) {
+    if (application.cloudProviders == null || application.cloudProviders.isEmpty()) {
+      return []
+    }
+
+    String applicationName = application.name
+    def cloudProviders = application.cloudProviders.split(",")
+    def validationErrors = cloudProviders
+      .findAll { cloudProviderNameConstraints.containsKey(it) }
+      .collect { provider ->
```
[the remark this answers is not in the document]
jtk54: Nice.

jtk54:
@ajordens The bug is misleading actually -- the issue is that they created a valid k8s v2 application with a dash in the name and then added another cloud provider that disallows dashes in app names, which caused issues for that provider. Hopefully that clarifies.

ajordens:
Gotcha, thanks for the clarification.

I don't think it hurts, so 👍 from me. It won't stop all downstream name-related issues and the cloud providers need to do a better job of sanitizing names on the resources they're creating.

PR #2609, opened 2019-01-18
fix(notifications): Propagate auth for application notifications

jervi:
Application and pipeline level notifications fails for protected apps, because orca is not propagating authentication to front50. That's sad if you want both security _and_ notifications. But I'm pleased to announce that with this PR, you can have your cake AND eat it too.

Root cause is https://github.com/spinnaker/front50/pull/408 (@ajordens FYI)


jervi:
Don’t merge this yet. We discovered a problem with some pipeline executions. I’ll update the PRthis

ajordens:
Ahh yes.

Let me know if you need a hand verifying any of this.

jervi:
The problem we had was because of a mismatch between the orca version and the echo version, and was unrelated to these changes. So this is ready to be merged if it is approved.

ajordens:
Thanks for the fix @jervi!

PR #2611, opened 2019-01-20
feat(retrofit): Support registering custom interceptors

ajordens:
The `RetrofitInterceptorProvider` interface provides an extension point
through which additional okhttp/okhttp3 interceptors can be registered.


orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/RetrofitConfiguration.groovy
```diff
@@ -42,33 +39,41 @@ import org.springframework.core.annotation.Order
 import retrofit.RestAdapter.LogLevel
 import retrofit.client.OkClient
 
-import java.util.concurrent.TimeUnit
-
 @Configuration
 @CompileStatic
 @Import(OkHttp3ClientConfiguration)
 @EnableConfigurationProperties
 class RetrofitConfiguration {
```
ajordens: Apparently a reformat came along for the ride ...

PR #2614, opened 2019-01-22
fix(webhooks): use error field consistently

marchello2000:
Always use `error` (never `errorMessage`) to report errors from webhooks stage
This is what UI looks for (improvements in that area will be PR'd shortly)


ajordens:
LGTM

PR #2622, opened 2019-01-25
fix(notifications): Support multiple app-level notifications

jervi:
Due to a bug in the comparison code, similar notifications of the same type was stripped away, so that only the first notification would be notified. In our case it was because `publisherName` was `null`, which made the `publisherMatches` criteria `true`, even though it was supposed to be `false`. I've fixed this by only considering non-null values.


orca-echo/src/test/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingExecutionListenerSpec.groovy
```diff
@@ -196,4 +196,31 @@ class EchoNotifyingExecutionListenerSpec extends Specification {
     }
     1 * echoService.recordEvent(_)
   }
+
+  def "handles cases with multiple notifications of the same type"() {
```
xavileon: maybe add an extra case where the fields are null, which was the source of the bug

jervi: They are `null`, because they are unset. But I guess I could have been more explicit about it

jervi: (Only `address`, `level`, `type` and `when` are set)

xavileon: right 👍

jervi:
@spinnaker/reviewers PTAL

PR #2628, opened 2019-01-29
chore(queue): De-escalate requeue messages, as this is normal behavior

robzienert:


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ContinueParentStageHandler.kt
```diff
@@ -17,7 +17,10 @@
 package com.netflix.spinnaker.orca.q.handler
 
 import com.netflix.spinnaker.orca.ExecutionStatus.NOT_STARTED
-import com.netflix.spinnaker.orca.ext.*
+import com.netflix.spinnaker.orca.ext.allAfterStagesComplete
+import com.netflix.spinnaker.orca.ext.allBeforeStagesSuccessful
+import com.netflix.spinnaker.orca.ext.anyBeforeStagesFailed
+import com.netflix.spinnaker.orca.ext.hasTasks
```
robfletcher: ¬_¬

PR #2633, opened 2019-01-30
feat(bake): if base name is empty string, remove from bake request

emjburns:
We have a user who is using SpEL to conditionally set the `baseName` in the bake. If the parameter of the pipeline has a value for baseName, it's set in the stage. If the parameter is empty, they send us empty string because there is no way to conditionally set a field in the UI. 

`"baseName": "${ parameters['customBaseName'] ?: '' }",`

It feels like the right behavior to remove this parameter from what gets turned into the BakeRequest. This shouldn't break anything because right now if you send an empty string for baseName our bakery errors out.


emjburns:
I'll investigate creating a spel helper, good suggestion.

PR #2635, opened 2019-01-30
fix(webhooks): give better error message for monitored webhooks

marchello2000:
Instead of bubbling semi-useless error message verbatim from JsonPath
Make it relevant and clear to the end user why monitoring for webhook failed
if the specified json path is not valid

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
@@ -38,6 +38,7 @@ class MonitorWebhookTask implements OverridableTimeoutRetryableTask {
 
   long backoffPeriod = TimeUnit.SECONDS.toMillis(1)
   long timeout = TimeUnit.HOURS.toMillis(1)
+  private static final String JSON_PATH_NOT_FOUND_ERR_FMT = "Unable to parse %s: JSON property '%s' not found in response body"
 
   @Override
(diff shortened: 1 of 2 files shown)


orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
```diff
@@ -38,6 +38,7 @@ class MonitorWebhookTask implements OverridableTimeoutRetryableTask {
 
   long backoffPeriod = TimeUnit.SECONDS.toMillis(1)
   long timeout = TimeUnit.HOURS.toMillis(1)
+  private static final String JSON_PATH_NOT_FOUND_ERR_FMT = "Unable to parse status: JSON property '%s' not found in response body"
```
ajordens: Not expert expect but looks like this error message is used for status and progress path failures ... perhaps just `JSON property '%s' not found in response body` would be sufficient?

Either way, much better than the exception message I reckon.

marchello2000:
looking into ci failures

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
marchello2000: Ah, you're right, thank you!

marchello2000: Made it parametrized. I think the `status` vs. `progress` is useful distinction

PR #2637, opened 2019-01-31
fix(mpt): Ensure that a merged template ends up with a `source` set

ajordens:


ajordens:
Working on something that needs to (reliably) know whether or not an `Execution` originated from a template or not.

Testing uncovered a situation where the `source` would not be set iff there was only _one_ template.

I'm not all that versed on MPT but does this make sense @jtk54 / @robzienert? Did a targeted fix towards v1 since that's all we have here.

robzienert:
Sometimes a source isn't going to be set, but this change seems innocuous enough. Cases where one wouldn't be is a plan operation where the template is being inlined with the config.

PR #2639, opened 2019-01-31
feat(core): Support pipeline correlation ids

robzienert:
This is to support de-duplicating pipelines from an internal tool (Gutenberg) which sends a ton of pipeline requests to us with a low timeout/retry cycle.


ajordens:
LGTM

PR #2640, opened 2019-02-01
fix(FindImageFromCluster): infer regions from deploy stages

marchello2000:
FindImageFromClusterTask now looks ahead to find all regions from subsequent deploy stages
This moves the failure point of a pipeline up when an image isn't found.
E.g. FindImage specifies to look for images in `us-east`, but a deploy right after was modified to
deploy to `us-east` AND `us-west`. Thus FindImage will locate the `us-east` image,
the deploy for that image will start but the deploy for `us-west` will fail.
This failure can cause regions to have different versions deployed.
To prevent this, we now will fail in the FindImage task before any deploy starts

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
@@ -26,9 +26,11 @@ import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder
 import com.netflix.spinnaker.orca.pipeline.TaskNode
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import com.netflix.spinnaker.orca.pipeline.tasks.artifacts.BindProducedArtifactsTask
+import com.netflix.spinnaker.orca.pipeline.util.RegionCollector
 import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
+import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
 import javax.annotation.Nonnull
@@ -41,6 +43,9 @@ import static com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner.STAG
 class BakeStage implements StageDefinitionBuilder {
 
   public static final String PIPELINE_CONFIG_TYPE = "bake"
+
+  @Autowired
+  RegionCollector regionCollector
 
   @Override
   void taskGraph(Stage stage, TaskNode.Builder builder) {
@@ -80,31 +85,8 @@ class BakeStage implements StageDefinitionBuilder {
     deployRegions.addAll(stage.context.regions as Set<String> ?: [])
 
     if (!deployRegions.contains("global")) {
-      deployRegions.addAll(stage.execution.stages.findAll {
-        it.type == "deploy"
-      }.collect {
-        Set<String> regions = it.context?.clusters?.inject([] as Set<String>) { Set<String> accum, Map cluster ->
-          if (cluster.cloudProvider == stage.context.cloudProviderType) {
-            accum.addAll(cluster.availabilityZones?.keySet() ?: [])
-          }
-          return accum
-        } ?: []
-        if (it.context?.cluster?.cloudProvider == stage.context.cloudProviderType) {
-          regions.addAll(it.context?.cluster?.availabilityZones?.keySet() ?: [])
-        }
-        return regions
-      }.flatten())
+      deployRegions.addAll(regionCollector.getRegionsFromChildStages(stage))
       // TODO(duftler): Also filter added canary regions once canary supports multiple platforms.
-      deployRegions.addAll(stage.execution.stages.findAll {
-        it.type == "canary"
-      }.collect {
-        Set<String> regions = it.context?.clusterPairs?.inject([] as Set<String>) { Set<String> accum, Map clusterPair ->
-          accum.addAll(clusterPair.baseline?.availabilityZones?.keySet() ?: [])
-          accum.addAll(clusterPair.canary?.availabilityZones?.keySet() ?: [])
-          return accum
-        } ?: []
-        return regions
-      }.flatten())
     }
 
     log.info("Preparing package `${stage.context.package}` for bake in ${deployRegions.join(", ")}")
--- a/orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStageSpec.groovy
+++ b/orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStageSpec.groovy
@@ -18,9 +18,11 @@ package com.netflix.spinnaker.orca.bakery.pipeline
 
 import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.pipeline.util.RegionCollector
 import groovy.time.TimeCategory
 import spock.lang.Specification
 import spock.lang.Unroll
+
 import static com.netflix.spinnaker.orca.test.model.ExecutionBuilder.pipeline
 import static com.netflix.spinnaker.orca.test.model.ExecutionBuilder.stage
 
@@ -34,11 +36,13 @@ class BakeStageSpec extends Specification {
           type = "deploy"
           name = "Deploy!"
           context = zones
+          refId = "2"
+          requisiteStageRefIds = ["1"]
         }
       }
     }
 
-    def bakeStage = new Stage(pipeline, "bake", "Bake!", bakeStageContext)
+    def bakeStage = new Stage(pipeline, "bake", "Bake!", bakeStageContext + [refId: "1"])
     def builder = Spy(BakeStage, {
       (0..1) * now() >> {
         use([TimeCategory]) {
@@ -46,6 +50,8 @@ class BakeStageSpec extends Specification {
         }
       }
     })
+
+    builder.regionCollector = new RegionCollector()
 
     when:
     def parallelContexts = builder.parallelContexts(bakeStage)
@@ -93,7 +99,8 @@ class BakeStageSpec extends Specification {
     }
 
     def bakeStage = pipeline.stageById("1")
-    def parallelStages = new BakeStage().parallelStages(bakeStage)
+    def parallelStages = new BakeStage(regionCollector: new RegionCollector()).parallelStages(bakeStage)
+
     parallelStages.eachWithIndex { it, idx -> it.context.ami = idx + 1 }
     pipeline.stages.addAll(parallelStages)
 
@@ -127,7 +134,7 @@ class BakeStageSpec extends Specification {
   private
   static List<Map> expectedContexts(String cloudProvider, String amiSuffix, String... regions) {
     return regions.collect {
-      [cloudProviderType: cloudProvider, amiSuffix: amiSuffix, type: BakeStage.PIPELINE_CONFIG_TYPE, "region": it, name: "Bake in ${it}"]
+      [cloudProviderType: cloudProvider, amiSuffix: amiSuffix, type: BakeStage.PIPELINE_CONFIG_TYPE, "region": it, name: "Bake in ${it}", refId: "1"]
     }
   }
 }
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
@@ -28,6 +28,7 @@ import com.netflix.spinnaker.orca.clouddriver.OortService
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.Location
 import com.netflix.spinnaker.orca.clouddriver.tasks.AbstractCloudProviderAwareTask
 import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.pipeline.util.RegionCollector
 import groovy.transform.Canonical
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
@@ -80,6 +81,9 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
   @Autowired
   ObjectMapper objectMapper
 
+  @Autowired
+  RegionCollector regionCollector
+
   @Canonical
   static class FindImageConfiguration {
     String cluster
@@ -118,6 +122,16 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
 
     Set<String> imageNames = []
     Map<Location, String> imageIds = [:]
+
+    // Supplement config with regions from subsequent deploy/canary stages:
+    def deployRegions = regionCollector.getRegionsFromChildStages(stage)
+
+    deployRegions.forEach {
+      if (!config.regions.contains(it)) {
+        config.regions.add(it)
+        log.info("Inferred and added region ($it) from deploy stage to FindImageFromClusterTask (executionId: ${stage.execution.id})")
+      }
+    }
 
     Map<Location, List<Map<String, Object>>> imageSummaries = config.requiredLocations.collectEntries { location ->
       try {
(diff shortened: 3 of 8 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -119,6 +119,29 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
     Set<String> imageNames = []
     Map<Location, String> imageIds = [:]
 
+    // Supplement config with regions from subsequent deploy steps:
+    def deployRegions = stage.execution.stages.findAll {
```
ajordens: Believe there should be a better way / helper method to only look at downstream stages.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -119,6 +119,29 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
...
+      if (it.context?.cluster?.cloudProvider == stage.context.cloudProviderType) {
+        regions.addAll(it.context?.cluster?.availabilityZones?.keySet() ?: [])
+      }
+      return regions
+    }.flatten()
+
+    deployRegions.forEach {
+      if (!config.regions.contains(it)) {
+        config.regions.add(it)
+        log.warn("Inferred and added region ($it) from deploy stage to FindImageFromClusterTask")
```
ajordens: Handy to get the `executionId` into the log message ... otherwise it's just lost in the noise.

(possibly also not a `warn`?)

marchello2000: makes sense will add and change

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
marchello2000: Took it from https://github.com/spinnaker/orca/blob/master/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy#L80

But also saw https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java#L55 which I can use. Will poke to see if there is anything better, let me know if you recall of the top of your head

marchello2000: Fixed it up, let me know what you think!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -422,6 +422,35 @@ public Task taskById(String taskId) {
     }
   }
 
+  /**
+   * Gets all descendant stages
+   */
+  public List<Stage> descendants() {
```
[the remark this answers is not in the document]
marchello2000: You're right, walking the topo sorted list should give us the correct answer but with enough branches there is potential for it to pick up extra deploy stages. Let me implement a proper graph search

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -275,10 +300,19 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
...
+    ArrayList<Location> missingLocations,
+    Map<Location, List<Map<String, Object>>> imageSummaries,
+    Map<String, Object> deploymentDetailTemplate,
+    FindImageConfiguration config,
+    String executionId
+  ) {
     for (Map image : images) {
       for (Location location : missingLocations) {
         if (imageSummaries[location] == null && image.amis && image.amis[location.value]) {
+          log.info("Resolved missing image in '$location.value' with '$image.imageName' ($executionId)")
```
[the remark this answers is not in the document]
marchello2000: 👍

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -164,11 +187,13 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
 
     if (!locationsWithMissingImageIds.isEmpty()) {
       // signifies that at least one summary was missing image details, let's retry until we see image details
-      log.warn("One or more locations are missing image details (locations: ${locationsWithMissingImageIds*.value}, cluster: ${config.cluster}, account: ${account})")
+      log.warn("One or more locations are missing image details (locations: ${locationsWithMissingImageIds*.value}, cluster: ${config.cluster}, account: ${account}, execution id: ${stage.execution.id})")
       return new TaskResult(ExecutionStatus.RUNNING)
     }
 
     if (missingLocations) {
+      log.info("Resolving images in missing locations: " + missingLocations.collect({it -> it.value}).join(","))
```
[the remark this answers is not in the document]
marchello2000: 👍

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -164,11 +187,13 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
 
     if (!locationsWithMissingImageIds.isEmpty()) {
       // signifies that at least one summary was missing image details, let's retry until we see image details
-      log.warn("One or more locations are missing image details (locations: ${locationsWithMissingImageIds*.value}, cluster: ${config.cluster}, account: ${account})")
+      log.warn("One or more locations are missing image details (locations: ${locationsWithMissingImageIds*.value}, cluster: ${config.cluster}, account: ${account}, execution id: ${stage.execution.id})")
```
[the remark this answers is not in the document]
marchello2000: 👍

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -422,6 +423,47 @@ public Task taskById(String taskId) {
...
+      while (!queue.isEmpty()) {
+        Stage stage = queue.pop();
+        if (!first) {
+          if ((stageFilter == null) || (stageFilter.test(stage))) {
+            descendents.add(stage);
+          }
+        }
+
+        first = false;
+        visitted.add(stage.refId);
```
robzienert: `visited`

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/RegionCollector.java
```diff
@@ -29,41 +28,45 @@
...
+      .filter(it -> "deploy".equals(it.getType()))
+      .forEach(deployStage -> {
+        List<Map> clusters = (List<Map>) deployStage.getContext().getOrDefault("clusters", new ArrayList<>());
 
-      for (Map cluster : clusters) {
-        if ((cluster.get("cloudProvider") != null) &&
-          cluster.get("cloudProvider").equals(stageCloudProviderType)) {
-          deployRegions.addAll(((Map) cluster.getOrDefault("availabilityZones", new HashMap())).keySet());
+        for (Map cluster : clusters) {
+          if ((cluster.get("cloudProvider") != null) &&
```
ajordens: Same null safe comparison as above could apply here ... 

```
if (stageCloudProviderType.equals(cluster.get("cloudProvider")) {
...
}
```

Not a big deal tho ... good use of `getOrDefault()`

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -80,31 +85,8 @@ class BakeStage implements StageDefinitionBuilder {
...
-            accum.addAll(cluster.availabilityZones?.keySet() ?: [])
-          }
-          return accum
-        } ?: []
-        if (it.context?.cluster?.cloudProvider == stage.context.cloudProviderType) {
-          regions.addAll(it.context?.cluster?.availabilityZones?.keySet() ?: [])
-        }
-        return regions
-      }.flatten())
+      deployRegions.addAll(regionCollector.getRegionsFromDescendentStages(stage))
```
ajordens: I dig it!

orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStageSpec.groovy
```diff
@@ -93,7 +99,9 @@ class BakeStageSpec extends Specification {
     }
 
     def bakeStage = pipeline.stageById("1")
-    def parallelStages = new BakeStage().parallelStages(bakeStage)
+    def bakeStageInstance = new BakeStage()
```
ajordens: should be able to `def parallelStages = new BakeStage(regionCollector: new RegionCollector().parallelStages(bakeStage)` if you wanted to preserve the existing form.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -422,6 +422,44 @@ public Task taskById(String taskId) {
...
+
+        List<Stage> childStages = stage.downstreamStages();
+
+        childStages
+          .stream()
+          .filter(s -> !visited.contains(s.refId))
+          .forEach(s -> queue.add(s));
+      }
+
+      return ImmutableList
```
ajordens: Might suggest not returning an Immutable or Unmodifiable list here ... the existing `downstreamStages()` doesn't and it's probably worth being consistent. 

We've been burnt one too many times with something being unmodifiable and blowing up at runtime when someone unexpectedly tries to add.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -422,6 +422,44 @@ public Task taskById(String taskId) {
     }
   }
 
+  /**
+   * Get all stages that are descendants of the current one
+   */
+  public List<Stage> descendants() {
```
ajordens: What do you think about `descendants()` alongside the existing `downstreamStages` methods?

The only thing I could think of would be calling it `allDownstreamStages()` or something to indicate similar intentions.

Not a strong opinion on this, and it's something that _could_ be refactored later.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/RegionCollector.java
```diff
@@ -0,0 +1,73 @@
...
+   * Traverses all descendant stages of a given stage looking for deploy/deploy canary stages
+   * When found, extracts the regions those stages deploy to.
+   *
+   * @param stage Stage for which to traverse the regions for
+   * @return union of all regions or an empty set
+   */
+  public @NotNull
+  Set<String> getRegionsFromDescendentStages(Stage stage) {
+    Set<String> deployRegions = new HashSet<>();
+    String stageCloudProviderType = stage.getContext().get("cloudProviderType") != null ?
```
ajordens: Might be a little tidier as: 

```
Map<String, Object> context = stage.getContext();
String cloudProvider = (String) context.getOrDefault("cloudProviderType", context.get("cloudProvider"))
```

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/RegionCollectorSpec.groovy
```diff
@@ -0,0 +1,107 @@
...
+
+  def "should find all regions from subsequent canary deploys"() {
+    when:
+    def regionSet = regionCollector.getRegionsFromDescendentStages(execution.stageByRef("1"))
+
+    then:
+    regionSet == allDeployRegions
+
+    where:
+    deployCluster1 = [
```
ajordens: semantics ... but most of what you've got in the `where:` block is stuff I would put in a `given:` block

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/RegionCollectorSpec.groovy
```diff
@@ -0,0 +1,107 @@
...
+        refId = "1"
+        context = [ cloudProvider: "aws"]
+      }
+
+      stage {
+        name = "canary"
+        type = "canary"
+        context = [
+          clusterPairs: [ [
+              baseline: deployCluster1,
```
ajordens: test could be significantly shortened if you had a helper method create the clusters ...

```
baseline: cluster(["us-west-2", "us-east-1"])
```

For a test like this, I think it's fine to use actual region names and then just reference them explicitly in your `then:` (vs. needing to compare against a variable)
...

```
regionSet == ["us-west-2", "us-east-1", "eu-west-1"] // this would be fine imo
```

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/RegionCollectorSpec.groovy
```diff
@@ -0,0 +1,107 @@
...
+
+import spock.lang.Specification
+
+import static com.netflix.spinnaker.orca.test.model.ExecutionBuilder.pipeline
+import static com.netflix.spinnaker.orca.test.model.ExecutionBuilder.stage
+
+class RegionCollectorSpec extends Specification {
+  RegionCollector regionCollector
+
+  def setup() {
```
ajordens: you don't really need the `setup()` ..

```
@Subject
def regionCollector = new RegionCollector()
```

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
marchello2000: thanks for helping me arrive at this solution!

orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStageSpec.groovy
marchello2000: nice, thanks!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: yeah, i think that would be more consistent esp since we have `ancestors` + `ancestorsOnly` (though it's an internal helper method). Anyway, I like, changing

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: I was following the pattern from `ancestors()` but don't have a preference, can also chance return type to `ImmutableList<>`? what would you prefer?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/RegionCollector.java
marchello2000: 👍

marchello2000: @ajordens : Btw, is it safe to assume that one of these keys (`cloudProviderType` or `cloudProvider`) will be set? I saw that `CloudProviderAware.getCloudProvider` always falls back on some default (`aws`) I should probably do the same in absence of strongly typed contexts?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/RegionCollector.java
marchello2000: can also guarantee that `stageCloudProviderType` is always set, and then flip, makes is much cleaner i think:
`cloudProviderType.equals(cluster.get(...))`

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/RegionCollectorSpec.groovy
marchello2000: nice thanks!

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/RegionCollectorSpec.groovy
marchello2000: yeah, you're right... fixing

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/RegionCollectorSpec.groovy
marchello2000: yeah, i think that makes sense. fixup

marchello2000:
@ajordens tested bake/find image and applied the changes you mentioned - so I think it's super close, thanks for reviewing this so many times!

ajordens:
👍 

Will merge once travis gives the a-ok.

PR #2645, opened 2019-02-07
fix(canary): capture exceptions from findImage in DeployCanaryStage

dreynaud:
The find image stage can generate a variety of exceptions that would
escape from findBaselineAmis.


dreynaud:
@marchello2000 generally speaking I try to avoid catch-all `try` blocks because they tend to obscure what we are trying to catch. In this case, the `findImage.execute` call does make remote calls and can fail as a matter of normal operations (e.g. if the baseline cluster does not exist), so I think it should be handled properly.

It's also possible for the `regions =` thing to fail (e.g. with an NPE if clusterPairs is null) but I don't find it shocking for this scenario to fail with an unexpected/internal error

marchello2000:
@dreynaud : got it. My instinct would be to wrap everything in try catch that can surface the error to the users since that makes it a bit more readable/user friendly albeit slightly misleading (the innerException is captured anyway).

PR #2646, opened 2019-02-07
chore(mysql): Adding database creation to the setup script

aravindmd:
adding database creation to the setup script


marchello2000:
Thanks for fixing it up

PR #2648, opened 2019-02-08
feat(clouddriver): Consider target capacity if supplied from `clouddriver`

ajordens:
AWS deployments are somewhat special in that they create an initial
asg of size 0/0/0 and then subsequently resize it to the target
capacity.

This PR protects against premature success by favoring the
clouddriver-supplied capacity if we see a 0/0/0 asg but expect a
non-0/0/0 asg.

The edge case only appears when the `ForceCacheRefresh` task is dropped
from the deploy stages in `orca`.

relates to spinnaker/clouddriver#3356

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
@@ -122,15 +122,16 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
         def isComplete = hasSucceeded(stage, serverGroup, serverGroup.instances ?: [], interestingHealthProviderNames)
         if (!isComplete) {
           Map newContext = getAdditionalRunningStageContext(stage, serverGroup)
-          if (seenServerGroup && !stage.context.capacitySnapshot) {
-            newContext.zeroDesiredCapacityCount = 0
-            newContext.capacitySnapshot = [
-              minSize        : serverGroup.capacity.min,
-              desiredCapacity: serverGroup.capacity.desired,
-              maxSize        : serverGroup.capacity.max
-            ]
-          }
           if (seenServerGroup) {
+            if (!stage.context.capacitySnapshot) {
+              newContext.zeroDesiredCapacityCount = 0
+              newContext.capacitySnapshot = newContext.capacitySnapshot ?: [
+                  minSize        : serverGroup.capacity.min,
+                  desiredCapacity: serverGroup.capacity.desired,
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/AbstractInstancesCheckTask.groovy
```diff
@@ -122,15 +122,16 @@ abstract class AbstractInstancesCheckTask extends AbstractCloudProviderAwareTask
...
-            newContext.capacitySnapshot = [
-              minSize        : serverGroup.capacity.min,
-              desiredCapacity: serverGroup.capacity.desired,
-              maxSize        : serverGroup.capacity.max
-            ]
-          }
           if (seenServerGroup) {
+            if (!stage.context.capacitySnapshot) {
+              newContext.zeroDesiredCapacityCount = 0
+              newContext.capacitySnapshot = newContext.capacitySnapshot ?: [
```
ajordens: Only change here is to allow a sub-class to override the `capacitySnapshot`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForUpInstancesTask.groovy
```diff
@@ -154,4 +167,38 @@ class WaitForUpInstancesTask extends AbstractWaitingForInstancesTask {
     // no health indicators is indicative of being down
     return !healths || healths.any { it.state == 'Down' || it.state == 'OutOfService' }
   }
+
+  private static Map<String, Integer> getServerGroupCapacity(Stage stage, Map serverGroup) {
```
ajordens: Will get some proper unit tests of various edge-cases in here ... wanted to get feedback on approach first.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForUpInstancesTask.groovy
```diff
@@ -154,4 +167,38 @@ class WaitForUpInstancesTask extends AbstractWaitingForInstancesTask {
...
+
+    if (serverGroup.capacity.max == 0 && initialTargetCapacity.max != 0) {
+      serverGroupCapacity = initialTargetCapacity
+    }
+
+    return serverGroupCapacity
+  }
+
+  private static Map<String, Integer> getInitialTargetCapacity(Stage stage, Map serverGroup) {
+    def katoTasks = (stage.context."kato.tasks" as List<Map<String, Object>>)?.reverse()
```
ajordens: Super clumsy to fetch "kato" results ... same comment as above re: edge case testing.

PR #2649, opened 2019-02-08
refactor(expressions): Use kork-expressions

jkschneider:
Removes duplication of expression logic that has now been moved to `kork-expressions`.


robzienert:
Looks like this has a conflict now.

robzienert:
Ping. Would be nice to see this in.

PR #2652, opened 2019-02-08
fix(webhooks): propogate body on failed calls

marchello2000:
Instead of throwing an exception, make sure to propagate task results on failed webhook calls.
Also, best effort JSON parse of the body (since it comes in as a string) for any SPeL
expressions to still work.


marchello2000:
@ajordens : removed the `JsonSlurper`, thanks for the tip!

PR #2653, opened 2019-02-08
fix(errors): Better error messages

marchello2000:
* Improve error messages in FindImage stage when images are inferred from subsequent deploys
* Don't surface intermediate errors during SpEL evaluation

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
@@ -125,10 +125,12 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
 
     // Supplement config with regions from subsequent deploy/canary stages:
     def deployRegions = regionCollector.getRegionsFromChildStages(stage)
+    Set<String> inferredRegions = new HashSet<>()
 
     deployRegions.forEach {
       if (!config.regions.contains(it)) {
         config.regions.add(it)
+        inferredRegions.add(it)
         log.info("Inferred and added region ($it) from deploy stage to FindImageFromClusterTask (executionId: ${stage.execution.id})")
       }
     }
@@ -170,7 +172,7 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
             return [(location): null]
           }
 
-          throw new IllegalStateException("Could not find cluster '$config.cluster' for '$account' in '$location.value'.")
+          throw new IllegalStateException("Could not find cluster '${config.cluster}' for '$account' in '${location.value}'.")
         }
         throw e
       }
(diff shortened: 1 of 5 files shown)


ajordens:
LGTM

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -205,14 +207,19 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
...
-            throw new IllegalStateException("Missing images in $stillUnresolved.value")
-          }
-        } else {
-          throw new IllegalStateException("Missing images in $unresolved.value")
+          unresolved = imageSummaries.findResults { it.value == null ? it.key : null }
         }
       }
+
+      if (unresolved) {
+        def errorMessage = "Missing image '${searchNames[0]}' in regions: $unresolved.value"
```
robzienert: We don't need to wrap `${unresolved.value}`?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
```diff
@@ -62,11 +62,16 @@ public ContextParameterProcessor(ContextFunctionConfiguration contextFunctionCon
   }
 
   public Map<String, Object> process(Map<String, Object> source, Map<String, Object> context, boolean allowUnknownKeys) {
+    ExpressionEvaluationSummary summary = new ExpressionEvaluationSummary();
+
+    return process(source, context, allowUnknownKeys, summary);
+  }
+
+  public Map<String, Object> process(Map<String, Object> source, Map<String, Object> context, boolean allowUnknownKeys, ExpressionEvaluationSummary summary) {
```
robzienert: Wouldn't mind seeing the arguments get wrapped so we're not going beyond 120.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
```diff
@@ -60,12 +62,28 @@ interface ExpressionAware {
...
+    // context. Otherwise, it's very confusing in the UI because the value is clearly correctly evaluated but
+    // the error is still shown
+    if (hasFailedExpressions()) {
+      val failedExpressions = this.context[PipelineExpressionEvaluator.SUMMARY] as MutableMap<String, *>
+
+      failedExpressions.keys.forEach( { expressionKey ->
+        if (evalSummary.wasAttempted(expressionKey) && !evalSummary.hasFailed(expressionKey)) {
+          failedExpressions.remove(expressionKey)
+        }
+      })
```
robzienert: Just a small little thing here, final arguments in Kotlin into a method don't need to be wrapped in parens.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
marchello2000: good catch, thanks!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
marchello2000: 👍

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
marchello2000: thanks!

PR #2654, opened 2019-02-11
fix(dependencies): update minutest dependency in orca-sql to fix build

claymccoy:
This kept Orca from building. I’m guessing that it wasn’t caught because everyone else (including travis) have a com.oneeyedmen:minutest:0.34.0 sitting in their .m2 cache.
It is described here (https://github.com/dmcg/minutest) that the group for the minutest project changed from ‘com.oneeyedmen’ to ‘dev.minutest’ and com.oneeyedmen:minutest:0.34.0 can no longer be referenced. The minor code changes are just what was needed when updating the dep.


ajordens:
LGTM

brandonnelson3:

ezimanyi:

ezimanyi:
Looks like the dependency was only added in 1.12, so no need to cherry pick earlier.

PR #2656, opened 2019-02-12
 feat(artifacts): SpEL evaluable expected artifact inputs

jkschneider:
Resolve expected artifact expressions in `ArtifactResolver` before attempting to match against incoming artifacts.

This PR also contains a polish commit that handles generics better in `ArtifactResolver`, resulting in fewer unchecked casts.


ezimanyi:
Looks like some tests need to pass the new argument to the `ArtifactResolver` constructor, but otherwise LGTM.

jkschneider:
Closing based on our conversation today (at least for now). It may still be that SpEL evaluated expected artifact inputs are useful as an alternative mechanism to avoid having to use Jinja to define some _a priori_ knowledge of how to synthesize an artifact from trigger context, but we can revisit a little later.

PR #2658, opened 2019-02-13
fix(logging context): add application name to the MDC

marchello2000:
Include application name from the execution to MDC during task execution
This way it's always there which is convenient for log analysis


marchello2000:
@robzienert What do you think about this? I was looking at some logs and thought this would be useful

robzienert:
You also need to remove this value in the `finally` block. Looks good aside.

marchello2000:
Doh, that's embarrassing, thanks & fixed

PR #2660, opened 2019-02-13
feat(redblack): pin min size of source server group

dreynaud:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy
```diff
@@ -126,8 +126,11 @@ class AwsDeployStagePreProcessor implements DeployStagePreProcessor {
   }
 
   private static boolean shouldPinSourceServerGroup(String strategy) {
-    // TODO-AJ consciously only enabling for rolling red/black -- will add support for other strategies after it's working
-    return strategy == "rollingredblack"
+    // TODO(dreynaud): consider adding support for highlander
+    // it would make sense to avoid the source server group getting scaled down in case of a long running
+    // deploy that needs to be canceled, but if we just add "highlander" in the supported strategies here
+    // the unpin task fails as it is started after the shrinkCluster task
+    return strategy in ["rollingredblack", "redblack"]
```
dreynaud: @ajordens no objection?

ajordens:
Does all the recover if pipeline fails work correctly?

That was a worry when it initially went in for RRB-only.

dreynaud:
Good point, I need to check that. Thanks!

Le jeu. 14 févr. 2019 à 14:41, Adam Jordens <notifications@github.com> a
écrit :

dreynaud:
> Did you verify that we recover in the R/B case if the deploy fails?
> (ie. Deploy fails and source will correctly get unpinned — some support did go in to tag a new asg that was pinned and then a background job as a last ditch effort to recover).
> 
> That was a worry and a reason why it was initially scoped to RRB.

Mhhh it does attempt to unpin it but then fails because of a 3min timeout? Not sure where that comes from:

```
2019-02-15 15:25:01.180 ERROR 5070 --- [    handlers-19] c.n.s.orca.q.handler.RunTaskHandler      : [] Error running DetermineHealthProvidersTask for pipeline[01D3SQERP4SACHZ1QS3SRCMCKW]
com.netflix.spinnaker.orca.exceptions.TimeoutException: Stage Unpin egg-redblack-v005 timed out after 3 minutes 1 second
```

ajordens:
I believe you figured out where that came from ... let's chat when you're back in the office.

Quite possible we should be explicitly overriding timeouts on these stages we expect to _always_ be running.

dreynaud:
With fcd7848b3a1cf2ae3702c3188c4c88763c824e43, I verified that I can fail a deploy because of a timeout, and it successfully unpins the source server group regardless at the end.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy
ghuntley: Ship it. https://twitter.com/FakeRyanGosling/status/1106714429247221761?s=09

PR #2664, opened 2019-02-14
feat(gremlin): Adds a Gremlin stage

mattrjacobs:
Gremlin is a fault-injection tool, and this addition wraps its API to allow for creation/monitoring/halting a fault-injection.


ajordens:
To match the suggestions from `gate`, can we:
- have a `orca-integrations-gremlin` artifact (just mv `orca-gremlin` to `orca-integrations-gremlin`)
- nest gremlin config under `integrations.gremlin`
- disabled by default (a `ConditionalOn` in your `@Configuration` -- it's important we find a way to make integrations easily discoverable but prefer that the out-of-the-box experience has minimal non-core components running by default)

orca-web/src/main/groovy/com/netflix/spinnaker/orca/Main.groovy
```diff
@@ -78,7 +79,8 @@ import org.springframework.scheduling.annotation.EnableAsync
   KayentaConfiguration,
   WebhookConfiguration,
   KeelConfiguration,
-  QosConfiguration
+  QosConfiguration,
+  GremlinConfiguration
```
ajordens: If you move `GremlinConfiguration` to `com.netflix.spinnaker.config`, it'll get picked up automatically and we don't need to touch anything outside of your `orca-gremlin` artifact.

This would be preferable.

mattrjacobs: I changed the package name to `com.netflix.spinnaker.config`, but I still need the `orca-web` module to depend on `orca-integrations-gremlin`.  I also believe that I need to still have an explicit reference to `GremlinConfiguration` (parallel to `KeelConfiguration` / `QosConfiguration`)

I think I addressed all of the other comments.

ajordens:
Changes look good to me, I'll merge after the TravisCI build completes.

PR #2672, opened 2019-02-21
feat(delivery): upsert and delete delivery configs through orca

emjburns:
Adding support for upserting and deleting delivery configs as tasks.
Making the `MonitorFront50Task` able to do more things than just monitor pipeline saves.
Thoughts on the keys for the stages?

A create stage body has no id supplied:
```
{
  "type": "upsertDeliveryConfig",
  "delivery": {
    "application": "emilyapp",
    "otherfields": "blah blah"
  }
}
```
An update has an id supplied
```
{
  "type": "upsertDeliveryConfig",
  "delivery": {
    "id": "331d7c36-695f-467a-97a0-2dafde036f9d",
    "application": "emilyapp",
    "otherfields": "blah blah"
  }
}
```
A delete stage body:
``` 
{
  "type": "deleteDeliveryConfig",
  "deliveryConfigId": "331d7c36-695f-467a-97a0-2dafde036f9d"
}
```


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/Front50Service.groovy
```diff
@@ -129,6 +130,18 @@ interface Front50Service {
   @POST("/serviceAccounts")
   Response saveServiceAccount(@Body ServiceAccount serviceAccount)
 
+  @GET("/deliveries/{id}")
+  Response getDeliveryConfig(@Path("id") String id)
```
marchello2000: How come we don't type the response, e.g. `Response<com.netflix.spinnaker.front50.model.delivery.Delivery>`

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeleteDeliveryConfigTask.java
```diff
@@ -0,0 +1,66 @@
...
+  public DeleteDeliveryConfigTask(Front50Service front50Service, ObjectMapper objectMapper) {
+    this.front50Service = front50Service;
+    this.objectMapper = objectMapper;
+    deliveryConfigUtils = new DeliveryConfigUtils(front50Service, objectMapper);
+  }
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    if (!stage.getContext().containsKey("deliveryConfigId")) {
```
marchello2000: if this were `stage.mapTo()` instead with a class that has `deliverConfigId` then if we ever change the name of the field we would get a compile-time error versus a confusing runtime error. Get to harness refactor, etc functionality that tools provide

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeleteDeliveryConfigTask.java
```diff
@@ -0,0 +1,66 @@
...
+
+    if (!config.isPresent()) {
+      log.debug("Config {} does not exist, considering deletion successful.", configId);
+      return new TaskResult(ExecutionStatus.SUCCEEDED);
+    }
+
+    try {
+      log.debug("Deleting delivery config: " + objectMapper.writeValueAsString(config.get()));
+    } catch (JsonProcessingException e) {
+      log.debug("Deleting malformed delivery config:" + config.get());
```
marchello2000: nit: this can probably not because it's malformed but rather some underlying issue json parser. Not sure if it's worth logging the original exception, we are deleting the config after all..

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeliveryConfigUtils.java
```diff
@@ -0,0 +1,51 @@
...
+    this.objectMapper = objectMapper;
+  }
+
+  public Optional<DeliveryConfig> getDeliveryConfig(String id) {
+    try {
+      DeliveryConfig deliveryConfig = readResponse(front50Service.getDeliveryConfig(id));
+      return deliveryConfig == null ? Optional.empty() : Optional.of(deliveryConfig);
+    } catch (RetrofitError e) {
+      //ignore an unknown (404) or unauthorized (403)
+      if (e.getResponse() != null && Arrays.asList(404, 403).contains(e.getResponse().getStatus())) {
```
marchello2000: if it can return 403, can it also return 401?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeliveryConfigUtils.java
```diff
@@ -0,0 +1,51 @@
...
+        throw e;
+      }
+    }
+  }
+
+  public DeliveryConfig readResponse(Response response) {
+    try {
+      return objectMapper.readValue(response.getBody().in(), DeliveryConfig.class);
+    } catch (IOException e) {
+      log.error("Unable to deserialize delivery config, reason: ", e.getMessage());
```
marchello2000: If this fails in production, we are back to the same problem of not being able to diagnose the root cause without cracking open logs. Would be nicer if the exception ended up in the context somehow

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorFront50Task.java
```diff
@@ -114,22 +97,76 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+      }
+
+      Long lastModifiedTime;
+      if (object.get().containsKey("updateTs")) {
+        lastModifiedTime = Long.valueOf(object.get().get("updateTs").toString());
+      } else {
+        lastModifiedTime = Long.valueOf(object.get().get("lastModified").toString());
+      }
+
+      if (lastModifiedTime < (startTime - gracePeriodMs)) {
```
marchello2000: speaking out loud: seems super hoaky because it relies on the various service instance to have reasonably synchronized timestamps... (e.g. `startTime` is set by one instance, `updateTs` by another, etc)

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/UpsertDeliveryConfigTask.java
```diff
@@ -0,0 +1,86 @@
...
+
+    Map<String, Object> outputs = new HashMap<>();
+    outputs.put("application", deliveryConfig.getApplication());
+
+    DeliveryConfig savedConfig;
+    try {
+      savedConfig = objectMapper.readValue(response.getBody().in(), DeliveryConfig.class);
+      outputs.put("deliveryConfig", savedConfig);
+    } catch (IOException e) {
+      log.error("Unable to deserialize saved delivery config, reason: ", e.getMessage());
```
marchello2000: again, if something goes wrong... must open logs to see what was the root cause

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/UpsertDeliveryConfigTask.java
```diff
@@ -0,0 +1,86 @@
...
+  @Autowired
+  public UpsertDeliveryConfigTask(Front50Service front50Service, ObjectMapper objectMapper) {
+    this.front50Service = front50Service;
+    this.objectMapper = objectMapper;
+    deliveryConfigUtils = new DeliveryConfigUtils(front50Service, objectMapper);
+  }
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
```
marchello2000: This has been discussed in a few tickets. Not suggesting we solve it here, just reminding/bringing up.
There is a issue with this style upsert in that "last write wins".
If we have two services get the config, tweak it, and upsert it one of them will lose. (obviously, this is an issue currently with a several of endpoints)

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/Front50Service.groovy
emjburns: I think that typed responses is a feature of retrofit 2, and we're on 1.9. I agree that typing of responses would be better, but we would have to upgrade Orca.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeleteDeliveryConfigTask.java
emjburns: It doesn't seem like typing the expected field brings us a ton of value - in `MonitorFront50Task` we type the context (`StageData`), but still need to check if the fields are null because if something changes in what we're sending the field doesn't get mapped. I'm happy to type this and continue with a good pattern though.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeliveryConfigUtils.java
emjburns: Could you explain this more? Is your desired state being able to diagnose everything by looking at the execution context and not the logs? Are there any errors that are best only in the logs? Any ideas on how to surface something like this in the execution context?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorFront50Task.java
emjburns: This is not my code or my "dirty hack", I'm just generalizing it to work for monitoring other front50 objects.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/UpsertDeliveryConfigTask.java
emjburns: There's probably a better story here with how to merge updates so you can just change one field, you don't overwrite other people, etc. However for simplicity here I just want the last write to win.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeleteDeliveryConfigTask.java
emjburns: This log statement is here more as a reaction to past issues. People sometimes ask us to revert deletions, and it's handy to be able to find that full object in the logs so that it's easy to revert. 

I'll log the error here just in case.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeliveryConfigUtils.java
emjburns: I think it just returns 403 but I'm not sure. I'll add 401 to this list.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/UpsertDeliveryConfigTask.java
emjburns: this is more an indication that we broke something. I don't think the user will be able to fix it - how would you suggest surfacing this in an appropriate way?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeleteDeliveryConfigTask.java
ajordens: Back-to-back logs also seem odd to me ... the try/catch is super narrowly scoped so I don't know that we need one logging the exception + `config.get()`.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeleteDeliveryConfigTask.java
```diff
@@ -38,29 +39,35 @@ public DeleteDeliveryConfigTask(Front50Service front50Service, ObjectMapper obje
...
     }
 
     try {
       log.debug("Deleting delivery config: " + objectMapper.writeValueAsString(config.get()));
     } catch (JsonProcessingException e) {
+      log.warn("Error serializing delivery config object: ", e);
       log.debug("Deleting malformed delivery config:" + config.get());
     }
-    Response response = front50Service.deleteDeliveryConfig(config.get().getApplication(), configId);
+    Response response = front50Service.deleteDeliveryConfig(config.get().getApplication(), stageData.deliveryConfigId);
```
ajordens: `deleteDeliveryConfig()` is going to throw an exception if it's non-200 (or maybe non-200/300) ... general pattern would be if you get a successful response than it's SUCCEEDED otherwise the exceptional case is TERMINAL.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/Front50Service.groovy
robzienert: Responses can be typed. Use `Response` if you need access to additional information such as headers or response status.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeliveryConfigUtils.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public DeliveryConfigUtils(Front50Service front50Service, ObjectMapper objectMapper) {
+    this.front50Service = front50Service;
+    this.objectMapper = objectMapper;
+  }
+
+  public Optional<DeliveryConfig> getDeliveryConfig(String id) {
+    try {
+      DeliveryConfig deliveryConfig = readResponse(front50Service.getDeliveryConfig(id));
+      return deliveryConfig == null ? Optional.empty() : Optional.of(deliveryConfig);
```
robzienert: `Optional.ofNullable(deliveryConfig)`

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeliveryConfigUtils.java
robzienert: I'd prefer this to b `log.error("Unable to deserialize delivery config", e)` -- then we'll get the full stacktrace.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/model/DeliveryConfig.java
```diff
@@ -0,0 +1,100 @@
+package com.netflix.spinnaker.orca.front50.model;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DeliveryConfig {
```
ajordens: I suspect we have `lombok` kicking around and you could `@Data` this and save some typing.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/UpsertDeliveryConfigStage.java
```diff
@@ -0,0 +1,20 @@
...
+
+import javax.annotation.Nonnull;
+
+@Component
+public class UpsertDeliveryConfigStage implements StageDefinitionBuilder {
+  @Override
+  public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
+    builder
+      .withTask("upsertDeliveryConfig", UpsertDeliveryConfigTask.class)
+      .withTask("monitorSave", MonitorFront50Task.class);
```
ajordens: `s/monitorSave/monitorUpsert` for consistent naming?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeleteDeliveryConfigTask.java
```diff
@@ -0,0 +1,73 @@
...
+  public DeleteDeliveryConfigTask(Front50Service front50Service, ObjectMapper objectMapper) {
+    this.front50Service = front50Service;
+    this.objectMapper = objectMapper;
+    deliveryConfigUtils = new DeliveryConfigUtils(front50Service, objectMapper);
+  }
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    StageData stageData = stage.mapTo(StageData.class);
```
ajordens: I like the `StageData` pattern

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorFront50Task.java
robzienert: We're all ntp'd, so timestamp drift should be quite minimal. It's not performing precise calculations here, so I think it's "fine" as is. The better way would be to have a deadline timestamp, but meh.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorFront50Task.java
```diff
@@ -138,5 +175,8 @@ public TaskResult execute(@Nonnull Stage stage) {
 
     @JsonProperty("pipeline.name")
     public String pipelineName;
+
+    @JsonProperty("deliveryConfig")
```
robzienert: I don't think this annotation is necessary, since the property name is the same.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/UpsertDeliveryConfigTask.java
```diff
@@ -0,0 +1,86 @@
...
+    if (!stage.getContext().containsKey("delivery")) {
+      throw new IllegalArgumentException("Key 'delivery' must be provided.");
+    }
+
+    //todo eb: base64 encode this if it will have expressions
+    DeliveryConfig deliveryConfig = objectMapper
+      .convertValue(stage.getContext().get("delivery"), new TypeReference<DeliveryConfig>(){});
+
+    log.debug("Received delivery config: " + deliveryConfig);
```
ajordens: Expecting this `debug` will eventually be removed

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/UpsertDeliveryConfigTask.java
ajordens: log message is missing a `{}`

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/UpsertDeliveryConfigTask.java
```diff
@@ -0,0 +1,86 @@
...
+
+    Response response;
+    if (deliveryConfigUtils.configExists(deliveryConfig.getId())) {
+      response = front50Service.updateDeliveryConfig(deliveryConfig.getId(), deliveryConfig);
+    } else {
+      response = front50Service.createDeliveryConfig(deliveryConfig);
```
ajordens: same comment as before about `Response` vs `DeliveryConfig`

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeleteDeliveryConfigTask.java
emjburns: @ajordens better if I remove the second logs and just ignore the `JsonProcessingException`?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/UpsertDeliveryConfigTask.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @Override
+  public long getBackoffPeriod() {
+    return 1000;
+  }
+
+  @Override
+  public long getTimeout() {
+    return TimeUnit.SECONDS.toMillis(30);
```
ajordens: Normally we don't retry on write tasks ... is there something in here that guarantees idempotent writes? (if so, I'd suggest a higher timeout to account for network timeouts)

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeleteDeliveryConfigTask.java
ajordens: Looking closer, I don't really see the try/catch around an attempted serialization to be all that relevant. 

Seems removable.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/UpsertDeliveryConfigTask.java
emjburns: Good point. I'll make it just a task!

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeleteDeliveryConfigTask.java
marchello2000: The main thing it buys is compile-time verification and easy "find all usages", etc. I like the `stageData` @ajordens recommended

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeliveryConfigUtils.java
marchello2000: I think, ideally, we surface the relevant errors to the end user so they can self-service issues. Thinking about error messages... but the immediate response is based on several recent support issues where folks ask us why a pipeline fails and we have to go to logs to see the exception and tell them what it said. Ideally we short circuit that. Will get back when I have some more thoughts about it

PR #2675, opened 2019-02-22
fix(redis): Fix intermittently failing test

ezimanyi:
Another test fails occasionally due to non-monotonic ULIDs over short timescales.


ajordens:
LGTM

PR #2677, opened 2019-02-26
fix(execution): Ensure that synthetic stages respect execution strategy

marchello2000:
Synthetic stages don't (always) get the parent context. Since the user
might set execution options (e.g. whether to fail or not the pipeline) on
the parent stage, those directives should be respected by synthetic stages.
With this change, recursively look up these directives on synthetic stages'
parent if they aren't explicitly passed down.

An example here is the deploy stages: if a synthetic generated by one of the
deploy strategies fails but the deploy stage specified to ignore failure, the
ignore behavior will not be respected.

### DIFF
--- a/orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
+++ b/orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
@@ -119,18 +119,33 @@ inline fun <reified O> Stage.mapTo(pointer: String): O = mapTo(pointer, O::class
 
 inline fun <reified O> Stage.mapTo(): O = mapTo(O::class.java)
 
-fun Stage.shouldFailPipeline(): Boolean =
-  context["failPipeline"] in listOf(null, true)
+fun Stage.failureStatus(default: ExecutionStatus = TERMINAL): ExecutionStatus {
+  // Look for `continuePipeline` or `failPipeline` attributes on parent stages in the event
+  // that they were not explicitly passed down as part of the synthetic stage context.
+  // Note: a value in a parent stage overrides non-existent value in synthetic child stage
+  fun getFromParent(param: String, defaultIfNotPresent: Boolean): Boolean {
+    var stage: Stage? = this
 
-fun Stage.shouldContinueOnFailure(): Boolean =
-  context["continuePipeline"] == true
+    while (stage != null) {
+      if (stage.context[param] != null) {
+        return (stage.context[param] == true)
+      }
 
-fun Stage.failureStatus(default: ExecutionStatus = TERMINAL) =
-  when {
-    shouldContinueOnFailure() -> FAILED_CONTINUE
-    shouldFailPipeline()      -> default
-    else                      -> STOPPED
+      stage = if (stage.syntheticStageOwner != null) stage.parent() else null
+    }
+
+    return defaultIfNotPresent
   }
+
+  var shouldContinueOnFailure = getFromParent("continuePipeline", false)
+  var shouldFailPipeline = getFromParent("failPipeline", true)
+
+  return when {
+    shouldContinueOnFailure -> FAILED_CONTINUE
+    shouldFailPipeline -> default
+    else -> STOPPED
+  }
+}
 
 fun Stage.isManuallySkipped(): Boolean {
   return context["manualSkip"] == true || parent?.isManuallySkipped() == true
--- a/orca-kotlin/src/test/kotlin/com/netflix/spinnaker/orca/ext/StageTest.kt
+++ b/orca-kotlin/src/test/kotlin/com/netflix/spinnaker/orca/ext/StageTest.kt
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
```diff
@@ -103,34 +103,47 @@ fun Stage.allAfterStagesSuccessful(): Boolean =
...
+
+      stage = if (stage.syntheticStageOwner != null) stage.parent() else null
+    }
 
-fun Stage.shouldContinueOnFailure(): Boolean =
-  context["continuePipeline"] == true
+    return default
+  }
+
+  // true if explicitly set to true
```
ajordens: Not sure this inline comment is necessary here?

orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
```diff
@@ -103,34 +103,47 @@ fun Stage.allAfterStagesSuccessful(): Boolean =
 fun Stage.anyBeforeStagesFailed(): Boolean =
   beforeStages().any { it.status in listOf(TERMINAL, STOPPED, CANCELED) }
 
-fun Stage.anyAfterStagesFailed(): Boolean =
```
ajordens: Why was this removed?

orca-kotlin/src/test/kotlin/com/netflix/spinnaker/orca/ext/StageTest.kt
```diff
@@ -0,0 +1,88 @@
...
+  describe("parent stage context is empty but child is set") {
+    val pipeline = pipeline {
+      application = "test"
+      stage {
+        refId = "1"
+        type = "parent"
+        status = ExecutionStatus.SUCCEEDED
+        context = mapOf(
+          "failPipeline" to false,
+          "continuePipeline" to true)
```
ajordens: conventions would say `)` on next line

orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
```diff
@@ -103,34 +103,47 @@ fun Stage.allAfterStagesSuccessful(): Boolean =
...
-  firstAfterStages().isNotEmpty()
-
 inline fun <reified O> Stage.mapTo(pointer: String): O = mapTo(pointer, O::class.java)
 
 inline fun <reified O> Stage.mapTo(): O = mapTo(O::class.java)
 
-fun Stage.shouldFailPipeline(): Boolean =
-  context["failPipeline"] in listOf(null, true)
+fun Stage.failureStatus(default: ExecutionStatus = TERMINAL): ExecutionStatus {
+  // Because not all stages pass the full context to their child
```
ajordens:
```
Look for `continuePipeline` or `failPipeline` attributes on parent stages in the event
that they were not explicitly passed down as part of the synthetic stage context.
```

orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
marchello2000: it wasn't used, so i figured I'd remove it

orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
```diff
@@ -103,34 +103,47 @@ fun Stage.allAfterStagesSuccessful(): Boolean =
...
-fun Stage.anyAfterStagesFailed(): Boolean =
-  afterStages().any { it.status in listOf(TERMINAL, STOPPED, CANCELED) }
-
 fun Stage.allAfterStagesComplete(): Boolean =
   afterStages().all { it.status.isComplete }
 
 fun Stage.hasTasks(): Boolean =
   tasks.isNotEmpty()
 
-fun Stage.hasAfterStages(): Boolean =
```
marchello2000: wasn't used

orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
marchello2000: thanks, i like this wording better

orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
marchello2000: yeah, i wanted to make clear why one is defaulted to `true` and one to `false`, but I renamed the `default` param and I think it's clearer now and no longer needs the comment

orca-kotlin/src/test/kotlin/com/netflix/spinnaker/orca/ext/StageTest.kt
marchello2000: 👍

robfletcher:
I think this will work as intended but I'm not sure it's the correct behavior. For example if we have a deploy stage that should continue on failure, that doesn't mean all its synthetic stages should do the same. For example, if the disable previous server group step fails we should probably not carry on to try to destroy it.

marchello2000:
Discussed offline, since this behavior is what happens in most cases today already (since most synthetic stages inherit the context of the parent) we are going to go forward with this behavior though, as @robfletcher points out, it's not 100% correct. We will be addressing the correctness separately.

orca-kotlin/src/main/kotlin/com/netflix/spinnaker/orca/ext/Stage.kt
marchello2000: will put back in

marchello2000:
closing since #2686 is a better solution

PR #2679, opened 2019-02-27
feat(webhook): Add support to whitelist domains for webhook stage

aravindmd:
* We don't allow webhooks to post to external endpoint , this change allows whitelisting a list of domains for webhook stage to post to external endpoint.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
@@ -32,6 +32,7 @@ public class UserConfiguredUrlRestrictions {
     private boolean rejectLocalhost = true;
     private boolean rejectLinkLocal = true;
     private List<String> rejectedIps = new ArrayList<>(); // can contain IP addresses and/or IP ranges (CIDR block)
+    private List<String> allowedDomains = new ArrayList<String>();
 
     public Builder withAllowedHostnamesRegex(String allowedHostnamesRegex) {
       setAllowedHostnamesRegex(allowedHostnamesRegex);
@@ -58,8 +59,13 @@ public class UserConfiguredUrlRestrictions {
       return this;
     }
 
+    public Builder withWhitelistedDomains(List<String> allowedDomains) {
+      setAllowedDomains(allowedDomains);
+      return this;
+    }
+
     public UserConfiguredUrlRestrictions build() {
-      return new UserConfiguredUrlRestrictions(Pattern.compile(allowedHostnamesRegex), allowedSchemes, rejectLocalhost, rejectLinkLocal, rejectedIps);
+      return new UserConfiguredUrlRestrictions(Pattern.compile(allowedHostnamesRegex), allowedSchemes, rejectLocalhost, rejectLinkLocal, rejectedIps, allowedDomains);
     }
   }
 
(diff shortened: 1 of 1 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -68,13 +74,15 @@ public UserConfiguredUrlRestrictions build() {
   private final boolean rejectLocalhost;
   private final boolean rejectLinkLocal;
   private final Set<String> rejectedIps;
+  private final List<String> whitelistedDomains;
```
marchello2000: I prefer `allowedDomains`

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -86,10 +94,9 @@ public URI validateURI(String url) throws IllegalArgumentException {
       if (!allowedSchemes.contains(u.getScheme().toLowerCase())) {
         throw new IllegalArgumentException("unsupported URI scheme " + url);
       }
-      if (!allowedHostnames.matcher(u.getHost()).matches()) {
-        throw new IllegalArgumentException("Host not allowed " + u.getHost() + ". Host much match " + allowedHostnames.toString() + ".");
+      if (!allowedHostnames.matcher(u.getHost()).matches() && !whitelistedDomains.contains(u.getHost())) {
```
marchello2000: This approach is just a simpler version of adding to the regex, e.g.
`'^(.+\.netflix(?:aws)?\.(?:net|com)$)|(myawesomedomain.com)'

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
ajordens: Could go with `allowedHostnames` to match the `allowedHostnamesRegex`.

I could see `allowedHostnames` just containing a list of strings that may themselves be regexes. This would allow for a simpler replacement of `allowedHostnamesRegex`.

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -32,6 +32,7 @@
     private boolean rejectLocalhost = true;
     private boolean rejectLinkLocal = true;
     private List<String> rejectedIps = new ArrayList<>(); // can contain IP addresses and/or IP ranges (CIDR block)
+    private List<String> allowedDomains = new ArrayList<String>();
```
ajordens: You don't need `String` on the right-hand side.

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -133,4 +140,7 @@ public boolean isRejectLocalhost() {
   public boolean isRejectLinkLocal() {
     return rejectLinkLocal;
   }
+
+  public List<String> getWhitelistedDomains() { return whitelistedDomains; }
```
ajordens: Should be separate lines for consistency.

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
ajordens: I would be fine with just modifying the regex as-is in our case and putting off this PR until sometime in the future when our regex becomes unwieldy.

aravindmd: Sounds good to me , i will close this PR for now.

PR #2680, opened 2019-02-27
fix(execution): Ensure exceptions in stage planning are captured

marchello2000:
Presently any exceptions that occur during stage planning are not captured
and therefore will not show up in the execution JSON/UI for end-user to
see. This can be very confusing as there is no explanation for why a pipeline
stage fails.


emjburns:
LGTM

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -87,7 +89,10 @@ class CompleteStageHandler(
           stage.status = status
           stage.endTime = clock.millis()
         } catch (e: Exception) {
-          log.error("Failed to construct after stages for $stage.id", e)
+          log.error("Failed to construct after stages for ${stage.name} ${stage.id}", e)
```
robzienert: `log.error("Failed blah blah for stage ${stage.name}", e, kv("stageId", stage.id))` is pretty neat-o too.

marchello2000: cool, thanks!

PR #2681, opened 2019-02-27
fix(MPT): Propagate notifications from pipeline config

marchello2000:
One can specify notification on the pipeline in deck (even if pipeline is templated)
However, those notifications are not respected because they don't end up in the
pipeline config. Copy them there so notifications work as expected.

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/handler/V1TemplateLoaderHandler.kt
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/handler/V1TemplateLoaderHandler.kt
@@ -37,6 +37,7 @@ class V1TemplateLoaderHandler(
 
   override fun handle(chain: HandlerChain, context: PipelineTemplateContext) {
     val config = objectMapper.convertValue(context.getRequest().config, TemplateConfiguration::class.java)
+    config.configuration.notifications.addAll(context.getRequest().notifications)
 
     // Allow template inlining to perform plans without publishing the template
     if (context.getRequest().plan && context.getRequest().template != null) {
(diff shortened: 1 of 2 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/handler/V1TemplateLoaderHandler.kt
```diff
@@ -37,6 +37,7 @@ class V1TemplateLoaderHandler(
 
   override fun handle(chain: HandlerChain, context: PipelineTemplateContext) {
     val config = objectMapper.convertValue(context.getRequest().config, TemplateConfiguration::class.java)
+    config.configuration.notifications.addAll(context.getRequest().notifications)
```
marchello2000: I don't think we need to dedupe notifications, but let me know if you disagree

robzienert: We probably do, actually. I'm not confident that orca / echo would know enough to dedupe further down.

marchello2000: ok. I am still not clear how `config.configuration.notification` would get filled since `deck` doesn't do that, but obviously possible via API

marchello2000: I guess the other comment is that one can create multiple notifications today that go to the same spot - they aren't deduped... But if you feel strongly I will add the dedupe logic here

marchello2000: resolved offline: leave as is for now

PR #2683, opened 2019-02-28
fix(triggers): surface error if build not found

emjburns:
If a user tries to start a pipeline manually from a jenkins trigger and the build is not found we fail super silently. Let's at least throw a new exception with the cause. 

This message will show up in a tiny tool tip in the UI. Not a great experience, but better than not showing the error at all.

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -266,7 +266,16 @@ class OperationsController {
 
   private void decorateBuildInfo(Map trigger) {
     if (trigger.master && trigger.job && trigger.buildNumber) {
-      def buildInfo = buildService.getBuild(trigger.buildNumber, trigger.master, trigger.job)
+      def buildInfo
+      try {
+        buildInfo = buildService.getBuild(trigger.buildNumber, trigger.master, trigger.job)
+      } catch (RetrofitError e) {
+        if (e.response?.status == 404) {
+          throw new IllegalStateException("Build " + trigger.buildNumber + " for " + trigger.master + "/" + trigger.job + " not found")
+        } else {
+          throw e
+        }
+      }
       if (buildInfo?.artifacts) {
         if (trigger.type == "manual") {
           trigger.artifacts = buildInfo.artifacts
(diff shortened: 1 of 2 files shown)


erikmunson:
🎉 thanks emily

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -266,7 +266,16 @@ class OperationsController {
 
   private void decorateBuildInfo(Map trigger) {
     if (trigger.master && trigger.job && trigger.buildNumber) {
-      def buildInfo = buildService.getBuild(trigger.buildNumber, trigger.master, trigger.job)
+      def buildInfo
+      try {
+        buildInfo = buildService.getBuild(trigger.buildNumber, trigger.master, trigger.job)
+      } catch (RetrofitError e) {
+        if (e.response?.status == 404) {
+          throw new IllegalStateException("Build " + trigger.buildNumber + " for " + trigger.master + "/" + trigger.job + " not found")
```
marchello2000: why not use string interpolation?

emjburns: Because groovy g strings sometimes behave weird and I don't like them so I often just use the + like a noob.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -266,7 +266,16 @@ class OperationsController {
...
+        } else {
+          throw e
```
marchello2000: i think it'd be nicer to do something like:
`throw new IllegalStageException("Failed to get the build ${...}...", e)`
that way we don't leak random exception without explanation.

(also this would be for `catch (Exception e)`

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
marchello2000: 👍 just curious - since I see diff patterns in the code base

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
emjburns: Do you mean instead of just throwing the exception without changing the message, adding some explanation?

marchello2000: yeah, say, you get retrofit error `Gateway Notfound`. If we just throw it as is, the UI will show `Gateway notfound` and the user will be scratching their head, wondering what does that mean. But, if we instead throw "We failed to get your build, btw, here is your inner exception:..." that gives them a ton more information. Also for anyone on support diagnosing will be pointed to the right place much faster with extra context

emjburns: Sounds good, I'll add that.

marchello2000:
LGTM

PR #2685, opened 2019-02-28
test(core): Basic startup test

ezimanyi:
Test that orca starts up with a very basic config, with just the baseUrl for each dependent service defined, and an in-memory execution queue.

### DIFF
--- a/orca-web/orca-web.gradle
+++ b/orca-web/orca-web.gradle
@@ -59,6 +59,7 @@ dependencies {
 
   spinnaker.group('test')
   testCompile project(":orca-test-groovy")
+  testCompile "com.netflix.spinnaker.keiko:keiko-mem:$keikoVersion"
 
   //this brings in the jetty GzipFilter which boot will autoconfigure
   runtime 'org.eclipse.jetty:jetty-servlets:9.2.11.v20150529'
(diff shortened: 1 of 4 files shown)


emjburns:
@ezimanyi is echo required? I thought halyard was using orca without w/o echo.

robzienert:
I'd definitely consider echo required. Halyard making changes to systems should have its events emitted somewhere.

ezimanyi:
Echo's on by default; I could get this to work as well by explicitly setting `echo.enabled: false` in the config file instead (and then removing the `@MockBean`).  You are right that I think Haylard doesn't have echo, but that's probably enough of an edge case to ignore in our base startup case?

emjburns:
FWIW I run orca locally w/o echo all the time. I agree that it's probably needed for a _good_ system, I just hesitate to say it's required for a startup test if it actually isn't.

emjburns:
👍 lgtm

ezimanyi:
That makes sense, and it makes me realize that igor/rosco are also definitely optional, so I update to just toggle them to false in the config and removed the mock beans.

I also realized that I hadn't fully mocked away the dependence on redis, so I added couple more flags and mock beans and was able to completely remove the redis url from the config.  (Prior to this change it was still trying to connect to redis and failing, but wasn't actually failing startup so the test passed.)

PR #2687, opened 2019-03-01
fix(templated-pipelines): handle a missing notifications field

skandragon:
This change should handle the case where notifications should not be null, but was in an existing template.

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/handler/V1TemplateLoaderHandler.kt
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/handler/V1TemplateLoaderHandler.kt
@@ -37,7 +37,7 @@ class V1TemplateLoaderHandler(
 
   override fun handle(chain: HandlerChain, context: PipelineTemplateContext) {
     val config = objectMapper.convertValue(context.getRequest().config, TemplateConfiguration::class.java)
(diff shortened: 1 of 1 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/handler/V1TemplateLoaderHandler.kt
```diff
@@ -37,7 +37,7 @@ class V1TemplateLoaderHandler(
 
   override fun handle(chain: HandlerChain, context: PipelineTemplateContext) {
     val config = objectMapper.convertValue(context.getRequest().config, TemplateConfiguration::class.java)
-    config.configuration.notifications.addAll(context.getRequest().notifications)
+    config.configuration.notifications.addAll(context.getRequest().notifications.orEmpty())
```
marchello2000: Thank you!!!

I am assuming this needs to be `?.orEmpty()`

marchello2000: (since, presumably you're guarding against `null` here?)

marchello2000: never mind i misread the docs :)

marchello2000:
LGTM

PR #2690, opened 2019-03-01
chore(clouddriver): rest helper for deploying clouddriver-sql with an empty cache

asher:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/CloudDriverCacheService.groovy
```diff
@@ -28,4 +29,6 @@ interface CloudDriverCacheService {
                             @Path("type") String type,
                             @Body Map<String, ? extends Object> data)
 
-}
\ No newline at end of file
+  @PUT("/admin/db/truncate/{namespace}")
```
ajordens: I suspect we'll want to put some authorization annotations on this API in clouddriver if it's being made accessible.

robzienert:
What's the word on this?

PR #2691, opened 2019-03-01
feat(spel): add manifestLabelValue helper function

maggieneterval:
@duftler proof of concept for using a helper function to resolve label values to enable blue/green deployments in which labels are dynamically patched into Services:

Used in patch manifest stage...

ez1g017frzr

...and resolved in pipeline execution:

ytgjism0dip

Would you recommend creating some sort of helper class for a ReplicaSet/Deployment manifest spec rather than checking each step of spec --> template --> metadata -->labels? This seems pretty verbose.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupport.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupport.java
@@ -24,6 +24,8 @@ import java.util.*;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Predicate;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.jayway.jsonpath.JsonPath;
+import com.jayway.jsonpath.PathNotFoundException;
 import com.netflix.spinnaker.orca.ExecutionStatus;
 import com.netflix.spinnaker.orca.pipeline.expressions.whitelisting.FilteredMethodResolver;
 import com.netflix.spinnaker.orca.pipeline.expressions.whitelisting.FilteredPropertyAccessor;
@@ -101,6 +103,7 @@ public class ExpressionsSupport {
         registerFunction(evaluationContext, "judgment", Object.class, String.class);
         registerFunction(evaluationContext, "judgement", Object.class, String.class);
         registerFunction(evaluationContext, "deployedServerGroups", Object.class, String[].class);
+        registerFunction(evaluationContext, "manifestLabelValue", Object.class, String.class, String.class, String.class);
       }
     } catch (NoSuchMethodException e) {
       // Indicates a function was not properly registered. This should not happen. Please fix the faulty function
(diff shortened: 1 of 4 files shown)


duftler:
I like the approach and find the patch manifest stage content very readable.

With respect to creating a helper class for walking spec->template->metadata->labels, you could just use something like `com.jayway.jsonpath.JsonPath` instead. It is already used in orca-webhook, so it wouldn't be introducing an additional dependency. You'd have to experiment a bit to confirm this, but I'm guessing the error messages it emits when a path can't be read are not quite as nice as the ones you've hand-crafted, but would suffice.

maggieneterval:
Thanks Matt, I'll add `com.jayway.jsonpath.JsonPath` and a spec for this and let you know when it's ready for a second look!

maggieneterval:
@duftler This is ready for another look. I will also open a spinnaker.github.io PR to document this function.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupport.java
```diff
@@ -367,6 +370,66 @@ static String judgment(Object obj, String id) {
     throw new IllegalArgumentException("An execution is required for this function");
   }
 
+  /**
+   * Gets value of given label key in manifest of given kind deployed by stage of given name
+   * @param obj #root.execution
+   * @param stageName the name of the stage to find
```
duftler: From reading the source, looks like this is "the name of a `deployManifest` stage to find", right?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupport.java
```diff
@@ -367,6 +370,66 @@ static String judgment(Object obj, String id) {
...
+    if (manifests == null || manifests.size() == 0) {
+      throw new SpelHelperFunctionException("No manifest could be found in the context of the specified stage");
+    }
+
+    Optional<Map> manifestOpt = manifests.stream()
+      .filter(m -> m.get("kind").equals(kind))
+      .findFirst();
+
+    if (!manifestOpt.isPresent()) {
+      throw new SpelHelperFunctionException("No manifest of specified kind could be found on the context of the specified stage");
```
duftler: Could include the specified `kind` in the error msg.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupport.java
```diff
@@ -367,6 +370,66 @@ static String judgment(Object obj, String id) {
...
+    }
+
+    Map manifest = manifestOpt.get();
+
```
duftler: Don't need the blank lines in between declarations here.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupportSpec.groovy
```diff
@@ -151,4 +178,34 @@ class ExpressionsSupportSpec extends Specification {
     then: "(deploy|createServerGroup|cloneServerGroup|rollingPush)"
     map.serverGroup == ["app-test-v001"]
   }
+
+  def "manifestLabelValue should resolve label value for manifest of given kind deployed by stage of given name"() {
+    expect:
+    ExpressionsSupport.manifestLabelValue(deployManifestPipeline, "Deploy ReplicaSet", "ReplicaSet", labelKey) == expectedLabelValue
+
+    where:
+    labelKey             | expectedLabelValue
```
duftler: Usually use `||` to separate given columns from expected columns.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupportSpec.groovy
```diff
@@ -151,4 +178,34 @@ class ExpressionsSupportSpec extends Specification {
     then: "(deploy|createServerGroup|cloneServerGroup|rollingPush)"
     map.serverGroup == ["app-test-v001"]
   }
+
+  def "manifestLabelValue should resolve label value for manifest of given kind deployed by stage of given name"() {
```
duftler: Probably want `@Unroll` here.

maggieneterval:
Thanks for the code review @duftler, just resolved all of your comments

PR #2697, opened 2019-03-05
fix(execution): a stage with redirect tasks gets marked terminal

claymccoy:
The logic for determining stage status seems overly strict.
The redirect execution status is used in a task loop. Without counting that as success how would something like RollingPushStage ever succeed? Or do the redirect statuses get modified somehow?
I could also make an argument that skipped tasks should not kick the stage out of opportunity to be a success.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
@@ -230,7 +230,7 @@ class CompleteStageHandler(
       allStatuses.contains(STOPPED)            -> STOPPED
       allStatuses.contains(CANCELED)           -> CANCELED
       allStatuses.contains(FAILED_CONTINUE)    -> FAILED_CONTINUE
-      allStatuses.all { it == SUCCEEDED }      -> SUCCEEDED
+      allStatuses.all { it == SUCCEEDED || it == REDIRECT }      -> SUCCEEDED
       afterStageStatuses.contains(NOT_STARTED) -> RUNNING // after stages were planned but not run yet
       else                                     -> {
         log.error("Unhandled condition for stage $id of $execution.id, marking as TERMINAL. syntheticStatuses=$syntheticStatuses, taskStatuses=$taskStatuses, planningStatus=$planningStatus, afterStageStatuses=$afterStageStatuses")


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -230,7 +230,7 @@ class CompleteStageHandler(
       allStatuses.contains(STOPPED)            -> STOPPED
       allStatuses.contains(CANCELED)           -> CANCELED
       allStatuses.contains(FAILED_CONTINUE)    -> FAILED_CONTINUE
-      allStatuses.all { it == SUCCEEDED }      -> SUCCEEDED
+      allStatuses.all { it == SUCCEEDED || it == REDIRECT }      -> SUCCEEDED
```
marchello2000: I am still wrapping my head around it - new to the `REDIRECT` status. But a few questions

Does this actually fix the problem?
1. It seems that the way it's used today `allStatuses` will never be all `REDIRECT`, am I missing something?
2. The `CheckForRemainingTerminationsTask` will eventually succeed: https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/rollingpush/CheckForRemainingTerminationsTask.groovy#L35

Do you have an example of a situation where this is an issue?

marchello2000: Never mind on 1. I misread the `all` closure

marchello2000: @robfletcher explained it to me a bit.
It seems that this change is not necessary because the redirect will be handled by `CompleteTaskHandler` (https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteTaskHandler.kt#L57)
So the `CompleteStageHandler` should never be called with any task status being a `REDIRECT`.
Do you see otherwise? Can you describe a situation/provide execution JSON where this is a problem?

ajordens:
Fair point on `SKIPPED` ... it does seem like a status that should be handled and not just go straight `TERMINAL`.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
claymccoy: You are right, I don't need this. The problem isn't the redirected task at the end of the loop, it is that I am also using the same task before the loop to set things up and that redirect never gets merged. I can make some changes to fix that. Thanks for pointing out the merging.

PR #2698, opened 2019-03-05
fix(expressions): Fix ConcurrentModificationException

marchello2000:
Fixes error introduced by me in https://github.com/spinnaker/orca/pull/2653

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
@@ -69,12 +69,19 @@ interface ExpressionAware {
     // context. Otherwise, it's very confusing in the UI because the value is clearly correctly evaluated but
     // the error is still shown
     if (hasFailedExpressions()) {
-      val failedExpressions = this.context[PipelineExpressionEvaluator.SUMMARY] as MutableMap<String, *>
+      try {
+        val failedExpressions = this.context[PipelineExpressionEvaluator.SUMMARY] as MutableMap<String, *>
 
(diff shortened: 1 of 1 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
```diff
@@ -69,12 +69,19 @@ interface ExpressionAware {
...
+        val failedExpressions = this.context[PipelineExpressionEvaluator.SUMMARY] as MutableMap<String, *>
 
-      failedExpressions.keys.forEach { expressionKey ->
-        if (evalSummary.wasAttempted(expressionKey) && !evalSummary.hasFailed(expressionKey)) {
+        val keysToRemove: List<String> = failedExpressions.keys.filter { expressionKey ->
+          (evalSummary.wasAttempted(expressionKey) && !evalSummary.hasFailed(expressionKey))
+        }.toList()
+
+        keysToRemove.forEach { expressionKey ->
           failedExpressions.remove(expressionKey)
```
marchello2000: this was dumb, but I did test it and I was surprised when it worked. So I left it as is. Trying to figure out if there are cases when it does work to satisfy my curiocity

emjburns:
LGTM

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
marchello2000: The answer is: if you modify the last element in the hashmap (as perceived by key iterator) all is honky dory because the `hasNext` method doesn't perform the `modCount` check. Makes sense I guess

PR #2701, opened 2019-03-05
refactor(core): Allow registering custom SpEL expression functions

robzienert:
Provides a strategy for extending Orca's SpEL expression language with new helper functions.
Additionally offers a more strongly typed and documented method of building these helpers, as
well as namespacing capabilities in case an organization wants to provide a canned group of
helper functions.

Moved `deployedServerGroups` and `manifestLabelValue` to use this new format.

Largely stolen from @ajordens prior art that was never merged.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionFunctionProvider.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionFunctionProvider.kt
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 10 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionFunctionProvider.kt
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.pipeline.expressions
+
+interface ExpressionFunctionProvider {
```
ajordens: kotlin in your src/main/java?

robzienert: That's what we've done with other `orca-core` kotlin code, yeah. Kinda like Java in our groovy src. :\

PR #2702, opened 2019-03-06
feat(artifacts): Add stage artifact resolver

jkschneider:
The new method is used to fully resolve a bound artifact on a stage that can either select an expected artifact ID for an expected artifact defined in a prior stage or as a trigger constraint OR define an inline expression-evaluable default artifact.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
@@ -37,14 +37,7 @@ import rx.schedulers.Schedulers;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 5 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -106,6 +102,26 @@ public ArtifactResolver(ObjectMapper objectMapper, ExecutionRepository execution
...
+   * Used to fully resolve a bound artifact on a stage that can either select
+   * an expected artifact ID for an expected artifact defined in a prior stage
+   * or as a trigger constraint OR define an inline expression-evaluable default artifact.
+   * @param stage The stage containing context to evaluate expressions on the bound artifact.
+   * @param id An expected artifact id. Either id or artifact must be specified.
+   * @param artifact An inline default artifact. Either id or artifact must be specified.
+   * @return A bound artifact with expressions evaluated.
+   */
+  public @Nullable Artifact getBoundArtifactForStage(Stage stage, @Nullable String id, @Nullable Artifact artifact) {
+    Artifact boundArtifact = id != null ? getBoundArtifactForId(stage, id) : artifact;
```
claymccoy: Oh, it necessary to evaluate the passed in Artifact too?

jkschneider: Yes.

jkschneider: Because the passed in Artifact will come from an inline default artifact which may have expressions:

image

PR #2705, opened 2019-03-06
fix(artifacts): handle bound artifact account missing

claymccoy:
the bound artifact should use the matched artifact account if it doesn’t have one


ajordens:
I'm an artifacts noob but code-wise, LGTM.

PR #2708, opened 2019-03-06
fix(kubernetes): Don't fail deploy for SpEL errors in manifest artifacts

willgorman:
Fixes spinnaker/spinnaker#3644
This will still display a warning to in the pipeline UI, just as when an evaluation error happens in a manifest in the stage as text. It allows the successful deployment of manifests that contain "expressions" that are intended to evaluated by the pod and not the pipeline (e.g. configmaps with Spring config, bash scripts with `${}` style environment variables, etc )

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java
@@ -125,10 +125,6 @@ public class DeployManifestTask extends AbstractCloudProviderAwareTask implement
               true
           );
 
-          if (manifestWrapper.containsKey("expressionEvaluationSummary")) {
-            throw new IllegalStateException("Failure evaluating manifest expressions: " + manifestWrapper.get("expressionEvaluationSummary"));
-          }
-
           return manifestWrapper.get("manifests");
         } catch (Exception e) {
           log.warn("Failure fetching/parsing manifests from {}", manifestArtifact, e);


maggieneterval:
Hey @willgorman -- the upcoming fix will most likely add an option to the Deploy Manifest stage to skip SpEL expression evaluation. Will this cover your use cases, or are there situations in which some of your expressions are meant for Spinnaker to evaluate while others are not, in which case you just don't want to fail the stage if the non-Spinnaker expressions fail?

willgorman:
We do have some actual pipeline expressions intended for Spinnaker to evaluate in the manifests as well, so disabling expressions entirely for the stage isn't ideal in our case.

maggieneterval:
Got it, thanks for the clarification @willgorman. Is there any reason you can't [escape the expressions not meant for Spinnaker](https://github.com/spinnaker/spinnaker/issues/2809#issuecomment-395510438)? For example:
`${ENV_VAR}` --> `${"${ENV_VAR}"}`

willgorman:
@maggieneterval I believe that I've seen those same errors  even when escaping in that way, but I'll run a few tests to confirm.

willgorman:
Ok, looks like I was mistaken or maybe thinking about behavior on a previous version.  Testing on 1.12.2 without this patch shows that when the expression isn't escaped it produces an error that fails the stage but when escaping it just displays the warning.  In our case we're deploying helm charts that we've created so adding escaping is a reasonable solution.

ezimanyi:
You may have run into #3028, which was around for a bit, but I believe I fixed in spinnaker/orca#2341

astlock:
AFAIU it's not fixed. I still can't install charts from helm stable. v1.12.5

ezimanyi:
Based on the above discussion that escaping will work for this use case, and work we discussed at the Kubernetes SIG today to make evaulating SpEL optional in the deploy manifest stage, closing this in favor of the other solutions.

PR #2710, opened 2019-03-07
feat(MPTv2): Support artifacts when executing v2 templated pipelines.

jtk54:


jtk54:

jtk54:
Manually porting.

ezimanyi:
@jtk54 : This is breaking regular non-templated pipelines with artifacts on both master and 1.12 now as we're now calling `resolveArtifacts` twice on the same pipeline, and it's not idempotent so the second call is failing for all cases other than the v2 templated case.

We could potentially make it idempotent, or somehow gate the second call to `resolveArtifacts` to only be for v2 templated pipelines---lets chat more Monday.  But for now I think I need to roll this back as all artifact workflow are completely broken and I need to get out a 1.12 patch release (and also don't want to leave master this broken).

jtk54:
Got it, thanks for looking into this.

PR #2712, opened 2019-03-08
feat(core): add save pipelines stage

claymccoy:
This stage will be used to extract pipelines from an artifact and save them. Spinnaker is a tool for deploying code, so when we treat pipelines as code it makes sense to use Spinnaker to deploy them.
Imagine you have your pipelines in a GitHub repo, and on each build you create an artifact that describes your pipelines. This CircleCI build is an example:
https://circleci.com/gh/claymccoy/canal_example/14#artifacts/containers/0

You can now create a pipleine that is triggered by that build, grab the artifact produced, and (with this new stage) extract the pipelines and save them.
It performs an upsert, where it looks for existing pipelines by app and name and uses the id to update in that case.

The format of the artifact is a JSON object where the top level keys are application names with values that are a list of pipelines for the app. The nested pipeline JSON is standard pipeline JSON. Here is an example:
https://14-171799544-gh.circle-artifacts.com/0/pipelines/pipelines.json

For now this simply upserts every pipeline in the artifact, but in the future it could allow you to specify a subset of apps and pipelines and effectively test and increase the scope of pipeline roll out. It (or a similar stage) could also save pipeline templates in the future as well.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/pipeline/SavePipelinesStage.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/pipeline/SavePipelinesStage.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2019 Pivotal, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.clouddriver.pipeline.pipeline;
+
+import com.netflix.spinnaker.orca.TaskResult;
+import com.netflix.spinnaker.orca.clouddriver.tasks.pipeline.CheckForRemainingPipelinesTask;
+import com.netflix.spinnaker.orca.clouddriver.tasks.pipeline.GetPipelinesFromArtifactTask;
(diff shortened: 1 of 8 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/pipeline/SavePipelinesStage.java
```diff
@@ -0,0 +1,59 @@
...
+import com.netflix.spinnaker.orca.front50.tasks.SavePipelineTask;
+import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder;
+import com.netflix.spinnaker.orca.pipeline.TaskNode.Builder;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SavePipelinesStage implements StageDefinitionBuilder {
```
ajordens: Would this be more aptly named `SavePipelinesFromArtifactStage`? 

I worry about confusion with the existing `SavePipelineStage`.

claymccoy: Yesk that makes sense.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/pipeline/GetPipelinesFromArtifactTask.java
```diff
@@ -0,0 +1,136 @@
...
+    return retrySupport.retry(() -> {
+      Response response = oort.fetchArtifact(artifact);
+      InputStream artifactInputStream;
+      try {
+        artifactInputStream = response.getBody().in();
+      } catch (IOException e) {
+        log.warn("Failure fetching pipelines from {}", artifact, e);
+        throw new IllegalStateException(e); // forces a retry
+      }
+      try (BufferedReader rd = new BufferedReader(new InputStreamReader(artifactInputStream))) {
```
jkschneider: Could do this all with `CharStreams.toString(artifactInputStream, Charsets.UTF_8)`.

PR #2713, opened 2019-03-08
fix(imagetagging): retry for missing namedImages

asher:
`ImageTagger` already retries fetching images by id, but not when subsequently fetching by image name. This poses a race condition for `UpsertImageTagsStage` due to the fact that clouddriver's `ImageCachingAgent` stores image metadata and image -> name mappings as distinct objects which are not updated atomically, at least in the case of the AWS provider.

If `UpsertImageTagsStage` executes in the narrow window where a newly baked image's `image` object has been stored but `imageName` has not, it will go terminal without retries. This PR enables retries for the lookup by name when that name was derived from a successful lookup by id. When `ImageTagger.findImages()` is called with image names, the existing behavior is preserved and the first lookup failure remains terminal.


ajordens:
LGTM

PR #2714, opened 2019-03-08
refactor(cf): Adopt artifacts model for CF deployments

jkschneider:
image


orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryServerGroupCreatorTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class CloudFoundryServerGroupCreatorTest {
+  @Test
+  void generateCloudFoundryManifestFromDirectInput() throws IOException {
+    String manifestPipelineJson = "{\n" +
```
claymccoy: These multiline Strings would look better in Kotlin or Groovy tests.

PR #2715, opened 2019-03-08
feat(core): add save pipelines stage

claymccoy:
This stage will be used to extract pipelines from an artifact and save them. Spinnaker is a tool for deploying code, so when we treat pipelines as code it makes sense to use Spinnaker to deploy them.
Imagine you have your pipelines in a GitHub repo, and on each build you create an artifact that describes your pipelines. This CircleCI build is an example:
https://circleci.com/gh/claymccoy/canal_example/14#artifacts/containers/0

You can now create a pipleine that is triggered by that build, grab the artifact produced, and (with this new stage) extract the pipelines and save them.
It performs an upsert, where it looks for existing pipelines by app and name and uses the id to update in that case.

The format of the artifact is a JSON object where the top level keys are application names with values that are a list of pipelines for the app. The nested pipeline JSON is standard pipeline JSON. Here is an example:
https://14-171799544-gh.circle-artifacts.com/0/pipelines/pipelines.json

For now this simply upserts every pipeline in the artifact, but in the future it could allow you to specify a subset of apps and pipelines and effectively test and increase the scope of pipeline roll out. It (or a similar stage) could also save pipeline templates in the future as well.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/pipeline/SavePipelinesFromArtifactStage.java
```diff
@@ -0,0 +1,59 @@
...
+      .withTask("savePipelinesCompleteTask", SavePipelinesCompleteTask.class);
+  }
+
+  @Component
+  public static class SavePipelinesCompleteTask implements com.netflix.spinnaker.orca.Task {
+
+    private final Logger log = LoggerFactory.getLogger(getClass());
+
+    @Override public TaskResult execute(Stage stage) {
+      log.info("Save Pipelines completed");
```
dreynaud: would it be possible to log the pipeline ids or names that have been saved? It would make that log line more useful

jtk54:
What value is being delivered by having the pipeline save happen inside Spinnaker? What sorts of workflows are we envisioning this feature would be used for? More importantly, what are those workflows attempting to achieve?

At an extremely high level, the value Spinnaker delivers to the artifacts/code it deploys is safety via version accounting and rollout+rollback strategies. None of that is supported in the management of Spinnaker pipelines themselves, so I'm skeptical about treating pipelines as something to deploy. My read on the goal of this is mainly to be able to express pipelines outside of Spinnaker. We already support this through `spin` in CI systems, and we should be very deliberate about where the line between CI and CD lies. If the goal is to define pipelines outside Spinnaker _and_ have them automatically synced into Spinnaker, we should have a larger design and planning discussion on how best to support that.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/pipeline/SavePipelinesFromArtifactStage.java
claymccoy: absolutely

claymccoy:
@jtk54 
Yes, the goal is indeed to define pipelines outside Spinnaker and have them automatically synced into Spinnaker. Let's discuss further.
We like treating pipelines as code and circumventing the UI. There are several efforts to do this. In most cases, something representing a pipeline is put into version control and when a commit happens, changes to the pipeline are pushed into Spinnaker. This syncing is usually done with some additional service like Dinghy or at the tail end of the build with the CLI. The pipeline syncing services that I have seen aren't useful to the community and seem unnecessary.  Having the CI deploy pipelines is what blurs the line of CI/CD. Having the CI create an artifact and Spinnaker deploy that artifact makes for a good CI/CD separation.
If we are treating pipelines as code, then this does seem like a deployment. And since Spinnaker specializes in deploying code, why not make a minor addition and allow Spinnaker to deploy its own pipelines as code?
It is true that there aren't rollout+rollback strategies for pipelines yet, and I think these would be great to have. Spinnaker can have the same version accounting and ability to trace a pipeline all the way back to a code commit if the trigger was from CI. There are a lot of interesting things we can do along those lines in the future. For now, we are merely gluing together some existing code by pulling pipelines out of artifacts and using the existing save pipeline stage. Spinnaker seems like the perfect tool for the job of deploying pipelines as code and was almost able to do it without this PR. Perhaps you'd be more comfortable if we feature flag the new stage in the UI as it evolves?

e96wic:
@claymccoy Do you have an example for a pipeline setup that uses your new stage? I tried to get this running with a github trigger but clouddriver fails to download the pipeline from the artifact. I think this pipeline is missing the option to specify an artifact account.

e96wic:
Like this: https://github.com/spinnaker/orca/blob/3fb0555d593c4a7e75dccf0330f404d12ed08b97/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java#L105

neelimamanubolu:
@claymccoy , I am facing the same issue , I too tried to get this running with a github trigger but clouddriver fails to download the pipeline from the artifact. Please find below the additional details : 

java.lang.IllegalArgumentException: An artifact account must be supplied to download this artifact: null
	at com.netflix.spinnaker.clouddriver.artifacts.ArtifactCredentialsRepository.getCredentials(ArtifactCredentialsRepository.java:58) ~[clouddriver-artifacts.jar:na]
	at com.netflix.spinnaker.clouddriver.artifacts.ArtifactDownloader.download(ArtifactDownloader.java:33) ~[clouddriver-artifacts.jar:na]
	at com.netflix.spinnaker.clouddriver.controllers.ArtifactController.lambda$fetch$0(ArtifactController.java:65) ~[main/:na]
	at org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBodyReturnValueHandler$StreamingResponseBodyTask.call(StreamingResponseBodyReturnValueHandler.java:110) ~[spring-webmvc-5.1.6.RELEASE.jar:5.1.6.RELEASE]
	at org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBodyReturnValueHandler$StreamingResponseBodyTask.call(StreamingResponseBodyReturnValueHandler.java:97) ~[spring-webmvc-5.1.6.RELEASE.jar:5.1.6.RELEASE]
	at org.springframework.web.context.request.async.WebAsyncManager.lambda$startCallableProcessing$4(WebAsyncManager.java:323) ~[spring-web-5.1.6.RELEASE.jar:5.1.6.RELEASE]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[na:1.8.0_212]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) ~[na:1.8.0_212]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) ~[na:1.8.0_212]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) ~[na:1.8.0_212]
	at java.lang.Thread.run(Thread.java:748) ~[na:1.8.0_212]

e96wic:
@neelimamanubolu have you tried this? https://github.com/spinnaker/spinnaker/issues/4224#issuecomment-480487905

neelimamanubolu:
@e96wic , I tried that.. Please check above comment..
For your reference::
PIPELINE JSON :: Removed "matchArtifact" from JSON object and added "artifactAccount".
Please check and revert back.

{
"artifactAccount": {
"id": "f8590624",
"name": "helmtest/manifest.yml",
"type": "github/file"
},

PR #2717, opened 2019-03-08
chore(jinja): Upgrade jinjava

ezimanyi:
The ModuleTag class breaks with jinjava >= 2.2.8 because of a fix in failsOnUnknownTokens.  We need to catch this error in cases where we expect that a token may be unknown. The test also uses True as a literal which should be true (ie, lowercase) which worked before but breaks with the upgrade.


ezimanyi:
Came across this because the autobump #/2716 is causing the `spinnaker-dependencies` version of jinjava to end up as the resolved library on the classpath. The version bump tests will pass after rebasing on this.

PR #2718, opened 2019-03-08
feat(spel): New `currentStage()` function

ajordens:
This fn provides a direct reference to the currently executing `Stage`.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.java
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.orca.pipeline.expressions;
 
 import com.netflix.spinnaker.orca.ExecutionStatus;
+import com.netflix.spinnaker.orca.pipeline.model.Execution;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.expression.EvaluationContext;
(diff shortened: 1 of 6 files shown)


ajordens:
Couple tests forthcoming but wanted to get the PR going before laptop went from 7% -> 0%.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.java
```diff
@@ -266,7 +272,18 @@ static String escapeSimpleExpression(String expression) {
    */
   private String includeExecutionParameter(String e) {
     String expression = e;
-    for (String fn : EXECUTION_AWARE_FUNCTIONS) {
+
+    // An expression aware function is any that takes an Execution as its first parameter
+    Set<String> expressionAwareFunctions = new HashSet<>(EXECUTION_AWARE_FUNCTIONS);
```
ajordens: Wouldn't mind splitting the other stage-related functions out of `ExpressionSupport` and out of the hardcoded `EXECUTION_AWARE_FUNCTIONS` constant ... was thinking that would be a subsequent PR (I don't _think_ anything should break!).

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/StageExpressionFunctionProvider.java
```diff
@@ -0,0 +1,69 @@
...
+            )
+        ))
+    );
+  }
+
+  /**
+   * @param execution the current execution
+   * @return the currently executing stage
+   */
+  public static Stage currentStage(Execution execution) {
```
ajordens: I'm not sure the history of the other functions taking an `Object` vs `Execution` explicitly ... this worked in my testing.

robzienert: Cool. I wasn't sure what the reason was for not using `Execution` either.

PR #2723, opened 2019-03-11
feature(ci): Fetch artifacts from CI builds

ezimanyi:
Closes spinnaker/spinnaker#3675

* refactor(ci): Convert BuildService and IgorService to Java

  Both of these files are trivial to convert to Java. Also, in `IgorService`, replace deprecated 
  `@EncodedPath` with the equivalent `@Path(encode = false)`

* refactor(ci): Clean up inheritance of CI stages

  It's confusing that Travis and Wercker stages extend Jenkins stages; make an abstract `CIStage` that these all extend. Also convert these stages to Java, and clean up some of the string interpolation that was hard to read.

* feature(ci): Fetch artifacts from CI builds

  Jenkins triggers support inflating artifacts from the build results based on a Jinja template specified in 
the build's properties. Add the same functionality to Jenkins stages.

  Use the property file defined in the CI stage to extract artifacts from the CI stage build.

* refactor(ci): Pull getProperties into its own task

  The prior commit added general support for retrying tasks that communicate with Igor; update `getProperties` to be its own task and have it use that new support.


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/CIStage.java
```diff
@@ -0,0 +1,100 @@
...
+  protected Class<? extends Task> waitForJobStartTaskClass() {
+    return MonitorQueuedJenkinsJobTask.class;
+  }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public void prepareStageForRestart(@Nonnull Stage stage) {
+    Object buildInfo = stage.getContext().get("buildInfo");
+    if (buildInfo != null) {
+      Map<String, Object> restartDetails =(Map<String, Object>) stage.getContext()
```
jtk54: nit: space after `=`.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/CIJobRequest.java
```diff
@@ -0,0 +1,45 @@
...
+@Getter
+public class CIJobRequest {
+  private final String master;
+  private final String job;
+  private final String propertyFile;
+  private final Integer buildNumber;
+  private final int consecutiveErrors;
+
+  public CIJobRequest(
+    @JsonProperty("master") String master,
```
jtk54: Is there a reason we have to annotate every param here? Wouldn't `@Data` and `@AllArgsConstructor` work?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/RetryableIgorTask.java
```diff
@@ -0,0 +1,82 @@
...
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+@RequiredArgsConstructor
+@Slf4j
+public abstract class RetryableIgorTask implements RetryableTask {
+  public long getBackoffPeriod() {
+    return TimeUnit.SECONDS.toMillis(2);
```
jtk54: Are these worth configuring? How do we handle Clouddriver tasks?

orca-igor/src/test/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildArtifactsTaskSpec.groovy
```diff
@@ -0,0 +1,98 @@
...
+    TaskResult result = task.execute(stage)
+    def artifacts = result.getOutputs().get("artifacts") as List<Artifact>
+
+    then:
+    1 * buildService.getArtifacts(BUILD_NUMBER, PROPERTY_FILE, MASTER, JOB) >> [testArtifact]
+    artifacts.size() == 1
+    artifacts.get(0).getName() == "my-artifact"
+  }
+
+  def "handles an emtpy artifact list"() {
```
jtk54: s/emtpy/empty

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/CIJobRequest.java
ezimanyi: I could not get any combination of Lombok annotations to get me an immutable class that would properly deserialize with Jackson.  The release notes for 1.16.22 (the next one after ours) mention better compatibility with Jackson, but I didn't want to expand the scope here to upgrading Lombok.

I guess I could have made it non-immutable so it could use the no-args constructor and setters (which is what I think mostly happens in our code)...but given that it never needs to be mutated that didn't feel like a great solution.

I'll at least add a comment as to why I did this.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/RetryableIgorTask.java
ezimanyi: Prior to this, it was just hard-coded in `MonitorJenkinsJobTask` (though I realize I slightly shortened the retry period, which I'll fix since I have no good reason to shorten it).

I don't think there's a lot of value in allowing this to be configurable...if someone had a good use case for why it should be different (and we shouldn't just change it for everyone) then I'd consider it, but absent that I don't think it's worth adding more config fields.

jtk54: Sounds good.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/CIJobRequest.java
jtk54: :+1: to comment.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/IgorService.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+import retrofit.http.*;
+
+import java.util.List;
+import java.util.Map;
+
+public interface IgorService {
+  @PUT("/masters/{name}/jobs/{jobName}")
+  public abstract String build(@Path("name") String master, @Path(encode = false, value = "jobName") String jobName, @QueryMap Map<String, String> queryParams, @Body String ignored);
```
emjburns: Line length for all these? Please split each arg onto its own line :)

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/CIStage.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import javax.annotation.Nonnull;
+import java.util.HashMap;
+import java.util.Map;
+
+@RequiredArgsConstructor
+@Slf4j
+public abstract class CIStage implements StageDefinitionBuilder, CancellableStage {
+  private final StopJenkinsJobTask stopJenkinsJobTask;
```
emjburns: Does this get initialized okay?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
```diff
@@ -0,0 +1,53 @@
...
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class GetBuildPropertiesTask extends RetryableIgorTask {
+  private final BuildService buildService;
```
emjburns: Does `@RequiredArgsConstructor` also do the dependency injection automatically?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/RetryableIgorTask.java
```diff
@@ -0,0 +1,82 @@
...
+      .put("consecutiveErrors", 0)
+      .build();
+    return new TaskResult(result.getStatus(), newContext, result.getOutputs());
+  }
+
+  private Map<String, Integer> errorContext(int errors) {
+    return Collections.singletonMap("consecutiveErrors", errors);
+  }
+
+  private boolean isRetryable(int statusCode) {
```
emjburns: I think we might want to retry once or twice on these two error codes. Maybe the error is ephemeral and will fix itself?

orca-igor/src/test/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildArtifactsTaskSpec.groovy
```diff
@@ -0,0 +1,98 @@
...
+    when:
+    TaskResult result = task.execute(stage)
+    def artifacts = result.getOutputs().get("artifacts") as List<Artifact>
+
+    then:
+    1 * buildService.getArtifacts(BUILD_NUMBER, PROPERTY_FILE, MASTER, JOB) >> []
+    artifacts.size() == 0
+  }
+
+  def "does not fetch artifacts if the property file is empty"() {
```
emjburns: Why should artifacts not be fetched if the property file is empty?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/IgorService.java
ezimanyi: Good point, will fix.  Also just realized that the `public abstract` part is redundant on an interface so will remove.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
ezimanyi: It doesn't directly, but if there's only a single constructor the Spring will automatically use it for autowiring.  (There is a way to ask lombok to annotate the constructor with the `onconstructor` argument, but I've never had to use it---I guess that would be necessary if someone added another constructor.)

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/CIStage.java
ezimanyi: Yup, responded to the comment below

orca-igor/src/test/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildArtifactsTaskSpec.groovy
ezimanyi: The way to specify the Jinja template to use to inflate artifacts is by passing a property in the property file, so there's nothing to do if we don't have a template.

It might make sense at some point to allow this to be configured more statically (maybe every job from the jenkins server uses a given template, or maybe the trigger defines it) but for now it's defined via an output property.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/RetryableIgorTask.java
ezimanyi: The two ones defined here (500, 503) are the ones where we do retry; any others (404, etc.) will not be retried.

I got the list from [here](https://github.com/spinnaker/orca/blob/f933bb73f90bde6b04c1e7b2cf2a0fbccdc69bb9/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/MonitorJenkinsJobTask.groovy#L103) though I dropped 404 as I think that was only in the list to handle cases where the build itself hadn't appeared yet.  (Looks like there were a few fixes to that file specifically to stop continually retrying fetching properties on a 404.)

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
emjburns: Oh gosh. I would love to make this kind of stuff explicit in the future - creating the constructor and adding the `@Autowired` annotation to it. It seems like an unnecessary mystery that doesn't buy us that much but could make it hard to spot bugs/changes in behavior. Especially as we're upgrading spring.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/RetryableIgorTask.java
emjburns: Ha! Great.

orca-igor/src/test/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildArtifactsTaskSpec.groovy
emjburns: Ahhh, I was missing that piece. Thanks for the explanation. So the whole template is in the property file, or it contains a reference to it?

ezimanyi: It's just a reference to it; you can either specify a built-in template by name (currently jar or jinja) or a custom template that you've configured in your `igor.yml`:
```
artifacts:
  templates:
  - name: <name of template>
    templatePath: <path to the template>
```
So basically the job itself says "this is how to extract an artifact", though I could also at some point in the future see the benfit of Spinnaker just knowing "this job always needs to get artifacts via this template".

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
ezimanyi: Fair enough, I sometimes try too hard to avoid writing a few extra lines :)

I think it should should be robust to Spring upgrades as it's  explicitly documented by Spring in both [1.5](https://docs.spring.io/spring-boot/docs/1.5.19.RELEASE/reference/htmlsingle/#using-boot-spring-beans-and-dependency-injection) and [2](https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-spring-beans-and-dependency-injection.html) that a single constructor automatically gets autowired.  But it is true that it adds extra magic that a reader needs to understand (and maybe we should avoid interactions between annotations like this).

PR #2725, opened 2019-03-11
fix(MPTv2): Supports artifact resolution for v2 MPTs.

jtk54:


jtk54:

PR #2733, opened 2019-03-12
fix(cloudformation): template should be a string

ethanfrogers:
previously cloudformation assumed that the template text would only be
JSON so it attempted to deserialize the artifact to a map.
cloudformation supports YAML or JSON templates so we should just pass
the template as a string all the way to the cloudformation tasks and let
any failures come from the aws sdk.


ethanfrogers:
@xavileon PATL!

ethanfrogers:
@xavileon closing as i plan on taking a different approach.

PR #2738, opened 2019-03-12
feat(expressions): adding #pipeline function

marchello2000:
Added #pipeline function for usage in SpEL expressions.
It returns the ID of the pipeline given the name of the pipeline
(within the same app only)

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.front50.pipeline;
+
+import com.netflix.servo.util.Strings;
+import com.netflix.spinnaker.kork.core.RetrySupport;
+import com.netflix.spinnaker.orca.front50.Front50Service;
+import com.netflix.spinnaker.orca.pipeline.expressions.ExpressionFunctionProvider;
+import com.netflix.spinnaker.orca.pipeline.expressions.SpelHelperFunctionException;
+import com.netflix.spinnaker.orca.pipeline.model.Execution;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+
+import static java.lang.String.format;
+
+@Component
+public class PipelineExpressionFunctionProvider implements ExpressionFunctionProvider {
+  // Static because it's needed during expression eval (which is a static)
+  private static Front50Service front50Service = null;
+
+  @Autowired
+  PipelineExpressionFunctionProvider(Optional<Front50Service> front50Service) {
+    if (front50Service.isPresent()) {
+      PipelineExpressionFunctionProvider.front50Service = front50Service.get();
+    }
+  }
+
+  @Nullable
+  @Override
+  public String getNamespace() {
+    return null;
+  }
+
+  @NotNull
+  @Override
+  public Collection<FunctionDefinition> getFunctions() {
+    return Arrays.asList(
+      new FunctionDefinition("pipelineId", Arrays.asList(
+        new FunctionParameter(
+          Execution.class, "execution", "The execution containing the currently executing stage"
+        ),
+        new FunctionParameter(
+          String.class, "pipelineName", "A valid stage reference identifier"
+        )
+      ))
+    );
+  }
+
+  /**
+   * Function to convert pipeline name to pipeline ID (within current application)
+   *
+   * @param execution    the current execution
+   * @param pipelineName name of the pipeline to lookup
+   * @return the id of the pipeline or null if pipeline not found
+   */
+  public static String pipelineId(Execution execution, String pipelineName) {
+    if (Strings.isNullOrEmpty(pipelineName)) {
+      throw new SpelHelperFunctionException("pipelineName must be specified for function #pipelineId");
+    }
+
+    if (front50Service == null) {
+      throw new SpelHelperFunctionException("front50 service is missing. It's required when using #pipelineId function");
+    }
+
+    try {
+      String currentApplication = execution.getApplication();
+
+      RetrySupport retrySupport = new RetrySupport();
+      Map<String, Object> pipeline = retrySupport.retry(() -> front50Service.getPipelines(currentApplication)
+          .stream()
+          .filter(p -> pipelineName.equals(p.getOrDefault("name", null)))
+          .findFirst()
+          .orElse(null),
+        3, 1000, true);
+
+      if (pipeline == null) {
+        throw new SpelHelperFunctionException(
+          format(
+            "Pipeline with name '%s' could not be found on application %s",
+            pipelineName,
+            currentApplication
+          ));
+      }
+
+      return (String) pipeline.get("id");
+    }
+    catch (SpelHelperFunctionException e) {
+      throw e;
+    }
+    catch (Exception e) {
+      throw new SpelHelperFunctionException("Failed to evaluate #pipelineId function", e);
+    }
+  }
+}


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -0,0 +1,97 @@
...
+   * @param execution the current execution
+   * @param pipelineName name of the pipeline to lookup
+   * @return the id of the pipeline or null if pipeline not found
+   */
+  public static String pipeline(Execution execution, String pipelineName) {
+    if ((pipelineName == null) || (pipelineName.isEmpty())) {
+      throw new SpelHelperFunctionException("pipelineName must be specified for function #pipeline");
+    }
+
+    if (front50Service == null) {
```
ajordens: Not sure this check is really worth it ... the `front50Service` will never really be null?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -0,0 +1,97 @@
...
+   * reference here
+   */
+  private static Front50Service front50Service;
+
+  @Autowired
+  PipelineExpressionFunctionProvider(Front50Service front50Service) {
+    PipelineExpressionFunctionProvider.front50Service = front50Service;
+  }
+
+
```
ajordens: Empty space

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -0,0 +1,97 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Map;
+
+@Component
+public class PipelineExpressionFunctionProvider implements ExpressionFunctionProvider {
+  /** Since expression evaluation happens in a static function we can't get access to front50 there
```
ajordens: Not sure this javadoc adds a lot of value ...

Do you mind merging the contents of `StageExpressionFunctionProvider` in here, no sense in having multiple providers providing similarly purposed functions.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -0,0 +1,97 @@
...
+      .stream()
+      .filter(p -> pipelineName.equals(p.getOrDefault("name", null)))
+      .findFirst()
+      .orElse(null);
+
+    if (pipeline != null) {
+      return (String)pipeline.getOrDefault("id", null);
+    }
+
+    return null;
```
ajordens: Perhaps worth throwing an exception that the pipeline couldn't be found rather than returning null?

ajordens: If you went the exception route ... `orElseThrow()` and then you can avoid the `pipeline != null` check.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -0,0 +1,97 @@
...
+  @Override
+  public String getNamespace() {
+    return null;
+  }
+
+  @NotNull
+  @Override
+  public Collection<FunctionDefinition> getFunctions() {
+    return Arrays.asList(
+      new FunctionDefinition("pipeline", Arrays.asList(
```
ajordens: Thoughts on the function name being something more specific and indicative of what you'll get back ala `pipelineId()`?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -0,0 +1,97 @@
...
+      throw new SpelHelperFunctionException("pipelineName must be specified for function #pipeline");
+    }
+
+    if (front50Service == null) {
+      throw new SpelHelperFunctionException("front50 service is missing. It's required when using #pipeline function");
+    }
+
+    String currentApplication = execution.getApplication();
+
+    Map<String, Object> pipeline = front50Service.getPipelines(currentApplication)
```
ajordens: It might help to think about retries here (or alternative ways to be less susceptible to a network failure blowing up the pipeline execution).

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
robzienert: I think it might be in a halyard case?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * Function to convert pipeline name to pipeline ID (within current application)
+   *
+   * @param execution the current execution
+   * @param pipelineName name of the pipeline to lookup
+   * @return the id of the pipeline or null if pipeline not found
+   */
+  public static String pipeline(Execution execution, String pipelineName) {
+    if ((pipelineName == null) || (pipelineName.isEmpty())) {
```
robzienert: `Strings.isNullOrEmpty`

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: I was trying to mimic #stage... will update, i prefer `pipelineId` as well

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: yeah, i thought there are cases when it's `null` during setup or something?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: yeah, i like it, thanks! i think it makes debugging expressions failures easier

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: oh, it's `String*s*` thank you!! i was looking for something under `String` to no avail...

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: sure happy to merge. should I also merge the "naked" `#stage` function that's in `ExpressionSupport.java`?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
ajordens: :)

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
ajordens: Sorry, I believe that to be true _but_ the way the `@Autowired` is setup on this class we'll fail to create the bean if `Front50Service` were ever not available.

You'd need to have an `Optional<Front50Service>` in the constructor (or perhaps better a `@ConditionalOnBean(Front50Service)`)

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: only caveat is that it'd be a bit weird since this file in in the `orca-front50`... should i move it elsewhere?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: 👍 thanks forgot about that

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: good point! is there a cache inside orca of pipelines? 'cuz we also eval expressions ALOT and if you have this function we will make the front50 call a bunch of times. maybe not really a problem just thinking out loud

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
ajordens: Oh hah! ... I didn't actually pick up that this was in `orca-front50` so let's leave them separate for now. 

Good call on being potentially weird.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
ajordens: At a glance, I don't think there's a cache of pipelines in Front50. 

It _might_ not hurt to have a local background refreshed cache of application -> list of pipelines (name + id).

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: decided to go with `Optional<>` because `ConditionalOnBean` wasn't working since it appears that this bean is created before front50, not sure how that's possible since it clearly depends on it... Anyway, used `Optional<>` and left the exception, I think that's better debuggability wise anyway

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: 👍

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: I will a simple retry for now, we can talk about the cache for another PR

marchello2000:
@robzienert / @ajordens ready for another round of review

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -0,0 +1,111 @@
...
+      Map<String, Object> pipeline = retrySupport.retry(() -> front50Service.getPipelines(currentApplication)
+          .stream()
+          .filter(p -> pipelineName.equals(p.getOrDefault("name", null)))
+          .findFirst()
+          .orElse(null),
+        3, 1000, true);
+
+      if (pipeline == null) {
+        throw new SpelHelperFunctionException(
+          String.format("Pipeline with name '%s' could not be found on application %s",
```
ajordens: My general preference for this would be something like ...

`import static ...String.format`

and then

```
format(
  "Pipeline with name ...",
  pipelineName,
  currentApplication
)
```

(specifically the line breaks and lining up braces)

marchello2000: gotcha, updating!

PR #2741, opened 2019-03-12
fix(mpt): temporarily pin back jinjava

asher:
jinjava >= 2.2.9 breaks some v1 pipeline evaluation due to a change in
unknown token behavior that needs to be handled.


ezimanyi:
@asher: I think this is actually going to need to be pinned, as a gradle was resolving to a newer `jinjava` in my testing, even when explicitly pulling in 2.2.3 because kork now has jinjava.

robzienert:
Yeah, you'll need to use `resolutionStrategy` with a `force = true` on the artifact, probably.

PR #2745, opened 2019-03-12
fix(evaluateVariables): enable EvaluateVariables stage to run in dryRun

marchello2000:
Variables are hard to get right, hence the EvaluateVariables stage, but
if it doesn't run in `dryRun` it makes quickly iterating on it harder.


marchello2000:
Thank you both! just rebased to get an up-to-date build

PR #2749, opened 2019-03-13
feat(core): Add support for a condition aware deploy preprocessor

jeyrschabu:
- adds ability to add a synthetic wait before a deploy stage
- pauses a deployment if certain conditions are not met
- provides visibility into currently unmet conditions

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline;
+
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.DeployStagePreProcessor;
+import com.netflix.spinnaker.orca.conditions.ConditionSupplier;
+import com.netflix.spinnaker.orca.kato.pipeline.support.StageData;
+import com.netflix.spinnaker.orca.pipeline.WaitForConditionStage;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.stereotype.Component;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Component
+@ConditionalOnExpression("${tasks.evaluateCondition.enabled:false}")
+@ConditionalOnBean(value = ConditionSupplier.class)
+public class ConditionAwareDeployStagePreprocessor implements DeployStagePreProcessor {
+  private final WaitForConditionStage waitForConditionStage;
+
+  @Autowired
+  public ConditionAwareDeployStagePreprocessor(
+    WaitForConditionStage waitForConditionStage
+  ) {
+    this.waitForConditionStage = waitForConditionStage;
+  }
+
+  @Override
+  public boolean supports(Stage stage) {
+    return true;
+  }
+
+  @Override
+  public List<StageDefinition> beforeStageDefinitions(Stage stage) {
+    final StageData stageData = stage.mapTo(StageData.class);
+    Map<String, Object> ctx = new HashMap<>();
+    ctx.put("region", stageData.getRegion());
+    ctx.put("cluster", stageData.getCluster());
+
+    StageDefinition stageDefinition = new StageDefinition();
+    stageDefinition.name = "Wait For Condition";
+    stageDefinition.context = ctx;
+    stageDefinition.stageDefinitionBuilder = waitForConditionStage;
+    return Collections.singletonList(stageDefinition);
+  }
+}
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/Condition.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/Condition.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.conditions;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+public class Condition {
+  private String name;
+  private String description;
+
+  @JsonCreator
+  public Condition(@JsonProperty("name") String name, @JsonProperty("description") String description) {
+    this.name = name;
+    this.description = description;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  public String getDescription() {
+    return description;
+  }
+
+  public void setDescription(String description) {
+    this.description = description;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    } if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    Condition that = (Condition) o;
+    return name.equals(that.name) && description.equals(that.description);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(name, description);
+  }
+}
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionConfigurationProperties.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionConfigurationProperties.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.conditions;
+
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.TimeUnit;
+
+@Component
+@ConfigurationProperties("tasks.evaluateCondition")
+public class ConditionConfigurationProperties {
+  private final DynamicConfigService configService;
+  private boolean enabled = false;
+  private Long backoffWaitMs = TimeUnit.MINUTES.toMillis(5);
+  private Long waitTimeoutMs = TimeUnit.MINUTES.toMillis(120);
+
+  public ConditionConfigurationProperties(DynamicConfigService configService) {
+    this.configService = configService;
+  }
+
+  public boolean isEnabled() {
+    return configService.getConfig(Boolean.class, "tasks.evaluateCondition", enabled);
+  }
+
+  public void setEnabled(boolean enabled) {
+    this.enabled = enabled;
+  }
+
+  public Long getBackoffWaitMs() {
+    return configService.getConfig(Long.class, "tasks.evaluateCondition.backoffWaitMs", backoffWaitMs);
+  }
+
+  public void setBackoffWaitMs(Long backoffWaitMs) {
+    this.backoffWaitMs = backoffWaitMs;
+  }
+
+  public long getWaitTimeoutMs() {
+    return configService.getConfig(Long.class, "tasks.evaluateCondition.waitTimeoutMs", waitTimeoutMs);
+  }
+
+  public void setWaitTimeoutMs(long waitTimeoutMs) {
+    this.waitTimeoutMs = waitTimeoutMs;
+  }
+}
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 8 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/Condition.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+package com.netflix.spinnaker.orca.conditions;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+public class Condition {
```
marchello2000: are we trying to go away from Lombok? I think it will generate all this boiler plate for you

orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/Condition.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public void setDescription(String description) {
+    this.description = description;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o)
+      return true;
```
marchello2000: I've debugged so many bugs where `{` was missing that I have an allergic reaction to it now

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @Override
+  public long getBackoffPeriod() {
+    return TimeUnit.MINUTES.toMillis(conditionsConfigurationProperties.getBackoffWaitMins());
+  }
+
+  @Override
+  public long getTimeout() {
+    return Integer.MAX_VALUE;
```
marchello2000: this seems dangerous...? I feel like we should have some reasonable default and then let the user override in the UI

orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/Condition.java
jeyrschabu: No Lombok.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
jeyrschabu: Curious why you think this is dangerous. It's assumed that the default stage timeout will kick in once reached.
Happy to change it to a much smaller value. (This value was inspired by the waitStage)

marchello2000: so maybe I am missing something, but i've been debugging something else related to timeouts and I have yet to find this mysterious default stage timeout :)
I am pretty sure pipelines generated by UI (without the manual "override timeout") will never have a default stage timeout (despite what deck makes you believe)
The waitStage has an an intrinsic timeout (the wait) so I feel like the `getTimeout` there is essentially the task timeout (though probably would be nicer to return the waitTime + some buffer to be safe).

Anyway, just seems that the timeouts are confusing and I think it's wrong for the task to never complete if the conditions aren't met, it should timeout eventually. If the there is a legit usecase to not timeout the user can override it in the pipeline definition

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
```diff
@@ -0,0 +1,105 @@
...
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    if (!conditionsConfigurationProperties.isEnabled() || !suppliers.isPresent()) {
+      log.warn("Conditions not configured. Skipping wait.");
+      stage.getContext().put("status", Status.SKIPPED);
+      return new TaskResult(ExecutionStatus.SUCCEEDED, stage.getContext());
+    }
+
+    final WaitForConditionContext waitForConditionContext = stage.mapTo(WaitForConditionContext.class);
+    if (waitForConditionContext.getStatus() == Status.SKIPPED) {
```
marchello2000: when does this `if` evaluate to `true`? Everywhere `status` is set to `Status.SKIPPED` the task completes with `SUCCEEDED` so this execute method shouldn't run again?
I mostly ask, because I am still trying to wrap my head around the execution flow in general. I feel like everytime I got it - I find something that i don't understand

jeyrschabu: `WaitForConditionContext.status` here is used as a marker. `status=SKIPPED` (when the user decides to skip the wait) would short-circuit the wait. Think of it as when a user decides to skip the remainder of `waitTime` in a wait stage. Happy to go over the general flow of tasks (retryable or not), stages, offline if you'd like.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
jeyrschabu: Ok, if stage timeouts aren't respected that has got to be a bug somewhere. In any case, I will update the timeout here to be shorter to handle the situation of a near infinite wait. I think 2 hours is reasonable default.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
```diff
@@ -0,0 +1,105 @@
...
+    Clock clock
+  ) {
+    this.conditionsConfigurationProperties = conditionsConfigurationProperties;
+    this.suppliers = suppliers;
+    this.clock = clock;
+  }
+
+  @Override
+  public long getBackoffPeriod() {
+    return TimeUnit.MINUTES.toMillis(conditionsConfigurationProperties.getBackoffWaitMins());
```
jeyrschabu: Thinking about this more: The backoff should be much smaller so to allow detecting a skipped execution.
I am going to update this so that the backoff is explicitly checked in the task execution instead.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
marchello2000: got it, that makes sense! thanks for the explanation
I assume the UI has a list of stages that can be skipped so that will need to be updated?

orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/Condition.java
emjburns: I don't actually know if we're trying to get away from lombok. I don't think we can ever actually get away from it, so I've actually started using it because it's less of a pain. 🤷‍♀️

emjburns:
LGTM

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+
+@Component
+@ConditionalOnBean(value = ConditionSupplier.class)
+public class ConditionAwareDeployStagePreprocessor implements DeployStagePreProcessor {
+  private static final Logger log = LoggerFactory.getLogger(ConditionAwareDeployStagePreprocessor.class);
+  private static final List<String> SUPPORTED_PROVIDERS = Arrays.asList("aws", "titus");
```
ajordens: Not sure we need to be restricting by provider, is there a reason?

orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionConfigurationProperties.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package com.netflix.spinnaker.orca.conditions;
+
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Component
+@ConfigurationProperties("conditions")
```
ajordens: Should perhaps follow the existing `tasks` pattern ...

```
tasks.evaluateCondition
```

orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionConfigurationProperties.java
```diff
@@ -0,0 +1,58 @@
...
+public class ConditionConfigurationProperties {
+  private Long backoffWaitSeconds = 60L;
```
ajordens: FWIW the existing task configs all specify things in `Ms` or `Millis`

orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionConfigurationProperties.java
```diff
@@ -0,0 +1,58 @@
...
+  private boolean enabled = false;
+  private final DynamicConfigService dynamicConfigService;
+  private Long waitTimeoutSeconds = 60 * 5L;
+
+  public ConditionConfigurationProperties(DynamicConfigService dynamicConfigService) {
+    this.dynamicConfigService = dynamicConfigService;
+  }
+
+  public boolean isEnabled() {
+    return dynamicConfigService.getConfig(Boolean.class, "conditions", enabled);
```
ajordens: `s/conditions/tasks.evaluateCondition`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
```diff
@@ -0,0 +1,72 @@
...
+
+    final StageData stageData = stage.mapTo(StageData.class);
+    return SUPPORTED_PROVIDERS.contains(stageData.getCloudProvider());
+  }
+
+  @Override
+  public List<StageDefinition> beforeStageDefinitions(Stage stage) {
+    StageDefinition stageDefinition = new StageDefinition();
+    stageDefinition.name = "Wait For Condition";
+    stageDefinition.context = new HashMap();
```
ajordens: No context to pass down?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Override
+  public long getTimeout() {
+    return TimeUnit.SECONDS.toMillis(conditionsConfigurationProperties.getWaitTimeoutSeconds());
+  }
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    if (!conditionsConfigurationProperties.isEnabled() || !suppliers.isPresent()) {
```
ajordens: We shouldn't even inject the conditions stage if it's disabled and/or there are no suppliers?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
```diff
@@ -0,0 +1,113 @@
...
+      final Status status = conditions.isEmpty() ? Status.SKIPPED : Status.WAITING;
+      tmp.put("status", status);
+      stage.getOutputs().put("conditions", conditions);
+      if (status == Status.WAITING) {
+        return new TaskResult(ExecutionStatus.RUNNING, tmp, stage.getOutputs());
+      }
+
+      return new TaskResult(ExecutionStatus.SUCCEEDED, tmp, stage.getOutputs());
+    } catch (Exception e) {
+      log.error("Error occurred while fetching for conditions to eval.", e);
```
ajordens: Don't think we need a log and exception thrown ... we'll get double logs.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
```diff
@@ -0,0 +1,113 @@
...
+  public long getTimeout() {
+    return TimeUnit.SECONDS.toMillis(conditionsConfigurationProperties.getWaitTimeoutSeconds());
+  }
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    if (!conditionsConfigurationProperties.isEnabled() || !suppliers.isPresent()) {
+      log.warn("Conditions not configured. Skipping wait.");
+      stage.getContext().put("status", Status.SKIPPED);
```
ajordens: Generally speaking we shouldn't directly be modifying the stage context, any additions should be returned in the `TaskResult`.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
```diff
@@ -0,0 +1,113 @@
...
+      return new TaskResult(ExecutionStatus.SUCCEEDED, stage.getContext());
+    }
+
+    final WaitForConditionContext ctx = stage.mapTo(WaitForConditionContext.class);
+    if (ctx.getStatus() == Status.SKIPPED) {
+      log.info("Skipping wait. Status: {}", ctx.getStatus());
+      return new TaskResult(ExecutionStatus.SUCCEEDED, stage.getContext(), stage.getOutputs());
```
ajordens: you don't need to be explicit about `stage.getOutputs()` and `stage.getContext()`.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    Instant now = clock.instant();
+    Duration backoff = Duration.ofSeconds(conditionsConfigurationProperties.getBackoffWaitSeconds());
+    Map<String, Object> tmp = new HashMap<>();
+    tmp.put("startTime", Optional.ofNullable(ctx.getStartTime()).orElse(now));
```
ajordens: Isn't the start time the start time of this stage or task ... do we need another?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
```diff
@@ -0,0 +1,113 @@
...
+      tmp.put("status", status);
+      stage.getOutputs().put("conditions", conditions);
+      if (status == Status.WAITING) {
+        return new TaskResult(ExecutionStatus.RUNNING, tmp, stage.getOutputs());
+      }
+
+      return new TaskResult(ExecutionStatus.SUCCEEDED, tmp, stage.getOutputs());
+    } catch (Exception e) {
+      log.error("Error occurred while fetching for conditions to eval.", e);
+      throw new IllegalStateException("Failure to retrieve conditions for evaluation.", e);
```
ajordens: Unclear if we want an error in a conditions stage to fail the pipeline?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
jeyrschabu: You know, not sure what additional info that will provide. Perhaps pass the context of the parent stage (whatever deploy stage).

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
jeyrschabu: What should the appropriate behavior be in your opinion? Should the pipeline not fail if for some reason a failure occurs while checking conditions?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
jeyrschabu: Good point. Can use the stage/task timeout.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
ajordens: Maybe some normalization of the cluster name and region it's being deployed into so that each supplier doesn't have to figure that out themselves?

I think of it as a way to describe the contract, the suppliers can say "if I need a region, it'll be defined under this key ... same with cluster".

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
jeyrschabu: meh, could do without asserting what provider is supported.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
jeyrschabu: Excellent point.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateConditionTask.java
ajordens: To start, I'd probably err on the side of it not blowing up the pipeline.

PR #2752, opened 2019-03-14
test(core): Remove front50 bean from startup test

ezimanyi:
When deploying to Kubernetes, halyard uses a bootstrap orca that doesn't have front50 enabled. This means that we need to keep orca starting without front50 around; set it to false in the test and remove the mock bean.


ezimanyi:
@emjburns : You (almost) predicted this [here](https://github.com/spinnaker/orca/pull/2685#issuecomment-468419016).  This test fails right now because deploying to kubernetes is broken by #2715, but should pass as soon as I get a fix in and rebase.

PR #2753, opened 2019-03-14
fix(core): Fix startup

ezimanyi:
Orca no longer starts up without front50 because a dependency on front50 was added to a task bean. Halyard needs an orca without front50 to bootstrap deploy to kubernetes.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
@@ -26,6 +26,7 @@ import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


ezimanyi:
Updated our startup test to catch this case in #2752

ezimanyi:
Going to merge this now to unbreak deploying to kubernetes

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
```diff
@@ -38,6 +39,7 @@
 import java.util.concurrent.TimeUnit;
 
 @Component
+@ConditionalOnProperty("front50.enabled")
```
asher: This was a breaking change, @ezimanyi can you update to default true?

ezimanyi: Really sorry about that, looking now...I was trying to fix a breaking change in https://github.com/spinnaker/orca/pull/2715 but obviously did not fix it properly.

PR #2754, opened 2019-03-14
fix(authz): Fix copying pipelines with managed service accounts

jervi:
Copying pipelines currently fails when managed service accounts are enabled. This commit fixes that by generating a pipeline id (UUID) in Orca before generating the managed service account.


jervi:
I've fixed overwriting managed service accounts and the cron triggers

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
```diff
@@ -89,6 +89,13 @@ public TaskResult execute(Stage stage) {
       updateServiceAccount(pipeline, serviceAccount);
     }
 
+    if (stage.getContext().get("pipeline.id") != null && pipeline.get("id") == null) {
+      pipeline.put("id", stage.getContext().get("pipeline.id"));
+
+      // We need to tell front50 to regenerate cron trigger id's
+      pipeline.put("regenerateCronTriggerIds", true);
```
jervi: https://github.com/spinnaker/front50/pull/488/files#diff-584c77ef2ccffac3251d9d570b65cba8R111

PR #2755, opened 2019-03-14
feat(pipelines): if saving multiple pipelines, continue on failure

claymccoy:
This is a minor change to SavePipelineTask. By default it will still fail with a TERMINAL status exactly as before. I’ve added a unit test for this original behavior. But if it detects that multiple pipelines are being saved, then the failure status will be FAILED_CONTINUE instead. This allows the SavePipelinesFromArtifactStage to attempt to save all the pipelines from the artifact and then give a summary of the results.


jtk54:
In the case where not all of the pipelines succeed, wouldn't we want to fail fast since the user will likely have to reinitiate the pipeline anyway for the ones that failed?

claymccoy:
This way I can give a summary of all the pipeline saves that failed, rather than forcing them to repeat the pipeline repeatedly to see the next problem. It is not time consuming to save a pipeline config to front50.

PR #2756, opened 2019-03-14
fix(clouddriver): Revert change to pin source capacity for redblack deploys

ajordens:


ajordens:
FYI @dreynaud 

Turns out it's caused a couple problems internally.

PR #2758, opened 2019-03-14
feat(core): Add support for Concourse triggers

jkschneider:
image

image


jkschneider:
This should have been merged last week and is needed for 1.13 as it collaborates with Concourse features in other Spinnaker services.

PR #2760, opened 2019-03-14
fix(ci): Fix cast error in CIStage

ezimanyi:
WaitForCompletion is a Boolean, but given that the prior groovy code accepted either a String or a Boolean, being defensive and handling both.

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/CIStage.java
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/CIStage.java
@@ -53,11 +53,14 @@ public abstract class CIStage implements StageDefinitionBuilder, CancellableStag
   }
 
   private boolean waitForCompletion(Stage stage) {
-    String waitForCompletion = (String) stage.getContext().get("waitForCompletion");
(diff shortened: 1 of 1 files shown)


ezimanyi:
Was doing some more testing while Travis ran, realized I had the default wrong; fixed it now.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/CIStage.java
```diff
@@ -53,11 +53,15 @@ public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
   }
 
   private boolean waitForCompletion(Stage stage) {
-    String waitForCompletion = (String) stage.getContext().get("waitForCompletion");
-    if (waitForCompletion != null && waitForCompletion.equals("false")) {
-      return false;
+    Object contextValue = stage.getContext().get("waitForCompletion");
```
ajordens: As an aside, you could avoid this with a `stage.mapTo()` and letting jackson handle proper type deserialization.

ezimanyi: Switched to `mapTo()` and added tests around this in #2768

PR #2761, opened 2019-03-14
feat(kubernetes): add expression evaluation options to bake and deplo…

maggieneterval:
…y manifest stages

Corresponding Deck PR: https://github.com/spinnaker/deck/pull/6696

The primary purpose of this change is to prevent Deploy Manifest stage failure for users attempting to deploy third-party Helm manifests that include `${}`-wrapped expressions, while allowing SpEL overrides to be explicitly evaluated just prior to baking. These options should probably be deprecated when expression evaluation is refactored in Spinnaker (see this [proposal](https://github.com/spinnaker/spinnaker/issues/3500)).

- Adds `evaluateOverrideExpressions` option to Bake Manifest stage. This defaults to false. When it is set to true, SpEL expressions evaluation of manifest overrides will occur during the Bake stage.
- Adds `skipExpressionEvaluation` option to Deploy Manifest stage for manifest artifacts. This defaults to false. When it is set to true, SpEL expression of the manifest artifact will not occur during the Deploy stage.
- Important caveat: Orca attempts to evaluate expressions in the entire pipeline context multiple times in logic that does not belong to these stages (for example, [here](https://github.com/spinnaker/orca/blob/master/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy#L167)). Therefore, there will still be "failed to evaluate" warnings for all `${}`-wrapped expressions not intended for Spinnaker.

Related issues:
- https://github.com/spinnaker/spinnaker/issues/2809
- https://github.com/spinnaker/spinnaker/issues/3028
- https://github.com/spinnaker/spinnaker/issues/3939
- https://github.com/spinnaker/spinnaker/issues/4109
- https://github.com/spinnaker/spinnaker/issues/3644
- https://github.com/spinnaker/spinnaker/issues/3335

Before:
2kX9K7zFvuJ

After:
2N55McgwoEf

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
@@ -17,7 +17,6 @@
 
 package com.netflix.spinnaker.orca.bakery.tasks.manifests;
 
-import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
 import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact;
@@ -28,6 +27,7 @@ import com.netflix.spinnaker.orca.bakery.api.BakeryService;
 import com.netflix.spinnaker.orca.bakery.api.manifests.helm.HelmBakeManifestRequest;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
 import com.netflix.spinnaker.orca.pipeline.util.ArtifactResolver;
+import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -62,12 +62,15 @@ public class CreateBakeManifestTask implements RetryableTask {
   @Autowired
   ObjectMapper objectMapper;
 
+  @Autowired
+  ContextParameterProcessor contextParameterProcessor;
+
   @Nonnull
   @Override
   public TaskResult execute(@Nonnull Stage stage) {
-    Map<String, Object> context = stage.getContext();
+    BakeManifestContext context = stage.mapTo(BakeManifestContext.class);
 
-    List<InputArtifactPair> inputArtifactsObj = objectMapper.convertValue(context.get("inputArtifacts"), new TypeReference<List<InputArtifactPair>>() {});
+    List<InputArtifactPair> inputArtifactsObj = context.getInputArtifacts();
     List<Artifact> inputArtifacts;
 
     if (inputArtifactsObj == null || inputArtifactsObj.isEmpty()) {
@@ -84,7 +87,7 @@ public class CreateBakeManifestTask implements RetryableTask {
           return a;
         }).collect(Collectors.toList());
 
-    List<ExpectedArtifact> expectedArtifacts = objectMapper.convertValue(context.get("expectedArtifacts"), new TypeReference<List<ExpectedArtifact>>() {});
+    List<ExpectedArtifact> expectedArtifacts = context.getExpectedArtifacts();
 
     if (expectedArtifacts == null || expectedArtifacts.isEmpty()) {
       throw new IllegalArgumentException("At least one expected artifact to baked manifest must be supplied");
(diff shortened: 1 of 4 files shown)


ethanfrogers:
@maggieneterval do we have a list of all the issues this would close out? it'd be useful to link them all here.

maggieneterval:
@ethanfrogers good call, just added a list of related issues!

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
```diff
@@ -96,11 +100,21 @@ public TaskResult execute(@Nonnull Stage stage) {
 
     String outputArtifactName = expectedArtifacts.get(0).getMatchArtifact().getName();
 
+    Map<String, Object> overrides = (objectMapper.convertValue(context.get("overrides"), new TypeReference<Map<String, Object>>() { }));
+    boolean evaluateOverrideExpressions = (boolean) context.get("evaluateOverrideExpressions");
```
ezimanyi: This will throw an NPE for stages that don't have this field present; should be a `Boolean` here, or perhaps better could use `stage.mapTo` with a defined model and have Jackson handle the mapping.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java
```diff
@@ -119,14 +119,17 @@ public TaskResult execute(@Nonnull Stage stage) {
           Map<String, Object> manifestWrapper = new HashMap<>();
           manifestWrapper.put("manifests", manifests);
 
-          manifestWrapper = contextParameterProcessor.process(
+          boolean evaluateManifestExpressions = !((boolean) task.get("skipExpressionEvaluation"));
```
ezimanyi: Same comment about NPE

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
ezimanyi: Here's a recent example of the `.mapTo` thing

ezimanyi: Would have been helpful if I added the link above: https://github.com/spinnaker/orca/blob/5e88d68f5bfc7588423f2afff8b3f896158269a3/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/RetryableIgorTask.java#L50

maggieneterval: Ah thank you, good catch! I will create models for both stage's contexts. I think after that I'll want to do a bit more testing of each stage just to make sure I didn't break anything -- do you know what time the release branch needs to be cut on Monday?

ezimanyi: Usually not until the afternoon, but I'm sure we can wait if this is almost ready.

maggieneterval: Perfect! Just pushed an update with the new context models. Let's hold off on merging this until I run through a bunch of test workflows when I'm back at my desktop Monday morning.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
```diff
@@ -96,12 +99,21 @@ public TaskResult execute(@Nonnull Stage stage) {
 
     String outputArtifactName = expectedArtifacts.get(0).getMatchArtifact().getName();
 
+    Map<String, Object> overrides = context.getOverrides();
+    if (context.getEvaluateOverrideExpressions()) {
```
ezimanyi: We'll still need a null check here, as `if(null)` is an NPE; alternately we could change the member variable on `BakeManifestContext` to a `boolean` and handle the null in the constructor.  (In this case, setting `false` if null, in the `Deploy` case setting `true`.)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java
```diff
@@ -119,14 +122,17 @@ public TaskResult execute(@Nonnull Stage stage) {
           Map<String, Object> manifestWrapper = new HashMap<>();
           manifestWrapper.put("manifests", manifests);
 
-          manifestWrapper = contextParameterProcessor.process(
+          Boolean skipExpressionEvaluation = context.getSkipExpressionEvaluation();
+          if (!skipExpressionEvaluation) {
```
ezimanyi: See above comment about null check

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
maggieneterval: Thanks for the wonderful code review and java schooling @ezimanyi! Just re-tested and added null checks to each class, so this should be ready for one final look :pray:

PR #2762, opened 2019-03-15
fix(scriptStage): add get properties task

emjburns:
The script stage also needs to fetch properties


ezimanyi:
Thanks for fixing this!  Retroactive LGTM.

PR #2766, opened 2019-03-18
fix(tests): Use JUnit vintage engine so Spock tests still run

jkschneider:
It's not clear to me why Spock's selection of JDK dynamic proxies for tests like `SpinnakerMetadataServerGroupTagGeneratorSpec` and `RestorePinnedServerGroupsPollerSpec` used to work without the correct visibility modifiers, but they seem to be necessary now?


jkschneider:

PR #2767, opened 2019-03-18
test(pipelinetemplates): Add tests to ModuleTag

ezimanyi:
I was trying to reproduce the particular case that broke on the Jinjava upgrade.  So far I've been unsuccessful, as these tests all pass both before and after the upgrade. But it's probably worth committing the tests anyway, just to prevent other issues in the future.


ezimanyi:
@robzienert : As noted above, I was unsuccessful in writing a test that failed on a Jinjava upgrade.  If you have any thoughts on how to reproduce, I'm happy to try some more.  That being said, it's not blocking me in any way to have this pinned to the old version, so removing the pin is not a priority and I'm happy to just leave it pinned.

robzienert:
Thanks for tackling this. I'll try to take some time into reproducing as well in the not-too-distant future.

PR #2774, opened 2019-03-19
fix(jenkins): Fix Jenkins trigger serialization

jkschneider:
* `BuildInfo<A>` needs to be reified to `JenkinsBuildInfo extends BuildInfo<JenkinsArtifact>` for deserialization to correctly deserialize into `JenkinsArtifact`. Otherwise, Jackson leaves the artifact as a `Map` and fails type checking that map against the `JenkinsArtifact` type on serialization.
* Made `JenkinsBuildInfo#name` optional, which I'm not sure if it really is, but was causing `BuildDetailExtractor` to silently swallow exceptions in tests.
* Made it clear that the deprecated method in `BuildDetailExtractor` was only intended for Jenkins build details.


jkschneider:

PR #2777, opened 2019-03-21
fix(MPTv2): Avoid resolving artifacts during v2 MPT plan.

jtk54:


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/V2PipelineTemplateController.java
```diff
@@ -38,14 +38,19 @@
 public class V2PipelineTemplateController {
 
   @Autowired
-  private OperationsController operationsController;
+  private ContextParameterProcessor contextParameterProcessor;
 
+  // pipelineProcess is not optional since it's gated by pipelineTemplates.enabled
```
ezimanyi: Not sure I understand this comment...you mean `pipelinePreprocessors`?

jtk54: Yes, I wanted to point out here that it's optionally autowired in the `OperationsController`, but we make it required in this class. I think it's likely more confusing than it's worth so I took it out.

jtk54:

PR #2779, opened 2019-03-21
fix(concourse): Fix concourse build info type

jkschneider:
The type needs to be concrete in order to be deserializable.


jkschneider:

PR #2780, opened 2019-03-21
fix(artifacts): Fix successful filter for find artifacts

ezimanyi:
When filtering to only include successful executions, the front-end sets the flag 'successful' in the stage config. Orca is incorrectly looking for a value 'succeeded', and thus ignores the value that was set on the front-end.

While 'succeeded' is slightly better named as it matches the execution status, changing this on the backend means that existing pipelines will work automatically.  (Whereas changing it on the front-end would require either a pipeline migrator or for anyone affected to re-configure the stage.)

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
@@ -73,13 +73,13 @@ public class FindArtifactFromExecutionTask implements Task {
 
   @Data
   private static class ExecutionOptions {
-    boolean succeeded;
+    boolean successful;
     boolean terminal;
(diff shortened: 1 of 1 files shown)


ezimanyi:
For reference, the code in deck that sets this flag is [here](https://github.com/spinnaker/deck/blob/09e96a2c31bd3a9d1e259438f9dceb88146d54ee/app/scripts/modules/core/src/pipeline/config/stages/findArtifactFromExecution/findArtifactFromExecutionConfig.html#L33).

sbwsg:
Is it possible that someone has figured out this workaround and correctly updated their stage JSON to `succeeded`?  Might it be worth checking on both `succeeded` and `successful` for backwards compat or do you think that's unnecessary?

ezimanyi:
@sbwsg : Good point, I'd hesitated around whether to accept both, but it's probably safer to do so.  I've updated this to accept either `successful` or `succeeded` and added a comment explaining why.

ezimanyi:

ezimanyi:

PR #2785, opened 2019-03-23
fix(triggers): Add Jenkins and Concourse build info types to allowed …

jkschneider:
@emjburns This solves the immediate problem you reported, but I'm not exactly sure what the second pipeline looked like. Would you mind testing this in your env?

> Failed to evaluate ${trigger.parentExecution.trigger.buildInfo.url}, returning raw value EL1021E: A problem occurred whilst attempting to access the property 'buildInfo': 'found getter for requested buildInfo but rejected due to return type class com.netflix.spinnaker.orca.pipeline.model.JenkinsBuildInfo'"
,
"logger_name":"com.netflix.spinnaker.orca.pipeline.expressions.ExpressionTransform"


emjburns:
LGTM - this fixes the problem we saw. Thanks!

PR #2787, opened 2019-03-25
feat(gcb): Add Google Cloud Build stage

ezimanyi:
Add a stage that triggers a build using Google Cloud Build. At this point, the stage only accepts a build definition directly in the stage, and does not wait for the build to complete.

### DIFF
--- a/orca-igor/orca-igor.gradle
+++ b/orca-igor/orca-igor.gradle
@@ -22,7 +22,7 @@ dependencies {
   compile project(":orca-front50")
   compile spinnaker.dependency('kork')
   compile spinnaker.dependency('bootAutoConfigure')
-  compile 'com.google.apis:google-api-services-cloudbuild:v1-rev836-1.25.0'
+  compile spinnaker.dependency('googleCloudBuild')
   testCompile project(":orca-test-groovy")
   testCompile spinnaker.dependency("springTest")
 }
--- a/orca-igor/orca-igor.gradle
+++ b/orca-igor/orca-igor.gradle
@@ -22,6 +22,7 @@ dependencies {
   compile project(":orca-front50")
   compile spinnaker.dependency('kork')
   compile spinnaker.dependency('bootAutoConfigure')
+  compile 'com.google.apis:google-api-services-cloudbuild:v1-rev836-1.25.0'
   testCompile project(":orca-test-groovy")
   testCompile spinnaker.dependency("springTest")
 }
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/GoogleCloudBuildStageDefinition.java
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/GoogleCloudBuildStageDefinition.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2019 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.igor.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.api.services.cloudbuild.v1.model.Build;
+import lombok.Getter;
+
+@Getter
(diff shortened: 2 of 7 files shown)


orca-igor/orca-igor.gradle
```diff
@@ -22,6 +22,7 @@ dependencies {
   compile project(":orca-front50")
   compile spinnaker.dependency('kork')
   compile spinnaker.dependency('bootAutoConfigure')
+  compile 'com.google.apis:google-api-services-cloudbuild:v1-rev836-1.25.0'
```
jtk54: Removed when we bump spinnaker-dependencies?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/GoogleCloudBuildStageDefinition.java
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.api.services.cloudbuild.v1.model.Build;
+import lombok.Getter;
+
+@Getter
+public class GoogleCloudBuildStageDefinition {
+  private final String account;
+  private final Build buildDefinition;
+
+  // There does not seem to be a way to auto-generate a constructor using our current version of Lombok (1.16.20) that
```
jtk54: Jackson needs getters and setters for the class fields, so `@Data` on the class usually works here.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/GoogleCloudBuildStage.java
```diff
@@ -0,0 +1,39 @@
...
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class GoogleCloudBuildStage implements StageDefinitionBuilder {
+  @Override
+  public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
+    GoogleCloudBuildStageDefinition stageDefinition = stage.mapTo(GoogleCloudBuildStageDefinition.class);
+    builder
+      .withTask("startGoogleCloudBuildTask", StartGoogleCloudBuildTask.class);
```
jtk54: Totally async right now -- will this eventually be enriched to parse build results or no?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartGoogleCloudBuildTask.java
```diff
@@ -0,0 +1,42 @@
...
+@Component
+@RequiredArgsConstructor
+public class StartGoogleCloudBuildTask implements Task {
+  private final IgorService igorService;
+
+  @Override
+  @Nonnull public TaskResult execute(@Nonnull Stage stage) {
+    GoogleCloudBuildStageDefinition stageDefinition = stage.mapTo(GoogleCloudBuildStageDefinition.class);
+    Operation result = igorService.createGoogleCloudBuild(stageDefinition.getAccount(), stageDefinition.getBuildDefinition());
+    return new TaskResult(ExecutionStatus.SUCCEEDED);
```
jtk54: Should this be dependent on the build result?

jtk54: Or is that just the async tracking task in GCP?

orca-igor/orca-igor.gradle
ezimanyi: Yup, exactly.  I'll remove it in this same PR after I rebase on top of the bump.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/GoogleCloudBuildStageDefinition.java
ezimanyi: Yeah, I really wanted an immutable class as the stage definition shouldn't be mutated (and won't actually be persisted if you do mutate it)...and this was the best way I could find to have Jackson deserialize into an immutable class.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/pipeline/GoogleCloudBuildStage.java
ezimanyi: Yup, absolutely---adding support for getting build status to igor, then having a task in orca to wait for that is next on my list.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartGoogleCloudBuildTask.java
ezimanyi: This task will succeed if the build is correctly triggered, then there will be a second "wait for results" task that will wait for the build to complete and succeed/fail based on the actual build result.

PR #2788, opened 2019-03-25
fix(sql): Fix intermittently failing tests

ezimanyi:
Some of the execution repository tests assert the ordering of executions retreived from the database, but ULIDs are only monotonic with millisecond resolution. The fix has been to sleep for 1ms beteween creating executions, which mostly fixed the tests but there are still occasional failures where two executions have
the same timestamp bits in their ULID.

To reduce these failures, just wait 5ms between adding executions to the database.


asher:
@robzienert I wonder if changing `SqlExecutionRepository` to always use `ULID().nextMonotonicValue` would be worth the requirement of tracking the last assigned ULID?

PR #2791, opened 2019-03-26
fix(queue): Ensure that after stages run with an authenticated context

ajordens:
This addresses a situation where the `DeployCanaryStage` failed to
plan after stages that involved a lookup against a restricted account.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
@@ -29,6 +29,7 @@ import com.netflix.spinnaker.orca.pipeline.model.Stage
 import com.netflix.spinnaker.orca.pipeline.model.Task
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
 import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor
+import com.netflix.spinnaker.orca.pipeline.util.StageNavigator
 import com.netflix.spinnaker.orca.q.*
 import com.netflix.spinnaker.q.Queue
 import org.springframework.beans.factory.annotation.Qualifier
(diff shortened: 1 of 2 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -78,7 +82,13 @@ class CompleteStageHandler(
               status = SKIPPED
             }
           } else if (status.isFailure) {
-            if (stage.planOnFailureStages()) {
+            var hasOnFailureStages = false
+
+            stage.withAuth {
+              hasOnFailureStages = stage.planOnFailureStages()
```
ajordens: I got some kotlin error when I nested all this within `stage.withAuth` ... something about `return@withStage` not being legitimate.

Broke this apart to get around it.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -64,7 +66,9 @@ class CompleteStageHandler(
             // check to see if this stage has any unplanned synthetic after stages
             var afterStages = stage.firstAfterStages()
             if (afterStages.isEmpty()) {
-              stage.planAfterStages()
+              stage.withAuth {
```
ajordens: `StartStageHandler` was already doing this which is why `DeployCanaryStage` half-worked.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
robfletcher: I always think that `return@something` syntax is a good sign that things need breaking apart so I'm 100% down with that.

marchello2000: reading through https://kotlinlang.org/docs/reference/returns.html and not understanding why you can't just do `return@withStage` here...  @robfletcher, if you know, would love to understand it.

PR #2793, opened 2019-03-26
feat(runJob): support kubernetes jobs

ethanfrogers:
supports preconfigured run job stages for the kubernetes v2 provider. i
had to break the config into provider specific configuration to work
around serialization issues.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/KubernetesPreconfiguredJobProperties.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/KubernetesPreconfiguredJobProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Netflix, Inc.
+ * Copyright 2019 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
@@ -16,14 +16,27 @@
 
 package com.netflix.spinnaker.orca.clouddriver.config;
 
+import com.google.common.collect.ImmutableList;
+import io.kubernetes.client.models.V1Job;
 import lombok.Data;
-import org.springframework.boot.context.properties.ConfigurationProperties;
+import lombok.EqualsAndHashCode;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
-@ConfigurationProperties("job.preconfigured")
 @Data
-public class JobConfigurationProperties {
-  List<TitusPreconfiguredJobProperties> titus;
-  List<KubernetesPreconfiguredJobProperties> kubernetes;
+@EqualsAndHashCode(callSuper = true)
+public class KubernetesPreconfiguredJobProperties extends PreconfiguredJobStageProperties {
+
+  private String account;
+  private String application;
+  private V1Job manifest;
+
+  public List<String> getOverridableFields() {
+    List<String> overrideableFields = Arrays.asList("account", "manifest", "application");
+    overrideableFields.addAll(super.getOverridableFields());
+    return overrideableFields;
+  }
+
 }
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/PreconfiguredJobStageProperties.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/PreconfiguredJobStageProperties.java
@@ -16,27 +16,34 @@
 
 package com.netflix.spinnaker.orca.clouddriver.config;
 
+import com.google.common.collect.ImmutableList;
+import jdk.nashorn.internal.ir.annotations.Immutable;
 import lombok.Data;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 @Data
+public abstract class PreconfiguredJobStageProperties {
 
-public class PreconfiguredJobStageProperties {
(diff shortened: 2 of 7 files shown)


ethanfrogers:
cc @robzienert @ezimanyi 

Hoping to get your feedback on approach, here. Particularly on how I had to implement provider specific `PreconfiguresJobStageProperties` to get around the deserialization issues that we talked about on Slack. `Map<String, Object> cluster` is too ambiguous for Spring to properly map lists so it just maps them as properties which throws thing off downstream. One thing to note (specifically for @robzienert) is this is backwards incompatible config change.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/PreconfiguredJobStageProperties.java
```diff
@@ -24,7 +24,7 @@
 
 @Data
 
-public class PreconfiguredJobStageProperties {
+public abstract class PreconfiguredJobStageProperties {
```
robzienert: If you're going to be using `@Data`, you shouldn't use public properties, since the accessors will be generated.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/PreconfiguredJobStageProperties.java
```diff
@@ -37,6 +37,8 @@
   public String credentials;
   public String region;
   public String propertyFile;
-  public Map<String, Object> cluster = new HashMap<>();
+//  public Map<String, Object> cluster = new HashMap<>();
```
robzienert: ?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/KubernetesPreconfiguredJobProperties.java
```diff
@@ -0,0 +1,37 @@
...
+
+@Data
+public class KubernetesPreconfiguredJobProperties extends PreconfiguredJobStageProperties {
+
+  public String account;
+  public String application;
+  public V1Job manifest;
+
+  public List<String> getAdditionalFields() {
+    return new ArrayList<>(Arrays.asList("account", "manifest", "application"));
```
robzienert: This wrapping of the `asList` result is unnecessary. If you're concerned about a consumer modifying it, then use `ImmutableList.of()`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/TitusPreconfiguredJobProperties.java
```diff
@@ -0,0 +1,33 @@
...
+package com.netflix.spinnaker.orca.clouddriver.config;
+
+import com.google.common.collect.ImmutableList;
+import lombok.Data;
+
+import java.lang.reflect.Array;
+import java.util.*;
+
+@Data
+public class TitusPreconfiguredJobProperties extends PreconfiguredJobStageProperties {
```
ezimanyi: I think you'll want a `@EqualsAndHashCode(callSuper = true)` to avoid a compile warning (and potentially incorrect `equals` behavior).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/KubernetesPreconfiguredJobProperties.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.common.collect.ImmutableList;
+import io.kubernetes.client.models.V1Job;
+import lombok.Data;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+@Data
+public class KubernetesPreconfiguredJobProperties extends PreconfiguredJobStageProperties {
```
ezimanyi: Same comment as below about `@EqualsAndHashCode`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.groovy
```diff
@@ -48,7 +48,9 @@ class PreconfiguredJobStage extends RunJobStage {
   }
 
   private Map<String, Object> overrideIfNotSetInContextAndOverrideDefault(Map<String, Object> context, PreconfiguredJobStageProperties preconfiguredJob) {
-    fields.each {
+    def fieldsToCheck = fields.collect()
```
ezimanyi: This is just following the pre-existing code, but I'm not a huge fan of the reflection used above...it might make more sense to a `getOverridableFields()` in the abstract class, then have the child classes override it with their additional fields and then call the super to add anything new.  That being said this is mostly pre-existing

ethanfrogers: Agreed, I think I actually prefer that. I guess I'll take this Travis outage as an opportunity to refactor this now 😃 .

ezimanyi:
Unrelated to this directly, but was not overly happy to see that the Groovy `PreconfiguredJobStage` that this is build on top of was added in an ureviewed, self-merged PR only a few months ago, despite the spinnakerbot warning about adding new Groovy classes (https://github.com/spinnaker/orca/pull/2540).

PR #2798, opened 2019-03-27
feat(core): Add support for more flexible execution preprocessors

robzienert:
We had an extension point for preprocessing Pipelines, but not Orchestrations. This PR adds the ability to have preprocessors for specific execution types, as well as generic preprocessors.

Also adds a new preprocessor for ensuring all executions have an application defined.


orca-core/src/main/java/com/netflix/spinnaker/orca/config/PreprocessorConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+import java.util.Optional;
+
+@Configuration
+@EnableConfigurationProperties(DefaultApplicationConfigurationProperties.class)
+@ComponentScan("com.netflix.spinnaker.orca.preprocessors")
+public class PreprocessorConfiguration {
+
+  @Bean
+  ExecutionPreprocessorRegistry executionPreprocessorRegistry(Optional<List<ExecutionPreprocessor>> preprocessors) {
```
jtk54: This is currently different from how `PipelinePreprocessors` are registered, right? Is this the pattern we want to support going forward?

orca-extensionpoint/orca-extensionpoint.gradle
```diff
@@ -0,0 +1,5 @@
+apply from: "$rootDir/gradle/spock.gradle"
+
+dependencies {
+  compile "com.google.code.findbugs:jsr305:3.0.2"
```
jtk54: Intentional? What is this dep buying us?

asher:
Is it still a long-term goal to have a single execution type, doing away with the orchestration distinction? If so, why not apply execution preprocessors to orchestrations without distinction?

orca-extensionpoint/orca-extensionpoint.gradle
cfieber: this is the (AFAIK) canonical dep for `javax.annotation.Nonnull`

ajordens:
This would still be neat-o (I'm adding a new `PreProcessor`)

robzienert:
@jtk54 sorry, been mia. this is preprocessing that occurs even before an execution becomes a real object. It allows you to fully massage the payload before the rest of the system works with it. Currently, this extension point is only available to pipelines (its how mpt plugs in), but this expands the capability to orchestrations as well (or in addition to). We have some security-/resilience-related preprocessors that  requires coverage of orchestrations as well. Make sense?

orca-core/src/main/java/com/netflix/spinnaker/orca/config/PreprocessorConfiguration.java
robzienert: Yes, correct.

robzienert:
@asher (+everyone, design idea inside) Hmm, you're probably right. I think I'd rather see a `type: ExecutionType` enum become a `String` and just use it as a (validated) partitioning key. With this, we'd still be able to organize certain executions into different areas of the UI (assuming `pipelines` and `orchestrations` are seeding types). 

1. Seem reasonable direction for an RFC? 
1. For this PR, should I reduce to just an `Execution`-level integration? Would definitely simplify things.

ajordens:
> @asher (+everyone, design idea inside) Hmm, you're probably right. I think I'd rather see a `type: ExecutionType` enum become a `String` and just use it as a (validated) partitioning key. With this, we'd still be able to organize certain executions into different areas of the UI (assuming `pipelines` and `orchestrations` are seeding types).
> 
> 1. Seem reasonable direction for an RFC?
> 2. For this PR, should I reduce to just an `Execution`-level integration? Would definitely simplify things.

I'm a +1 for the PR as-is ... but will leave it to you to decide whether it's worth effort to simplify now vs. when we get around to consolidating the two types.

robzienert:
@ajordens Thanks. I'll simplify & merge this weekend!

PR #2799, opened 2019-03-27
fix(imagetagging): asset foundImage count >= upstreamImageIds count

asher:


asher:
When there are upstream imageId's (such as in the output of a Bake stage), `ImageTagger` will retry up to the task timeout until those images can be successfully looked up via clouddriver in order to map the `imageIds` to a set of `imageNames`. Clouddriver is then queried for each `imageName` and `ImageTagger` will retry up to the task timeout until each imageName resolves, however there was no assertion that the number of images returned by name actually match the number of upstream imageIds.

It is common for a bake to produce images in multiple regions with a common name. If it takes longer for clouddriver to cache a region than others, the image lookup by name may return a subset of the just-baked images. This adds an assertion that will retry the by-name lookups until the found image count is not < the number of upstream `imageIds`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/image/ImageTagger.java
```diff
@@ -102,6 +108,14 @@ protected Collection findImages(Collection<String> imageNames, Collection<String
       foundImages.add(objectMapper.convertValue(matchedImage, matchedImageType));
     }
 
+    if (foundImages.size() < upstreamImageIds.size()) {
+      throw new ImageNotFound(
+        format("Only found %d images for names %s but %d upstreamImageIds",
```
marchello2000: nit: this could show up in the UI, right? If so, maybe something a bit digesteable for the average user would be better, e.g.
`Only found %d images to tag but %d were specified (%s)`

asher: This is a special exception type that results in the message being logged and the task requeued. It isn’t a failure and shouldn’t surface to deck or execution context. Will clarify the language for the sake of our logs however.

asher: This is a special exception type that results in the message being logged and the task requeued. It isn’t a failure and shouldn’t surface to deck or execution context. Will clarify the language for the sake of our logs however.

marchello2000: i see, I thought that the last one of these would get logged together with the timeout status/exception, but now i can't find that code

PR #2801, opened 2019-03-28
fix(clouddriver): Enable the parameter of allowDeleteActive

mybayern1974:
There is the logic of “allowDeleteActive” in orca but seems it’s hardcoded to false. Azure has this specific scenario of deleting active instances in the ShrinkClusterTask in the Red/Black strategy, so enable this parameter. Assume other cloud providers that have been worked without interacting with this parameter won’t be impacted since this change makes its value as “false” as what it is used to be if this parameter is not set explicitly anywhere else.


robfletcher:
LGTM

PR #2802, opened 2019-03-28
fix(preconfiguredJob): add tests preconfig job

ethanfrogers:
adds tests for the preconfigured job stage for upcoming groovy -> java
refactor.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/KubernetesPreconfiguredJobProperties.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/KubernetesPreconfiguredJobProperties.java
@@ -34,7 +34,7 @@ public class KubernetesPreconfiguredJobProperties extends PreconfiguredJobStageP
   private V1Job manifest;
 
   public List<String> getOverridableFields() {
-    List<String> overrideableFields = Arrays.asList("account", "manifest", "application");
+    List<String> overrideableFields = new ArrayList<>(Arrays.asList("account", "manifest", "application"));
     overrideableFields.addAll(super.getOverridableFields());
     return overrideableFields;
   }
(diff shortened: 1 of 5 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/KubernetesPreconfiguredJobProperties.java
```diff
@@ -33,8 +33,10 @@
   private String application;
   private V1Job manifest;
 
+  public KubernetesPreconfiguredJobProperties(){}
```
ezimanyi: This should be generated by default

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.groovy
```diff
@@ -22,17 +22,21 @@ import com.netflix.spinnaker.orca.clouddriver.exception.PreconfiguredJobNotFound
...
-import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
 @Component
 class PreconfiguredJobStage extends RunJobStage {
 
-  @Autowired(required=false)
   private JobService jobService
 
+  public PreconfiguredJobStage() {}
```
ezimanyi: Is the no-arg constructor used?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStageSpec.groovy
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
ezimanyi: s/Netflix/Armory/

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.groovy
ethanfrogers: There used to be a `@Autowired(required=false)` on `jobService` so I figured I need it? How do you handle non-required beans with constructor injection?

ezimanyi: Ah, I see, good question---the way to do that is to inject an `Optional<JobService> optionalJobService` into the constructor, then have the constructor do something like `this.jobService = optionalJobService.orElse(null)` in the constructor.

PR #2803, opened 2019-03-28
fix(MPTv2): Fix pipeline triggers for v2 templated pipelines.

jtk54:
The listener for pipeline triggers was not planning templated pipelines and hence not triggering on pipeline completion.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy
@@ -16,32 +16,50 @@
 
 package com.netflix.spinnaker.orca.front50.spring
 
+
 import com.netflix.spinnaker.fiat.shared.FiatStatus
 import com.netflix.spinnaker.orca.ExecutionStatus
+import com.netflix.spinnaker.orca.extensionpoint.pipeline.PipelinePreprocessor
 import com.netflix.spinnaker.orca.front50.DependentPipelineStarter
 import com.netflix.spinnaker.orca.front50.Front50Service
 import com.netflix.spinnaker.orca.listeners.ExecutionListener
 import com.netflix.spinnaker.orca.listeners.Persister
 import com.netflix.spinnaker.orca.pipeline.model.Execution
+import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor
+import com.netflix.spinnaker.orca.pipelinetemplate.V2Util
 import com.netflix.spinnaker.security.User
 import groovy.transform.CompileDynamic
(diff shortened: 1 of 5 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy
```diff
@@ -16,32 +16,55 @@
...
 import groovy.transform.CompileDynamic
 import groovy.util.logging.Slf4j
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Component
+
 import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.PIPELINE
 
 @Slf4j
 @CompileDynamic
+@Component
```
ezimanyi: I think moving this from an explicit `@Bean` to a `@Component` removed the gating on `@ConditionalOnExpression('${front50.enabled:true}')`

Also this moves us from constructor injection (which is preferred) to value injection...even moving this to a component you could still just autowire the constructor.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy
```diff
@@ -51,8 +74,17 @@ class DependentPipelineExecutionListener implements ExecutionListener {
     }
 
     def status = convertStatus(execution)
+    def allPipelines = front50Service.getAllPipelines()
+    if (pipelinePreprocessors) {
+      // Resolve templated pipelines if enabled.
+      allPipelines = allPipelines.findAll { it.type == 'templatedPipeline'}
+      .collect { pipeline ->
+        V2Util.planPipeline(contextParameterProcessor, pipelinePreprocessors, pipeline)
```
ezimanyi: Do we need to gate this to only get called on v2 templates, or is it a no-op for v1?

jtk54: We should no-op for v1 to preserve behavior.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy
jtk54: Agreed on both, thanks for pointing that out.

jtk54:

jtk54:
I'll cook up a port for 1.13.

jtk54:
Actually I think this fix was also built on https://github.com/spinnaker/orca/pull/2777, so I need to pick that one first.

jtk54:

PR #2804, opened 2019-03-28
refactor(preconfiguredJob): groovy -> java

ethanfrogers:
refactor preconfigured job stage from groovy to java


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.java
```diff
@@ -0,0 +1,90 @@
...
+      .orElseThrow(() -> new PreconfiguredJobNotFoundException(stage.getType()));
+
+    stage.setContext(overrideIfNotSetInContextAndOverrideDefault(stage.getContext(), preconfiguredJob));
+    super.taskGraph(stage, builder);
+  }
+
+  private Map<String, Object> overrideIfNotSetInContextAndOverrideDefault(Map<String, Object> context, PreconfiguredJobStageProperties preconfiguredJob) {
+    Map<String, Object> preconfiguredMap = objectMapper.convertValue(preconfiguredJob, Map.class);
+    preconfiguredJob.getOverridableFields().stream().forEach(field -> {
+        Object contextValue = context.containsKey(field) ? context.get(field) : null;
```
ezimanyi: Indentation looks a bit off here.  Also I *think* this is equivalent to
```
if (!context.containsKey(field) || preconfiguredJob.containsKey(field)) {
  context.put(field, preconfiguredJob.get(field);
}
```
(though that would assume that the maps don't actually contain keys which point to null).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.java
```diff
@@ -0,0 +1,90 @@
...
+        Object preconfiguredValue = preconfiguredMap.containsKey(field) ? preconfiguredMap.get(field) : null;
+      if (contextValue == null || preconfiguredValue != null) {
+        context.put(field, preconfiguredValue);
+      }
+    });
+
+    preconfiguredJob.getParameters().stream().forEach( parameter -> {
+      if (parameter.getDefaultValue() != null) {
+        Eval.xy(context, parameter.getDefaultValue(), String.format("x.%s = y.toString()", parameter.getMapping()));
```
ezimanyi: This (pre-existing) string `Eval` stuff is kind of scary and also a bit confusing.  I'm 99% sure this is just equivalent to the much safer and more readable:
```
context.put(parameter.getMapping(), parameter.getDefaultValue().toString())
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.java
```diff
@@ -0,0 +1,90 @@
...
+        .forEach( entry -> {
+          String parameterKey = entry.getKey();
+          String parameterValue = entry.getValue();
+          PreconfiguredJobStageParameter parameter = preconfiguredJob.getParameters().stream()
+            .filter( p ->  p.getName().equals(parameterKey))
+            .findFirst()
+            .orElse(null);
+
+          if (parameter != null) {
+            Eval.xy(context, parameterValue, String.format("x.%s = y.toString()", parameter.getMapping()));
```
ezimanyi: Same comment as above about the pre-existing `Eval`.  This looks like it's just:
```
context.put(parameter.getMapping(), parameterValue.toString())
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.java
ethanfrogers: I'll experiment with it. If `Map.put` can do deeply nested keys that should be fine. `Eval.xy` allows us to replace things like `manifest.spec.template.spec.containers[0].env[0].value`.

ezimanyi: Oh, I had no idea that was supported by that `Eval`...yeah, that makes this more difficult.  If at all possible would be great to find a more robust solution.  (This would seem to allow things like assigning `metaClass` or other reserved things to the parameter map, which is not great.)

ethanfrogers: Yeah, it's actually pretty handy because you can just define how a parameter maps to any of the fields. Could you provide an example for how you could assign `metaClass` using the things above?

Also, maybe we can just take the approach that preconfigured jobs are managed by the trusted operator? User supplied values are always `toString`ed so it seems safe to me.

ethanfrogers: Artifact replacers in Clouddriver do something similar, however, they convert the map to a document and then use `JsonPath` to replace things.

ezimanyi: I dug a bit into this and don't think it's actually injectable from an end-user perspective as the uesr-supplied values are bound to expressions by `Eval.xy`...the inject is in the `parameter.getMapping()` which is operator-supplied as you say.

What I did run into while testing this out is that I got an NPE when I tried to assign more deeply nested maps unless the `context` already had the full structure of the map.  That is, if I start with an empty context and try:
```
Eval.xy(context, 'abc', "x.abc.def = y")
```
I get an NPE, but it's fine if I've already initialized `context = [abc: [:]]`.  Presumably something is already setting up the structure of the context before we get to this code?

I'd probably prefer something more explicit than eval of the string...but realize it's pre-existing and might be difficult to clean up.  To what extent are the new kubernetes jobs relying on this deeply-nested assignment behavior?

ethanfrogers: Right, I guess it's assumed that you're mapping values directly onto an existing map which, in this case, is supplied by the preconfigured job.

Kubernetes jobs aren't relying on this _themselves_ but the preconfigured portion of them is (Titus included). This is used to allow the user to supply single values that are "overlayed" onto the configuration before it's executed by the providers run job interface.

ethanfrogers: For posterity: The concern here is that you could craft a `mapping` which actually runs malicious code. While it's generally accepted that the operator trusted we're going to put some additional checks in place to ensure that mapping is actually a JSON path (as intended).

cfieber:
@ethanfrogers not sure if you and @ezimanyi came to a consensus on this but I'm definitely :+1:  on getting rid of the groovy but would really like to see a better story than the groovy Eval... (it would be nice to get to a place where groovy is not a compile time dependency)

ethanfrogers:
@cfieber we've definitely reached a consensus on getting rid of the Groovy but not necessarily on how to eliminate the Eval and preserve the functionality (or intended functionality, rather). I'll probably take another wack at this next week.

robzienert:
@ethanfrogers ping

ethanfrogers:
@robzienert pong. This hasn't been at the top of my priority list, unfortunately. Eric and I chatted a while ago about how to get rid of `Eval` but didn't come up with a good solution. I'll try to take another look at this today if I can.

robzienert:
Feel free to re-open when you get around to it - just doing some cleanup.

PR #2805, opened 2019-03-28
fix(cf): Clone server group allows for the setting of environment var…

jasonjchu:
spinnaker/spinnaker#4211


jkschneider:
We need to make the necessary change in clouddriver instead.

PR #2811, opened 2019-04-03
Upgrade to Gradle 5.0

robfletcher:
As a prelude to consuming the BOM from Kork, I need to upgrade the Gradle version.


orca-queue/orca-queue.gradle
```diff
@@ -25,6 +25,7 @@ dependencies {
   compile "org.funktionale:funktionale-partials:1.2"
   compile spinnaker.dependency("logstashEncoder")
   compile "com.netflix.spinnaker.keiko:keiko-spring:$keikoVersion"
+  compile "javax.ws.rs:jsr311-api:1.1.1"
```
robfletcher: Was getting pulled in transitively by Eureka before but not now.

PR #2812, opened 2019-04-04
feat(exp): deployedServerGroups now grabs deployments from deploy result

claymccoy:
The deployment result is the place that cloud providers can put info about their deployments, but it is deeply nested on the context and hard to find. That can be eased by appending it to the info returned from the existing deployedServerGroups property/function.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/DeployedServerGroupsExpressionFunctionProvider.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/DeployedServerGroupsExpressionFunctionProvider.java
@@ -15,6 +15,7 @@
  */
 package com.netflix.spinnaker.orca.pipeline.expressions.functions;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.netflix.spinnaker.orca.ExecutionStatus;
 import com.netflix.spinnaker.orca.pipeline.expressions.ExpressionFunctionProvider;
 import com.netflix.spinnaker.orca.pipeline.model.Execution;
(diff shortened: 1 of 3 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/DeployedServerGroupsExpressionFunctionProvider.java
```diff
@@ -83,13 +87,32 @@ public String getNamespace() {
             deployDetails.put("serverGroup", serverGroups.get(0));
           }
 
+          DeploymentContext deploymentContext = stage.mapTo(DeploymentContext.class);
+          List<Map<String, Object>> deployments = Optional.ofNullable(deploymentContext.tasks).orElse(emptyList()).stream()
+            .flatMap(task -> Optional.ofNullable(task.results).orElse(emptyList()).stream())
+            .flatMap(result -> Optional.ofNullable(result.deployments).orElse(emptyList()).stream())
+            .collect(Collectors.toList());
+          deployDetails.put("deployments", deployments);
```
emjburns: Just a question, why are you using optionals here? Does it provide any advantages over checking if something exists and only adding deployments if they exist?

jkschneider: I suppose it is `Optional.ofNullable(foo).orElse(emptyList()).stream()` or `(foo == null ? emptyList() : foo).stream()`. Stylistic preference?

emjburns: I don't have a big preference. I think if statements are slightly more clear in general, but feel free to leave as is.

PR #2813, opened 2019-04-04
fix(travis): Support timestamp in JenkinsBuildInfo (provided by Travis)

jervi:
(Travis is piggybacking on the Jenkins trigger implementation in Orca)

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/JenkinsTrigger.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/JenkinsTrigger.kt
@@ -57,7 +57,8 @@ constructor(@param:JsonProperty("name") override val name: String?,
             @param:JsonProperty("result") override val result: String?,
             @param:JsonProperty("artifacts") override val artifacts: List<JenkinsArtifact>?,
             @param:JsonProperty("scm") override val scm: List<SourceControl>?,
-            @param:JsonProperty("building") override var building: Boolean = false)
+            @param:JsonProperty("building") override var building: Boolean = false,
+            @param:JsonProperty("timestamp") val timestamp: Long?)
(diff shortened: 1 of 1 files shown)


jervi:
@spinnaker/reviewers PTAL 🙂

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/JenkinsTrigger.kt
```diff
@@ -57,7 +57,8 @@ constructor(@param:JsonProperty("name") override val name: String?,
             @param:JsonProperty("result") override val result: String?,
             @param:JsonProperty("artifacts") override val artifacts: List<JenkinsArtifact>?,
             @param:JsonProperty("scm") override val scm: List<SourceControl>?,
-            @param:JsonProperty("building") override var building: Boolean = false)
+            @param:JsonProperty("building") override var building: Boolean = false,
+            @param:JsonProperty("timestamp") val timestamp: Long?)
```
robfletcher: Is it really nullable or will Jenkins always supply one?

ajordens:
Slightly off-topic but `(Travis is piggybacking on the Jenkins trigger implementation in Orca)` ... is there a path to making this not the case? 

Strikes me as strange, is there a reason it has to be this way or is it legacy?

jervi:
It is legacy I think. The Travis implementation was in the beginning only in Igor; a Travis master appeared in the rest of Spinnaker as a Jenkins master named “travis-something”. @gardleopard would know more, I think.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/JenkinsTrigger.kt
jervi: I don’t really know if Jenkins supplies this field at all... Travis always supplies it.

PR #2814, opened 2019-04-04
refactor(MPTv2): Change nomenclature from version to tag.

jtk54:


jtk54:
Yes, we did store the version in the MPT blob in Front50. That being said, we didn't surface the `version` functionality at all in tooling so I'd be surprised if this was used. Additionally, we can't simply migrate the `version` entry to `tag` since we also restricted the space of tags as well, so I'm inclined to let this go.

PR #2815, opened 2019-04-04
fix(provider/cf): Bind clone manifest artifacts

jkschneider:
There was some AWS-specific logic in `CloneServerGroupTask` that adds launch descriptions to the task list for a clone. Separately, there was a need for us to bind an expected artifact for a manifest that is an input to the CF clone operation. I took this opportunity to extract the cloud provider specific modifications to this task to a `CloneDescriptionDecorator`.

image

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/clone/CloneServerGroupTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/clone/CloneServerGroupTask.groovy
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.orca.clouddriver.tasks.servergroup
+package com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.clone
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.orca.ExecutionStatus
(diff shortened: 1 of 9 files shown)


jkschneider:

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/clone/CloneServerGroupTask.groovy
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.orca.clouddriver.tasks.servergroup
+package com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.clone
```
marchello2000: @jkschneider : any reason why this was changed to a different package? This breaks the upgrade scenario when you have a bunch of messages in the queue with `"taskType": "com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.CloneServerGroupTask",`
and you upgrade to this orca (same for downgrade)

do we have a mechanism to coerce types that I can fix before moving this back to the old package?

(cc: @ajordens)

marchello2000: Looking and I think we can use `SerializationMigrator` to fix this up

marchello2000: @jkschneider FYI: i am going to move this back as the migrator is not a good solution (since we will not be able to roll back from it if need be). I will CC you on the PR, shortly

PR #2817, opened 2019-04-04
fix(expressions): make sure trigger is a map

marchello2000:
Some places uses `ContextParameterProcessor.buildExecutionContext` which (correctly) converts the
`trigger` in the pipeline to an object (instead of keeping it as `Trigger` class).
However, when tasks run, they use `.withMergedContext` which opts to build it's own SpEL
executionContext (in `.augmentContext`). This is needed so that fancy look up in ancestor stages works.
Fix `.augmentContext` to convert the trigger to an object/map.

This addresses issues evaluating expressions like
`${trigger.buildNumber ?: #stage('s1').context.buildNumber}`` when no `buildNumber` is present on the `trigger`

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesTask.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesTask.java
@@ -36,6 +36,11 @@ public class EvaluateVariablesTask implements Task {
   public EvaluateVariablesTask() {
   }
 
+  /**
+   * Copies previously evaluated expressions to the outputs map for consumption by subsequent stages.
+   * The variables aren't evaluated here because would've been evaluated already by a call to
+   * e.g. ExpressionAware.Stage#withMergedContext
+   */
   @Nonnull
   @Override
   public TaskResult execute(@Nonnull Stage stage) {
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
@@ -16,7 +16,10 @@
 
 package com.netflix.spinnaker.orca.q.handler
 
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.convertValue
 import com.netflix.spinnaker.orca.exceptions.ExceptionHandler
+import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper
 import com.netflix.spinnaker.orca.pipeline.expressions.ExpressionEvaluationSummary
 import com.netflix.spinnaker.orca.pipeline.expressions.PipelineExpressionEvaluator
 import com.netflix.spinnaker.orca.pipeline.model.Execution
@@ -35,6 +38,9 @@ interface ExpressionAware {
   val contextParameterProcessor: ContextParameterProcessor
   val log: Logger
     get() = LoggerFactory.getLogger(javaClass)
+  private val mapper: ObjectMapper
+    get() = OrcaObjectMapper.newInstance()
+
 
   fun Stage.withMergedContext(): Stage {
     val evalSummary = ExpressionEvaluationSummary()
@@ -130,7 +136,9 @@ interface ExpressionAware {
 
   private fun StageContext.augmentContext(execution: Execution): StageContext =
     if (execution.type == PIPELINE) {
-      this + mapOf("trigger" to execution.trigger, "execution" to execution)
+      this + mapOf(
+        "trigger" to mapper.convertValue<Map<String, Any>>(execution.trigger),
+        "execution" to execution)
     } else {
       this
     }
--- a/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesTaskSpec.groovy
+++ b/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesTaskSpec.groovy
@@ -26,7 +26,7 @@ class EvaluateVariablesTaskSpec extends Specification {
   @Subject
   task = new EvaluateVariablesTask()
 
-  void "Should correctly evaulate variables"() {
+  void "Should correctly copy evaluated variables"() {
     setup:
     def stage = stage {
       refId = "1"
--- a/orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
+++ b/orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
@@ -44,7 +44,6 @@ import org.jetbrains.spek.api.dsl.on
 import org.jetbrains.spek.api.lifecycle.CachingMode.GROUP
 import org.jetbrains.spek.subject.SubjectSpek
 import org.threeten.extra.Minutes
-import java.lang.RuntimeException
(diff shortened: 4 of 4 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
```diff
@@ -130,7 +136,9 @@ interface ExpressionAware {
 
   private fun StageContext.augmentContext(execution: Execution): StageContext =
```
marchello2000: ideally, we opt out of this method and use `ContextParameterProcessor.buildExecutionContext` but that doesn't generate `StageContext` preventing us from doing recursive lookups... An investigation for another day

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesTask.java
```diff
@@ -36,6 +36,11 @@
   public EvaluateVariablesTask() {
   }
 
+  /**
+   * Note: Because the variables are evaluated by (pretty much all) task handlers
+   * i.e. in ExpressionAware.Stage#withMergedContext
+   * all the task needs to do is to copy the variables to the outputs
```
emjburns: A key part of this stage is the "copy" - because variables are evaluated and then saved as the non-expression version. Is there a way to make this comment clearer?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
```diff
@@ -130,7 +136,9 @@ interface ExpressionAware {
...
     if (execution.type == PIPELINE) {
-      this + mapOf("trigger" to execution.trigger, "execution" to execution)
+      this + mapOf(
+        "trigger" to mapper.convertValue(execution.trigger, object : TypeReference<Map<String, Any>>() {}),
```
emjburns: do you need this `object` bit?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesTask.java
marchello2000: that's what I wanted to convey, clearly that didn't work :)
How about:
```
This task copies previously evaluated expressions to the outputs map for consumption by subsequent stages.
It does not need to evaluate those variables because they would've been evaluated already by a call to e.g. ExpressionAware.Stage#withMergedContext
```

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
marchello2000: good q, i think i can use the `kotlin` `convertValue<T>` variant

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesTask.java
emjburns: I like this wording better but I think "it does not need to evaluate those variables" is not really true - they are evaluated in that stage, they're just not evaluated with a direct call. They're evaluated in behind the scenes magic.

marchello2000: Doesn't "...because they would've been evaluated already by..." say that?

marchello2000: how about:

This task copies previously evaluated expressions to the outputs map for consumption by subsequent stages.
The variables aren't evaluated here because would've been evaluated already by a call to e.g. ExpressionAware.Stage#withMergedContext

marchello2000:
added a test for good measure

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesTask.java
emjburns: That's great. I know it's super explicit, but hopefully it will help future debuggers.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesTask.java
```diff
@@ -36,6 +36,11 @@
   public EvaluateVariablesTask() {
   }
 
+  /**
+   * Copies previously evaluated expressions to the outputs map for consumption by subsequent stages.
+   * The variables aren't evaluated here because would've been evaluated already by a call to
+   * e.g. ExpressionAware.Stage#withMergedContext
+   */
```
robzienert: maybe better put on the class level docs?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
robzienert: worth adding as a todo  comment so we dont lose context after this pr is forgotten?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
```diff
@@ -35,6 +38,9 @@ interface ExpressionAware {
   val contextParameterProcessor: ContextParameterProcessor
   val log: Logger
     get() = LoggerFactory.getLogger(javaClass)
+  private val mapper: ObjectMapper
+    get() = OrcaObjectMapper.newInstance()
```
robzienert: mmm, do we really want to `new` up a new object mapper each time? i think just assigning this once would be better

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesTask.java
marchello2000: i was going back the forth on this :)... happy to move it

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
marchello2000: yeah, my bad, thanks!

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
maggieneterval: Hi friends! Sorry to surface bad memories, but I came across this PR while investigating [this bug](https://github.com/spinnaker/spinnaker/issues/4932), which reports that we are failing to evaluate SpEL in Kubernetes manifests downloaded as artifacts. Since we are generating the execution context for the [mid-task SpEL evaluation](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java#L132) using `ContextParameterProcessor.buildExecutionContext`, which @marchello2000 correctly points out does not return `StageContext`, the evaluation fails. I've verified that changing the return type of that method from `Map<String, Object>` to `StageContext` resolves the issue, but wanted to check if there had been additional discussion offline about this before I submitted a PR. Thanks in advance!

Love,
Sleepless in SpEL-hell

PR #2820, opened 2019-04-05
fix(pipelines): Fixed interpretation of missing startTime during inst…

ewiseblatt:
…rumentation

Fixes https://github.com/spinnaker/spinnaker/issues/4244

There are conditions in which CompleteStageHandler can be called without
having a startTime set. Often this is because an exception was thrown
before the startTime was set, but there are various guards within
StartStageHandler as well.

Previously when this happened, a StartTime of 0 was used. This is
particularly bad because the 40-years worth of nanoseconds that is
spit out of spectator is beyond the resolution of "double" for metric
stores that use double for time. This fix changes the assumption
to be the end time, resulting in assuming 0 time. Not that this could
possibly be a different type of issue if the underlying problem was
something else, such as StartStageHandler not even being called for some
reason.

In addition I set the startTime at the beginning of the try block where
exceptions may be thrown in order to ensure there is a startTime when
going through that control flow. I think this might be messy depending
on how time and state are used because it has a startTime but is not in
the RUNNING state. There does not seem to be an appropriate state definition
for this particular phase and am not sure the implications of adding one.
I also dont know if moving the startTime so it includes this extra check
has other implications. If so then the startTime could be reset where it
is currently set.

@robfletcher


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
```diff
@@ -77,14 +77,14 @@ class StartStageHandler(
...
         } else {
           try {
+            // Set the startTime in case we throw an exception.
+            stage.startTime = clock.millis()
+
             stage.withAuth {
               stage.plan()
             }
 
-            stage.status = RUNNING
```
marchello2000: what happened to this assignment? Pretty sure it's a) needed and b) cause of test failures

robfletcher: Very needed, yes.

robfletcher: You will also need to call `repository.storeStage(stage)` before doing the planning otherwise any exception thrown will result in the timestamp being lost.

marchello2000:
other than missing `stage.status = RUNNING` lgtm, I don't think setting `startTime` earlier will have any consequences and agree with you it would be nice to set `startTime` somewhere in a generic way that doesn't seem to exist.

@robfletcher correct me if I am wrong

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
ewiseblatt: My bad. I'd moved this and a copy of store stage above the withAuth then decided not to and forgot to put it back.

ewiseblatt:
Note I kept the RUNNING where it was after the withAuth but not before the
origin save with the time.
Is that OK? I'm not sure how that is used or the implications. Please
confirm.

jtk54:
@ewiseblatt I think setting the startTime prior to setting RUNNING is OK since all this state is wrapped in a message that's pushed to the queue.

ewiseblatt:
My concern was more if the exception is thrown and it is in the RUNNING
state but never reached that original code point.
I dont know how the states are used and whether someone knows RUNNING means
it completed the earlier step that is now failing.

robfletcher:
It feels odd to me to not have the state and the startTime set together but it's probably very low impact as that RUNNING state is going to get either set or replaced with TERMINAL in very short order.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -149,9 +149,17 @@ class CompleteStageHandler(
         } ?: id
       }
 
+    // If startTime was not set, then assume this was instantaneous.
+    // In practice 0 startTimes can happen if there is
+    // an exception in StartStageHandler or guard causing skipping.
+    // Without a startTime, we cannot record a meaningful time,
```
ezimanyi: Would it make sense to just skip recording the metric entirely if we don't have a start time, rather than record a placeholder value of 0?

ewiseblatt: There's still value in the counts. If the time were 0 because of some other error condition, the final status would be TERMINAL (or maybe NOT_STARTED if it were some non-error reason). Otherwise you would want another counter indicating errors in this counter to raise awareness to a situation where stages were not executing. That's even more complicated.

ezimanyi: Yeah, I guess it does provide visibility into the counts without the complexity of another metric, makes sense.

ewiseblatt:
I needed to make a change to satisfy the tests (and fix the cardinality checks in some tests). Please take another look.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
```diff
@@ -77,12 +77,14 @@ class StartStageHandler(
...
           try {
+            // Set the startTime in case we throw an exception.
+            stage.startTime = clock.millis()
+            repository.storeStage(stage)
+
             stage.withAuth {
               stage.plan()
             }
-
             stage.status = RUNNING
```
marchello2000: I think it's actually better to have this together with `stage.startTime` as @robfletcher noted. Looking at `CompleteStageHandler` (https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt#L56) it seems that it will be harmless. And it will be more accurate

marchello2000:
@ewiseblatt should we merge this PR?

PR #2822, opened 2019-04-08
feat(conditions): Adding support for config based conditions

jeyrschabu:
- support for config based conditions
- support for config based whitelisted clusters

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionConfigurationProperties.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionConfigurationProperties.java
@@ -17,9 +17,11 @@
 package com.netflix.spinnaker.orca.conditions;
 
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.stereotype.Component;
 
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 @Component
@@ -29,7 +31,10 @@ public class ConditionConfigurationProperties {
   private boolean enabled = false;
   private Long backoffWaitMs = TimeUnit.MINUTES.toMillis(5);
   private Long waitTimeoutMs = TimeUnit.MINUTES.toMillis(120);
+  private List<String> clusters;
+  private List<String> activeConditions;
 
+  @Autowired
   public ConditionConfigurationProperties(DynamicConfigService configService) {
     this.configService = configService;
   }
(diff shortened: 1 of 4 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionConfigurationProperties.java
```diff
@@ -29,7 +31,10 @@
   private boolean enabled = false;
   private Long backoffWaitMs = TimeUnit.MINUTES.toMillis(5);
   private Long waitTimeoutMs = TimeUnit.MINUTES.toMillis(120);
+  private List<String> clusters;
+  private List<String> conditions;
```
emjburns: What is the format of these conditions/clusters?

orca-core/src/test/groovy/com/netflix/spinnaker/orca/conditions/ConfigurationBackedConditionSupplierSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+
+    when:
+    def result = conditionSupplier.getConditions(stage)
+
+    then:
+    result.size() == numberOfConditions
+
+    where:
+    ctx                 | clusters        | configuredConditions | numberOfConditions
+    [cluster: "foo"]    | []              | []                   | 0
```
emjburns: so the behavior when there are no clusters is nothing happens?

emjburns:
I screwed up the github review dialog  (and meant to just comment) but basically I have some questions about how this is designed to be used.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/conditions/ConfigurationBackedConditionSupplierSpec.groovy
jeyrschabu: In this assertion, the resulting number of conditions is 0 because there aren't any configured conditions. Defining a list of clusters in the config is analogous to whitelisting. Let me know if this is unclear

orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionConfigurationProperties.java
emjburns: A string for a condition doesn't seem very resilient. Or is it just a message to print?

orca-core/src/test/groovy/com/netflix/spinnaker/orca/conditions/ConfigurationBackedConditionSupplierSpec.groovy
emjburns: What happens if there are 0 clusters but 2 conditions?

orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionConfigurationProperties.java
jeyrschabu: This is just a message to print. Aimed to be used for testing.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/conditions/ConfigurationBackedConditionSupplierSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+
+    then:
+    result.size() == numberOfResultingConditions
+
+    where:
+    ctx                 | clusters        | activeConditions     | numberOfResultingConditions
+    [cluster: "foo"]    | []              | []                   | 0
+    [cluster: "foo"]    | ["foo", "bar"]  | []                   | 0
+    [cluster: "foo"]    | ["bar"]         | [ "c1", "c2"]        | 0
+    [cluster: "foo"]    | ["foo", "bar"]  | [ "c1", "c2"]        | 2
```
emjburns: can you add the test for zero clusters but two conditions?

PR #2825, opened 2019-04-09
chore(cf): allow deployment strategies to work across different regions

Jammy-Louie:
* allow deployment strategies to work across different regions
* relax constraints so that account does not have to be the same credentials.

spinnaker/spinnaker#4258

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
@@ -169,7 +169,8 @@ class TargetServerGroup {
     static Location locationFromStageData(StageData stageData) {
       try {
         List zones = stageData.availabilityZones?.values()?.flatten()?.toArray()
-        return resolveLocation(stageData.namespace, stageData.region, zones?.getAt(0))
+        String region = stageData.destination?.region?:stageData.region
+        return resolveLocation(stageData.namespace, region, zones?.getAt(0))
       } catch (e) {
         throw new IllegalArgumentException("Incorrect location specified for ${stageData}: ${e.message}")
       }
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/StageData.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/StageData.groovy
@@ -44,6 +44,7 @@ class StageData {
   Boolean useSourceCapacity
   Boolean preferSourceCapacity
   Source source
+  Destination destination
 
   @Deprecated
   long delayBeforeDisableSec
@@ -65,9 +66,6 @@ class StageData {
   }
 
   String getAccount() {
-    if (account && credentials && account != credentials) {
-      throw new IllegalStateException("Cannot specify different values for 'account' and 'credentials' (${application})")
-    }
     return account ?: credentials
   }
 
@@ -113,6 +111,12 @@ class StageData {
     String serverGroupName
     Boolean useSourceCapacity
     Boolean preferSourceCapacity
+  }
+
+  @EqualsAndHashCode
+  @ToString
+  static class Destination {
+    String region
   }
 
   static class PipelineBeforeCleanup {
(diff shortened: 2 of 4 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
```diff
@@ -169,7 +169,8 @@ class TargetServerGroup {
     static Location locationFromStageData(StageData stageData) {
       try {
         List zones = stageData.availabilityZones?.values()?.flatten()?.toArray()
-        return resolveLocation(stageData.namespace, stageData.region, zones?.getAt(0))
+        String region = stageData.destination?.region?:stageData.region
```
jkschneider: Isn't this used to determine where to locate the server group that we are cloning? I think "target" is misleading in this context.

Jammy-Louie: when I did a search for the usages of `locationFromStageData` it appears this method is only used here [AbstractDeployStrategyStage](https://github.com/spinnaker/orca/blob/90e24bd548777c2c0bc1f55163ec2a4f15d3f131/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/AbstractDeployStrategyStage.groovy)

on L130
https://github.com/spinnaker/orca/blob/90e24bd548777c2c0bc1f55163ec2a4f15d3f131/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/AbstractDeployStrategyStage.groovy#L130

and L230

https://github.com/spinnaker/orca/blob/90e24bd548777c2c0bc1f55163ec2a4f15d3f131/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/AbstractDeployStrategyStage.groovy#L230

setting the region to destination to the destination region would mean we are applying the deployment strategy to the where the clone destination is.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/StageData.groovy
```diff
@@ -65,9 +66,6 @@ class StageData {
   }
 
   String getAccount() {
-    if (account && credentials && account != credentials) {
```
ajordens: Motivation for removing this sanity check?

Jammy-Louie: The clone operation we wanted to perform could be done across Cloud Foundry foundations which would mean we want to perform a clone from different accounts and regions.

we were using `account` as the destination and `credentials` as the the source of the clone. The check prevents the usage of deployment strategy for our clone operation.

jkschneider: We are trying to achieve a clone server group stage that can lift an app from one account to another and use typical deployment strategies _a la_ highlander.

`CloneServerGroupStage` has a series of tasks and inheritance that impose these constraints:
1. Determines the server group it is going to clone _from_ with `credentials`
2. Determines which server group to monitor in `WaitForUpInstancesTask` with this (looks at `account` before `credentials`): https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/clouddriver/utils/CloudProviderAware.java#L51
3. `CloneServerGroupStage` extends from `AbstractDeployStrategyStage`. When a deployment strategy is present on the stage configuration, the strategy handling logic calls this method `StageData#getAccount()`

It is necessary for us to use both `account` to identify the target account and `credentials` to define the source account without requiring significant changes to the way the stage is constructed.

Does this make sense?

jkschneider:
After a discussion with @ajordens, we'll close this for now in favor of exploring the use of `source` instead. `destination.account` would map to `credentials`, `destination.region` would map to `region` on the root of the stage context.

We'll look at having `CloneServerGroupTask` prefer `source` and fall back on `credentials`/`region` on providers that haven't historically supported cross-account cloning.

PR #2826, opened 2019-04-09
fix(cloneservergrouptask): undo the move of CloneServerGroupTask

marchello2000:
This change undoes the move of CloneServerGroupTask to a different package
(this was introduced in #2815).

Today, tasks can not be moved once they have been created because Orca will
be unable to deserialize the tasks already in the queue created under a
different package.


marchello2000:
As discussed with @robzienert - we can't use a migration here because that will prevent us from rolling back this release should we need to.

marchello2000:
In order to get a release out today, I am going to merge, please comment if you see anything weird and I will address it after merge

PR #2827, opened 2019-04-09
feat(cf): Fetch created service key via SpEL

stuart-pollock:
spinnaker/spinnaker#4260

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProvider.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProvider.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2019 Pivotal, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.pipeline.expressions.functions;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.TreeTraversingParser;
+import com.netflix.spinnaker.orca.ExecutionStatus;
+import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper;
+import com.netflix.spinnaker.orca.pipeline.expressions.ExpressionFunctionProvider;
+import com.netflix.spinnaker.orca.pipeline.model.Execution;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import lombok.Data;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+import java.util.*;
+import java.util.function.Predicate;
+
+@Component
+public class ServiceKeyExpressionFunctionProvider implements ExpressionFunctionProvider {
+  private static final String CREATE_SERVICE_KEY_STAGE_NAME = "createServiceKey";
+  private static final ObjectMapper objectMapper = OrcaObjectMapper.newInstance();
+
+  @Nullable
+  @Override
+  public String getNamespace() {
+    return null;
+  }
+
+  @NotNull
+  @Override
+  public Collection<FunctionDefinition> getFunctions() {
+    return Collections.singletonList(
+      new FunctionDefinition("cfServiceKey", Arrays.asList(
+        new FunctionParameter(Execution.class, "execution", "The execution within which to search for stages"),
+        new FunctionParameter(String.class, "idOrName", "A stage name or stage ID to match")
+      ))
+    );
+  }
+
+  public static Map<String, Object> cfServiceKey(Execution execution, String idOrName) {
+    return execution.getStages()
+      .stream()
+      .filter(matchesServiceKeyStage(idOrName))
+      .findFirst()
+      .map(stage -> {
+        Map<String, Object> serviceKeyDetails = new HashMap<>();
+
+        Optional.ofNullable(stage.getContext().get("kato.tasks"))
+          .ifPresent(k -> {
+            List<Map<String, Object>> katoTasks = (List<Map<String, Object>>) k;
+            try {
+              ServiceKeyKatoTask katoTask = objectMapper.readValue(
+                new TreeTraversingParser(objectMapper.valueToTree(katoTasks.get(0)), objectMapper),
+                ServiceKeyKatoTask.class);
+              serviceKeyDetails.putAll(katoTask.getResultObjects().get(0).getServiceKey());
+            } catch (IOException e) {
+            }
+          });
+
+          return serviceKeyDetails;
+        })
+      .orElse(Collections.emptyMap());
+  }
+
+  private static Predicate<Stage> matchesServiceKeyStage(String idOrName) {
+    return stage -> CREATE_SERVICE_KEY_STAGE_NAME.equals(stage.getType()) &&
+      stage.getStatus() == ExecutionStatus.SUCCEEDED &&
+      (Objects.equals(idOrName, stage.getName()) || Objects.equals(idOrName, stage.getId()));
+  }
+
+  @Data
+  private static class ServiceKeyKatoTask {
+    private List<ServiceKeyResult> resultObjects;
+
+    @Data
+    private static class ServiceKeyResult {
+      private Map<String, Object> serviceKey;
+    }
+  }
+}
--- a/orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProviderTest.java
+++ b/orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProviderTest.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2019 Pivotal, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.pipeline.expressions.functions;
+
+import com.google.common.collect.ImmutableMap;
+import com.netflix.spinnaker.orca.ExecutionStatus;
+import com.netflix.spinnaker.orca.pipeline.expressions.ExpressionFunctionProvider.FunctionDefinition;
+import com.netflix.spinnaker.orca.pipeline.model.Execution;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import org.junit.jupiter.api.Test;
+
+import java.util.*;
+
+import static com.netflix.spinnaker.orca.ExecutionStatus.RUNNING;
+import static com.netflix.spinnaker.orca.ExecutionStatus.SUCCEEDED;
+import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.PIPELINE;
+import static java.util.Collections.*;
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ServiceKeyExpressionFunctionProviderTest {
+  @Test
+  void getFunctionsShouldReturnOneFunctionWithTheCorrectNameWhichHasTwoParameters() {
+    ServiceKeyExpressionFunctionProvider functionProvider = new ServiceKeyExpressionFunctionProvider();
+    Collection<FunctionDefinition> functionDefinitions = functionProvider.getFunctions();
+
+    assertThat(functionDefinitions.size()).isEqualTo(1);
+    functionDefinitions.stream().findFirst().ifPresent(functionDefinition -> {
+      assertThat(functionDefinition.getName()).isEqualTo("cfServiceKey");
+      assertThat(functionDefinition.getParameters().get(0).getType()).isEqualTo(Execution.class);
+      assertThat(functionDefinition.getParameters().get(1).getType()).isEqualTo(String.class);
+    });
+  }
+
+  @Test
+  void serviceKeyShouldResolveForValidStageType() {
+    String user = "user1";
+    String password = "password1";
+    String url = "mysql.example.com";
+    Map<String, Object> serviceKey = new ImmutableMap.Builder<String, Object>()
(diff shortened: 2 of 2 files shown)


orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProviderTest.java
```diff
@@ -0,0 +1,128 @@
...
+  @Test
+  void serviceKeyShouldReturnEmptyMapForNoResult() {
+    Map<String, Object> katoTaskMapWithoutResults = new ImmutableMap.Builder<String, Object>()
+      .put("id", "task-id")
+      .put("status", SUCCEEDED)
+      .put("history", Collections.emptyList())
+      .put("resultObjects", Collections.emptyList())
+      .build();
+    Execution execution = new Execution(PIPELINE, "stage-name-1", "application-name");
+    Map<String, Object> contextWithoutServiceKey = new HashMap<>();
```
marchello2000:
```
    Map<String, Object> contextWithoutServiceKey = new HashMap<>();
    contextWithoutServiceKey.put("parentStage", "parent-stage");
    contextWithoutServiceKey.put("account", "account-name");
    contextWithoutServiceKey.put("credentials", "my-account");
    contextWithoutServiceKey.put("region", "org > space");
```
is repeated a bunch of times, can make a helper function to make test more readable

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProvider.java
```diff
@@ -0,0 +1,103 @@
...
+        new FunctionParameter(String[].class, "ids", "A list of stage name or stage IDs to search")
+      ))
+    );
+  }
+
+  public static Map<String, Object> serviceKey(Execution execution, String... id) {
+    return execution.getStages()
+      .stream()
+      .filter(matchesServiceKeyStage(id))
+      .findFirst()
```
marchello2000: can there only be one `createServiceKeyStage`?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProvider.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+          });
+
+          return serviceKeyDetails;
+        })
+      .orElse(Collections.emptyMap());
+  }
+
+  private static Predicate<Stage> matchesServiceKeyStage(String... id) {
+    List<String> idsOrNames = Arrays.asList(id);
```
marchello2000: can/should be `final`

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProvider.java
```diff
@@ -0,0 +1,103 @@
...
+  @Override
+  public String getNamespace() {
+    return null;
+  }
+
+  @NotNull
+  @Override
+  public Collection<FunctionDefinition> getFunctions() {
+    return Collections.singletonList(
+      new FunctionDefinition("serviceKey", Arrays.asList(
```
marchello2000: my concern here is that this is the first function that is (sort of) provider specific. Perhaps worth prefixing it with `cf`, e.g. `cfServiceKey`? cc: @robzienert

marchello2000: Also, don't forget to document it here: https://github.com/spinnaker/spinnaker.github.io/blob/master/reference/pipeline/expressions/index.md

robzienert: This was actually the point of to `getNamespace` method. So cloud provider spel helpers could be prefixed with the provider name. Then calling the method would look like so:

```
${#cloudfoundry.serviceKey(...)}
```

All this said, I don't like our provider specific logic in `orca-core` but that's currently the what we do for the other spel helpers today so I figure something out for think it's fine for now. This seems a perfect candidate later on for plugins?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProvider.java
stuart-pollock: IntelliJ complains:  `As private methods cannot be meaningfully overridden, declaring them final is redundant.`.  We've been suggested to follow IntelliJ's suggestions in the past.  It's also consistent with https://github.com/spinnaker/orca/blob/820148ef29850cdce320aa0af3cad484065f4989/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/DeployedServerGroupsExpressionFunctionProvider.java#L93.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProvider.java
stuart-pollock: There may be multiple `createServiceKey` stages.  Each will have its own name, and we expect them to be addressed primarily by name rather than by id.  We have removed the spread operator.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProvider.java
marchello2000: i meant the `idsOrNames`

stuart-pollock: We've eliminated the variable in response to another of your suggestions.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProvider.java
ajordens: +1 on the namespacing, that was the intention.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProvider.java
marchello2000: got it, so the above should be `"id"` instead of `"ids"` (in the params)

marchello2000:
looks good, just add the namespace and change `ids` to `id`

orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+    public String getNamespace() {
+      return "testNamespace";
+    }
+
+    @NotNull
+    @Override
+    public Collection<FunctionDefinition> getFunctions() {
+      return Collections.singletonList(
+        new FunctionDefinition("testFunctionInNamespace", singletonList(
+          new FunctionParameter(Execution.class, "functionParam", "Function param")
```
marchello2000: from the usage of how you have it above, this should be:
```
Arrays.asList(
        new FunctionParameter(Execution.class, "execution", "Execution param"),
        new FunctionParameter(String.class, "functionParam", "Function param")
))
```

instead of `singletonList`

orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransformTest.java
```diff
@@ -0,0 +1,95 @@
...
+  void includeExecutionParameterDoesNotInsertNullNamespaceInExpressionFunctionProvider() {
+    String expression = "${#testFunctionNoNamespace(\"param\")}";
+    String expectedResult = "${#testFunctionNoNamespace( #root.execution, \"param\")}";
+
+    String result = expressionTransform.includeExecutionParameter(expression);
+
+    assertThat(result).isEqualTo(expectedResult);
+  }
+
+  private static class TestExpressionFunctionProviderWithNamespace implements ExpressionFunctionProvider {
```
marchello2000: nit: you could fold `TestExpressionFunctionProviderWithNamespace ` and `TestExpressionFunctionProviderWithoutNamespace` into a single class and then pametrize the `namespace` and the `params`

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.java
```diff
@@ -270,21 +270,31 @@ static String escapeSimpleExpression(String expression) {
...
-  private String includeExecutionParameter(String e) {
+  // Visible for testing
+  String includeExecutionParameter(String e) {
     String expression = e;
 
     // An expression aware function is any that takes an Execution as its first parameter
     Set<String> expressionAwareFunctions = new HashSet<>(EXECUTION_AWARE_FUNCTIONS);
-    expressionFunctionProviders.forEach(p -> {
-      p.getFunctions().forEach(f -> {
+    for (ExpressionFunctionProvider p : expressionFunctionProviders) {
```
marchello2000: i guess `namespace` was only implemented half-way, thank you for fixing it

marchello2000:
LGTM, but see if @ajordens has any other comments.

orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProviderTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+import static com.netflix.spinnaker.orca.ExecutionStatus.RUNNING;
+import static com.netflix.spinnaker.orca.ExecutionStatus.SUCCEEDED;
+import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.PIPELINE;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
+import static java.util.Collections.singletonMap;
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ServiceKeyExpressionFunctionProviderTest {
```
ajordens: For future reference, I still believe we can author tests in groovy (we just would prefer non-tests be java/kotlin).

Unit tests in strict java are still a little verbose for my liking (don't need to do this as part of this PR but consider it next time).

ajordens:
LGTM after the namespace `_` vs `.` clarification.

orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransformTest.java
stuart-pollock: 👍

orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransformTest.java
stuart-pollock: We think that the tests would benefit from being clear and explicit, since they serve essentially as documentation for the code.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionTransform.java
```diff
@@ -270,21 +270,31 @@ static String escapeSimpleExpression(String expression) {
...
+      String namespace = p.getNamespace();
+      for (ExpressionFunctionProvider.FunctionDefinition f : p.getFunctions()) {
         if (!f.getParameters().isEmpty() && f.getParameters().get(0).getType() == Execution.class) {
-          expressionAwareFunctions.add(f.getName());
+          String functionName = f.getName();
+          if (namespace != null) {
+            String registeredFunction = namespace + "_" + functionName;
+            String nameSpaceFunction = namespace + "." + functionName;
+            expression = expression.replaceAll("#" + nameSpaceFunction + "\\(", "#" + registeredFunction + "(");
```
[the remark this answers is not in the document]
stuart-pollock: We feel that the consistency of using ‘.’ instead of ‘_’ as a delimiter in the incoming SpEL expression is consistent with the use of ‘.’ as a delimiter in Javascript (for example), and will thus lead to a better user experience.  Having the pattern of using ‘.’ in Deck (and in pipeline JSON), then transforming the ‘.’ into ‘_’ under the covers in Orca feels like a small technical and conceptual price to pay for a syntax which will pay dividends in the future.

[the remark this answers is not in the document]
robzienert: Seems reasonable to me. Can we get rid of `_`? So there is a true One Way to do things?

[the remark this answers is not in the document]
stuart-pollock: Unfortunately, I do not know of a way to tell the SpEL parser that the namespacing dots should not be considered a delimiter which introduces another token.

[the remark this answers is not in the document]
ajordens: My preference is to only support one way of using a namespace'd fn and it sounds like that needs to be `_` delimitated (rather than have to worry about converting back and forth between implementation and documentation).

Will merge after.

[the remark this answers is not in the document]
stuart-pollock: There is already some precedent for using . to delimit some scope at https://github.com/spinnaker/deck/blob/0333b67defc62ae77cf5e48096ce7ea4715c1a5b/app/scripts/modules/core/src/widgets/spelText/SpelAutocompleteService.ts#L54..L55.  I think that changing those names without understanding its impact is beyond the scope of our namespacing change.  Changing our implementation to support _ would make this inconsistent.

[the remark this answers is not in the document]
stuart-pollock: We have decided to remove namespacing from our change, in lieu of @marchello2000 's initial suggestion of `cfServiceKey` instead of either `cloudfoundry_serviceKey` or `cloudfoundry.serviceKey`.  If `_` is going to be used for namespacing, then nothing would really prevent us from using `_` as part of the function name in Orca.  This invites the question of why we would use namespaces, since there are now two ways to define a function which is exposed in Deck with `_` contained within its name.

orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ServiceKeyExpressionFunctionProviderTest.java
claymccoy: Is it okay to mix Groovy and Kotlin in the test section of a subproject? One reason that we are avoiding adding Groovy, even in tests, is to clear the way to use more Kotlin in the future and there are apparently issues with mixing Groovy and Kotlin.
Also, there are a lot of ways to write Kotlin tests and none of them are as succinct as Spock. Can you point to ideal examples? We typically use Kotlin, JUnit, assert4j, and mockito, but I've seen variance in the Spinnaker projects.

ajordens: I believe it is fine ... I also suspect we should be more strict on what ends up living in a `-core` artifact.

PR #2828, opened 2019-04-09
feat(webhook): retry on name resolution failures

dreynaud:
This is adding a new narrow use case where we retry, instead of retrying
on broader exceptions. CreateWebhookTask was already quite specific about
the failure modes that are retryable, so this change is consistent with
that pattern. The reason for being conservative is that we don't want to
potentially cause side effects where we hit the same webhook multiple
times in a row.

I chose to rely on orca's built in scheduling mechanism (instead of
internal retries within the task) as it seems to be properly configured
already in terms of backoff and timeout.

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
@@ -59,14 +59,19 @@ class CreateWebhookTask implements RetryableTask {
     def response
     try {
       response = webhookService.exchange(stageData.method, stageData.url, stageData.payload, stageData.customHeaders)
-    } catch(IllegalArgumentException e) {
+    } catch (IllegalArgumentException e) {
       if (e.cause instanceof UnknownHostException) {
         String errorMessage = "name resolution failure in webhook for pipeline ${stage.execution.id} to ${stageData.url}, will retry."
         log.warn(errorMessage, e)
         outputs.webhook << [error: errorMessage]
         return new TaskResult(ExecutionStatus.RUNNING, outputs)
+      } else {
+        String errorMessage = "an exception occurred in webhook to ${stageData.url}: ${e}"
+        log.error(errorMessage, e)
+        outputs.webhook << [error: errorMessage]
+        return new TaskResult(ExecutionStatus.TERMINAL, outputs)
       }
-    } catch(HttpStatusCodeException e) {
+    } catch (HttpStatusCodeException e) {
       def statusCode = e.getStatusCode()
 
       outputs.webhook << [statusCode: statusCode, statusCodeValue: statusCode.value()]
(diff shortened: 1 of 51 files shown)


erikmunson:
👍 nifty

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -59,7 +59,14 @@ class CreateWebhookTask implements RetryableTask {
     def response
     try {
       response = webhookService.exchange(stageData.method, stageData.url, stageData.payload, stageData.customHeaders)
-    } catch (HttpStatusCodeException e) {
+    } catch(IllegalArgumentException e) {
```
marchello2000: it's fascinating that it throws `IllegalArgumentException`....

ajordens: couple missing spaces before the `(`

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -59,7 +59,14 @@ class CreateWebhookTask implements RetryableTask {
...
+      if (e.cause instanceof UnknownHostException) {
+        String errorMessage = "name resolution failure in webhook for pipeline ${stage.execution.id} to ${stageData.url}, will retry."
```
ajordens: If this error is going to be user facing I'd suggest stripping the `stage.execution.id` from it, that's confusing.

could just tack it onto the log ... `log.warn("${errorMessage} (executionId: ${stage.execution.id}", e)`

ajordens: Tough to say whether logging the entire stack trace is going to be super noisy and whether we only really need `e.message` for an `UnknownHostException`

🤷‍♂️

ajordens: Thoughts about the same treatment for the `MonitorWebhookTask`?

In theory it's equally susceptible.

marchello2000: I just came back to add a note about `MonitorWebhookTask` but you're already on, thanks @ajordens! :)

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
dreynaud: well it's really `validateURI` that wraps any exception in an `IllegalArgumentException`

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
dreynaud: @ajordens I don't think it's supposed to be user facing, if it does persist the task would fail with a timeout I think

Good catch about the `MonitorWebhookTask`, will add it

dreynaud:
Just realized I introduced an issue, the catch block is now silently swallowing `IllegalArgumentException`s that are not caused by name resolution failures. Fix incoming!

marchello2000:
lgtm

PR #2829, opened 2019-04-10
 feat(provider/kubernetes): Add traffic options to deploy manifest

ezimanyi:
* refactor(provider/kubernetes): Use constructor injection 

  Refactor DeployManifestTask to use constructor injection, and change impelentation-specific fields to be private and final.

* feat(provider/kubernetes): Add traffic options to deploy manifest 

  Add functionality to the deploy manifest task to handle the fields in the new trafficManagement section. In particular, pass along any specified service as well as whether to send traffic to new workloads.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestContext.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestContext.java
@@ -19,8 +19,10 @@ package com.netflix.spinnaker.orca.clouddriver.tasks.manifest;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.Getter;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Optional;
 
 @Getter
 public class DeployManifestContext extends HashMap<String, Object> {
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestContext.java
```diff
@@ -37,12 +40,42 @@ public DeployManifestContext(
...
   ){
     this.source = source;
     this.manifestArtifactId = manifestArtifactId;
     this.manifestArtifactAccount = manifestArtifactAccount;
     this.skipExpressionEvaluation = skipExpressionEvaluation;
+    this.trafficManagement = trafficManagement;
     this.requiredArtifactIds = requiredArtifactIds;
   }
+
+  @Getter
```
jtk54: Is there a reason to do `@Getter` and `@JsonProperty` rather than `@Data`?

ezimanyi: Immutability---as we create this from the input map and never modify it (and actually modifications will never be persisted) I didn't want to expose setters.

jtk54: Got it -- SGTM.

PR #2830, opened 2019-04-11
fix(k8s): Deploy manifest now accepts stage-inlined artifacts

jkschneider:
In 1.13, the K8S deploy manifest stage displays the new artifacts UI when the feature flag is enabled. The JSON structure it produces for the stage is not digestable by Orca. As a refresher on the structural changes done such that this is backwards compatible:

1. `manifestArtifact` can (optionally) have an `artifactAccount` now. With the new artifacts UI, you _must_ specify an artifact account up front in the expected artifact definition, so any artifacts created with the new UI will have it. `manifestArtifactAccount` is now a backwards compatible fallback for existing pipelines.
2. `requiredArtifacts` contains a list of bind artifacts that contain either an `expectedArtifactId` or a fully-defined `artifact`. The UI presents any artifacts defined in the old model (still present in `requiredArtifactIds`) along with any artifacts defined in the new model side-by-side.

image

Here's an example of what I mean by stage-inlined artifacts in the context of the k8s deploy manifest stage:

image

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java
@@ -44,6 +44,8 @@ import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
+import static java.util.Collections.emptyList;
+
 @Component
 @RequiredArgsConstructor
 @Slf4j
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java
```diff
@@ -70,25 +72,25 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+      if (manifestArtifact == null) {
         throw new IllegalArgumentException("No manifest artifact was specified.");
       }
 
-      String manifestArtifactAccount = context.getManifestArtifactAccount();
-      if (manifestArtifactAccount == null) {
-        throw new IllegalArgumentException("No manifest artifact account was specified.");
+      // Once the legacy artifacts feature is removed, all trigger expected artifacts will be required to define
+      // an account up front.
+      if(context.getManifestArtifactId() != null && manifestArtifact.getArtifactAccount() == null) {
```
ezimanyi: Should we be checking `context.getManifestArtifactAccount() != null` here?  I suppose that we expect `context.getManifestArtifactId()` to be `null` iff `context.getManifestArtifactAccount()` is `null` but it might be clearer to check the relevant field.

jkschneider: I'm afraid of a dangling `manifestArtifactAcount` as folks migrate pipelines. So trying to be extra safe.

ezimanyi: Ok, makes sense :+1:

jkschneider:
> Under what conditions would we need to evaluate SpEL in an artifact at the time of stage evaluation?
> 
> I ask mostly because in general we're trying to constraint SpEL evaluation to occur a the beginning of a pipeline run, as having multiple passes of evaluation has led to a lot of bugs and fragile code---so I'm hesitant to add another pass of evaluation here.

It is possible for artifacts to depend on preceding stage execution context, so can't be evaluated at the beginning of the pipeline:

image

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java
jkschneider: Good point. Updated.

jkschneider:
> I'm still not overly excited about explicitly supporting tight coupling between stages like that

I think I agree. Thanks for the link to the proposal. It is unclear to me how variables can be referred to in a way that is any less prone to ambiguity than SpEL, as you'd need to demarcate a variable with some set of symbols.

> Is there any concern that an artifact might contain `$` signs and that evaluating SpEL on the artifact might fail?

Theoretically there could be, but I'm not aware of any right now.

jkschneider:

rverma-nikiai:
@jkschneider Facing issue with saving pipeline using spin cli with 1.7.2 and 1.14 but able to create pipeline using ui. It looks like there is some issue with inheritance piece which is not handled correctly. Is this related to this issue?

PR #2832, opened 2019-04-11
fix(expressions): precompute trigger.parameters

marchello2000:
If trigger.parameters uses spel expressions and then, you use
some of those parameters in subsequent spel expression (e.g. in a stage)
bad things will happen, e.g.
say, you have:
```
trigger.parameters.commitSha = "${trigger.properties.commitSha}"
```

then, a stage has
```
${parameters[\"commitSha\"].substring(0,6)}"
```

the result will be "${trig" instead of the first six chars of SHA

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
@@ -81,7 +81,7 @@ public class ContextParameterProcessor {
 
     Map<String, Object> result = expressionEvaluator.evaluate(
       source,
-      precomputeValues(context),
+      precomputeValues(context, summary),
       summary,
       allowUnknownKeys
     );
@@ -117,7 +117,7 @@ public class ContextParameterProcessor {
     return isNotEmpty(value) && value.contains("${");
   }
 
-  private Map<String, Object> precomputeValues(Map<String, Object> context) {
+  private Map<String, Object> precomputeValues(Map<String, Object> context, ExpressionEvaluationSummary summary) {
     Object rawTrigger = context.get("trigger");
     Trigger trigger;
     if (rawTrigger != null && !(rawTrigger instanceof Trigger)) {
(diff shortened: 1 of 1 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
```diff
@@ -127,7 +127,9 @@ public static boolean containsExpression(String value) {
     }
 
     if (trigger != null && !trigger.getParameters().isEmpty()) {
-      context.put("parameters", trigger.getParameters());
+      // Eval trigger.parameters in case they have SpEL expressions
+      context.put("parameters",
```
ajordens: I don't know for certain ... but I recall expressions in trigger parameters have been contentious/error prone historically (purely from memory).

Unfortunately I don't know if there are existing expectations where the expression hasn't been evaluated yet. 

Is this a fix for a regression or new feature? I suppose someone could explicitly use an `Eval Expression` stage if they knew it was an expression?

Also ... I'd write this as

```
context.put(
  "parameters", expressionEvaluator.evaluate(trigger.getParameters(), context, summary, true)
);
```

or

```
context.put(
  "parameters", 
  expressionEvaluator.evaluate(trigger.getParameters(), context, summary, true)
);
```

(break the first parameter on to it's own line when longer than 120 characters)

marchello2000: thanks! Yes, I am torn about whether this should even be merged (as you said about past expectations). My plan is to basically extract all current expressions and run through this new code and make sure they produce same results as before.

This was reported today as someone was trying to simplify their parameter list in a pipeline. It's also been apparently reported before and we just told folks to use a workaround - put the whole expression into the param.

you can't use an eval expression stage (actually this is how it was reported).
e.g. you can't:
```
trigger.parameters.commitSha = "${trigger.properties.commitSha}"
```
and then use:
```
${parameters[\"commitSha\"].substring(0,6)}"
```

instead, you must:
```
trigger.parameters.commitSha = "${trigger.properties.commitSha.substring(0,6)}"
```

with this change, you can break up them up and use eval vars stage.

But yes, will do tons of testing before merging

PR #2833, opened 2019-04-12
fix(repository): don't generate invalid grammar

marchello2000:
When performing a search with a pipeline name that doesn't exist
sqlrepository will generate a query with an empty IN clause:
```
SELECT * FROM ... WHERE IN ()
```

this change short circuits this evaluation.

Additionally, set H2 into `MODE=MYSQL` so that we can actually catch
this invalid SQL in unit tests

### DIFF
--- a/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
+++ b/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
@@ -19,11 +19,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.config.TransactionRetryProperties
 import com.netflix.spinnaker.kork.core.RetrySupport
 import com.netflix.spinnaker.orca.ExecutionStatus
-import com.netflix.spinnaker.orca.ExecutionStatus.BUFFERED
-import com.netflix.spinnaker.orca.ExecutionStatus.CANCELED
-import com.netflix.spinnaker.orca.ExecutionStatus.NOT_STARTED
-import com.netflix.spinnaker.orca.ExecutionStatus.PAUSED
-import com.netflix.spinnaker.orca.ExecutionStatus.RUNNING
+import com.netflix.spinnaker.orca.ExecutionStatus.*
 import com.netflix.spinnaker.orca.pipeline.model.Execution
 import com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType
 import com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.ORCHESTRATION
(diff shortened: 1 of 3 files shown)


marchello2000:
@ajordens this is the fix for the bad grammar exception we saw.
I want to see if I can generalize it somehow, but putting this PR out there for initial eyes

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -440,6 +423,10 @@ class SqlExecutionRepository(
     buildTimeEndBoundary: Long,
     executionCriteria: ExecutionCriteria
   ): List<Execution> {
+    if (pipelineConfigIds.isEmpty()) {
```
ajordens: Seems reasonable, I'd also consider adding this to `retrieveAllPipelinesForPipelineConfigIdsBetweenBuildTimeBoundary()` since it's a public method on the interface (even though the current impl does delegate to this fn).

orca-sql/src/test/groovy/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepositorySpec.groovy
```diff
@@ -491,4 +491,17 @@ class SqlExecutionRepositorySpec extends ExecutionRepositoryTck<SqlExecutionRepo
     storeLimit = 6
   }
 
+  def "doesn't fail on empty configIds"() {
+    when:
```
ajordens: could change this to:

```
expect:
repository.retrieveAllPipelinesForPipelineConfigIdsBetweenBuildTimeBoundary(
  ...
) == 0
```

orca-sql/src/test/java/com/netflix/spinnaker/orca/sql/SqlTestUtil.java
```diff
@@ -33,17 +33,17 @@
...
 
-import static org.jooq.SQLDialect.H2;
+import static org.jooq.SQLDialect.MYSQL;
 import static org.jooq.conf.RenderNameStyle.AS_IS;
 
 public class SqlTestUtil {
 
   public static TestDatabase initDatabase() {
-    return initDatabase("jdbc:h2:mem:orca_test");
+    return initDatabase("jdbc:h2:mem:orca_test;MODE=MYSQL");
```
ajordens: ?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
marchello2000: 👍

orca-sql/src/test/groovy/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepositorySpec.groovy
marchello2000: neat, thanks

orca-sql/src/test/java/com/netflix/spinnaker/orca/sql/SqlTestUtil.java
marchello2000: so this was needed otherwise the test wouldn't actually throw bad sql grammar exception (without my fix)
this puts H2 into MySQL compatibility mode and then it throws as expected

ajordens: 👍

ajordens:
LGTM -- merge away1

marchello2000:
Refactored a bit to use the native jook `in` operator which obviates the need for any special casing of empty lists. And it's safer

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -443,14 +426,17 @@ class SqlExecutionRepository(
     val select = jooq.selectExecutions(
       PIPELINE,
       conditions = {
-        val inClause = "config_id IN (${pipelineConfigIds.joinToString(",") { "'$it'" }})"
-        val timeCondition = it.where(inClause)
+        var conditions = it.where(
+          field("config_id").`in`(*pipelineConfigIds.toTypedArray())
```
marchello2000: this cute syntax is needed because kotlin for whatever reason can't disambiguate the calls

marchello2000:
For those curious here is the kotlin "bug" that necessitated the weird syntax.
https://youtrack.jetbrains.com/issue/KT-31045

marchello2000:
> Change to `in` syntax looks great, thanks!

yes, much cleaner this way, thanks for the suggestion!

PR #2835, opened 2019-04-12
feat(preconfiguredJobs): produces artifacts

ethanfrogers:
allow preconfigured jobs to produce artifact. kubernetes jobs default to
true. can be overridden.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/KubernetesPreconfiguredJobProperties.java
```diff
@@ -33,6 +33,10 @@
   private String application;
   private V1Job manifest;
 
+  public KubernetesPreconfiguredJobProperties() {
+    this.setProducesArtifacts(true);
```
ethanfrogers: @ezimanyi I'm not sure how make this do what i want - I want to set this to true by default for Kubernetes job properties but still make it overridable by the user configuration.

ethanfrogers: Nevermind! Seems like this works as expected :)

PR #2836, opened 2019-04-12
feat(canary) - Add support for time shifting canary scopes

perek:
Currently canaries initiated from Orca must always have the same start/end time for all analysis scopes (baseline, canary, etc). This PR adds support for setting an offset value which will time shift the analysis allowing for comparing metrics against a pre-deployment baseline. This feature is intended to be used for monitoring metrics during deployments, and enabling canary analysis for applications which do not have the ability the do a full canary deployment.

Added unit tests and confirmed functionality. This PR is backwards compatable, and has some associated UI changes in Deck and Deck-Kayenta repos.


maggieneterval:
@perek I think spinnakerbot wants you to change your commit to use a colon instead of a dash: `feat(canary): Add support for time shifting canary scopes`.

This LGTM, adding @ezimanyi and @skandragon for approval. Once this is in we can merge your UI work.

ezimanyi:
@perek : I'll merge this once conflicts are resolved; the 1.14 branch cuts on Monday, so would need to be by then for this to make 1.14.  Thanks!

perek:
@ezimanyi I have rebased my branch

AbdulRahmanAlHamali:
Hello, has this ever been tested end-to-end?
I'm looking here:

https://github.com/spinnaker/kayenta/blob/c2d7251d5222023cd3993c4adbd45db123c7eba2/kayenta-standalone-canary-analysis/src/main/java/com/netflix/kayenta/standalonecanaryanalysis/orca/stage/SetupAndExecuteCanariesStage.java#L238

And I see that Kayenta does not really distinguish time of the baseline from that of the canary. Is there something I'm missing? Or does this feature just not work?

perek:
@AbdulRahmanAlHamali  the offset is applied in the interval: https://github.com/perek/orca/blob/master/orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/pipeline/RunCanaryIntervalsStage.kt#L118

perek:
the offset is passed to kayenta, which converts the offset times.

perek:
@AbdulRahmanAlHamali https://github.com/spinnaker/orca/blob/master/orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/pipeline/RunCanaryIntervalsStage.kt#L182

It is taken care of here. I am not sure what this standalone analysis module is? Is this some plugin to allow non pipeline canaries? The feature was only designed with orca in mind

AbdulRahmanAlHamali:
@perek, indeed. I was confused. I think I'll submit something to the standalone version then to support the offset. Thanks!

AbdulRahmanAlHamali:
submitted here:
https://github.com/spinnaker/kayenta/pull/824/files

PR #2837, opened 2019-04-12
fix(titus) : Add some additional logging for delete operation

aravindmd:
- Adding additional logging for debugging titus delete operation , I see a few executions missing 
 ` "kato.task.id": {
          "id": ""
        },
        "terminate.instance.ids": [
          ""
        ] `  
as part of the execution details , and i suspect the data returned from clouddriver might not have the instance that needs to be deleted and the task is marked successful , adding this additional logging to help debug this issue. Let me know if there is a better way

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/instance/TerminatingInstanceSupport.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/instance/TerminatingInstanceSupport.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.instance
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup
 import com.netflix.spinnaker.orca.clouddriver.utils.CloudProviderAware
 import com.netflix.spinnaker.orca.clouddriver.utils.OortHelper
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/instance/TerminatingInstanceSupport.groovy
```diff
@@ -83,9 +87,12 @@ class TerminatingInstanceSupport implements CloudProviderAware {
     String location = stage.context.region
 
     def tsg = oortHelper.getTargetServerGroup(account, serverGroupName, location, cloudProvider)
+    log.info("Target Server Group from oort : {}", objectMapper.writeValueAsString(tsg))
```
ajordens: This seems like an overkill amount of logging to me ... I don't understand what the symptom is that we're trying to dig into?

ajordens: If we need anything, it seems like we only need instance ids? (and perhaps at debug)

aravindmd: The issue i see is that when a titus `terminateInstances` task is called , it succeeds but fails to terminate the instance , my guess is this check fails ->(https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/TerminateInstancesTask.groovy#L61 ) and it just returns a success and i wanted to get the details of what the  `List<TerminatingInstance>` are from `clouddriver` and added these additional logging . Let me know what you think.

PR #2839, opened 2019-04-16
fix(aws/imagetag): ensure matched images by name includes all upstream ids

asher:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/image/ImageTagger.java
```diff
@@ -134,6 +136,22 @@ public OperationContext(List<Map<String, Map>> operations, Map<String, Object> e
     }
   }
 
+  /**
```
emjburns: This is a great comment, thank you!

PR #2842, opened 2019-04-17
feat(MPTv2): Inverts inherit -> exclude in template inheritance.

jtk54:
Previously by default, template consumers would have to specifically opt in to including triggers, parameters, and notifications from the parent template. This is an inversion of what is expected by
template consumers and has resulted in a bunch of confusion and misuse. We've changed the default to inherit by default and have users manually opt in to exclude the fields targeted by "inherit".

@dorbin FYI for documentation.


jtk54:
Possibly worth adding a pipeline migrator for this, thoughts?

louisjimenez:
Cool, I do think this inheritance modal is more intuitive for pipeline configs.

A migrator could be helpful if it prevents this change from breaking in use pipeline configs. I'm thinking of something like a templated trigger, which in turn references a template variable with no default value, may now be automatically inherited in a pipeline config. That config could be missing a value for the variable and have a misconfigured trigger.

jtk54:
Makes sense to me, I already dropped a note in the pipeline-templates slack channel.

PR #2843, opened 2019-04-17
fix(cloudformation): Scope force update by account and region

xavileon:
When force updating the cloud formation cache on the
AmazonCloudFormationCachingAgent, it can take some time if the number of
accounts is quite big because the on demand update iterates over all
accounts for a given type. This makes the force refresh task to fail
because of a timeout.

This patch sends scoping information to clouddriver so the caching
agents can skip doing the update if the force refresh does not involve
its own account and or region, making on demand update more efficient.

Related PR: https://github.com/spinnaker/clouddriver/pull/3576


xavileon:
@spinnaker/reviewers PTAL

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/CloudFormationForceCacheRefreshTaskSpec.groovy
```diff
@@ -36,4 +37,29 @@ class CloudFormationForceCacheRefreshTaskSpec extends Specification {
...
+
+
+    when:
+    task.execute(stage)
+
+    then:
+    1 * task.cacheService.forceCacheUpdate('aws', CloudFormationForceCacheRefreshTask.REFRESH_TYPE, data)
+
+    where:
+    credentials   | regions       | data
```
ajordens: minor nit ... expected sides of a spock table should be prefixed with `||` (vs. `|`)

```
credentials      | regions      || expectedData
```

ajordens:
Committed a couple stylistic changes ... will merge when travis gives the 👍

xavileon:
Thanks @ajordens!

PR #2847, opened 2019-04-18
fix(logging): Correctly propagate stage IDs for logging

marchello2000:
Currently, the stageID isn't sent over when we make e.g. clouddriver calls.
This makes tracing the operation of a given stage VERY difficult.
Add stageIDs to the MDC context for propagation with HTTP headers


marchello2000:
> What is the thinking around joining into one string vs putting each of as separate attribute? Not sure I have a strong opinion was just curious...

I originally had as a `X-SPINNAKER-REQUEST-ID` but @robzienert pointed out that we want the request id to be unique upon retries, which makes sense. So we settled on this pattern. I think it makes sense, but I can also see having separate headers. This seems slightly more concise to me - but also don't have a strong preference

cfieber:
:shipit:

PR #2848, opened 2019-04-18
fix(provider/azure): Failed to disable azure server group when rollback

neil-yechenwei:
Background:
DisableServerGroup stage failed to disable server group and threw error server group could not be found after ResizeServerGroup stage when rollback.
Fix:
After tested, DisableServerGroup stage can disable server group successfully after execute the ForceCacheRefresh task as the last task in ResizeServerGroup stage like other stages (EnableServerGroupStage, DisableServerGroupStage).


neil-yechenwei:
@duftler , a kindly ping to review this PR. Please help to review it.

ajordens:
I'm not so sure about this ... nor is this an `azure`-specific change.

The description talks about disable and resize but this change is only in `ResizeServerGroup`. Neither of these stages deal with creating a server group so I'm unsure why you'd get an error about one being not found.

duftler:
@neil-yechenwei Please add additional context and examples so we can understand the issue.

PR #2849, opened 2019-04-18
fix(MPTv2): Restricts config var scope based on declared template vars.

jtk54:
Fixes https://github.com/spinnaker/spinnaker/issues/4158.


jtk54:
Makes sense, I have a nice new function to unit test.

PR #2851, opened 2019-04-19
fix(FindImageFromCluster): only infer regions from deploy for aws

asher:


asher:
fixes https://github.com/spinnaker/spinnaker/issues/4299

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTaskSpec.groovy
```diff
@@ -250,6 +258,10 @@ class FindImageFromClusterTaskSpec extends Specification {
         ]
       ]
     ]
+
+    cloudProvider || findCalls
```
marchello2000: curious why do you need `||` vs `|`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -128,7 +128,7 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
     Set<String> inferredRegions = new HashSet<>()
 
     deployRegions.forEach {
-      if (!config.regions.contains(it)) {
+      if (!config.regions.contains(it) && cloudProvider == 'aws') {
```
marchello2000: prob would be more readable if it were
```
if (cloudProvider == 'aws') {
      deployRegions.forEach {
            if (!config.regions.contains(it) && cloudProvider == 'aws') {
....
}
```
or, maybe just have the conditional on line 127, that way, we don't even bother looking at child stages unless it's `aws`?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTaskSpec.groovy
asher: They're interchangeable. There's a convention to use || to separate an output or result parameter vs. input but not sure if it really makes sense here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
asher: Agree, no point iterating `deployRegions` if !aws, will update

asher:

PR #2855, opened 2019-04-22
fix(orca): if build stage fails and prop file exists, try and fetch

emjburns:
This fixes a "regression" that was introduced with the ci stage refactor. Previously we would try and get the properties file if the jenkins job failed. This pr adds a failure stage so that if a property file is defined, and the ci stage fails, we will try and get the property file. This failure stage may fail, but that seems fine because the stage itself already failed.

A user of ours is asking for this behavior.

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
@@ -41,7 +41,12 @@ public class GetBuildPropertiesTask extends RetryableIgorTask {
       return TaskResult.SUCCEEDED;
     }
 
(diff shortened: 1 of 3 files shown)


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
```diff
@@ -41,7 +41,12 @@
       return TaskResult.SUCCEEDED;
     }
 
-    Map<String, Object> properties = buildService.getPropertyFile(stageDefinition.getBuildNumber(), stageDefinition.getPropertyFile(), stageDefinition.getMaster(), stageDefinition.getJob());
+    Map<String, Object> properties = buildService.getPropertyFile(
```
emjburns: This formatting bothered me so I fixed it.

emjburns:
@ezimanyi same, I initially thought it was a new feature request!

PR #2857, opened 2019-04-22
fix(cf): Move cfServiceKey from orca-core to orca-integrations-cloudfoundry

ajordens:


ajordens:
Potentially looking to establish some better patterns around cloudprovider-specific code in `orca` (and keeping it out of the core artifacts wherever possible).

This is an initial port of a recent PR and not an attempt at excising all cloudprovider-specific code.

ajordens:
cc: @jkschneider -- Any obvious pitfalls/objections to starting down this path? 

General idea being that it's not just `cloudfoundry` (but I had made this mental note when reviewing a recent PR). 

We encouraged the gremlin folks to also go down this path with their recent work.

orca-integrations-cloudfoundry/orca-integrations-cloudfoundry.gradle
```diff
@@ -0,0 +1,18 @@
...
+  compileOnly spinnaker.dependency("lombok")
+  annotationProcessor spinnaker.dependency("lombok")
+
+  compile project(":orca-core")
+
+  testCompile spinnaker.dependency("junitJupiterApi")
+  testCompile spinnaker.dependency("assertj")
+
+  testRuntime spinnaker.dependency("junitJupiterEngine")
+  testRuntime "org.junit.vintage:junit-vintage-engine:${spinnaker.version('jupiter')}"
```
jkschneider: I'd strike vintage in any new module. It only exists where there are a mix of Spock/JUnit 5 tests in other modules.

PR #2858, opened 2019-04-22
fix(webhooks): Avoid a nasty HTTP 500 fetching preconfigured webhooks

ajordens:
This PR offers protection if `fiat` were unavailable when attempting
to fetch preconfigured webhooks.

The current behavior results in an HTTP 500 even if there were no
restricted preconfigured webhooks.

The proposed behavior would result in the unrestricted subset being
returned with an error logged.


ajordens:
@ezimanyi It looks like you touched this code most recently ... popped up during a chaos exercise today where `fiat` was disabled.

Looks like `deck` fetches preconfigured webhooks fairly often and this was littering the JS console with 500's (even tho for us we don't actually have preconfigured webhooks).

PR #2859, opened 2019-04-22
refactor(conditions): Do not inject waitForCondition on no conditions

jeyrschabu:
- Updated interface to specify cluster, region and account
- Make injecting the stage optional on if there are active conditions

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
@@ -16,29 +16,37 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline;
 
+import com.netflix.spinnaker.orca.clouddriver.pipeline.conditions.Condition;
+import com.netflix.spinnaker.orca.clouddriver.pipeline.conditions.ConditionSupplier;
+import com.netflix.spinnaker.orca.clouddriver.pipeline.conditions.WaitForConditionStage;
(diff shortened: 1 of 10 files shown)


jeyrschabu:
Currently `waitForConditionStage` gets injected before a deploy even if there are no conditions,
resulting in a NOOP `waitForCondition` that succeeds right away showing up in the UI.
To avoid confusion for our users, this PR ensures that the `waitForConditionStage` is not injected if there aren't active conditions to wait on.

orca-core/src/main/java/com/netflix/spinnaker/orca/conditions/ConditionSupplier.java
```diff
@@ -26,5 +25,5 @@
   /**
    * returns a list of currently unmet conditions.
    */
-  List<Condition> getConditions(Stage stage);
+  List<Condition> getConditions(String cluster, String region, String account);
```
ajordens: As an aside, perhaps this interface should move out of `orca-core` and into `orca-clouddriver` since it's deploy-centric anyhow?

jeyrschabu: +1

ajordens: One thing that just came to mind ... have these tasks been run in any of our environments? May need to do some cleanup to account for the tasks changing package.

Otherwise porting this code to the other package still makes sense to me.

jeyrschabu: Yeah In prestaging from a test pipeline. Clarification question re:cleanup to account, are you referring to cleaning the tasks from the database?

ajordens: Ya, that should be all that's needed. At some point we'll get to fixing the deserialization such that we don't care so much about tasks changing package.

PR #2860, opened 2019-04-23
chore(gradle): Convert orca to use kork-bom

marchello2000:
* Use kork-bom
* Remove usages of servo
* Minor todo fixes along the way (around .gradle files)

Note: since `spinnaker-dependencies` isn't used anymore (all versioning is done via `kork-bom`) the tooling around automatically creating spinnaker-dependency bump will be temporarily broken. @cfieber has already addressed some of it, but full implementation will be when all services use `kork-bom` (which is actively being worked on)


gradle/kotlin.gradle
```diff
@@ -29,11 +29,12 @@ configurations.all {
...
-    languageVersion = "1.2"
+    languageVersion = "1.3"
     jvmTarget = "1.8"
   }
 }
 
-kapt {
-  mapDiagnosticLocations = true
-}
+//TODO: Ask Fletch
```
emjburns: paging @robfletcher 😆

marchello2000: haha, thanks, I forgot about it :)

PR #2861, opened 2019-04-24
refactor(headers): Update spinnaker headers

marchello2000:
Update the headers to match https://github.com/spinnaker/kork/pull/270

### DIFF
--- a/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingStageListener.groovy
+++ b/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingStageListener.groovy
@@ -24,6 +24,7 @@ import com.netflix.spinnaker.orca.pipeline.model.Stage
 import com.netflix.spinnaker.orca.pipeline.model.Task
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
 import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor
+import com.netflix.spinnaker.security.AuthenticatedRequest
 import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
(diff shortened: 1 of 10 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -305,7 +305,6 @@ class RunTaskHandler(
 
   private fun Stage.withLoggingContext(taskModel: com.netflix.spinnaker.orca.pipeline.model.Task, block: () -> Unit) {
     try {
-      MDC.put("application", this.execution.application)
```
marchello2000: this is now included automatically here https://github.com/spinnaker/orca/pull/2861/files#diff-574feb3cd5fd21d4373f7a819aa5a401R65

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingStageListener.groovy
```diff
@@ -127,12 +126,12 @@ class EchoNotifyingStageListener implements StageListener {
       }
 
       try {
-        MDC.put(SPINNAKER_EXECUTION_ID, stage.execution.id);
-        MDC.put(SPINNAKER_USER, stage.execution?.authentication?.user ?: "anonymous")
+        MDC.put(AuthenticatedRequest.SpinnakerHeader.EXECUTION_ID.header, stage.execution.id)
+        MDC.put(AuthenticatedRequest.SpinnakerHeader.USER.header, stage.execution?.authentication?.user ?: "anonymous")
```
ajordens: Unrelated to this PR ... but sure would be nice to swap that `stage.execution.authentication.user` out for the more appropriate `AuthenticationAware` logic (ie. look at ancestors for a `AuthenticatedStage` stage). 

I don't believe there's anything in `echo` that actually needs the correct authenticated user which would explain why we haven't run into an issue with this previously.

This PR looks good otherwise.

marchello2000: thanks. Let me look into it. Also, will pull in new kork now and update this PR

PR #2862, opened 2019-04-24
feat(core): ability to resolve targeted SG on different accounts and regions

Jammy-Louie:
spinnaker/spinnaker#4258

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/SourceResolver.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/SourceResolver.groovy
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.core.JsonParseException
 import com.fasterxml.jackson.databind.JsonMappingException
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.orca.clouddriver.OortService
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.Location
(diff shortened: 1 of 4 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/SourceResolver.groovy
```diff
@@ -43,8 +44,33 @@ class SourceResolver {
   StageData.Source getSource(Stage stage) throws RetrofitError, JsonParseException, JsonMappingException {
     def stageData = stage.mapTo(StageData)
     if (stageData.source) {
-      // has an existing source, return it
-      return stageData.source
+      // targeting a source in a different account and region
```
ajordens: Looks like we're really just trying to decide whether we're targeting a source cluster? (it may or may not be in a different account?)

If so, can we get away with:

```
if (stageData.source.clusterName && stage.context.target) {
   ...
}
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/StageData.groovy
```diff
@@ -113,6 +113,7 @@ class StageData {
     String serverGroupName
     Boolean useSourceCapacity
     Boolean preferSourceCapacity
+    String targetCluster
```
ajordens: I'd suggest `clusterName` vs `targetCluster` to fit with the existing convention of `serverGroupName`. 

I'm not sure that the `target` prefix adds particular meaning?

PR #2866, opened 2019-04-26
chore(conditions): Adding logging

jeyrschabu:
- added more logging around pausing deploys

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/conditions/EvaluateConditionTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/conditions/EvaluateConditionTask.java
@@ -82,7 +82,9 @@ public class EvaluateConditionTask implements RetryableTask {
     Duration backoff = Duration.ofMillis(conditionsConfigurationProperties.getBackoffWaitMs());
     Instant startTime = getStartTime(stage);
     Instant now = clock.instant();
-    if (ctx.getStatus() != null && startTime.plus(backoff).isAfter(now)) { // if status is null, then
+    if (ctx.getStatus() != null && startTime.plus(backoff).isAfter(now)) {
+      log.info("Paused deployment for execution {}, cluster",
+        stage.getExecution().getId(), ctx.getCluster());
       return new TaskResult(
         ExecutionStatus.RUNNING,
         Collections.singletonMap("status", Status.WAITING)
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/conditions/EvaluateConditionTask.java
```diff
@@ -82,7 +82,9 @@ public TaskResult execute(@Nonnull Stage stage) {
     Duration backoff = Duration.ofMillis(conditionsConfigurationProperties.getBackoffWaitMs());
     Instant startTime = getStartTime(stage);
     Instant now = clock.instant();
-    if (ctx.getStatus() != null && startTime.plus(backoff).isAfter(now)) { // if status is null, then
+    if (ctx.getStatus() != null && startTime.plus(backoff).isAfter(now)) {
+      log.info("Paused deployment for execution {}, cluster",
```
marchello2000: i think you're missing a param placeholder in the format string: 
```
log.info("Paused deployment for execution {}, cluster {}",
```

marchello2000: Also, this would log every 3 seconds for every execution today, since the backoff is 3s, right? probably a bit excessive

ajordens: Agree that it may be excessive ... 

Perhaps `log.debug("Deployment to {} has been conditionally paused (executionId: {})", ..., ...);`

jeyrschabu: @marchello2000 I did not see your comment. Yeah good catch. I will put that in the next PR

jeyrschabu: Like that format better too

jeyrschabu: @marchello2000 the task backoff is 3s but the there is an additional configured waitBackoff that is longer, so the log message wouldn't be printed every 3s

jeyrschabu:
feedback addressed in this PR: https://github.com/spinnaker/orca/pull/2867

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/conditions/EvaluateConditionTask.java
jeyrschabu: Actually, that would be every 3s. Made an incorrect assumption on stage startTime vs task startTime. That would be a subject of another PR.

PR #2867, opened 2019-04-26
chore(conditions): Adding better log messages

jeyrschabu:
-  update log messages


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/conditions/EvaluateConditionTask.java
```diff
@@ -83,8 +83,8 @@ public TaskResult execute(@Nonnull Stage stage) {
     Instant startTime = getStartTime(stage);
     Instant now = clock.instant();
     if (ctx.getStatus() != null && startTime.plus(backoff).isAfter(now)) {
-      log.info("Paused deployment for execution {}, cluster",
-        stage.getExecution().getId(), ctx.getCluster());
+      log.debug("Deployment to {} has been conditionally paused (executionId: {})",
```
marchello2000: Can't wait for https://github.com/spinnaker/kork/pull/270  ;)  to be merged and used so we don't have to include `executionId`s in our log messages.

PR #2868, opened 2019-04-26
feat(core): Delegate task/stage lookup to `TaskResolver` and `StageResolver` respectively

ajordens:
The `*Resolver` implementations will be able to look at both raw
implementation classes _as well as_ any specified aliases.

This supports (currently unsupported!) use cases that would be made
easier if a `Task` or `StageDefinitionBuilder` could be renamed.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/StageResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/StageResolver.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca;
+
+import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder;
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+import static java.lang.String.format;
+
+/**
+ * {@code StageResolver} allows for {@code StageDefinitionBuilder} retrieval via bean name or alias.
+ * <p>
+ * Aliases represent the previous bean names that a {@code StageDefinitionBuilder} registered as.
+ */
+public class StageResolver {
+  private final Map<String, StageDefinitionBuilder> stageDefinitionBuilderByAlias = new HashMap<>();
+
+  public StageResolver(Collection<StageDefinitionBuilder> stageDefinitionBuilders) {
+    for (StageDefinitionBuilder stageDefinitionBuilder : stageDefinitionBuilders) {
+      stageDefinitionBuilderByAlias.put(stageDefinitionBuilder.getType(), stageDefinitionBuilder);
+      for (String alias : stageDefinitionBuilder.aliases()) {
+        if (stageDefinitionBuilderByAlias.containsKey(alias)) {
+          throw new DuplicateStageAliasException(
+              format(
+                  "Duplicate stage alias detected (alias: %s, previous: %s, current: %s)",
+                  alias,
+                  stageDefinitionBuilderByAlias.get(alias).getClass().getCanonicalName(),
+                  stageDefinitionBuilder.getClass().getCanonicalName()
+              )
+          );
+        }
+
+        stageDefinitionBuilderByAlias.put(alias, stageDefinitionBuilder);
+      }
+    }
+  }
+
+  /**
+   * Fetch a {@code StageDefinitionBuilder} by {@param type} or {@param typeAlias}.
+   *
+   * @param type      StageDefinitionBuilder type
+   * @param typeAlias StageDefinitionBuilder alias (optional)
+   * @return the StageDefinitionBuilder matching {@param type} or {@param typeAlias}
+   * @throws NoSuchStageDefinitionBuilderException if StageDefinitionBuilder does not exist
+   */
+  @Nonnull
+  public StageDefinitionBuilder getStageDefinitionBuilder(@Nonnull String type, String typeAlias) {
+    StageDefinitionBuilder stageDefinitionBuilder = stageDefinitionBuilderByAlias.getOrDefault(
+        type, stageDefinitionBuilderByAlias.get(typeAlias)
+    );
+
+    if (stageDefinitionBuilder == null) {
+      throw new NoSuchStageDefinitionBuilderException(type, stageDefinitionBuilderByAlias.keySet());
+    }
+
+    return stageDefinitionBuilder;
+  }
+
+  class DuplicateStageAliasException extends IllegalStateException {
+    DuplicateStageAliasException(String message) {
+      super(message);
+    }
+  }
+
+  class NoSuchStageDefinitionBuilderException extends IllegalArgumentException {
+    NoSuchStageDefinitionBuilderException(String type, Collection<String> knownTypes) {
+      super(
+          format(
+              "No StageDefinitionBuilder implementation for %s found (knownTypes: %s)",
+              type,
+              String.join(",", knownTypes)
+          )
+      );
+    }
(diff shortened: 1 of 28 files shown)


ajordens:
Wanted to get some feedback on the general idea ... otherwise consider this a PoC.

orca-core/src/main/java/com/netflix/spinnaker/orca/StageResolver.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class StageResolver {
+  private final Collection<StageDefinitionBuilder> stageDefinitionBuilders;
+
+  public StageResolver(Collection<StageDefinitionBuilder> stageDefinitionBuilders) {
+    this.stageDefinitionBuilders = stageDefinitionBuilders;
+  }
+
+  public StageDefinitionBuilder getStageDefinitionBuilder(String type, String typeAlias) {
+    return stageDefinitionBuilders
```
ajordens: This is 100% lifted from the `StageDefinitionBuilderFactory` ... idea would be that a `StageDefinitionBuilder` could define 0..* aliases that would also be checked.

These aliases would reflect previous bean names for a particular `StageDefinitionBuilder` if it's ever been renamed.

orca-core/src/main/java/com/netflix/spinnaker/orca/Task.java
```diff
@@ -18,11 +18,15 @@
...
+import java.util.List;
 
 public interface Task {
   @Nonnull TaskResult execute(@Nonnull Stage stage);
 
   default void onTimeout(@Nonnull Stage stage) {}
 
   default void onCancel(@Nonnull Stage stage) {}
+
+  default List<String> aliases() { return Collections.emptyList(); }
```
ajordens: General idea is that a `Task` could be renamed and moved between packages.

These changes would be reflected as `aliases`.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/TestTaskResolver.kt
```diff
@@ -0,0 +1,27 @@
...
+
+package com.netflix.spinnaker.orca.q.handler
+
+import com.netflix.spinnaker.orca.Task
+import com.netflix.spinnaker.orca.TaskResolver
+
+
+class TestTaskResolver : TaskResolver(emptyList()) {
+  override fun getTaskClass(taskTypeIdentifier: String): Class<out Task> {
+    return Class.forName(taskTypeIdentifier) as Class<out Task>
```
ajordens: I'd actually probably just pull this up to the `TaskResolver` as a catch-all fallback to the original behavior.

The test cases are a little different because they're just using interfaces rather than actual classes, thus the `Class.forName()` is needed.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/migration/TaskTypeDeserializer.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.JsonDeserializer
+import com.fasterxml.jackson.databind.exc.InvalidFormatException
+import com.netflix.spinnaker.orca.Task
+import com.netflix.spinnaker.orca.TaskResolver
+import com.netflix.spinnaker.orca.pipeline.model.Execution
+
+class TaskTypeDeserializer(
```
ajordens: I don't yet know enough about deserializers to know if this is sufficient scoped (vs say implementing a deserializer for the entirety of a `RunTask` message)

orca-core/src/main/java/com/netflix/spinnaker/orca/StageResolver.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static java.util.stream.Collectors.toList;
+
+public class StageResolver {
```
robzienert: I'd like to see classes without an interface documented going forward. Not sure what level of detail we want to shoot for, though.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/orca/q/redis/migration/TaskTypeDeserializer.kt
robzienert: This should work just fine - I like it at this granularity.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/config/RedisOrcaQueueConfiguration.kt
```diff
@@ -41,13 +44,16 @@ import java.util.*
...
+  @Autowired
+  fun redisQueueObjectMapper(mapper: ObjectMapper,
+                             objectMapperSubtypeProperties: ObjectMapperSubtypeProperties,
+                             taskResolver: TaskResolver) {
     mapper.apply {
       registerModule(KotlinModule())
       registerModule(
         SimpleModule()
           .addDeserializer(ExecutionType::class.java, ExecutionTypeDeserializer())
+          .addDeserializer(Class::class.java, TaskTypeDeserializer(taskResolver))
```
robzienert: Did you mean to use `Class` here? Is this right?

orca-core/src/main/java/com/netflix/spinnaker/orca/TaskResolver.java
```diff
@@ -0,0 +1,48 @@
...
+  public TaskResolver(Collection<Task> tasks) {
+    for (Task task : tasks) {
+      taskByAlias.put(task.getClass().getCanonicalName(), task);
+      for (String alias: task.aliases()) {
+        taskByAlias.put(alias, task);
+      }
+    }
+  }
+
+  public Task getTask(String taskTypeIdentifier) {
```
robzienert: Thoughts on trying to enforce usage of `@Nonnull`, etc on args / returns within Java?

orca-core/src/main/java/com/netflix/spinnaker/orca/StageResolver.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  public StageDefinitionBuilder getStageDefinitionBuilder(String type, String typeAlias) {
+    StageDefinitionBuilder stageDefinitionBuilder = stageDefinitionBuilderByAlias.getOrDefault(
+        type, stageDefinitionBuilderByAlias.get(typeAlias)
+    );
+
+    if (stageDefinitionBuilder == null ) {
+      throw new ExecutionRunner.NoSuchStageDefinitionBuilder(type, stageDefinitionBuilderByAlias.keySet());
```
robzienert: 👍 🎉 specific exceptions

orca-core/src/main/java/com/netflix/spinnaker/orca/StageResolver.java
ajordens: I can get behind that, I'll document this interface.

orca-core/src/main/java/com/netflix/spinnaker/orca/TaskResolver.java
ajordens: I'm not in the habit of doing so but don't see the harm.

orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/config/RedisOrcaQueueConfiguration.kt
ajordens: It was working with this but I need to play around a bit more as it does look a bit strange.

ajordens:
I hadn't gotten around to duplicate aliases in the original PR but followed up here with some exceptions.

I'd rather fail early on duplicate alias usage as it's unexpected.

ajordens:
Took a clean up pass through this, believe it's in decent shape.

Would appreciate another pass @robzienert.

cfieber:
I like the idea, but I am concerned about the interplay between these resolvers and the application context as the source of stage builders and tasks.

I'd be tempted to ditch the resolvers dealing with instances and move them purely to class based (discover classes via scanning and annotation) and always go to the applicationContext to get an instance (this could be handled within the resolvers by giving them the applicationContext)

Otherwise we are in a potentially weird state if we have non-singleton beans or any other strange beanery.

ajordens:
`RunTaskHandler` already holds on to a set of all `Task` references so not sure any twiddling I'd do in these resolvers would materially change things downstream.

Inclined to merge as-is and circle around on beanery as that comes up.

PR #2869, opened 2019-04-26
 refactor(provider/kubernetes): Add tests and simplify cache refresh

ezimanyi:
* test(provider/kubernetes): Add tests to ManifestForceCacheRefresh 

  Most of the functionality in ManifestForceCacheRefreshTask is not tested. Add significant test coverage to this class to prepare for a bug fix/refactor.

* refactor(provider/kubernetes): Replace nested maps with an object 

  The ManifestForceCacheRefreshTask currently keeps track of its manifests as a Map<String, List<String>> which leads to a lot of complex iteration over this structure. Create a new class ScopedManifest and flatten this structure into a List<ScopedManifest> so that it's much easier to follow the processing that the class is doing.

* refactor(provider/kubernetes): Add account to ScopedManifest 

  Rather than thread account through multiple calls in this class, add it to the ScopedManifest class and set the account on each manifest when we create the initial list.

  This also makes the helper class Details identical to ScopedManifest, so replace instances of Details with ScopedManifest.

  Finally, replace pendingRefreshProcessed which both returns a status and mutates the stage context with getRefreshStatus, which only returns a status. Leave mutation of the context to the caller, checkPendingRefreshes. This allows us to avoid needing to mutate a ScopedManifest and keep the class immutable.

* refactor(provider/kubernetes): Track ScopedManifests directly 

  Now that we have a simple data class to represent a manifest to refresh that implements equals and hashCode, we don't need to manually serialize it with toManifestIdentifier. Just directly add these manifests to the collections we're using to track them.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
@@ -106,16 +106,15 @@ public class ManifestForceCacheRefreshTask extends AbstractCloudProviderAwareTas
     }
 
     String cloudProvider = getCloudProvider(stage);
-    String account = getCredentials(stage);
     StageData stageData = fromStage(stage);
     stageData.deployedManifests = getDeployedManifests(stage);
 
-    if (refreshManifests(cloudProvider, account, stageData)) {
+    if (refreshManifests(cloudProvider, stageData)) {
       registry.timer(durationTimerId.withTags("success", "true", "outcome", "complete"))
         .record(duration, TimeUnit.MILLISECONDS);
       return new TaskResult(SUCCEEDED, toContext(stageData));
     } else {
-      TaskResult taskResult = checkPendingRefreshes(cloudProvider, account, stageData, startTime);
+      TaskResult taskResult = checkPendingRefreshes(cloudProvider, stageData, startTime);
 
       // ignoring any non-success, non-failure statuses
       if (taskResult.getStatus().isSuccessful()) {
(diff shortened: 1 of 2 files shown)


ezimanyi:
The motivation behind this is that I need to make a subtle change to the force cache refresh task, so that it waits a polling cycle before checking if a cache refresh request exists, but wanted to first make the code more robust and tested.  That actual change will come in a separate (smaller) PR.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
```diff
@@ -204,69 +199,63 @@ private boolean pendingRefreshProcessed(PendingRefresh pendingRefresh, Set<Strin
...
-          }
-
-          stageData.getRefreshedManifests().add(id);
-        } catch (Exception e) {
-          log.warn("Failed to refresh {}: ", id, e);
-          allRefreshesSucceeded[0] = false;
-          stageData.errors.add(e.getMessage());
+    List<ScopedManifest> manifests = manifestsNeedingRefresh(stageData);
+
+    boolean allRefreshesSucceeded = true;
```
maggieneterval: I'm confused as to why `allRefreshesSucceeded` was previously `boolean[]`, if we were only every toggling the first element from true to false... :woman_shrugging:

ezimanyi: It was a workaround to the fact that you can only use a `final` or "effectively final" variable from within a lambda function.  So if we just used a `boolean` we'd need to set it to `final`, which of course means that we can't set it.  The workaround is to create a single-element array that is `final`...which makes the reference final, but still allows us to change the value it's pointing to.

I find this pattern pretty confusing (and also had no idea what it was doing the firs time I saw it)...removing it was one of the motivations behind my reworking some of this file.

maggieneterval: Whoah I never would have guessed that...thanks again for turning me reviewing your code into a java lesson for me!

PR #2871, opened 2019-04-26
fix(provider/kubernetes): Don't poll immediately after cache refresh

ezimanyi:
* test(provider/kubernetes): Re-order statements in tests 

  This commit has no functional effect, it is just going to make the tests much easier to read in the next commit, where I change the order between checking on pending cache refresh requests and sending new ones.

* refactor(provider/kubernetes): Change control flow in refresh 

  Instead of having refreshManifests and checkPendingRefreshes contain logic for determinig if the task is done, have them focus on mutating refreshedManifests and deployedManifests as they take actions.

  Then add allManifestsProcessed to check whether we're done; which is just checking if all deployed manifests have been processed. This removes the need to track the state of all manifests in the mutating functions and allows us to consolidate the return value of the task to one place.

* fix(provider/kubernetes): Don't poll immediately after cache refresh 

  We currently poll clouddriver to get the status of a cache refresh immediately after requesting the cache refresh, and schedule a re-refresh is we don't see the request we just requested.

  For users with a read-only clouddriver pointed at a replica of redis, there will be some replication lag before the pending refresh appears in the cache, which will cause us to keep re-scheduling the same cache refresh.

  To address this, wait one polling cycle before checking on the status of a pending refresh. Do this by changing the order of operations so that we first check on any pending requests from the last cycle, then schedule and needed new requests.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
@@ -108,9 +108,9 @@ public class ManifestForceCacheRefreshTask extends AbstractCloudProviderAwareTas
     StageData stageData = fromStage(stage);
     stageData.deployedManifests = getDeployedManifests(stage);
 
-    refreshManifests(cloudProvider, stageData);
-
     checkPendingRefreshes(cloudProvider, stageData, startTime);
+
+    refreshManifests(cloudProvider, stageData);
 
     if (allManifestsProcessed(stageData)) {
       registry.timer(durationTimerId.withTags("success", "true", "outcome", "complete"))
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
```diff
@@ -212,38 +214,29 @@ private RefreshStatus getRefreshStatus(PendingRefresh pendingRefresh, long start
       .collect(Collectors.toList());
   }
 
-  private boolean refreshManifests(String provider, StageData stageData) {
+  /**
+   * Requests and on-demand cache refresh for any manifest without an refresh requests that is either pending or
```
maggieneterval:
```suggestion
   * Requests an on-demand cache refresh for any manifest without any refresh requests that is either pending or
```

ezimanyi: :+1: , fixed

PR #2872, opened 2019-04-26
refactor(TaskResult): Add a TaskResultBuilder and use it everywhere

plumpy:


plumpy:
cc @robzienert @robfletcher 

I definitely don't know my way around orca yet, but the confusion between context and outputs wasn't helping me. I think this is a pretty safe change, though.

In particular, can I call your attention to the Javadoc I added to 'context' and 'outputs' in TaskResult? Are those descriptions accurate?

Thanks! Sorry for the 200-file code review.

plumpy:
Also, FYI, the majority of this was done with IntelliJ's Structural Search and Replace. It doesn't work with Kotlin files (?!) so I had to do those by hand as well as a few of the more esoteric Groovy cases.

PR #2877, opened 2019-04-29
chore(conditions): Adding metrics around deploy pauses

jeyrschabu:
- added a metric around deploy pauses


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/conditions/EvaluateConditionTask.java
```diff
@@ -120,6 +127,13 @@ public TaskResult execute(@Nonnull Stage stage) {
     }
   }
 
+  private void recordDeployPause(WaitForConditionContext ctx) {
+    registry.counter(
```
jeyrschabu: @dreynaud wondering if a gauge is more appropriate

dreynaud: lgtm like this with a counter! 👍

PR #2880, opened 2019-04-29
refactor(ci): Generify RetryableIgorTask

ezimanyi:
* refactor(ci): Generify RetryableIgorTask 

  I'd like to re-use the logic in RetryableIgorTask for some GCB tasks, but it requires that the stage map to a CIStageDefinition. Make the stage definition a parameter to the class so it can be re-used.

  At some point it might make sense to make this even more general than just for igor tasks, but at least this makes it a bit more general than it is now.


ezimanyi:
Added one more commit, as I realized this class wasn't properly handling network errors (as opposed to HTTP status errors), which we should also retry:

 * fix(ci): RetryableIgorTask should also retry on network errors 

  We're currently immediately looking up the response status, which will NPE (and fail the task) if the error is a network error. We should retry network errors as these are likely to succeed on retry.

PR #2881, opened 2019-04-29
feat(core): Allow `ExecutionPreprocessor` to discriminate on type

ajordens:
This supports existing use cases around pipeline-centric preprocessors.

### DIFF
--- a/orca-extensionpoint/src/main/java/com/netflix/spinnaker/orca/extensionpoint/pipeline/ExecutionPreprocessor.java
+++ b/orca-extensionpoint/src/main/java/com/netflix/spinnaker/orca/extensionpoint/pipeline/ExecutionPreprocessor.java
@@ -26,10 +26,15 @@ public interface ExecutionPreprocessor {
   /**
    * Returns whether or not the preprocess can handle the inbound execution.
    */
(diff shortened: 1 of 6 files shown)


orca-extensionpoint/src/main/java/com/netflix/spinnaker/orca/extensionpoint/pipeline/ExecutionPreprocessor.java
```diff
@@ -26,10 +26,15 @@
...
 
   /**
    * Allows modification of an execution configuration.
    */
   @Nonnull Map<String, Object> process(@Nonnull Map<String, Object> execution);
+
+  enum Type {
+    PIPELINE,
+    ORCHESTRATION
+  }
```
robzienert: I explicitly left a type off this to support cases where type isn't applicable. Maybe add an `ANY`?

ajordens: The PR passes in a `Type` at the call site so I think it's appropriate to be either `PIPELINE` or `ORCHESTRATION` (vs the preprocessor declaring what type it supports).

PR #2882, opened 2019-04-30
feat(gcb): Fetch artifacts produced by a GCB build

ezimanyi:
* refactor(gcb): Refactor monitor task to use generic retry logic

  Now that RetryableIgorTask is generic, MonitorGoogleCloudBuildTask can extend it instead of having its own retry logic.                  

* feat(gcb): Fetch artifacts produced by a GCB build

* fix(gcb): Add longer timeout to GCB polling stage

  We're currently using the default timeout of 1 minute, which will likely be too short.  As a starting point, use the timeout and backoff period we use for other CI jobs, though we can change these in the future if needed.

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/GoogleCloudBuildStageDefinition.java
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/GoogleCloudBuildStageDefinition.java
@@ -20,22 +20,26 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.Getter;
 
 import java.util.Map;
+import java.util.Optional;
 
 @Getter
-public class GoogleCloudBuildStageDefinition {
+public class GoogleCloudBuildStageDefinition implements RetryableStageDefinition {
   private final String account;
   private final GoogleCloudBuild buildInfo;
   private final Map<String, Object> buildDefinition;
+  private final int consecutiveErrors;
 
   // There does not seem to be a way to auto-generate a constructor using our current version of Lombok (1.16.20) that
   // Jackson can use to deserialize.
   public GoogleCloudBuildStageDefinition(
     @JsonProperty("account") String account,
     @JsonProperty("buildInfo") GoogleCloudBuild build,
-    @JsonProperty("buildDefinition") Map<String, Object> buildDefinition
+    @JsonProperty("buildDefinition") Map<String, Object> buildDefinition,
+    @JsonProperty("consecutiveErrors") Integer consecutiveErrors
   ) {
     this.account = account;
     this.buildInfo = build;
     this.buildDefinition = buildDefinition;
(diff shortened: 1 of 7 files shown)


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetGoogleCloudBuildArtifactsTask.java
```diff
@@ -0,0 +1,54 @@
...
+  private final IgorService igorService;
+
+  @Override
+  public @Nonnull TaskResult tryExecute(@Nonnull GoogleCloudBuildStageDefinition stageDefinition) {
+    List<Artifact> artifacts = igorService.getGoogleCloudBuildArtifacts(
+      stageDefinition.getAccount(),
+      stageDefinition.getBuildInfo().getId()
+    );
+    Map<String, List<Artifact>> outputs = Collections.singletonMap("artifacts", artifacts);
+    return new TaskResult(ExecutionStatus.SUCCEEDED, Collections.emptyMap(), outputs);
```
jtk54: What is the inlined `emptyMap()`? May want to inline a comment.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/GoogleCloudBuildStageDefinition.java
```diff
@@ -20,22 +20,26 @@
...
+  private final int consecutiveErrors;
 
   // There does not seem to be a way to auto-generate a constructor using our current version of Lombok (1.16.20) that
   // Jackson can use to deserialize.
   public GoogleCloudBuildStageDefinition(
     @JsonProperty("account") String account,
     @JsonProperty("buildInfo") GoogleCloudBuild build,
-    @JsonProperty("buildDefinition") Map<String, Object> buildDefinition
+    @JsonProperty("buildDefinition") Map<String, Object> buildDefinition,
+    @JsonProperty("consecutiveErrors") Integer consecutiveErrors
```
jtk54: Is there a particular reason this `consecutiveErrors` is `Integer` vs `int`?

ezimanyi: It's `Integer` in the constructor as it might be null if it's unset in the JSON; we convert to a primitive `int` before setting it on the object so consumers can just use a `int`.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetGoogleCloudBuildArtifactsTask.java
ezimanyi: This will get addressed by #2872 which is converting this to a `builder` so it will be clear what each parameter means.

PR #2885, opened 2019-04-30
chore(conditions): Adding a flag to skip wait

jeyrschabu:
- added config property tasks.evaluateCondition.skipWait
- when skipWait=true, all paused deployments will proceed

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/conditions/ConditionConfigurationProperties.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/conditions/ConditionConfigurationProperties.java
@@ -78,4 +78,8 @@ public class ConditionConfigurationProperties {
   public void setActiveConditions(List<String> activeConditions) {
     this.activeConditions = activeConditions;
   }
+
+  public boolean isSkipWait() {
+    return configService.getConfig(Boolean.class, "tasks.evaluateCondition.skipWait", false);
+  }
 }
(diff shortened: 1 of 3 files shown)


jeyrschabu:
`asks.evaluateCondition.skipWait=true` allows an operator to effectively un-pause deployments.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/conditions/ConditionConfigurationProperties.java
```diff
@@ -78,4 +79,12 @@ public void setClusters(List<String> clusters) {
   public void setActiveConditions(List<String> activeConditions) {
     this.activeConditions = activeConditions;
   }
+
+  public boolean isSkipWait() {
+    return configService.getConfig(Boolean.class, "tasks.evaluateCondition.skipWait", skipWait);
+  }
+
+  public void setSkipWait(boolean skipWait) {
+    this.skipWait = skipWait;
```
dreynaud: if the only use of this is for testing, I would prefer this style:

```
// remove the skipWait field and setSkipWait method
  public boolean isSkipWait() {
    return configService.getConfig(Boolean.class, "tasks.evaluateCondition.skipWait", false);
  }

// in the test, mock the response with
1 * conditionsConfigurationProperties.isSkipWait() >> true
```

jeyrschabu: Reasonable

PR #2886, opened 2019-04-30
chore(cf): Move CF specific code to its own module

eleftherias:
spinnaker/spinnaker#4340


ajordens:
Fan of the effort but changing the package names for `Tasks` is going to cause a problem when deserializing (because message handlers in `orca` hold on to a `Class<Task>` reference.

We recently added some support for `Task` and `StageDefinitionBuilder` to define aliases with the idea of being able to rename/repackage implementations.

```
@Task.Aliases(
  "com.netflix.spinnaker.orca.clouddriver.tasks.providers.cf.CloudFoundryDeployServiceTask"
)
```

We're using in one spot internally so far.

ajordens:
Alternative would be to shift the code and leave the renaming/repackaging for a subsequent PR.

jmelchio:
@ajordens we'll take a look at this to make sure we fully understand the issue.

Jammy-Louie:
closing for now. will review in the future

PR #2888, opened 2019-04-30
refactor(logging): improve logging for AbstractWaitForClusterWideClou…

marchello2000:
…ddriverTask

Add specificity to the log message so you can tell which task is actually waiting for something to happen.
So.. instead of:
```
Pipeline 01D9QSYBNK7ENG0BSM7V9V07ZZ is looking for [us-east-1->serverlabmvulfson-dev-v044]
Server groups matching AbstractWaitForClusterWideClouddriverTask$DeployServerGroup(region:us-east-1, name:serverlabmvulfson-dev-v044) ...
```
we will have:
```
Pipeline 01D9R3HMZHZ82H0RNKG1D8JBRW:WaitForClusterDisableTask looking for server groups: [us-east-1->serverlabmvulfson-dev-v047] found: [[instances...
```


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/AbstractWaitForClusterWideClouddriverTask.groovy
```diff
@@ -68,7 +68,8 @@ abstract class AbstractWaitForClusterWideClouddriverTask extends AbstractCloudPr
       def isMatch = it.region == deployServerGroup.region && it.name == deployServerGroup.name
       isMatch
     })
-    log.info "Server groups matching $deployServerGroup : $matchingServerGroups"
+
+    log.info("${this.getClass().getSimpleName()}: Server groups matching $deployServerGroup : $matchingServerGroups")
```
ajordens: Curious ... was the default `logger_name` wrong and is that a byproduct of using `@Slf4j` that could have been solved by explicitly doing a `LoggerFactory.getLogger(getClass())`?

It'd be nice to not litter log messages with the classes names when they're also represented in the logger context.

marchello2000: yeah, it was always AbstractWaitForClusterWideClouddriverTask, let me use the `getLogger(getClass())` that's way better, thanks!

PR #2889, opened 2019-05-01
feat(kayenta): pass the accountId to Kayenta for deployments

skandragon:
Things that did not change:

* For non-deployments, all scope data is passed through unchanged.

Changes:
* Attempt to retrieve the accountId from Mort for the deployment we have just made, if any.
* If it is present, set it in the `extendedScopeParams` under the key `accountId` for both control and experiment.
* Stop doing a very late override in the stage that runs the canary report; instead, this stage will now pass along exactly the scopes provided in its context.
* This pushes all the heavy lifting of "is this a deployment or not" as well as temporal planning to the `RunCanaryIntervalsStage`.


skandragon:
@robfletcher, I added you as this is Kotlin and you wrote the original.  Please comment on anything, but especially the areas where I could Kotlinize things better.

@dufter, please pay special attention to the accountId fetching from Mort and its subsequent handling.

orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/pipeline/RunCanaryIntervalsStage.kt
```diff
@@ -46,6 +47,18 @@ class RunCanaryIntervalsStage(private val clock: Clock) : StageDefinitionBuilder
   override fun taskGraph(stage: Stage, builder: TaskNode.Builder) {
   }
 
+  private fun getDeployDetails(stage: Stage) : DeployedServerGroupContext? {
+    val deployedServerGroupsStage = stage.parent?.execution?.stages?.find {
```
ajordens: IMO, you'll want to use `StageNavigator` and not just look for the first matching stage in the execution.

The navigator will look up the graph and allow you to grab the "closest" stage.

skandragon: I'll look into this, thanks

ajordens: Scratch that comment ... misunderstanding around the shape of this graph.

PR #2892, opened 2019-05-02
fix(logs): Allow anonymous calls

marchello2000:
Explicitly allow anonymous calls for dependent pipeline starter querying front50
and ephemeral cluster poller.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pollers/EphemeralServerGroupsPoller.java
```diff
@@ -146,36 +146,40 @@ protected void tick() {
...
-
-          return ephemeralServerGroupTag;
-        })
-        .findFirst()
-        .orElse(null)
-      )
-      .filter(Objects::nonNull)
-      .filter(t -> t.expiry.isBefore(ZonedDateTime.now(ZoneOffset.UTC)))
-      .collect(Collectors.toList());
+    try {
```
marchello2000: i did something funky in kork that we have to catch these exceptions. i will fix it separately

PR #2894, opened 2019-05-03
fix(logging): use logger.GetFactory over @slf4j

marchello2000:
This gets the proper class/logger name instead of using the parent `AbstractWaitForClusterWideClouddriverTask`
fixup for: https://github.com/spinnaker/orca/pull/2888


marchello2000:
Thanks, @ajordens  - works well

ajordens:
Great, thanks for the followup!

PR #2895, opened 2019-05-06
feat(injectArtifact): inject artifacts stage

ethanfrogers:
the `InjectArtifactStage` fetches an artifact, parses it and returns it
to the stage context. this can be useful for when jobs produce artifacts
that aren't deployable but may influence the flow of a pipeline or have
data that can be used in other downstream stages.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/artifacts/InjectArtifactStage.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/artifacts/InjectArtifactStage.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2019 Armory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline.artifacts;
+
+import com.netflix.spinnaker.orca.clouddriver.tasks.artifacts.InjectArtifactTask;
+import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder;
+import com.netflix.spinnaker.orca.pipeline.TaskNode;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import org.springframework.stereotype.Component;
+
+@Component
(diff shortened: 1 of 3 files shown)


ethanfrogers:
cc @bpowell

ezimanyi:
I'm not sure I fully understand the expected use case for this stage.  One of the main purposes of artifacts was that there was a well-defined way of passing information between stages, replacing the need for tight-coupling between stages where stage A puts something in the context in a pre-arranged location then stage B reads.  In general, the pattern here would be that the downstream stage that needs information from an artifact should be responsible for reading the artifact and acting accordingly.

If there's a specific use case for this stage, I'd be interested in digging into that more to see if there's a way we can meet that use case without needing to drop an entire artifact into the stage context.

brianwolfe:
On our side, we basically want to have a mechanism for a job to return a few lightweight variables while avoiding the parsing of logs. I'm concerned about clouddriver pulling logs in addition to its other responsibilities. Clouddriver performance is limiting us right now, adding log pulling and parsing to clouddriver seems like an unnecessary risk when we actually just want to be able to extract a small JSON packet.

By putting it into a custom artifact type, I thought we could explicitly limit the artifact size on upload and download without interfering with developer's use of logs to debug their jobs. The usage of an artifact in our use-case would be to protect clouddriver. It's easier to understand a hard limit on file size than to specify a limit on the number of log lines processed or equivalent.

The particular pattern that I'm seeing is that wanted to be able to tweak the behavior of a particular scaleManifest or deployManifest stage. This came up when selecting/baking the correct artifact to deploy to a canary baseline, for instance. This came in the form of a job that determined exactly what artifact we should be referencing in a downstream stage. For our case, we weren't creating the artifact in that stage, we just were just de-referencing it. Maybe it would be easier if we were to just create it and specify that the job must output the artifact.

In another case, a job was creating a new asset that returned a UUID. The logs were just fairly voluminous to let the developer debug problems, but the UUID was pretty lightweight. We then had to pass the UUID to a downstream job. This use-case was a little more complex, I'll have to think about if there's a way to map it to Spinnaker more easily.

ezimanyi:
I just re-read the design doc, and see a bit better now how this fits in.  I think my main concern here is around having a new "Inject Artifact" stage rather than have this be part of the "Run Job" stage.

For the particular case of running a job, it seems reasonable to have the job put some structured output in a file somewhere and have Spinnaker read that in (similar to how it would read a specified file from Jenkins to inject properties).

I'm less sold on wanting to surface this as a top-level stage.  I see a fair amount of confusion around artifact workflows, and worry that adding a new top-level stage for this particular use case will add more confusion than the benefit it provides.  I'd love to chat in a bit more detail about the specific use case for wanting to inject an artifact that wasn't a job into the pipeline context---is this something that is know prior to the pipeline starting and could be solved by having pipeline parameters?

If there is a way to solve that use case without a "consume artifact" stage then it might be better to just have this be a task in the "Run job" stage.  Though if after thinking about it in detail there really isn't another way of solving that use case, then this might be reasonable.

ethanfrogers:
That's fair. I'm all for less confusion around artifacts. I hadn't considered simply adding this to the end of Run Job. I'll do a little experimenting to see how easily we can do that.

As for usecases _outside_ of Run Job - perhaps we can have @brianwolfe discuss their needs at the next Kubernetes SIG.

PR #2903, opened 2019-05-09
feat(runJob/consumeArtifact): consume artifacts

ethanfrogers:
At the end of the run job stage, if there is an artifact to be consumed,
consume it and dump it's output into the stage outputs. for kubernetes,
this can be used as an alternative way to consume data produced by the
job.


ethanfrogers:
@ezimanyi @brianwolfe took a different approach to for jobs to consume artifacts. This is definitely better than having a dedicated stage.

ethanfrogers:
## Configure Artifact



## Configuring Consume Artifact


## Execution Details


## Result

PR #2907, opened 2019-05-09
chore(*): Rename time.kt to timeutil.kt to keep intelliJ happy

marchello2000:


orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/ExecutionLatch.kt
```diff
@@ -43,7 +43,7 @@ class ExecutionLatch(private val predicate: Predicate<ExecutionComplete>)
     }
   }
 
-  fun await() = latch.await(1, TimeUnit.SECONDS)
+  fun await() = latch.await(2, TimeUnit.SECONDS)
```
ezimanyi: Thank you!  I was just trying to figure out why that test was intermittently failing.

marchello2000: no problem. I still don't know why travis has troubles with this test - 1s should be plenty... but we just decided to bump it 🤞

PR #2909, opened 2019-05-10
New endpoint to list correlated executions

robfletcher:
I want to avoid writing redundant resource history in Keel so rather than blindly submitting tasks and relying on Orca to de-dupe using the correlation id I'd like to first see if there are any in-flight tasks and if so do nothing in Keel.

### DIFF
--- a/gradle/groovy.gradle
+++ b/gradle/groovy.gradle
@@ -26,4 +26,5 @@ dependencies {
   testImplementation("cglib:cglib-nodep")
   testImplementation("org.objenesis:objenesis")
   testImplementation("org.codehaus.groovy:groovy-all")
+  testRuntimeOnly("org.junit.vintage:junit-vintage-engine")
 }
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/CorrelatedTasksController.java
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/CorrelatedTasksController.java
@@ -1,0 +1,51 @@
+package com.netflix.spinnaker.orca.controllers;
+
+import java.util.List;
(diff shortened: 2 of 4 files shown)


gradle/groovy.gradle
```diff
@@ -26,4 +26,5 @@ dependencies {
   testImplementation("cglib:cglib-nodep")
   testImplementation("org.objenesis:objenesis")
   testImplementation("org.codehaus.groovy:groovy-all")
+  testRuntimeOnly("org.junit.vintage:junit-vintage-engine")
```
robfletcher: Spock needs this if using JUnit 5 to run tests in Gradle

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/CorrelatedTasksController.java
```diff
@@ -0,0 +1,51 @@
...
+      .<Execution>builder()
+      .add(getCorrelated(correlationId, PIPELINE))
+      .add(getCorrelated(correlationId, ORCHESTRATION))
+      .build()
+      .filter(Objects::nonNull)
+      .map(Execution::getId)
+      .collect(toList());
+  }
+
+  private Execution getCorrelated(String correlationId, ExecutionType pipeline) {
```
marchello2000: nit: probably more understandable to rename `pipeline` -> `executionType`
```suggestion
  private Execution getCorrelated(String correlationId, ExecutionType executionType) {
```

robfletcher: D'oh, product of using extract method, etc. in IntelliJ. Fixed.

PR #2911, opened 2019-05-13
feat(cloudformation): add stack as output of CFN stage

xavileon:
The outputs of a CFN stack are available in the stack object, however we
were not exposing them as the result of the CFN stage. This is usefull
if one stack depends on the outputs of another stack, so we can chain
them.

This patch returns the stack object as a result of a successfull stage,
so it's available in the context for later stages.


xavileon:
@ethanfrogers you may be interested?

xavileon:
@spinnaker/reviewers PTAL

xavileon:
@spinnaker/reviewers is this ok to be merged?

PR #2915, opened 2019-05-13
chore(*): Apply codestyles to orca

ezimanyi:
* chore(*): Fix a few things that break the code formatter 

  * Remove wildcard imports from kotlin files
  * Fix a single line that was intended by an odd number of spaces

* chore(*): Apply codestyles to orca 

  Run ./gradlew spotlessApply on the entire project

* chore(*): Turn on codestyle enforcement


ezimanyi:
Cherry-picks will be much easier if we do this before rather than after cutting the 1.14 branch.

PR #2917, opened 2019-05-16
chore(clouddriver): Use kork's `selectableService`

jeyrschabu:
- Remove duplicated code for `selectableService`


jeyrschabu:
When defining clouddriver shards configs should be updated to use kork's package:

From: `selectorClass: com.netflix.spinnaker.orca.clouddriver.config.<SelectorName>`
To: `selectorClass: com.netflix.spinnaker.kork.web.selector.<SelectorName>`

PR #2918, opened 2019-05-16
fix(MPTv1): Fix template expansion

marchello2000:
With kork-bom, we have picked up a new version of jinjava (despite the force=true
directive to gradle which doesn't work in bom scenarios where first one wins)
Anyway, old jinjava would handle missing variables by evaluating the expression to false;
new jinjava throws.
When a stage has a `{when}` condition that uses a variable that was part of the stage config
that variable doesn't get added to the context until we process the stage the second time!
Add those variables to the context on the first time round to make template expansion happy


marchello2000:
going to merge this as it's blocking a few folks, please give feedback and I will address it

PR #2919, opened 2019-05-16
fix(executionprocessor): add @Order to execution processors

marchello2000:
This allows an execution processor to specify order of invocation.
This is useful (/ needed) when an preprocessor needs to operate on a
fully instantiated pipeline. If the pipeline is templated it won't be
fully instantiated till the templatepreprocessor runs.


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePreprocessor.kt
```diff
@@ -27,11 +27,13 @@ import com.netflix.spinnaker.orca.pipelinetemplate.handler.SchemaVersionHandler
 import com.netflix.spinnaker.orca.pipelinetemplate.v2schema.model.V2PipelineTemplate
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.core.annotation.Order
 import org.springframework.stereotype.Component
 import javax.annotation.Nonnull
 import javax.annotation.PostConstruct
 
 @Component("pipelineTemplatePreprocessor")
+@Order(1)
```
robzienert: I don't think we want orders to be the same. Sounds like a neat recipe for indeterminate behavior.

marchello2000: it's what it is now anyway, but yeah, i can change it

PR #2920, opened 2019-05-16
fix(boot2): camel to kebab for fast properties

emjburns:
Looks like boot2 is not happy with the dynamic spring properties being camel case. Switching to kebab.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/conditions/ConditionConfigurationProperties.java
```diff
@@ -82,6 +82,6 @@ public void setActiveConditions(List<String> activeConditions) {
   }
 
   public boolean isSkipWait() {
-    return configService.getConfig(Boolean.class, "tasks.evaluateCondition.skipWait", false);
+    return configService.getConfig(Boolean.class, "tasks.evaluate-condition.skipWait", false);
```
cfieber: `skip-wait` too

PR #2921, opened 2019-05-16
fix(mptv1): revert back Jinjava to 2.2.3 

marchello2000:
Also, making some related calls anonymous to not be noisy
Also, had to pull specific version of guava since kork doesn't specify it


orca-pipelinetemplate/orca-pipelinetemplate.gradle
```diff
@@ -8,7 +8,10 @@ dependencies {
   implementation(project(":orca-front50"))
   implementation(project(":orca-clouddriver"))
 
-  implementation("com.hubspot.jinjava:jinjava")
+  implementation("com.hubspot.jinjava:jinjava:2.2.3") {
```
marchello2000: oh, darn, i need to remove this

PR #2924, opened 2019-05-17
perf(sql): Add indexes on correlated execution ids for faster deletes

robzienert:


orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -195,8 +195,19 @@ class SqlExecutionRepository(
     jooq.transactional {
       val correlationField = if (type == PIPELINE) "pipeline_id" else "orchestration_id"
       val (ulid, _) = mapLegacyId(it, type.tableName, id)
-      it.delete(table("correlation_ids")).where(field(correlationField).eq(ulid)).execute()
-      it.delete(type.stagesTableName).where(field("execution_id").eq(ulid)).execute()
+      it.delete(table("correlation_ids")).where(field("id").eq(
+        it.select(field("id")).from("correlation_ids")
```
asher: I think the lists of `correlation_ids.id` and `pipeline|orchestration_stages.ids` should be fetched as individual queries outside of the transactional, followed by a transaction that deletes from the 3 tables by primary key with a single `execute()`.

PR #2926, opened 2019-05-20
fix(bake/rosco): fix regression in snake casing of rosco specific bake properties

cfieber:

PR #2927, opened 2019-05-20
fix(bake/rosco): fix regression in snake casing of rosco specific bake properties

cfieber:
cherry-pick of #2926 with merge conflicts fixed


cfieber:
@maggieneterval not sure what the process is when the spinnakerbot cherry-pick fails but this ports #2926 from master for release-1.14.x

PR #2929, opened 2019-05-21
Allow min/max/desired to be strings (cos expressions)

robfletcher:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForUpInstancesTask.groovy
```diff
@@ -120,7 +119,7 @@ class WaitForUpInstancesTask extends AbstractWaitingForInstancesTask {
     Integer targetDesiredSize
 
     if (useConfiguredCapacity(stage, currentCapacity)) {
-      targetDesiredSize = ((Map<String, Integer>) stage.context.capacity).desired
+      targetDesiredSize = ((Map<String, Integer>) stage.context.capacity).desired as Integer
```
ajordens: Doh strings as numbers :(

Alternative to this would be to stage.mapTo("capacity", some concrete capacity object), we're doing that in a handful of spots.

robfletcher: That is ultimately the right way to do it, yes.

PR #2933, opened 2019-05-22
fix(kubernetes): force cache refresh after deploy stage artifact cleanup

maggieneterval:
Addresses: https://github.com/spinnaker/spinnaker/issues/4422
Related PR: https://github.com/spinnaker/clouddriver/pull/3711

When using Kubernetes V2 Rollout Strategies with the `max-version-history` annotation, Clouddriver sometimes tries to disable a previously-deleted ReplicaSet because we are not refreshing the cache after the Deploy Manifest stage's `CleanupArtifactsTask`. Adds `MonitorKatoTask`, `PromoteManifestKatoOutputsTask`, and `ManifestForceCacheRefreshTask` in order to prevent this.

Previously, each stage had at most one `PromoteManifestKatoOutputsTask`, which composes the outputs of each kato task's outputs by taking the *first* non-null output value at each relevant key. In this case, our cleanup task's `outputs.manifestNamesByNamespace` was never being surfaced because this task would only surface the first kato task's (the Deploy task's) `outputs.manifestNamesByNamespace`. This adds a `manifestNamesByNamespaceToRefresh` key, which will be populated by the most recent task's `manifestNamesByNamespace` and cleared at every run of `ManifestForceCacheRefreshTask`. Continues to also check `manifestNamesByNamespace` when finding manifests to refresh because not all `ManifestForceCacheRefreshTask`s are preceded by `PromoteManifestKatoOutputsTask`s.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestAware.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestAware.java
@@ -27,7 +27,12 @@ import java.util.logging.Logger;
 public interface ManifestAware {
   default Map<String, List<String>> manifestNamesByNamespace(Stage stage) {
     Map<String, List<String>> result =
-        (Map<String, List<String>>) stage.getContext().get("outputs.manifestNamesByNamespace");
+        (Map<String, List<String>>)
+            stage
+                .getContext()
+                .get(
+                    PromoteManifestKatoOutputsTask.outputKey(
+                        PromoteManifestKatoOutputsTask.MANIFESTS_BY_NAMESPACE_KEY));
     if (result != null) {
       return result;
     }
@@ -43,5 +48,23 @@ public interface ManifestAware {
     }
 
     return result;
+  }
+
+  default Map<String, List<String>> manifestsToRefresh(Stage stage) {
+    Map<String, List<String>> result =
+        (Map<String, List<String>>)
+            stage
+                .getContext()
+                .get(PromoteManifestKatoOutputsTask.MANIFESTS_BY_NAMESPACE_TO_REFRESH_KEY);
+    if (result != null
+        && (boolean)
+            stage
+                .getContext()
+                .get(
+                    PromoteManifestKatoOutputsTask
+                        .SHOULD_REFRESH_MANIFESTS_BY_NAMESPACE_TO_REFRESH_KEY)) {
+      return result;
+    }
+    return manifestNamesByNamespace(stage);
   }
 }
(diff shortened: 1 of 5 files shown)


maggieneterval:
@ezimanyi ready for another look after the holiday weekend, though not really sure this approach is any better :see_no_evil:

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
```diff
@@ -288,6 +288,8 @@ private StageData fromStage(Stage stage) {
     Set<ScopedManifest> processedManifests = new HashSet<>();
 
     Set<String> errors = new HashSet<>();
+
+    Map<String, List<String>> manifestNamesByNamespaceToRefresh = null;
```
maggieneterval: Ugh sorry I'm breaking the tests, I guess I can't just set this to null here...do you know if there is a reason this task is using a static class and its own special `fromStage` method instead of `stage.mapTo`? Will fix this up on Tuesday :sob:

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestAware.java
```diff
@@ -26,8 +26,29 @@
 
 public interface ManifestAware {
   default Map<String, List<String>> manifestNamesByNamespace(Stage stage) {
-    Map<String, List<String>> result =
-        (Map<String, List<String>>) stage.getContext().get("outputs.manifestNamesByNamespace");
+    return manifestNamesByNamespace(stage, false);
+  }
+
+  default Map<String, List<String>> manifestNamesByNamespace(Stage stage, Boolean isRefresh) {
```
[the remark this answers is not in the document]
maggieneterval: Ah good catch, thanks for pointing out anyway!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestForceCacheRefreshTask.java
```diff
@@ -112,6 +112,9 @@ public TaskResult execute(@Nonnull Stage stage) {
     StageData stageData = fromStage(stage);
     stageData.deployedManifests = getDeployedManifests(stage);
 
+    // Clear manifests to be refreshed
+    stageData.manifestNamesByNamespaceToRefresh = null;
```
maggieneterval: Okay I'm a dummy, this is breaking the tests because members of an immutable map can't be null, and the outputs of a task are an immutable map:
https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/TaskResult.java#L51
Going to sleep on this one, we'll need a different way to somehow mark this manifest as refreshed. Otherwise, when we have a subsequent disable stage as part of red/black, we'll be trying to refresh any manifests removed as part of the cleanup rather than the manifest to disable (because enable/disable force cache refresh tasks are not preceded by the promote outputs task that adds the deployed manifests to the refresh key).

maggieneterval: @ezimanyi since we cannot null out context keys, I added a flag, `shouldRefreshManifestNamesByNamespaceToRefresh`, which is set to true in the `PromoteManifestKatoOutputsTask` when a task produces manifest outputs, and then set to false in the `ManifestForceCacheRefreshTask` when those manifests are refreshed in the cache. While kind of disgusting, this should keep this change fully backwards compatible.

maggieneterval:

PR #2936, opened 2019-05-24
fix(config): fix bakery port in orca-web/config

zqfan:
rosco service listens port 8087, 8089 is listened by echo instead


ezimanyi:
Thanks, that looks like it was a typo---though I think in essentially all cases this is going to be overridden by a custom `orca.yml` (whether halyard-generated or manually generated), which is likely why this hasn't caused any issues.

PR #2937, opened 2019-05-24
perf(orca): reuse ObjectMapper instances

mattnworb:
creating a new ObjectMapper instance for each Stage or other per-request
contexts can be wasteful; ObjectMapper is safe to reuse across threads
(after it is configured). See https://github.com/spinnaker/spinnaker/issues/4367 for more details.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
@@ -33,7 +33,7 @@ import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.
  */
 trait DeploymentDetailsAware {
 
-  private ObjectMapper pipelineObjectMapper = OrcaObjectMapper.newInstance()
+  private ObjectMapper pipelineObjectMapper = OrcaObjectMapper.getInstance()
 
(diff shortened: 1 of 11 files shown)


robfletcher:
Are there any places where the `ObjectMapper` is customized for particular use cases? Just wondering why we had it as a factory in the first place.

mattnworb:
@robzienert I did find a few places in the Orca codebase where the new ObjectMapper instance returned by `OrcaObjectMapper.newInstance()` is mutated / configured in different ways:

https://github.com/spinnaker/orca/blob/cd6b6cad218757a735e4b16d983127c2bc345a58/orca-integrations-gremlin/src/main/java/com/netflix/spinnaker/orca/config/GremlinConfiguration.kt#L41-L45

https://github.com/spinnaker/orca/blob/cd6b6cad218757a735e4b16d983127c2bc345a58/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/config/KeelConfiguration.kt#L62-L66

https://github.com/spinnaker/orca/blob/2b9de2785bdfaacba2f7b94d2c55aeed9ad74c9e/orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/config/KayentaConfiguration.kt#L67-L69

I left all of those unchanged in this PR, since from what I could tell, those call sites were not creating one ObjectMapper per request / unit of work etc. (and plus they seem to have legitimate reasons to configure the ObjectMapper instance differently than other regions of code).

PR #2938, opened 2019-05-24
perf(artifacts): only load prior artifacts when needed

mattnworb:
When a pipeline is triggered via `POST /orchestrate`,
OperationsController will resolve artifacts for the pipeline. The
behavior of `ArtifactResolver.resolveArtifacts(Map)` will load all
previous executions of the pipeline in order to find prior artifacts of
the pipeline (from the most recent execution), even if the
`priorArtifacts` list is never accessed (because the pipeline trigger
contains all expected artifacts).

This change tries to avoid the unnecessary loading of previous
executions / prior artifacts by wrapping the calls to
getArtifactsForPipelineId() etc in a `Supplier` so that the data (which
could be voluminous if old pipeline execution cleanup is not enabled) is
only loaded when truly needed.

To make the change a little bit smaller, I refactored the signature for
the public `resolveSingleArtifact` method to remove the `priorArtifacts`
parameter which was always null in all call sites.

See https://github.com/spinnaker/spinnaker/issues/4367 for more details.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
@@ -59,7 +59,7 @@ public class FindArtifactFromExecutionTask implements Task {
         artifactResolver.getArtifactsForPipelineId(pipeline, executionOptions.toCriteria());
 
     Artifact match =
-        artifactResolver.resolveSingleArtifact(expectedArtifact, priorArtifacts, null, false);
+        artifactResolver.resolveSingleArtifact(expectedArtifact, priorArtifacts, false);
 
     if (match == null) {
       outputs.put(
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
@@ -35,7 +35,6 @@ import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository.Execu
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -47,7 +46,6 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import lombok.Data;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -393,12 +391,6 @@ public class ArtifactResolver {
     ArtifactResolutionException(String message, Throwable cause) {
       super(message, cause);
     }
-  }
-
-  @Data
-  public static class ResolveResult {
-    Set<Artifact> resolvedArtifacts = new HashSet<>();
-    Set<ExpectedArtifact> unresolvedExpectedArtifacts = new HashSet<>();
   }
 
   private static Func2<Execution, Execution, Integer> startTimeOrId =
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
@@ -23,6 +23,7 @@ import static org.apache.commons.lang3.StringUtils.isEmpty;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Suppliers;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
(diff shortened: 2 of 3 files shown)


mattnworb:
Using a memoizing `Supplier` from Guava here seemed like an option for changing the behavior while minimizing the amount of code to change, since these methods are called from a few other places within Orca, and to deal with the fact that the `priorArtifacts` are read a few method calls down from where they are first loaded - but this trades off complexity with size of the diff.

I'm not sure on Spinnaker's stance on using Guava, and a memoizing Supplier might seem too confusing - I would be happy to refactor this to be cleaner if desired.

ezimanyi:
@mattnworb : I just read through your analysis on spinnaker/spinnaker#4367---great find on the loading of all executions into memory during artifact resolution.

I think there may be a way to implement your #2 solution---avoiding the need to fetch and hydrate every pipeline.  The `ExecutionCriteria` class has a `pageSize` argument that (if I'm reading correctly) will actually propagate all the way down to redis and limit the returned results to the `pageSize`.  We store the executions in redis as a sorted set (based on `execution.buildTime`) which is what allows an efficient retrieval of the N most recent executions.  (I don't think that the change from `startTime` sorting to `buildTime` sorting would be significant, though we'd want to think a bit more about that.)

If that #2 solution does indeed work, it probably makes sense to go with that solution (which would also help anyone who does rely on prior artifacts).  In that case it may not be worth the complexity to defer fetching the latest execution as fetching a single execution is probably negligible.

Let me know your thoughts---thanks!

mattnworb:
@ezimanyi I think that is a great idea and you are spot on that reducing the amount of executions loaded in that step should resolve my original concern and make this PR redundant. I can take a crack at this in a new PR.

I agree there would be implications to change through how the list of executions is sorted in order for ArtifactResolver to get the most recent execution. Perhaps that question could be sidestepped though if a new sorted set was added to track the executions based on `startTime`, and then a new method could be added to ExecutionRepository to get the most recent execution - how would that sound?

mattnworb:
Related to the above point on sorting by `Execution.getStartTime()` versus `Execution.getBuildTime()` (how the IDs are scored in the sorted set `pipeline:executions:<pipelineId>` in Redis):

I'm having trouble finding where in the Orca codebase that the `startTime` is ever set - `buildTime` appears to be set to the current timestamp when the execution is launched in https://github.com/spinnaker/orca/blob/c5b55476e094b91258024c1e1c8bc96210801764/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/PipelineBuilder.java#L78, but I can't find where `startTime` is set - any pointers on if I am missing something?

ezimanyi:
It looks like setting the start time is actually handled by the repository-specific code.

Redis:
https://github.com/spinnaker/orca/blob/c5b55476e094b91258024c1e1c8bc96210801764/orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java#L301

SQL:
https://github.com/spinnaker/orca/blob/c5b55476e094b91258024c1e1c8bc96210801764/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt#L179

ezimanyi:
I'd hesitate to  add another zrange on `startTime` (which would add more complexity and also would potentially require a backfill for existing pipelines) unless we're really sure that we need to sort by `startTime` rather than `buildTime`.

I think the only time that this could make a difference is if somehow there were two pipelines queued up in `NOT_STARTED` state that then started in the opposite order they were queued...but I'm fairly sure that orca respects the order they were queued so this wouldn't happen.

If for some reason we do conclude that we need to sort by `startTime` we could probably even add a limit of ~10-20 (by `buildTime`) on the query, then sort by `startTime` in Java, which would protect us from enormous execution histories and would only return an incorrect result in the craziest of edge cases.  (I believe this is actually what the UI does to show the most recent piplines.)

mattnworb:
@ezimanyi 

> I think there may be a way to implement your #2 solution---avoiding the need to fetch and hydrate every pipeline. The `ExecutionCriteria` class has a `pageSize` argument that (if I'm reading correctly) will actually propagate all the way down to redis and limit the returned results to the `pageSize`. We store the executions in redis as a sorted set (based on `execution.buildTime`) which is what allows an efficient retrieval of the N most recent executions. (I don't think that the change from `startTime` sorting to `buildTime` sorting would be significant, though we'd want to think a bit more about that.)

I took a look at the SQL based implementation of ExecutionRepository and, if I am understanding the use of jooq correctly, it looks to me like it will sort the results of the query by `id`, not `buildTime`:

https://github.com/spinnaker/orca/blob/b487729658144318b59436b428b710f1d5a5ec4a/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt#L295-L297

Would that change the feasibility of this idea, of having ArtifactResolver call `executionRepository.retrievePipelinesForPipelineConfigId(pipelineId, criteria)` with a small `pageSize` argument in order to avoid loading the full (or actually the default pageSize of 3500) list of stored Executions?

I could be understanding wrong, but it seems like `Execution.id` will be sorted in the same order as `startTime`/`buildTime` anyway in most cases, since the id is a ULID, but that could be a strange assumption for `ArtifactResolver` to make of `executionRepository.retrievePipelinesForPipelineConfigId(pipelineId, criteria)` since there is nothing documented about the ordering of the returned results.

There is a `sortType` field in `ExecutionCriteria`, but `SqlExecutionRepository.retrievePipelinesForPipelineConfigId(pipelineId, criteria)` does not use it.

mattnworb:
superseded by https://github.com/spinnaker/orca/issues/2955

PR #2940, opened 2019-05-28
feat(aws/lambda): Add update lambda stage.

andrewbackes:
CRUD operations for lambdas are already available in Clouddriver [here](https://github.com/spinnaker/clouddriver/tree/master/clouddriver-lambda). This stage allows users access to the update lambda operation from within a pipeline.

## Usage
After you have Lambda enabled in Clouddriver, you can hit Orca with something like:
```
curl -X POST \
  http://localhost:8083/orchestrate \
  -H 'Content-Type: application/json' \
  -d '
  {
      "application": "myapp",
      "description": "updateLambda Test",
      "stages": [
        {
            "type": "updateLambdaFunction",
            "region": "us-west-2",
            "functionName": "arn:aws:lambda:us-west-2:<acct>:function:<function-name>",
            "description": "sample",
            "credentials": "aws-acct",
            "handler": "lambda_function.lambda_handler",
            "memory": 512,
            "role": "arn:aws:iam::<acctno>:role/service-role/test",
            "runtime": "python3.6",
            "timeout": "68",
            "tags": [{"key":"value"}]
        }
      ] 
  }
'
```

Edit: Added usage instructions

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/UpdateLambdaFunctionStage.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/UpdateLambdaFunctionStage.java
@@ -30,7 +30,7 @@ public class UpdateLambdaFunctionStage implements StageDefinitionBuilder {
   @Override
   public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
     builder
-      .withTask(updateLambdaFunctionTask.TASK_NAME, updateLambdaFunctionTask.class)
-      .withTask("monitorLambdaFunction", MonitorKatoTask.class);
+        .withTask(updateLambdaFunctionTask.TASK_NAME, updateLambdaFunctionTask.class)
+        .withTask("monitorLambdaFunction", MonitorKatoTask.class);
   }
 }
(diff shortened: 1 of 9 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/UpdateLambdaFunctionStage.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.annotation.Nonnull;
+import org.springframework.stereotype.Component;
+
+@Component
+public class UpdateLambdaFunctionStage implements StageDefinitionBuilder {
+  public static final String PIPELINE_CONFIG_TYPE = "updateLambdaFunction";
+
+  @Override
+  public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
+    builder.withTask(updateLambdaFunctionTask.TASK_NAME, updateLambdaFunctionTask.class);
```
asher: You probably want to follow with `MonitorKatoTask` here.

andrewbackes: That's a good catch, thank you.

PR #2943, opened 2019-05-29
feat(aws/lambda): Lambda stage can handle arbitrary ops

andrewbackes:
The `updateLambdaFunction` stage that was added before only supported updating lambda function config. It didn't allow users to update the lambda function code, create or delete a lambda. This modification will allow users to forward arbitrary operations to Clouddriver.

## Usage
The instructions [here](https://github.com/spinnaker/clouddriver/tree/master/clouddriver-lambda) describe how to enable lambda and available operations. Then you can hit Orca with something like:

```curl -X POST \
  http://localhost:8083/orchestrate \
  -H 'Content-Type: application/json' \
  -d '
  {
      "application": "andrewtest",
      "description": "lambda Test",
      "stages": [
        {
            "type": "lambdaFunction",
            "operation": "updateLambdaFunctionCode",
            "region": "us-east-1",
            "functionName": "andrew-lambda-handler",
            "credentials": "aws-dev",
            "s3Bucket": "armory-lambda-dev",
            "s3Key": "handler.zip",
            "publish": "true"
        }
      ] 
  }
'
```
The required fields are `operation` and `type`.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/lambda/lambdaFunctionTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/lambda/lambdaFunctionTask.java
@@ -27,15 +27,22 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import javax.annotation.Nonnull;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+/*
+  lambdaFunctionTask creates a Clouddriver operation using the field labeled
+  'operation'. The context of the task are added to the operation body. Users
+  will have to know what Clouddriver is expecting to use this task. It
+  essentially just passes an arbitrary operation over to Clouddriver.
+*/
 @Component
-public class updateLambdaFunctionTask extends AbstractCloudProviderAwareTask implements Task {
+public class lambdaFunctionTask extends AbstractCloudProviderAwareTask implements Task {
 
   @Autowired KatoService katoService;
 
-  public static final String TASK_NAME = "updateLambdaFunctionConfiguration";
+  public static final String TASK_NAME = "lambdaFunction";
 
   @Nonnull
   @Override
@@ -44,8 +51,14 @@ public class updateLambdaFunctionTask extends AbstractCloudProviderAwareTask imp
 
     Map<String, Object> task = new HashMap<>(stage.getContext());
 
+    String operationName = (String) task.get("operation");
+    if (!StringUtils.isNotBlank(operationName)) {
+      throw new IllegalArgumentException(
+        "Field 'operation' is missing and must be present in stage context");
+    }
+
     Map<String, Map> operation =
-        new ImmutableMap.Builder<String, Map>().put(TASK_NAME, task).build();
+        new ImmutableMap.Builder<String, Map>().put(operationName, task).build();
 
     TaskId taskId =
         katoService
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/lambda/lambdaFunctionTask.java
```diff
@@ -27,15 +27,22 @@
...
+  essentially just passes an arbitrary operation over to Clouddriver.
+*/
 @Component
-public class updateLambdaFunctionTask extends AbstractCloudProviderAwareTask implements Task {
+public class lambdaFunctionTask extends AbstractCloudProviderAwareTask implements Task {
 
   @Autowired KatoService katoService;
 
-  public static final String TASK_NAME = "updateLambdaFunctionConfiguration";
+  public static final String TASK_NAME = "lambdaFunction";
```
robzienert: Is `TASK_NAME` used anywhere?

robzienert:
Also, if you're changing the Task name, you will need to create a migrator so that in-flight tasks during upgrade are correctly handled.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/lambda/lambdaFunctionTask.java
andrewbackes: Yeah, it's used in the stage [here](https://github.com/spinnaker/orca/pull/2943/files#diff-d93c048a1464581cd7051d6b3ab58794R33)

andrewbackes:
Do you think that's really necessary? I don't think anybody is using this stage yet.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/lambda/lambdaFunctionTask.java
```diff
@@ -44,8 +51,14 @@ public TaskResult execute(@Nonnull Stage stage) {
 
     Map<String, Object> task = new HashMap<>(stage.getContext());
 
+    String operationName = (String) task.get("operation");
```
asher: Are there other required parameters for `lambdaFunctionTask` to succeed? Seems likely from a quick look at the related clouddriver code. It may be better to have a private pojo class in here with constraint annotations (like `@NotNull`), then use `stage.mapTo()` and check for validation errors on the resulting object.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/lambda/lambdaFunctionTask.java
```diff
@@ -44,8 +51,14 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+    if (!StringUtils.isNotBlank(operationName)) {
```
asher: The double negative is a bit confusing, `if (StringUtils.isBlank(operationName))` is > readable

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/lambda/lambdaFunctionTask.java
robzienert: I like the idea of Orca being lean on validation and type definitions of clouddriver operations unless it's critical to performing orchestration logic. Clouddriver is the authority of the payload structure, and so should also be the authority on acceptable values.

robzienert:
Sure, since the task isn't used yet, seems fine to change things up. Good point.

PR #2944, opened 2019-05-29
feat(webhook): support yaml parsing

xavileon:
The Webhook stage supports parsing of the response body when it's a json
string, so it can be embedded as Objects into the context. This patch
adds support for parsing YAML strings as well.


xavileon:
It has unexpected side effects, closing to find a better approach.

PR #2945, opened 2019-05-29
feat(core): Support client-provided execution IDs

robzienert:
The V2 trigger pipeline endpoint is currently broken (it doesn't return a valid ref for people to lookup progress). This is one of 3 changes (other two in echo, gate), allowing clients to pass an execution ID (in addition to or in substitution of) a correlation ID. This will allow Gate to generate a pipeline execution ID and return it to the client without blocking on Orca to actually process the request.

echo: https://github.com/spinnaker/echo/pull/563
gate: https://github.com/spinnaker/gate/pull/811


ajordens:
Belated LGTM.

PR #2946, opened 2019-05-30
feat(pipeline/expressions): support yaml parsing

xavileon:
This patch adds two helper functions `readYaml` and
`yamlFromUrl` to transfrom yaml strings (either
embedded or from retrieved from an URL) to their
object representation.


orca-core/orca-core.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
   implementation("org.apache.commons:commons-lang3")
   implementation("javax.servlet:javax.servlet-api:4.0.1")
   implementation("com.jayway.jsonpath:json-path:2.2.0")
+  implementation("org.yaml:snakeyaml")
```
marchello2000: where does the version for `snakyaml` come from?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupport.java
```diff
@@ -254,6 +257,20 @@ static Object readJson(String text) {
     }
   }
 
+  /**
+   * Attempts to read yaml from a text String. Will throw a parsing exception on bad yaml
+   *
+   * @param text text to read as yaml
+   * @return the object representation of the yaml text
+   */
+  static Object readYaml(String text) {
```
marchello2000: ideally, these functions should live in a functionProvider - that's the new, more modular way:

Example:
https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/StageExpressionFunctionProvider.java

you also won't need to do the registration above.

marchello2000:
Thanks for adding this!
Please also document here: https://github.com/spinnaker/spinnaker.github.io/blob/master/reference/pipeline/expressions/index.md

orca-core/orca-core.gradle
xavileon: from kork? I'm not very familiar with how the new dependency definition works :/ I saw there are other dependencies without version, just thought it would be ok not to pin.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupport.java
xavileon: oh cool! will move to that. But how does that work with the allowUnknownKeys filtering? Do you mean only move the readYaml (and readJson/readProperties) and leave the yamlFromUrl registered on demand as this patch today?

xavileon:
@marchello2000 happy to move to the new approach! just a couple of question on on how to proceed

orca-core/orca-core.gradle
marchello2000: Apologies for the delay -  was out of town.
Interesting, I don't see it specified in `kork`...  somehow it gets into kork-bom. oh well

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/ExpressionsSupport.java
marchello2000: I see what you're saying, thanks for catching that. I guess leave it as is for now then. No point splitting up the few line function out

orca-core/orca-core.gradle
marchello2000: looks like it comes from `spring-boot-dependencies` (pulled in by `kork`: https://search.maven.org/artifact/org.springframework.boot/spring-boot-dependencies/2.1.4.RELEASE/pom)

xavileon: thanks for the pointer 👍

xavileon:
Added documentation in https://github.com/spinnaker/spinnaker.github.io/pull/1389

scottfrederick:
> ideally, these functions should live in a functionProvider - that's the new, more modular way

> happy to move to the new approach! just a couple of question on on how to proceed

@xavileon This move to an `ExpressionFunctionProvider` was done as part of a refactoring that moved common expression-handling code to `kork-expressions` (https://github.com/spinnaker/orca/pull/3071).

PR #2948, opened 2019-05-30
fix(cleanup): remove 2nd check on thresholdDays in old pipeline cleanup

mattnworb:
Fixes https://github.com/spinnaker/spinnaker/issues/4467

OldPipelineCleanupPollingNotificationAgent is checking the age of the
pipeline execution twice before removing it:

1. the `Observable<Execution>` returned from
`executionRepository.retrievePipelinesForApplication(app)` is filtered
to only executions that have a completed status and where startTime (or
buildTime, if startTime is not set) is before `now - thresholdDays days`

2. After converting the `Observable<Execution>` to a `List`, sorting it,
and retaining `minimumPipelineExecutions` elements, each execution is
only deleted if `days between startTime and now > thresholdDays`.

The first check is true for executions as soon as they are a moment
older than `executionDays` days, but the second requires executions to be
older than `executionDays + 1` days (due to using `>`).

Since the second check seems redundant, I've simplify removed it.

My assumption here is that it is a bug that a configuration setting like
`thresholdDays: N` actually only removes executions older than `N+1`
days.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/notifications/scheduling/OldPipelineCleanupPollingNotificationAgent.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/notifications/scheduling/OldPipelineCleanupPollingNotificationAgent.java
@@ -28,7 +28,11 @@ import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository;
 import java.time.Clock;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


mattnworb:
I am trying to add a test for this change, since the current test only checks that an execution 10 days old is removed when thresholdDays is 5, but I am having trouble getting the spock tests to even run.

If I change `OldPipelineCleanupPollingNotificationAgentSpec` in a way where the logic/thresholds no longer make sense, it does not ever seem to fail - or even run - when I run

```
$ ./gradlew clean build
```

Am I invoking gradle wrong?

```
$ ./gradlew clean build
...
BUILD SUCCESSFUL in 4m 0s
240 actionable tasks: 234 executed, 6 up-to-date

$ ls -1 orca-core/build/test-results/test/
TEST-com.netflix.spinnaker.orca.pipeline.model.JenkinsTriggerTest.xml
binary

$ ls -1 orca-core/build/reports/tests/test/classes/
com.netflix.spinnaker.orca.pipeline.model.JenkinsTriggerTest.html
```

I am not sure if spock tests have results saved somewhere else, but this almost seems like only the single Java/JUnit5-based test is being executed in a build.

mattnworb:
I added a commit here to refactor `OldPipelineCleanupPollingNotificationAgentSpec` to demonstrate the bug here by failing (once https://github.com/spinnaker/spinnaker/issues/4469 is resolved). The existing test masks the bug because it sets up seven pipelines that are each one day apart and sets the clock to be at the 10th day - after removing the `minimumPipelineExecutions` amount of executions from the list of candidate executions to delete, all of them are older than `thresholdDays+1` which hides the bug in only deleting them if they pass the `startTime - clock > thresholdDays` check.

PR #2949, opened 2019-05-31
fix(sql): optimize reading executions by pipeline id

asher:
This PR fixes an issue where calls to `/applications/{application}/pipelines` can result in a full `pipelines` table scan per each of the applications pipeline configuration ids if they have more than N executions, where N is determined by the sql query optimizer at runtime.

### DIFF
--- a/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
+++ b/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
@@ -268,16 +268,35 @@ class SqlExecutionRepository(
     pipelineConfigId: String,
     criteria: ExecutionCriteria
   ): Observable<Execution> {
-    val select = jooq.selectExecutions(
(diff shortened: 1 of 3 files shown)


orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -268,16 +268,30 @@ class SqlExecutionRepository(
...
-      PIPELINE,
-      conditions = {
-        it.where(field("config_id").eq(pipelineConfigId))
-          .statusIn(criteria.statuses)
-      },
-      seek = {
-        it.orderBy(field("id").desc()).limit(criteria.pageSize)
-      }
-    )
+    val select = if (criteria.statuses.isEmpty() || criteria.statuses.size == ExecutionStatus.values().size) {
```
robzienert: I wouldn't mind a fast comment here about why this conditional exists. I won't recall in 6 months & would be nice to not git blame dive.

ajordens:
Thanks for digging into this one!

PR #2951, opened 2019-05-31
fix(pipelines): save pipelines task shouldn’t reuse pipeline id

claymccoy:
When save pipelines task is used multiple times in a stage, it now reuses the pipeline ID. This is problematic because if new pipelines are created they will reuse a previous pipeline ID and overwrite it rather than creating a new pipeline. I think the real problem here is the use of “pipeline.id” in the stage context for multiple purposes (both of which make sense independently).

1) SaveServiceAccountTask generates a new pipeline.id if one doesn’t exist for use with regenerateCronTriggerIds. It effectively changes the contract from a new pipeline not having an id, to it being generated beforehand.

2) Later in this class, after a pipeline is updated or created, its id is stored in “pipeline.id” to aid with refresh logic.

But the two of these together keeps the context “pipeline.id” from the last saved pipeline, and reuses it in the case that the pipeline.id is missing (when it should be created).
It may be that these two usages should have different names in the stage context. In that case this loop wouldn’t happen. Since SavePipelinesFromArtifactStage is the only stage where multiple pipelines are saved, and it does not currently use managed service accounts, it seems easier to just avoid the new regenerateCronTriggers logic when isSavingMultiplePipelines is true. Changing the model can have unintended consequences.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
@@ -82,8 +82,12 @@ public class SavePipelineTask implements RetryableTask {
     if (serviceAccount != null) {
       updateServiceAccount(pipeline, serviceAccount);
     }
-
-    if (stage.getContext().get("pipeline.id") != null && pipeline.get("id") == null) {
+    final Boolean isSavingMultiplePipelines =
+        (Boolean)
+            Optional.ofNullable(stage.getContext().get("isSavingMultiplePipelines")).orElse(false);
+    if (stage.getContext().get("pipeline.id") != null
+        && pipeline.get("id") == null
+        && !isSavingMultiplePipelines) {
       pipeline.put("id", stage.getContext().get("pipeline.id"));
 
       // We need to tell front50 to regenerate cron trigger id's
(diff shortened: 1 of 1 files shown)


fkyc:
@claymccoy thanks for making the fix! This will unblock us for the time being.

I have some thoughts around how we could improve on the solution longer term.

Currently, `SaveServiceAccountTask` and `SavePipelineTask` seem to be coupled like so:
- `SaveServiceAccountTask` needs a pipeline ID in order to generate a service account name. However, `SavePipelinesTask` hasn't ran yet, so it needs to generate its own pipeline ID.
- `SavePipelineTask` needs a service account name to update triggers in order to properly save all information to Front50. Front50 generates a pipeline ID for new pipelines.

Perhaps I need more context, but I'm not sure why we would ever want to decouple these tasks when creating or updating pipelines. Maybe we can merge the two tasks together?

Couple of benefits I can think of:
1. We wouldn't have pipeline ID generation logic in two different services (Front50 and Orca).
2. We wouldn't have to remember to put `SaveServiceAccountTask` before `SavePipelineTask` to support service accounts.
3. `SavePipelineTask` can continue to operate under the assumption that new pipelines don't have an ID, which seems more natural.

Would love to get more opinions on this @claymccoy @jtk54 @jervi

claymccoy:
That all makes sense to me. I'm not an expert on managed service accounts though and I don't have an environment configured to test it. Hopefully, I can get this fix in soon to unblock you and any case where multiple save pipeline tasks are used in a stage.

jervi:
Yeah, as I mentioned on Slack, I think this solution is ok for now, but I would also like to make a proper fix for the automatic service accounts, and as a part of that, stop generating pipeline ID's in Orca when copying pipelines. Anyway, good job @claymccoy

PR #2960, opened 2019-06-04
fix(findImage) : Fix child deploy to use correct find image when clou…

aravindmd:
Given the following pipelines:
- Parent
  -  Find Image
  - Run Child Pipeline
- Child
   - Find Image
  -  Deploy

You would probably expect the child deploy to deploy the image found by the child.
This is the case when parent and child both find AMIs, but when the child pipeline is finding a container image and deploying to Titus, the deploy ends up using the AMI found by the parent

Fix : Adding a check to make sure the cloud-providers match while getting ancestors

I have tested the following cases and it looks good
Case 1 
- Parent
  -  Find Image (aws) 
  - Run Child Pipeline
- Child
   - Find Image (aws)
  -  Deploy (aws)

Case 2
- Parent
  -  Find Image (aws)
  - Run Child Pipeline
- Child
   - Find Image (titus)
  -  Deploy (titus)

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
@@ -84,7 +84,10 @@ trait DeploymentDetailsAware {
   List<Stage> getAncestors(Stage stage, Execution execution) {
     if (stage?.requisiteStageRefIds) {
       def previousStages = execution.stages.findAll {
-        it.refId in stage.requisiteStageRefIds
+        // Include cloudProvider check to avoid confusion with multi-provider,
+        // in some cases ancestors can be one of any multi-provider
+        // Eg parent->aws and child->titus
+        it.refId in stage.requisiteStageRefIds && it.context.cloudProvider == stage.context.cloudProvider
       }
       def syntheticStages = execution.stages.findAll {
         it.parentStageId in previousStages*.id
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
@@ -81,7 +81,7 @@ trait DeploymentDetailsAware {
     }
   }
 
-  boolean checkCloudprovider(Stage stage, Stage execution){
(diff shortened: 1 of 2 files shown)


marchello2000:
looks like a few tests failed: in `com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.CreateServerGroupTaskSpec`

aravindmd:
Taking a look at the failed tests

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
```diff
@@ -84,7 +84,10 @@ trait DeploymentDetailsAware {
   List<Stage> getAncestors(Stage stage, Execution execution) {
     if (stage?.requisiteStageRefIds) {
       def previousStages = execution.stages.findAll {
-        it.refId in stage.requisiteStageRefIds
+        // Include cloudProvider check to avoid confusion with multi-provider,
+        // in some cases ancestors can be one of any multi-provider
+        // Eg parent->aws and child->titus
+        it.refId in stage.requisiteStageRefIds && it.context.cloudProvider == stage.context.cloudProvider
```
marchello2000: this makes sense. But.. I am still confused why this is necessary since the assumption is that we find the nearest ancestor, right?

With this fix, what happens in this case:

1. Parent (aws)
    a. Find Image
    b. Run Child Pipeline
2. Child (titus)
    a. Find Image
    b. Deploy
3. Grandchild (titus)
    a. Find Image
    b. Deploy

will 3b deploy image from 3a or 2a?

aravindmd: For the above case it deploys the image from 3a , i tested a similar setup with 3rd pipeline being an aws find image and deploy which works fine. I'm going to test a few more cases before merging this change.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
```diff
@@ -81,10 +81,21 @@ trait DeploymentDetailsAware {
     }
   }
 
+  boolean checkCloudprovider(Stage stage, Stage execution){
```
marchello2000: nit: maybe `isCloudProviderEqual()` would be more descriptive

emjburns:
@aravindmd please add a test for this!

PR #2965, opened 2019-06-05
feat(admin): admin endpoints to retrieve executions

marchello2000:
Currently, when there is an outage and the Spinnaker operators needs to
locate pipelines affected by the outage one must go straight to the executions DB.
This adds two endpoint points:
* one to retrieve executions (pipelines & orchestrations) that ran during a given time window with a specific status (e.g. `TERMINAL`)
* one to retrieve executions that had a stage of particular type complete with a specific status (e.g. `TERMINAL`) in a given time window

Admin EPs so not exposed in `gate`

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
@@ -240,9 +240,10 @@ class DualExecutionRepository(
     updatedAtEnd: Long
   ): List<Execution> {
 
-    return primary.retrieveExecutionsWithStatusInTimeWindow(executionType, status, updatedAtStart, updatedAtStart)
-      .plus(previous.retrieveExecutionsWithStatusInTimeWindow(executionType, status, updatedAtStart, updatedAtStart)
-    )
+    return Observable.merge(
+      Observable.from(primary.retrieveExecutionsWithStatusInTimeWindow(executionType, status, updatedAtStart, updatedAtStart)),
+      Observable.from(previous.retrieveExecutionsWithStatusInTimeWindow(executionType, status, updatedAtStart, updatedAtStart))
+    ).toList().toBlocking().single()
   }
 
   override fun retrieveExecutionsWithSpecificStageTypesInTimeWindow(
@@ -253,9 +254,10 @@ class DualExecutionRepository(
     updatedAtEnd: Long
   ): List<Execution> {
 
-    return primary.retrieveExecutionsWithSpecificStageTypesInTimeWindow(executionType, status, stageType, updatedAtStart, updatedAtStart)
-      .plus(previous.retrieveExecutionsWithSpecificStageTypesInTimeWindow(executionType, status, stageType, updatedAtStart, updatedAtStart)
-      )
+    return Observable.merge(
+      Observable.from(primary.retrieveExecutionsWithSpecificStageTypesInTimeWindow(executionType, status, stageType, updatedAtStart, updatedAtStart)),
+      Observable.from(previous.retrieveExecutionsWithSpecificStageTypesInTimeWindow(executionType, status, stageType, updatedAtStart, updatedAtStart))
+    ).toList().toBlocking().single()
   }
 
   override fun retrieveOrchestrationsForApplication(
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
@@ -233,6 +233,31 @@ class DualExecutionRepository(
     )
   }
 
+  override fun retrieveExecutionsWithStatusInTimeWindow(
+    executionType: Execution.ExecutionType,
+    status: String,
+    updatedAtStart: Long,
+    updatedAtEnd: Long
+  ): List<Execution> {
+
+    return primary.retrieveExecutionsWithStatusInTimeWindow(executionType, status, updatedAtStart, updatedAtStart)
+      .plus(previous.retrieveExecutionsWithStatusInTimeWindow(executionType, status, updatedAtStart, updatedAtStart)
+    )
+  }
+
+  override fun retrieveExecutionsWithSpecificStageTypesInTimeWindow(
+    executionType: Execution.ExecutionType,
+    status: String,
+    stageType: String,
+    updatedAtStart: Long,
+    updatedAtEnd: Long
+  ): List<Execution> {
+
+    return primary.retrieveExecutionsWithSpecificStageTypesInTimeWindow(executionType, status, stageType, updatedAtStart, updatedAtStart)
+      .plus(previous.retrieveExecutionsWithSpecificStageTypesInTimeWindow(executionType, status, stageType, updatedAtStart, updatedAtStart)
+      )
+  }
+
   override fun retrieveOrchestrationsForApplication(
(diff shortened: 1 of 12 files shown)


orca-sql/src/main/resources/db/changelog/20190604-stages-updatedat-index.yml
```diff
@@ -0,0 +1,24 @@
...
+  - changeSet:
+      id: 20190604-update-stages-indices
+      author: mvulfson
+      changes:
+      - createIndex:
+          indexName: pipeline_stages_updated_at_idx
+          tableName: pipeline_stages
+          columns:
+          - column:
+              name: updated_at
```
marchello2000: @asher , just confirming that secondary indices can be non-unique (and if anything else needs to change here - my first migration)

orca-sql/src/test/groovy/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepositorySpec.groovy
```diff
@@ -501,4 +503,42 @@ class SqlExecutionRepositorySpec extends ExecutionRepositoryTck<SqlExecutionRepo
       new ExecutionCriteria().setPageSize(1).setSortType(BUILD_TIME_ASC)
     ).size() == 0
   }
+
+  @Unroll
+  def "can retrieve #type by status and time window"() {
```
marchello2000: i wanted to add a unit test for the other function (`retrieveExecutionsWithSpecificStageTypesInTimeWindow`) but `H2` doesn't support `JSON_EXTRACT` so it wasn't working. Let me know if there a way to make it work

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -308,6 +308,53 @@ class SqlExecutionRepository(
...
+    updatedAtEnd: Long
+  ): List<Execution> {
+    // SELECT tP.*
+    //   FROM pipeline_stages AS tS
+    //   INNER JOIN pipelines AS tP
+    //     ON tS.execution_id = tP.id
+    //   WHERE JSON_EXTRACT(tS.body, "$.type") = "deploy" AND tS.updated_at > 0 AND tS.updated_at < 10000000000000 AND tS.status = "SUCCEEDED";
+
+    return jooq.select(
+      field("tP.id").`as`("id"), field("tP.body").`as`("body"))
```
marchello2000: This almost makes me want to just write out the SQL with `?` parameters... let me know if there is a better way

robzienert: We _definitely_ do not want to use `JSON_EXTRACT` as part of a `WHERE` condition: This will cause a full table scan. You'll need to add a column for the stage type and index it.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
```diff
@@ -74,6 +80,29 @@ class AdminController {
     forceExecutionCancellationCommand.forceCancel(executionType, executionId, canceledBy)
   }
 
+  @RequestMapping(value = "/executions", method = RequestMethod.GET)
+  AdminExecutionSearchResult getExecutions(@RequestParam(value = "status", required = true) String status,
+  @RequestParam(value = "startTime", required = true) long startTime,
+  @RequestParam(value = "endTime", required = true) long endTime) {
```
robzienert: This formatting is mega funky.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
marchello2000: I am pretty sure `explain` locally was not showing a full table scan, but I will have to double check. It will have to parse a bunch of the statuses but it's going to be limited to the time window and the status clauses (both of which are required and with the index created on `updated_at` it should be pretty fast. but I will dig some more to make sure

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
marchello2000: yeah, wtf? just checked that nothing was left behind by spotless.. sigh

marchello2000: this is borderline unreadable

robzienert: Groovy does not get formatted, because Groovy is a tire fire.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
robzienert: `JSON_EXTRACT` doesn't provide any magic: It requires inspection of a blob field, JSON deserialization and then filtering (probably via a temp table, now that I think of it, instead of a table scan). Try running that against the read replica in main and you'll have a bad time.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
```diff
@@ -233,6 +233,31 @@ class DualExecutionRepository(
...
+  override fun retrieveExecutionsWithStatusInTimeWindow(
+    executionType: Execution.ExecutionType,
+    status: String,
+    updatedAtStart: Long,
+    updatedAtEnd: Long
+  ): List<Execution> {
+
+    return primary.retrieveExecutionsWithStatusInTimeWindow(executionType, status, updatedAtStart, updatedAtStart)
+      .plus(previous.retrieveExecutionsWithStatusInTimeWindow(executionType, status, updatedAtStart, updatedAtStart)
+    )
```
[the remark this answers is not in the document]
marchello2000: ok

orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
```diff
@@ -616,6 +617,27 @@ public void delete(@Nonnull ExecutionType type, @Nonnull String id) {
...
   }
 
+  @Override
+  public List<Execution> retrieveExecutionsWithStatusInTimeWindow(
+      Execution.ExecutionType executionType,
+      String status,
+      long updatedAtStart,
+      long updatedAtEnd) {
+    throw new NotImplementedException(
+        "retrieveExecutionsWithStatusInTimeWindow Not supported in RedisExecutionRepository");
```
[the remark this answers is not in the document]
marchello2000: yeah, we were going back and forth on it. will convert to `emptyList`

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
marchello2000: roger that, fixing

robzienert:
Closing for now. Feel free to reopen if you find time again to revisit.

PR #2966, opened 2019-06-06
fix(runJob): cancel underlying job on cancel stage

emjburns:
Currently, when you cancel a run job stage it just leaves the underlying job running. That's unexpected behavior. 

This PR adds support for canceling the underlying job when you cancel the stage.

@ezimanyi @ethanfrogers I'm looking for some help verifying that this data structure also works for kubernetes (specifically that `jobStatus.name` is the name of the kubernetes job and the same name used in this [destroyJob](https://github.com/spinnaker/clouddriver/blob/29c5c1435ab0c750c4b491208f600647a3e98901/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/job/DestroyKubernetesJobAtomicOperation.groovy) atomic operation).

This PR uses https://github.com/spinnaker/clouddriver/pull/3762 for Titus support.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
@@ -16,20 +16,36 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.job;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.orca.CancellableStage;
 import com.netflix.spinnaker.orca.clouddriver.tasks.artifacts.ConsumeArtifactTask;
+import com.netflix.spinnaker.orca.clouddriver.tasks.job.DestroyJobTask;
 import com.netflix.spinnaker.orca.clouddriver.tasks.job.MonitorJobTask;
 import com.netflix.spinnaker.orca.clouddriver.tasks.job.RunJobTask;
 import com.netflix.spinnaker.orca.clouddriver.tasks.job.WaitOnJobCompletion;
+import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper;
 import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder;
 import com.netflix.spinnaker.orca.pipeline.TaskNode;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
 import com.netflix.spinnaker.orca.pipeline.tasks.artifacts.BindProducedArtifactsTask;
 import java.util.HashMap;
 import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 @Component
-public class RunJobStage implements StageDefinitionBuilder {
+public class RunJobStage implements StageDefinitionBuilder, CancellableStage {
+
+  private final Logger log = LoggerFactory.getLogger(getClass());
+  private final DestroyJobTask destroyJobTask;
+  private final ObjectMapper objectMapper = OrcaObjectMapper.newInstance();
+
+  @Autowired
+  public RunJobStage(DestroyJobTask destroyJobTask) {
+    this.destroyJobTask = destroyJobTask;
+  }
 
   @Override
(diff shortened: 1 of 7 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
```diff
@@ -16,20 +16,40 @@
...
 
 @Component
-public class RunJobStage implements StageDefinitionBuilder {
+public class RunJobStage implements StageDefinitionBuilder, CancellableStage {
+
+  private final Logger log = LoggerFactory.getLogger(getClass());
+  private final DestroyJobTask destroyJobTask;
+  private final ObjectMapper objectMapper = OrcaObjectMapper.newInstance();
+
+  public RunJobStage() {
```
cfieber: why is this needed? I suspect it is just so existing tests don't need to be updated... I'd lean towards just updating the tests to pass in a stub if that is the case

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
```diff
@@ -56,6 +76,43 @@ public void taskGraph(Stage stage, TaskNode.Builder builder) {
...
+      destroyContext.put("cloudProvider", context.getCloudProvider());
+      destroyContext.put("region", context.getJobStatus().getRegion());
+      destroyContext.put("credentials", context.getCredentials());
+
+      stage.setContext(destroyContext);
+
+      destroyJobTask.execute(stage);
+    } catch (Exception e) {
+      log.error(
+          String.format(
```
marchello2000: nit: some weird parens here. also, for my education, is `RunJob` titus specific? if not, probably worth making this log message more generic

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
emjburns: You are correct, I'll fix the tests so this is not needed.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
emjburns: Oops good call. Run job is for titus and kubernetes, so I'll make this generic. I'll also fix the parens

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStageContext.java
```diff
@@ -0,0 +1,85 @@
...
+  public Map<String, Object> other() {
+    return other;
+  }
+
+  @JsonAnySetter
+  public void set(String name, Object value) {
+    other.put(name, value);
+  }
+
+  public JobStatus getJobStatus() {
```
ezimanyi: Any reason not to annotate the class with `@Getter` and `@Setter` instead of manually defining these?

emjburns: Do you know if this works `@JsonAnyGetter` for the `other()` properties?  That's why I didn't do it here

ezimanyi: I don't know for sure; my guess is that it would work because serialization/deserialization shouldn't care how the `getXXX` and `setXXX` methods were generated.  I think it just looks for methods with those names, then if absent would fall back to the `@JsonAny...` annotated methods.  But I also am not 100% sure of this so definitely could be wrong here...

emjburns: I'm going to leave this as it is for now, although I think you may be right.

emjburns:
Got confirmation from a kubernetes run job user that `jobStatus.name` is present and is the correct name to cancel the job with.

ethanfrogers:
LGTM!

PR #2968, opened 2019-06-07
fix(RRB): Add dedicated pin stage

marchello2000:
For RRB we start by pinning the source SG so that it doesn't scale down during deploy (and end with unpinning it).
However, if for whatever reason some of the instances are taken out of discovery/targetgroup we will timeout waiting for the capacity match on this operation.
Since this pin/unpin operations are purely "cautionary" and don't require us to actually match the capacity, don't wait for the capacity match so we don't fail for this non-critical reason

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/PinServerGroupStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/PinServerGroupStage.groovy
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/PinServerGroupStage.groovy
```diff
@@ -0,0 +1,46 @@
...
+ */
+@Component
+@Slf4j
+class PinServerGroupStage extends TargetServerGroupLinearStageSupport {
+  public static final String TYPE = getType(PinServerGroupStage)
+
+  @Override
+  protected void taskGraphInternal(Stage stage, TaskNode.Builder builder) {
+    builder
+      .withTask("determineHealthProviders", DetermineHealthProvidersTask)
```
dreynaud: Capturing offline chat: is `determineHealthProviders` still needed? 

Plus please add some motivation for why the scaling processes changes are not needed. Otherwise LGTM! 👍

marchello2000: removed the `determineHealthProviders` and updated comments

PR #2971, opened 2019-06-10
fix(core): Mark mptv2 items as inherited

louisjimenez:
Marks inherited items (triggers, notifications, parameters) as "inherited" so that the UI can distinguish between what is on the pipeline config and editable versus what is read only from the template.
https://github.com/spinnaker/spinnaker/issues/4451

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v2schema/V2SchemaExecutionGenerator.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v2schema/V2SchemaExecutionGenerator.java
@@ -23,6 +23,7 @@ import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.graph.v2.transform.V
 import com.netflix.spinnaker.orca.pipelinetemplate.v2schema.model.V2PipelineTemplate;
 import com.netflix.spinnaker.orca.pipelinetemplate.v2schema.model.V2TemplateConfiguration;
 import java.util.*;
+import java.util.stream.Collectors;
 
 public class V2SchemaExecutionGenerator implements V2ExecutionGenerator {
 
@@ -76,7 +77,7 @@ public class V2SchemaExecutionGenerator implements V2ExecutionGenerator {
       pipeline.put(
           "notifications",
           TemplateMerge.mergeDistinct(
-              (List<HashMap<String, Object>>) template.getPipeline().get("notifications"),
+              markInherited(getTemplateCollection(template, "notifications")),
               configuration.getNotifications()));
     }
   }
@@ -93,7 +94,7 @@ public class V2SchemaExecutionGenerator implements V2ExecutionGenerator {
       pipeline.put(
           "parameterConfig",
           TemplateMerge.mergeDistinct(
-              (List<HashMap<String, Object>>) template.getPipeline().get("parameterConfig"),
+              markInherited(getTemplateCollection(template, "parameterConfig")),
               configuration.getParameters()));
     }
   }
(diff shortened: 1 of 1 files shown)


louisjimenez:
Related PR for deck: https://github.com/spinnaker/deck/pull/7102

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v2schema/V2SchemaExecutionGenerator.java
```diff
@@ -110,8 +111,25 @@ private void addTriggers(
       pipeline.put(
           "triggers",
           TemplateMerge.mergeDistinct(
-              (List<HashMap<String, Object>>) template.getPipeline().get("triggers"),
+              markInherited(getTemplateCollection(template, "triggers")),
               configuration.getTriggers()));
     }
   }
+
+  private static List<HashMap<String, Object>> markInherited(
```
maggieneterval:
```suggestion
  private static List<Map<String, Object>> markInherited(
```

maggieneterval: I think we typically want to have return types be interfaces unless there's a good reason for them to be implementations

maggieneterval: Separately, I feel like the name `markInherited` incorrectly implies that we are mutating the input. I don't have a good alternative name though...

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v2schema/V2SchemaExecutionGenerator.java
```diff
@@ -110,8 +111,25 @@ private void addTriggers(
...
+    return templateCollection.stream()
+        .map(
+            templateItem -> {
+              templateItem.put("inherited", true);
+              return templateItem;
+            })
+        .collect(Collectors.toList());
+  }
+
+  private static List<HashMap<String, Object>> getTemplateCollection(
```
maggieneterval:
```suggestion
  private static List<Map<String, Object>> getTemplateCollection(
```

maggieneterval:
Adding @duftler as well since Eric is OOO :smile_cat:

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v2schema/V2SchemaExecutionGenerator.java
maggieneterval: `getInheritanceDesignatedTemplateCollections`? Probably too verbose, ignore me :see_no_evil:

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v2schema/V2SchemaExecutionGenerator.java
louisjimenez: Ah good call out, definitely, thanks!

louisjimenez: hmmm now I recall why I did this, configuration.getParameters returns a HashMap and mergeDistinct expects both arguments to be of the same type

maggieneterval: ooooh gotcha :+1:

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v2schema/V2SchemaExecutionGenerator.java
louisjimenez: 👍 I agree that markInherited is a bit confusing. I am renaming and refactoring to make it clearer that it mutates with an `inherited` key

maggieneterval: Nice LGTM :100:

louisjimenez:

PR #2972, opened 2019-06-10
fix(gcb): Bind artifacts produced from GCB stage

ezimanyi:
Fixes spinnaker/spinnaker#4503

* fix(spel): Fix NPE in SpEL evaluation 

  Evaluating SpEL on a null object throws an NPE; if the input object is null, just return null as the evaluated expression.

* fix(gcb): Bind artifacts produced from GCB stage 

  Because we didn't include the BindProducedArtifacts task in the GCB stage, produced artifacts were not matched to expected artifacts, which prevented correctly using them downstream.


ezimanyi:
The first commit was just something I ran into while trying to reproduce the issue that made debugging harder; the second commit actually fixes the bug.

I'd like to pull this into the next 1.14 patch release.

maggieneterval:

PR #2974, opened 2019-06-11
fix(Orca): Feature: option to add a delay before polling starts in Webhook stage #3450

neelimamanubolu:
Hi,
As suggested added 404 status for the webhook stage. Please find the below link for refernce.

https://github.com/spinnaker/spinnaker/issues/3450

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
@@ -101,7 +101,7 @@ class CreateWebhookTask implements RetryableTask {
         return TaskResult.builder(ExecutionStatus.TERMINAL).context(outputs).build()
       }
 
-      if (statusCode.is5xxServerError() || statusCode.value() == 429) {
+      if (statusCode.is5xxServerError() || statusCode.value() == 429 || statusCode.value() == 404) {
         String errorMessage = "error submitting webhook for pipeline ${stage.execution.id} to ${stageData.url}, will retry."
         log.warn(errorMessage, e)
 
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy
@@ -17,7 +17,7 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.providers.aws
 
-import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerGroupStage
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.PinServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.AbstractDeployStrategyStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.DeployStagePreProcessor
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup
@@ -39,7 +39,7 @@ class AwsDeployStagePreProcessor implements DeployStagePreProcessor {
   ApplySourceServerGroupCapacityStage applySourceServerGroupSnapshotStage
 
   @Autowired
-  ResizeServerGroupStage resizeServerGroupStage
+  PinServerGroupStage pinServerGroupStage
 
   @Autowired
   TargetServerGroupResolver targetServerGroupResolver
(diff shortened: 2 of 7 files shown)


orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -101,7 +101,7 @@ class CreateWebhookTask implements RetryableTask {
         return TaskResult.builder(ExecutionStatus.TERMINAL).context(outputs).build()
       }
 
-      if (statusCode.is5xxServerError() || statusCode.value() == 429) {
+      if (statusCode.is5xxServerError() || statusCode.value() == 429 || statusCode.value() == 404) {
```
marchello2000: I believe the issue is talking about the monitoring version of the webhook task:
https://github.com/spinnaker/orca/blob/master/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy#L92

I am a little concerned about blindly retrying on 404 - I think in more cases than not 404 is an irrecoverable error rather than retryable. 

Adding this to the monitor side is a little more palatable but I'd rather we don't add it here

marchello2000:
Also, the title is misleading. `option to add a delay before polling starts in Webhook stage` - this change doesn't do that

marchello2000:
@neelimamanubolu : I am going to revert this PR for now, please see comments and put up another PR, I am happy to review it then!

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
neelimamanubolu: Hi Mark Vulfson ,
Kindly please go through @Eric Zimanyi comments for the issue " Adding two parameters in the UI to create delay before polling starts in Webhook stage spinnaker/spinnaker#4470 ".
Original issue is **"Feature: option to add a delay before polling starts in Webhook stage spinnaker/spinnaker#3450"**.
We proposed solution that we will go in the form of issue #4470, Eric suggested to go with two parts.. 1. add 404 status and 2.add delay.
We done 404 status as the first attempt and trying to add delay in the next step.

neelimamanubolu:
> Also, the title is misleading. `option to add a delay before polling starts in Webhook stage` - this change doesn't do that

As posted earlier this title is the suggested solution for the issue #3450. For your refrence pls go through the comments from @Eric  in the "Adding two parameters in the UI to create delay before polling starts in Webhook stage #4470 ".

Still if u find the commit doesnt suit then  we can discuss still more brief about this.

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
marchello2000: @neelimamanubolu I see what @ezimanyi is saying. But note that as I mentioned earlier, the changes you made do not implement his suggestion. His idea is to retry on 404 in the `MonitorWebhookTask` not `WebhookTask`. I would be more OK with retrying on 404 on the MonitorTask (though I still think, and @ezimanyi notes in his comment as well, this is a breaking change.)

marchello2000:
> > Also, the title is misleading. `option to add a delay before polling starts in Webhook stage` - this change doesn't do that
> 
> As posted earlier this title is the suggested solution for the issue #3450. For your refrence pls go through the comments from @eric in the "Adding two parameters in the UI to create delay before polling starts in Webhook stage #4470 ".
> 
> Still if u find the commit doesnt suit then we can discuss more brief about this.

I understand that, but think about someone going through the commits 6 months from now trying to understand why something happened. The person will read the title "add a delay before starting in webhook stage" but when they look at the code it has nothing to do with a delay. All it does is retry on 404s.
It's fine to add all those details to the body of the commit message, but the title should be a short description of what the change does

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
neelimamanubolu: In MonitorWebhookTask, changes are not reflecting, @ezimanyi  gave suggestion to add like 5xx and 429, but not exactly in the same class. So we created in WebhookTask , tested and fially made PR.

neelimamanubolu:
> > > Also, the title is misleading. `option to add a delay before polling starts in Webhook stage` - this change doesn't do that
> > 
> > 
> > As posted earlier this title is the suggested solution for the issue #3450. For your refrence pls go through the comments from @eric in the "Adding two parameters in the UI to create delay before polling starts in Webhook stage #4470 ".
> > Still if u find the commit doesnt suit then we can discuss more brief about this.
> 
> I understand that, but think about someone going through the commits 6 months from now trying to understand why something happened. The person will read the title "add a delay before starting in webhook stage" but when they look at the code it has nothing to do with a delay. All it does is retry on 404s.
> It's fine to add all those details to the body of the commit message, but the title should be a short description of what the change does


Ok..Mark, as requested will raise another PR for the other issue "Adding two parameters in the UI to create delay before polling starts in Webhook stage #4470 ".

PR #2980, opened 2019-06-12
fix(findImage) : Fix findImage bug introduced in #2960

aravindmd:
- fix bug introduced in #2960  , add the `clouddriver` check after synthetic stages are computed

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
@@ -42,7 +42,7 @@ trait DeploymentDetailsAware {
     Closure callback) {
     Stage previousStage = getPreviousStageWithImage(stage, targetRegion, targetCloudProvider)
     def result = [:]
-    if (previousStage) {
+    if (previousStage && isCloudProviderEqual(stage, previousStage)) {
       if (previousStage.context.containsKey("amiDetails")) {
         def amiDetail = previousStage.context.amiDetails.find {
           !targetRegion || it.region == targetRegion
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/DeploymentDetailsAware.groovy
```diff
@@ -42,7 +42,7 @@ trait DeploymentDetailsAware {
     Closure callback) {
     Stage previousStage = getPreviousStageWithImage(stage, targetRegion, targetCloudProvider)
     def result = [:]
-    if (previousStage) {
+    if (previousStage && isCloudProviderEqual(stage, previousStage)) {
```
jeyrschabu: Might be able to get away with just ` previousStage.context.cloudProvider == stage.context.cloudProvider` here and get rid of the function.

aravindmd: i had to throw in the function to add those `null` checks , some tests pass `null` for `cloudprovider`

marchello2000: what happens if:
```
Stage A          ->       Stage B                -> Stage C
findimage(aws)        findImage(titus)        deploy(aws)
```

won't stage C find image from stage B only to discard it because the cloud providers don't match?
In other words if the getPreviousStageWithImage returned the wrong image and we have to continue searching this change won't work? (I am probably missing something)

aravindmd: this case works fine , i can discuss this with you offline.

PR #2981, opened 2019-06-12
fix(Orca): Open Adding two parameters in the UI to create delay before polling starts in Webhook stage #4470

neelimamanubolu:
Hi, As discussed, created another PR for the same changes for the proposed issue.

Please find below issue for reference :

https://github.com/spinnaker/spinnaker/issues/4470


ezimanyi:
Closing per above discussion.

PR #2982, opened 2019-06-12
feat(k8s): Support new artifact model in Patch Manifest stage

jkschneider:
Dependency for https://github.com/spinnaker/deck/pull/7109, as that UI change includes some additive changes to the Patch Manifest model.

This PR also:
* Extracts the common logic for Deploy and Patch Manifest.
* Deserializes both stages' contexts to a new type for better type safety going forward.
* Text input for Patch Manifest is now expression evaluated like text input for Deploy Manifest was previously (consistency).
* Simplifies the data model in a couple places, e.g. using `@Data` rather than `@Getter` + manual constructors and using `@JsonProperty` on enums rather than manual `fromKey()` etc.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestContext.java
```diff
@@ -17,88 +17,45 @@
...
-  private final Artifact manifestArtifact;
-  private final String manifestArtifactAccount;
-  private final Boolean skipExpressionEvaluation;
-  private final TrafficManagement trafficManagement;
-  private final List<String> requiredArtifactIds;
-  private final List<BindArtifact> requiredArtifacts;
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class DeployManifestContext extends ManifestContext {
+  @Nullable private List<String> manifests;
```
ezimanyi: When using the inline artifact editor, the manifests are actually converted to JSON before being stored in the stage, so I believe this needs to be `List<Object>`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
```diff
@@ -0,0 +1,198 @@
...
+                  log.warn("Failure fetching/parsing manifests from {}", manifestArtifact, e);
+                  // forces a retry
+                  throw new IllegalStateException(e);
+                }
+              },
+              10,
+              200,
+              true); // retry 10x, starting at .2s intervals);
+    } else if (context.getRawManifest() != null) {
+      rawManifests = Collections.singletonList(yamlParser.get().loadAll(context.getRawManifest()));
```
ezimanyi: Manifests that are stored directly on the stage in `manifests` are already in JSON so we shouldn't be deserializing YAML here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestContext.java
```diff
@@ -17,88 +17,45 @@
...
-        this.strategy = ManifestStrategyType.fromKey(strategy);
-      }
+      @JsonProperty("none")
+      None
     }
   }
+
+  @Override
+  public String getRawManifest() {
+    return manifests == null ? null : manifests.stream().findFirst().orElse(null);
```
ezimanyi: Why are we only returning the first manifest here?  The inline editor supports multiple manifests separated by `---` and I believe this will cause us to deploy only the first.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
Jammy-Louie: it appears the previous implementation also deserializes the directly stored manifest on the stage in order to perform `ExpressionEvaluation` if `skipExpressionEvaluation` was not set to true

ezimanyi: Where is that happening?  I'm pretty sure this will error as we'd be parsing JSON as YAML, but maybe I'm missing a translation somewhere---in any case, it's important to test this works for both inline manifests and artifacts.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestContext.java
```diff
@@ -17,88 +17,45 @@
...
-public class DeployManifestContext extends HashMap<String, Object> {
-  private final String source;
-  private final String manifestArtifactId;
-  private final Artifact manifestArtifact;
-  private final String manifestArtifactAccount;
-  private final Boolean skipExpressionEvaluation;
-  private final TrafficManagement trafficManagement;
-  private final List<String> requiredArtifactIds;
-  private final List<BindArtifact> requiredArtifacts;
+@Data
```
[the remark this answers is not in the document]
Jammy-Louie: I've updated this so DeployManifestContext and PatchManifestContext is immutable

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestContext.java
Jammy-Louie: 👍

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
Jammy-Louie: I was incorrect. I made the appropriate changes to maintain previous functionality

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestContext.java
Jammy-Louie: 👍

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStage.java
```diff
@@ -58,11 +59,91 @@ public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
...
-    DeployManifestContext context = stage.mapTo(DeployManifestContext.class);
-    TrafficManagement trafficManagement = context.getTrafficManagement();
+    TrafficManagement trafficManagement =
+        stage.mapTo(DeployManifestContext.class).getTrafficManagement();
     if (trafficManagement.isEnabled()) {
-      manifestStrategyStagesAdder.addAfterStages(
-          trafficManagement.getOptions().getStrategy(), graph, context);
+      switch (trafficManagement.getOptions().getStrategy()) {
+        case RedBlack:
+          disableOldManifests(stage.getContext(), graph);
```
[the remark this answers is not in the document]
Jammy-Louie: 👍

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestContext.java
```diff
@@ -17,88 +17,65 @@
...
-    @Getter
-    public static class Options {
-      private final boolean enableTraffic;
-      private final List<String> services;
-      private final ManifestStrategyType strategy;
+    public enum ManifestStrategyType {
+      @JsonProperty("redblack")
+      RedBlack,
+
+      @JsonProperty("highlander")
```
ezimanyi: This changes the serialized form of the `ManifestStrategyType`; prior to this it was capitalized.  This will break deploying in clouddriver, as clouddriver is expecting this to be capitalized as it was prior to the change.

Jammy-Louie: good catch.
I've reverted it back to how it was previously.

ezimanyi:
Looks like the last commit introduced a compilation error---I think some references to the enum need to be updated too.

Jammy-Louie:
@ezimanyi sorry about that

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStage.java
```diff
@@ -58,11 +59,93 @@ public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
...
 
   public void afterStages(@Nonnull Stage stage, @Nonnull StageGraphBuilder graph) {
-    DeployManifestContext context = stage.mapTo(DeployManifestContext.class);
-    TrafficManagement trafficManagement = context.getTrafficManagement();
+    TrafficManagement trafficManagement =
+        stage.mapTo(DeployManifestContext.class).getTrafficManagement();
     if (trafficManagement.isEnabled()) {
-      manifestStrategyStagesAdder.addAfterStages(
-          trafficManagement.getOptions().getStrategy(), graph, context);
+      switch (trafficManagement.getOptions().getStrategy()) {
```
ezimanyi: The logic here is wrong; we should not be deleting old manifests in `RED_BLACK`.  The existing logic is:
Disable and delete for highlander:
https://github.com/spinnaker/orca/blob/12d98dff16f6e02f5529ac22e9389f66b057ecd0/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestHighlanderStrategy.java#L33

Disable only for Red/Black:
https://github.com/spinnaker/orca/blob/12d98dff16f6e02f5529ac22e9389f66b057ecd0/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestRedBlackStrategy.java#L33

I think somewhere in the refactor this was broken.

Jammy-Louie: 👍

ezimanyi:
@Jammy-Louie : This broke our integration tests with a deserialization failure on the patch manifest stage.  I'm getting the same error any time I try to run a patch manifest stage as well---can you please take a look?

```
java.lang.IllegalArgumentException: Unable to map context to class com.netflix.spinnaker.orca.clouddriver.tasks.manifest.PatchManifestTask$PatchManifestContext at com.netflix.spinnaker.orca.pipeline.model.Stage.mapTo(Stage.java:453)
```

This is with an inline patch body (rather than an artifact); not sure if you tested with both cases.  Thanks!

PR #2984, opened 2019-06-12
fix(webhooks): addresses issue 3450 - introduce a delay before polling webhook

marchello2000:
Add additional parameters to the monitored webhook to allow:
1. waiting some number of seconds before polling starts
2. retrying on specific HTTP status codes

Still needs deck counterpart

see [3450](https://github.com/spinnaker/spinnaker/issues/3450)

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
@@ -17,6 +17,7 @@
 
 package com.netflix.spinnaker.orca.webhook.tasks
 
+import com.google.common.base.Strings
 import com.jayway.jsonpath.JsonPath
 import com.jayway.jsonpath.PathNotFoundException
 import com.netflix.spinnaker.orca.ExecutionStatus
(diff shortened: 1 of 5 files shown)


orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
```diff
@@ -53,48 +53,54 @@ class MonitorWebhookTask implements OverridableTimeoutRetryableTask {
...
 
-  static requiredParameters = ["statusEndpoint", "statusJsonPath"]
-
   @Override
   TaskResult execute(Stage stage) {
-    def missing = requiredParameters.findAll { !stage.context.get(it) }
-    if (!missing.empty) {
-      throw new IllegalStateException("Missing required parameter${missing.size() > 1 ? 's' : ''} '${missing.join('\', \'')}'")
-    }
+    StageData stageData = stage.mapTo(StageData)
```
marchello2000: did some minor refactor here to bring into ~2018... :)
`stageData` and no more throws from here -> return TERMINAL instead

marchello2000:
cc: @neelimamanubolu

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
```diff
@@ -53,48 +53,54 @@ class MonitorWebhookTask implements OverridableTimeoutRetryableTask {
...
 
-    String statusEndpoint = stage.context.statusEndpoint
-    String statusJsonPath = stage.context.statusJsonPath
-    String progressJsonPath = stage.context.progressJsonPath
-    String successStatuses = stage.context.successStatuses
-    String canceledStatuses = stage.context.canceledStatuses
-    String terminalStatuses = stage.context.terminalStatuses
-    def customHeaders = stage.context.customHeaders
+    if (stageData.statusEndpoint == null || stageData.statusJsonPath == null) {
```
ezimanyi: Has this changed the behavior if either `statusEndpoint` or `statusJsonPath` is the empty string?  I think before it would throw an exception before where as now it will not.

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
```diff
@@ -53,48 +53,54 @@ class MonitorWebhookTask implements OverridableTimeoutRetryableTask {
...
+      String errorMessage = "an exception occurred in webhook monitor to ${stageData.statusEndpoint}: ${e}"
+      log.error(errorMessage, e)
+      Map<String, ?> outputs = [webhook: [monitor: [:]]]
+      outputs.webhook.monitor << [error: errorMessage]
+      return TaskResult.builder(ExecutionStatus.TERMINAL).context(outputs).build()
     } catch (HttpStatusCodeException  e) {
       def statusCode = e.getStatusCode()
-      if (statusCode.is5xxServerError() || statusCode.value() == 429) {
-        log.warn("error getting webhook status from ${statusEndpoint}, will retry", e)
+      def statusValue = statusCode.value()
```
ezimanyi: I wonder if there's a way to pull this logic below when we map HTTP statuses onto task statuses.  It's a bit confusing that we do some status mapping up here in a catch block, and then other mapping below...which is pre-existing though now we're adding more logic here.

orca-webhook/src/test/groovy/com/netflix/spinnaker/orca/webhook/pipeline/WebhookStageSpec.groovy
```diff
@@ -0,0 +1,70 @@
...
+    expectedWaitTaskCount * builder.withTask("waitBeforeMonitorWebhook", WaitTask)
+
+    then:
+    expectedMonitorTaskCount * builder.withTask("monitorWebhook", MonitorWebhookTask)
+
+    stage.context.waitTime == expectedWaitTimeInContext
+
+    where:
+    waitForCompletion | waitTime || expectedWaitTimeInContext | expectedWaitTaskCount | expectedMonitorTaskCount
+    true              | 10       || 10                        | 1                     | 1
```
ezimanyi: Can you add a test for what happens if the wait time is input as a string?

marchello2000: thanks, good catch!

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
marchello2000: yes, thanks!

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
marchello2000: will take a peek tomorrow

marchello2000: yeah, i hear you - this method has about 10 too many `return`s... Not sure how to cleanly refactor these without thinking a bunch more about it. Do you have some ideas that come to you?

ezimanyi: Maybe just store the status in this block, then fall back to the status logic below?  Though I'm not sure if any of the intermediate logic still running would be a problem.  If it's not a simple change, no worries...just an idea to improve it if reasonable in scope.

marchello2000: the problem is that then you can't run all the "response.xxx" code. maybe something for next time

ezimanyi: Sounds good!

PR #2985, opened 2019-06-14
feat(plugins): start of plugin system

bpowell:
This is a first step at creating a plugin system for Orca. We started to move out of `orca-core` all things that are required (and dependencies of those requirements) to `orca-api`. As we continue down the path of plugins, we plan on refactoring this to only have what's required for plugin developers in `orca-api`. That also means tightening down the dependencies that are inside of `orca-api`.


orca-core/src/main/java/com/netflix/spinnaker/orca/config/OrcaConfiguration.java
```diff
@@ -75,7 +75,8 @@
   "com.netflix.spinnaker.orca.pipeline.util",
   "com.netflix.spinnaker.orca.preprocessors",
   "com.netflix.spinnaker.orca.telemetry",
-  "com.netflix.spinnaker.orca.notifications.scheduling"
+  "com.netflix.spinnaker.orca.notifications.scheduling",
+  "com.netflix.spinnaker.plugin"
```
robzienert: What's the purpose of this?

bpowell: To be able to component scan plugins. Right now if you create a project including `orca-api` and implement StageDefintionBuilder, it will be component scanned if you drop the resulting jar into `/opt/spinnaker/plugins`. This is just a simple way to test plugins for right now

andrewbackes: I think the intended scope of this PR is to pull out the building blocks that can be used by plugin developers. It would probably be best to exclude the 'loader' code, especially since it is really just in there for temporary testing.

ezimanyi:
How did you decide which modules to pull into `orca-api`, and are the public interfaces of all classes that have been moved now considered extension points that can't be modified without potentially breaking plugins?  It looks like the plan on spinnaker/spinnaker#4181 was that there would be a small set of extension points that would be considered as external contracts---this seems like a larger set of classes than I'd expect to become contracts in an initial plugin implementation.

Also, is there a slightly more detailed plan for how to execute on this than spinnaker/spinnaker#4181?  I'm trying to make sure I see how moving things to `orca-api` fits into the overall plan, what other work will follow, and what authoring a plugin would look like.

cfieber:
This feels significantly larger than I'd have expected. As an example having the SpEL evaluator seems way out of scope for the -api module, OrcaObjectMapper, etc. I'm not sure all the concrete Trigger implementation types should be in api (as opposed to just a Trigger interface).

I don't think this will be successful just relocating files to a new package, I think there needs to be some effort to carve out the right abstraction and move that over and leave the implementation details behind.

We ultimately want this artifact to be pretty stable so ideally it is just model classes and interfaces and not the implementation details of how pipeline evaluation and execution happens.

bpowell:
@ezimanyi @cfieber we pulled out everything that `StageDefintionBuilder` imported (and the things those classes imported). This is not the end goal, this is a stepping stone for us to refactor parts of `orca-api` to only contain the public contract for plugin developers.

We have this doc: https://docs.google.com/document/d/1-n5wGif28Glao_NLkSGlv5LzAeC_i1clntjMhJVfIQI/edit and then if you check in Spinnaker Slack, we have a #plugins channel where we have been discussing issues as they pop up.

cfieber:
Yep, makes sense, I just think there is more to do before we drop this in master. I'd rather see some refactoring done so less stuff comes along in the tangle of extracting stage definition builder.

PR #2988, opened 2019-06-17
fix(jobs): fix race condition in override

ethanfrogers:
fixes a case where parameter overrides would override the base job
configuration as well as the job in the context. this is because the
previous implementation would assign a reference to `context[it]`
instead of a copy. when the parameter overriding was done it would not
only modify the fields in context but also the base configuration. this
presented itself when running > 1 job in parallel. by focing a new copy
(via `converValue` for simplicity) we get a fresh object reference we
can assign.

fixes spinnaker/spinnaker#4487


ethanfrogers:

PR #2992, opened 2019-06-18
fix(task): FindArtifactFromExecutionTask no longer matches its own artifacts

perek:
Background
Currently the FindArtifactFromExecution stage resolves all artifacts for a pipeline and searches back to front for a matching artifact. This is undesirable when resolving from the same pipeline as the resolver will always resolve the same artifact produced from the previous FindArtifactFromExecution stage.

Change
Add new method getArtifactsForPipelineIdStageRef, which resolves artifacts for a given stage ref. This method is used to remove matched artifacts from the pipeline artifact resolver. By removing the artifacts found from the previous stage execution the task will not match artifacts produced by this stage.

Test
Added unit test for ArtifactResolver. Ran test suite. Writing UT for the task proved difficult with all the dependencies, since it currently has no UT i left it like that.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
@@ -73,7 +73,7 @@ public class FindArtifactFromExecutionTask implements Task {
     if (match == null) {
       outputs.put(
           "exception",
-          "No artifacts matching " + expectedArtifact + " found among " + priorArtifacts);
+          "No artifact matching " + expectedArtifact + " found among " + priorArtifacts);
       return TaskResult.builder(ExecutionStatus.TERMINAL)
           .context(new HashMap<>())
           .outputs(outputs)
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
@@ -53,7 +53,8 @@ public class FindArtifactFromExecutionTask implements Task {
     List<Artifact> priorArtifacts;
     // never resolve artifacts from the same stage in a prior execution
     // we will get the set of the artifacts and remove them from the collection
-    String pipelineConfigId = Optional.ofNullable(stage.getExecution().getPipelineConfigId()).orElse("");
+    String pipelineConfigId =
+        Optional.ofNullable(stage.getExecution().getPipelineConfigId()).orElse("");
     if (pipelineConfigId.equals(pipeline)) {
       priorArtifacts =
           artifactResolver.getArtifactsForPipelineIdWithoutStageRef(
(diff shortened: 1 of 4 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
```diff
@@ -58,13 +58,21 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+    }
+
     Artifact match =
         artifactResolver.resolveSingleArtifact(expectedArtifact, priorArtifacts, null, false);
 
     if (match == null) {
       outputs.put(
           "exception",
-          "No artifact matching " + expectedArtifact + " found among " + priorArtifacts);
+          "No artifacts matching " + expectedArtifact + " found among " + priorArtifacts);
```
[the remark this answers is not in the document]
perek: this was the linter running @ezimanyi

perek:
Addressed comments

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
```diff
@@ -58,13 +58,21 @@ public TaskResult execute(@Nonnull Stage stage) {
     List<Artifact> priorArtifacts =
         artifactResolver.getArtifactsForPipelineId(pipeline, executionOptions.toCriteria());
 
+    // never resolve artifacts from the same stage in a prior execution
+    // we will get the set of the artifacts and remove them from the collection
+    if (stage.getExecution().getPipelineConfigId().equals(pipeline)) {
```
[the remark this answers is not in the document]
perek: done

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -199,16 +210,36 @@ public ArtifactResolver(
...
-            .orElse(null);
+    Execution execution = getExecutionForPipelineId(pipelineId, criteria);
 
     return execution == null ? Collections.emptyList() : getAllArtifacts(execution);
   }
 
-  public void resolveArtifacts(@Nonnull Map<String, Object> pipeline) {
+  public @Nonnull List<Artifact> getArtifactsForPipelineIdWithoutStageRef(
+      @Nonnull String pipelineId, @Nonnull String stageRef, @Nonnull ExecutionCriteria criteria) {
+    List<Artifact> artifacts = getArtifactsForPipelineId(pipelineId, criteria);
```
ezimanyi: Would it maybe be a bit cleaner (and a bit faster) to just directly call the new function you wrote `getAllArtifacts()` with a filter excluding the stage `it -> !stageRef.equals(it.getRefId())` than to get all artifacts, then all artifacts that match, and subtract the sets?

perek: only thing we lose is the log line, but i guess we could omit the use of that.

ezimanyi: Yeah, I don't think the log here is too important (at least not important enough to justify the extra code)

PR #2997, opened 2019-06-19
feat(kubernetes): Skip loading events for manifest retrieval in WaitF…

mtweten:
This change makes use of the changes in https://github.com/spinnaker/clouddriver/pull/3796 to skip loading events when retrieving manifests in WaitForManifestStableTask.

We ran into this scenario when we had a lot of deployment pipelines running concurrently, and each deployment involved a decent number of manifests (~30). This task iterates through the manifests involved in the deployment and calls the retrieve manifest API for each one, which (when live manifest mode is enabled) causes a lot of full event loads to occur, which tanks (in our case) the kubernetes API server. 

WaitForManifestStableTask does not actually need any event information to do its job, so in this particular case we can skip loading events and prevent the extra load on the kubernetes api server.


mtweten:
Alright, addressed review comments and the corresponding clouddriver PR has also been updated.

PR #2998, opened 2019-06-19
perf(kubernetes): WaitForManifestStableTask - don't check already com…

mtweten:
…pleted manifests

This change simply uses the existing persisted information (in the context) about stable manifests and failed manifests to prevent rechecking known stable/failed manifests.

When WaitForManifestStableTask executes, it checks all manifests involved in a deployment, even if those manifests have been checked previously and known to be stable or failed. 

This can have some compounding effects when you have a deployment with a lot of manifests, of which most become stable immediately (e.g. configmaps, networkpolicy, etc), and others take many minutes to become stable (e.g. deployments). 

This leads to unnecessary sustained pressure on the kubernetes API server, especially when many concurrent deployments are occurring. It's exacerbated further by the issue described in https://github.com/spinnaker/orca/pull/2997

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableContext.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableContext.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2019 Cerner Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.tasks.manifest;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import lombok.Getter;
+
+@Getter
+public class WaitForManifestStableContext extends HashMap<String, Object> {
+
+  private final Optional<List<String>> messages;
+  private final Optional<Exception> exception;
+  private final Optional<List<Map<String, String>>> stableManifests;
+  private final Optional<List<Map<String, String>>> failedManifests;
+  private final Optional<List> warnings;
+
+  // There does not seem to be a way to auto-generate a constructor using our current version of
+  // Lombok (1.16.20) that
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTask.java
```diff
@@ -63,19 +64,61 @@ public long getTimeout() {
   public TaskResult execute(@Nonnull Stage stage) {
     String account = getCredentials(stage);
     Map<String, List<String>> deployedManifests = manifestNamesByNamespace(stage);
-    List<String> messages = new ArrayList<>();
-    List<String> failureMessages = new ArrayList<>();
-    List<Map<String, String>> stableManifests = new ArrayList<>();
-    List<Map<String, String>> failedManifests = new ArrayList<>();
-    List warnings = new ArrayList<>();
+
+    List<String> messages =
```
ezimanyi: The pattern we've started using to avoid all this casting is to define a model for the stage context, then use `stage.mapTo()` at the beginning of the stage to get the model.  (A good recent example is `DeployManifestTask`.)  I think the model here should be fairly easy to write, just `messages`, `stableManifests`, `failedManifests`, `warnings`, and `exception` and then all this casting would go away.

mtweten:
@ezimanyi Thanks for the suggestion - that helps clean things up a lot. I went ahead and took a crack at it, and also implemented some basic tests just to make sure the context was getting persisted and read as expected.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableContext.java
```diff
@@ -0,0 +1,75 @@
...
+  private final Optional<Exception> exception;
+  private final Optional<List<Map<String, String>>> stableManifests;
+  private final Optional<List<Map<String, String>>> failedManifests;
+  private final Optional<List> warnings;
+
+  // There does not seem to be a way to auto-generate a constructor using our current version of
+  // Lombok (1.16.20) that
+  // Jackson can use to deserialize.
+  public WaitForManifestStableContext(
+      @JsonProperty("messages") Optional<List<String>> messages,
```
mtweten: Went the route of using Optionals for the getters here since it makes it nice to either get the existing context field or initialize to a new array list in the main task logic using orElseGet.

mtweten: In retrospect I can also just do that logic in this constructor - set to the deserialized value if non-null or initialize a new array list otherwise). Really no reason to do it in the task logic. Thoughts?

ezimanyi: Yeah, I wonder if just having the constructor do `Optional.ofNullable(messages).orElse(...)` would be a bit cleaner---but up to you, I think this is fine too.  (I didn't realize that Jackson had support for using `Optional` in this way, but it must as your test case is exercising this.)

mtweten: Alright, made the context cleaner to consume. In the context I still make use of the Jackson Optional support though, since it essentially does the `Optional.ofNullable(...)` part for me.

PR #3005, opened 2019-06-21
feat(core): Support new artifact model in bake Manifest stage

Jammy-Louie:


Jammy-Louie:
Corresponds to the Deck PR https://github.com/spinnaker/deck/pull/7138

PR #3007, opened 2019-06-25
feat(gce): Add StatefullyUpdateBootDisk{Task,Stage}

plumpy:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/StatefullyUpdateBootImageStage.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private final DynamicConfigService dynamicConfigService;
+
+  @Autowired
+  StatefullyUpdateBootImageStage(DynamicConfigService dynamicConfigService) {
+    this.dynamicConfigService = dynamicConfigService;
+  }
+
+  @Override
+  public void taskGraph(Stage stage, TaskNode.Builder builder) {
```
ezimanyi: The parent has `stage` `@Nonnull` so this probably should as well

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/gce/StatefullyUpdateBootImageTask.java
```diff
@@ -0,0 +1,97 @@
...
+  private final TargetServerGroupResolver resolver;
+
+  @Autowired
+  StatefullyUpdateBootImageTask(KatoService katoService, TargetServerGroupResolver resolver) {
+    this.katoService = katoService;
+    this.resolver = resolver;
+  }
+
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
```
ezimanyi: Return should be `@Nonnull` to match parent

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/StatefullyUpdateBootImageStage.java
```diff
@@ -0,0 +1,59 @@
+/*
```
ezimanyi: This isn't really important, but the format of this header looks different than what we usually use (two `*` on every line).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/StatefullyUpdateBootImageStage.java
plumpy: Oh, right. In google3, everything is assumed @Nonnull by default (unless annotated @Nullable) so I removed them. Old habits die hard :)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/StatefullyUpdateBootImageStage.java
plumpy: Oops, I need to get my copyright plugin squared away.

PR #3011, opened 2019-06-26
fix(bake): Lookup artifact details from all upstream stages

marchello2000:
Imagine following setup:
```
Pipeline A
 -> Jenkins job (produces foo.*.deb)
 -> Run Pipeline B
    -> Jenkins job (produces bar.*.deb)
    -> Bake package foo
```

In this case, the package lookup for bake stage will fail to lookup artifact
details for `foo.deb` to pass to the bakery. Thus the bakery is free to pick the
latest artifact matching the name which can be the wrong artifact.

This change allows package look up to traverse up parent pipeline stages, similar
to `FindImage` tasks

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
@@ -30,6 +30,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TreeTraversingParser;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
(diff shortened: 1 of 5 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -402,6 +403,106 @@ public Task taskById(String taskId) {
...
+  /**
+   * Computes all ancestor stages, including those of parent pipelines
+   *
+   * @return list of ancestor stages
+   */
+  public List<Stage> ancestorsWithParentPipelines() {
+    return Stage.getAncestors(this, this.execution);
+  }
+
+  private static List<Stage> getAncestors(Stage stage, Execution execution) {
```
marchello2000: note: this is basically a conversion from groovy (from DeploymentDetailsAware.groovy) and move into the `Stage` so it can be reused by both findimage and bake.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -402,6 +403,106 @@ public Task taskById(String taskId) {
...
+              .findFirst();
+
+      if (parentPipelineStage.isPresent()) {
+        allPriorStages.addAll(getAncestors(parentPipelineStage.get(), parentPipelineExecution));
+      } else {
+        List<Stage> parentPipelineStages = new ArrayList<>(parentPipelineExecution.getStages());
+        parentPipelineStages.sort(
+            new Comparator<Stage>() {
+              @Override
+              public int compare(Stage s1, Stage s2) {
```
marchello2000: i tried doing `Comparator.comparingNullLast(Comparator.comparing(stage::getEndTime))` but it was throwing NPE, if you know what I was doing wrong, would love to simplify this to use an off the shelf comparator

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -402,6 +403,106 @@ public Task taskById(String taskId) {
...
+                + stage.getId()
+                + " with parent "
+                + stage.getParentStageId());
+      }
+
+      allPriorStages.add(parent.get());
+      allPriorStages.addAll(getAncestors(parent.get(), execution));
+    } else if ((execution.getType() == PIPELINE)
+        && (execution.getTrigger() instanceof PipelineTrigger)) {
+      PipelineTrigger parentTrigger = (PipelineTrigger) execution.getTrigger();
```
emjburns: This goes only one level back, correct? So your package must be created either in the existing pipeline or its parent, but not its grandparent.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
aravindmd: do we have any context on why this time comparison is required ?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
aravindmd: can we add a test case for this case if not present already : parent -> child -> grandchild

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -30,6 +30,7 @@
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TreeTraversingParser;
+import com.google.common.base.Strings;
```
emjburns: Do we need this dependency?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
asher: I think this will go back multiple levels, getAncestors() recurses:

```
      if (parentPipelineStage.isPresent()) {
        allPriorStages.addAll(getAncestors(parentPipelineStage.get(), parentPipelineExecution));
      } else {
```

This could be problematic from a resource consumption standpoint, especially in cases I've seen within the community where users have massive chains of dependent pipelines.

I'd like to see parent pipeline traversal limited to fulfilling a search condition. So instead of "get all stages across all parents, then filter for $condition", "traverse ancestors until $condition".

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: yes, [`Strings.isNullOrEmpty`](https://github.com/spinnaker/orca/pull/3011/files#diff-b9a1511802a5c213abc5e0d0e977f8e6R437) comes from there

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: Yes, @asher is correct - this should recurse until root execution is found.

@asher : can you elaborate on this comment (we can chat in person as well - no rush). I agree it would be more efficient to only recurse until the condition is satisfied. But other than the efficiency are there any other reasons for it that I am not understanding?

Because the stages are returned in order (e.g. closest ancestor first) so the search downstream will find the closest match (assuming they traverse in order which currently all callers do).

To be clear: I like your suggestion and I think it will actually simplify some `findImage` code - just want to make sure I understand it completely since I am not super familiar with various pipeline designs you've seen from the community

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: this is a great mystery... hence my comment about this being a "port to java". I wonder if it's there for unit tests?
Or, more likely, this is a relic of the past that could be removed? @ajordens / @duftler do either of you happen to know the history here? (https://github.com/spinnaker/orca/pull/874)

marchello2000:
@asher : made the function take a predicate and terminate as soon as match is found - a little more code but the consumption side nicer (not to mention it's more efficient, thanks!)
Also added some "grandparent" tests

PR #3016, opened 2019-06-28
feat(ecs): Add support for task definition artifacts

allisaurus:
Add support for deploying an ECS Task Definition as an artifact.

* Implements https://github.com/spinnaker/spinnaker/issues/4070
* Fixes https://github.com/spinnaker/spinnaker/issues/3341

This feature allows Task Definition attributes, including multiple containers, to be defined in an artifact (JSON file) and used in conjunction with pipeline images and other server group settings to deploy to ECS.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
@@ -16,12 +16,16 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.providers.ecs
 
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.ServerGroupCreator
 import com.netflix.spinnaker.orca.kato.tasks.DeploymentDetailsAware
 import com.netflix.spinnaker.orca.pipeline.model.DockerTrigger
 import com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType
 import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.pipeline.util.ArtifactResolver
 import groovy.util.logging.Slf4j
+import javax.annotation.Nullable
 import org.springframework.stereotype.Component
 
 @Slf4j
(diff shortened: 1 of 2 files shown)


allisaurus:
Associated PRs:
https://github.com/spinnaker/deck/pull/7162
https://github.com/spinnaker/clouddriver/pull/3825

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
```diff
@@ -101,4 +93,72 @@ class EcsServerGroupCreator implements ServerGroupCreator, DeploymentDetailsAwar
...
+    if (!description.imageId) {
+      description.imageId = buildImageId(description.registry, description.repository, description.tag)
+    }
+
+    return description.imageId
+  }
+
+  private static class TaskDefinitionArtifact {
+    @Nullable public String artifactId
+    @Nullable public Artifact artifact
```
clareliguori: What if these are both null? What is the behavior if the task artifact resolver cant find an artifact?

allisaurus: Looks like the resolver will continue try to map it into an object and potentially return a null object. I found that [elsewhere ](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryServerGroupCreator.java#L80) a null check is done after calling it, I can add a similar check for this.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
```diff
@@ -101,4 +93,73 @@ class EcsServerGroupCreator implements ServerGroupCreator, DeploymentDetailsAwar
...
+      containerToImageMap.put(name, resolvedImageAddress)
+    }
+    return containerToImageMap
+  }
+
+  private String getImageAddressFromDescription(Map<String, Object> description, Stage givenStage) {
+    if (description.fromContext) {
+      if (givenStage.execution.type == ExecutionType.ORCHESTRATION) {
+        // Use image from specific "find image from tags" stage
+        def imageStage = givenStage.findAncestor({
```
clareliguori: Merge conflict fail: I think this is supposed to be "ancestorsWithParentPipelines.find" now based on the diff in line 52

allisaurus: Hmm seems to be ok now? Not sure if something upstream changed to permit that.

PR #3017, opened 2019-07-01
Fix generated pom.xml to omit invalid lombok declaration

Aloren:
compileOnly dependency should not be added to generated pom.xml,
see discussion: https://discuss.gradle.org/t/publishing-plugin-should-respect-compileonly-configuration/22903/2
see related commit: nebula-plugins/nebula-publishing-plugin@a5432aa

Currently all published `pom.xml` contain:
```
    <dependency>
      <groupId>org.projectlombok</groupId>
      <artifactId>lombok</artifactId>
      <version/>
      <scope>provided</scope>
    </dependency>
```
which is considered invalid by Maven and thus all transitive dependencies of such artifacts are omitted.


Aloren:
changed commit message as requested by @spinnakerbot

PR #3019, opened 2019-07-02
feat(core): Support deserializing `Execution.initialConfig`

ajordens:
Also adds support for using `initialConfig.enabled` as a short circuit
of the `EnabledPipelineValidator` check.

This seems reasonable as a user can already submit pipeline json against
an arbitrary pipeline config id (that may or may not exist).

Providing an ability to explicitly override the enabled check seems
harmless.


robzienert:
Oh, huh. That's what `initialConfig` is for?

PR #3020, opened 2019-07-02
feat(plugins): start of new plugin system

bpowell:
Factored out all of the things required to create a Stage into `orca-api`. `orca-api` still has dependencies on jackson, but I think we can do the proxying/mixins of `Stage` and `Execution` classes in a later PR to get the ball rolling on plugins.


orca-api/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
```diff
@@ -352,142 +351,4 @@ public static Execution newOrchestration(String application) {
   public static Execution newPipeline(String application) {
     return new Execution(PIPELINE, application);
   }
-
```
robzienert: What was the rationale behind making these root-level classes?

bpowell: To make things a little easier to deal with. I'm not that big of a fan of having static inner classes. This allows for things to be split a little bit better going forward. If we don't need some of those classes in the `orca-api`, we can put them back in `orca-core`

bpowell: @robzienert ^^

robzienert: I'm pretty gun-shy about inner classes, too. In this case however, the reason they are inner classes is because they are types that define a group of related attributes of a type, in this case `Execution`. They're not intended to be reused--and wouldn't necessarily fit--outside of `Execution`, which is why they're not root-level.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/StageDefinitionBuilder.java
```diff
@@ -42,7 +42,7 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
-public interface StageDefinitionBuilder {
+public interface StageDefinitionBuilder extends SimpleStageDefinitionBuilder {
```
ajordens: Is there now a bunch of duplication w/ `SimpleStageDefinitionBuilder`?

orca-api/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
ajordens: This one seems problematic to me as it's significantly and unnecessarily blown out the size of this PR.

I'd like this style of change to be done independently.

bpowell: @robzienert @ajordens moved everything back into `Execution`. Can we get a marge please?

ethanfrogers:
@robzienert @ajordens what's the status on this PR? Any chance we can push it along?

orca-core-tck/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepositoryTck.groovy
```diff
@@ -628,14 +626,14 @@ abstract class ExecutionRepositoryTck<T extends ExecutionRepository> extends Spe
...
-    [RUNNING, SUCCEEDED] | 1     | PIPELINE      | 1
-    [RUNNING, SUCCEEDED] | 1     | ORCHESTRATION | 1
+    []                   | 0     | Execution.ExecutionType.PIPELINE      | ExecutionStatus.values().size()
+    []                   | 0     | Execution.ExecutionType.ORCHESTRATION | ExecutionStatus.values().size()
+    [RUNNING, SUCCEEDED] | 0     | Execution.ExecutionType.PIPELINE      | 2
+    [RUNNING, SUCCEEDED] | 0     | Execution.ExecutionType.ORCHESTRATION | 2
+    []                   | 1     | Execution.ExecutionType.PIPELINE      | 1
+    []                   | 1     | Execution.ExecutionType.ORCHESTRATION | 1
+    [RUNNING, SUCCEEDED] | 1     | Execution.ExecutionType.PIPELINE      | 1
+    [RUNNING, SUCCEEDED] | 1     | Execution.ExecutionType.ORCHESTRATION | 1
```
robzienert: It would be nice if all these refactored import statements got rolled back. It's still a hassle to review the PR as-is.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/StageDefinitionBuilder.java
robzienert: I'd also like to get an answer on this as well. What's the purpose of `SimpleStageDefinitionBuilder` if this one extends it and remains unchanged?

robzienert:
@ethanfrogers It is being pushed along. There's still outstanding feedback / questions.

orca-core-tck/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepositoryTck.groovy
bpowell: 👍 will do

PR #2963, opened 2019-06-05
fix(pipelines): Fix pipeline triggers for some templated pipelines

jervi:
Pipeline triggers and pipeline stages were broken for templated pipelines that use a dynamic source that references an artifact. Resolved by resolving the artifacts a bit earlier (before the preprocessors).

PR #3023, opened 2019-07-04
fix(pipelines): expectedArtifacts inheritance from dynamic parent templates

xabierlaiseca:
When expected artifacts are defined in a template that is sourced dynamically,
expected artifacts are not added to the execution.

Expected artifacts can be defined in `TemplatedPipelineRequest`,
`TemplateConfiguration` and `PipelineTemplate` objects, so a merge strategy
is required. Given expected artifacts have a unique `id`, artifacts with
same `id` in these object will be overridden. The artifacts with higher
priority are the ones in `TemplatedPipelineRequest`, then `TemplateConfiguration`
and finally `PipelineTemplate`.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
@@ -100,7 +100,6 @@ class DependentPipelineStarter implements ApplicationContextAware {
 
     def trigger = pipelineConfig.trigger
     //keep the trigger as the preprocessor removes it.
-    def expectedArtifacts = pipelineConfig.expectedArtifacts
 
     if (parentPipelineStageId != null) {
       pipelineConfig.receivedArtifacts = artifactResolver?.getArtifacts(parentPipeline.stageById(parentPipelineStageId))
@@ -108,12 +107,8 @@ class DependentPipelineStarter implements ApplicationContextAware {
       pipelineConfig.receivedArtifacts = artifactResolver?.getAllArtifacts(parentPipeline)
     }
 
-    def artifactError = null
-    try {
-      artifactResolver?.resolveArtifacts(pipelineConfig)
-    } catch (Exception e) {
-      artifactError = e
-    }
+    // This is required for template source with jinja expressions
+    trigger.artifacts = pipelineConfig.receivedArtifacts
 
     for (ExecutionPreprocessor preprocessor : executionPreprocessors.findAll {
       it.supports(pipelineConfig, ExecutionPreprocessor.Type.PIPELINE)
@@ -121,12 +116,19 @@ class DependentPipelineStarter implements ApplicationContextAware {
       pipelineConfig = preprocessor.process(pipelineConfig)
     }
 
+    pipelineConfig.trigger = trigger
+
+    def artifactError = null
+
+    try {
+      artifactResolver?.resolveArtifacts(pipelineConfig)
+    } catch (Exception e) {
+      artifactError = e
+    }
+
     if (pipelineConfig.errors != null) {
       throw new ValidationException("Pipeline template is invalid", pipelineConfig.errors as List<Map<String, Object>>)
     }
-
-    pipelineConfig.trigger = trigger
-    pipelineConfig.expectedArtifacts = expectedArtifacts
 
     // Process the raw trigger to resolve any expressions before converting it to a Trigger object, which will not be
     // processed by the contextParameterProcessor (it only handles Maps, Lists, and Strings)
--- a/orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarterSpec.groovy
+++ b/orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarterSpec.groovy
@@ -584,4 +584,140 @@ class DependentPipelineStarterSpec extends Specification {
     result.stages[0].refId == "wait1"
     result.stages[0].name == "Wait for 5 seconds"
   }
+
+  def "should trigger v1 templated pipelines with dynamic source using inherited expectedArtifacts"() {
+    given:
+    def triggeredPipelineConfig = [
+      id: "triggered",
+      type: "templatedPipeline",
+      config: [
+        configuration: [
+          inherit: ["expectedArtifacts", "triggers"]
+        ],
+        pipeline: [
+          application: "covfefe",
+          name: "Templated pipeline",
+          template: [
+            source: "{% for artifact in trigger.artifacts %}{% if artifact.type == 'spinnaker-pac' && artifact.name == 'wait' %}{{ artifact.reference }}{% endif %}{% endfor %}"
+          ]
+        ],
+        schema: "1"
+      ]
+    ]
+    def triggeredPipelineTemplate = mapper.convertValue([
+      schema: "1",
+      id: "barebones",
+      configuration: [
+        expectedArtifacts: [[
+                              defaultArtifact: [
+                                customKind: true
+                              ],
+                              id: "helm-chart",
+                              displayName: "helm-chart",
+                              matchArtifact: [
+                                customKind: true,
+                                type: "http/file",
+                                name: "artifact-name"
+                              ],
+                              useDefaultArtifact: false,
+                              usePriorArtifact: false
+        ]]
+      ],
+      stages: [[
+                 id: "bake-manifest",
+                 type: "bakeManifest",
+                 name: "Bake manifest",
+                 config: [
+                   templateRenderer: "HELM2",
+                   inputArtifacts: [[
+                                      account: "my-account",
+                                      id: "helm-chart"
+                                    ]],
+                   expectedArtifacts: [[
+                                         id: "baked-manifest",
+                                         matchArtifact: [
+                                           kind: "base64",
+                                           name: "baked-manifest",
+                                           type: "embedded/base64"
+                                         ],
+                                         useDefaultArtifact: false
+                                       ]],
+                   namespace: "a-namespace",
+                   outputName: "baked-manifest"
+                 ]
+               ]]
+
+    ], PipelineTemplate)
+
+    Artifact testArtifact = new Artifact(
+      type: "http/file",
+      name: "artifact-name",
+      customKind: true,
+      reference: "a-reference"
+    )
+    def parentPipeline = pipeline {
+      name = "parent"
+      trigger = new DefaultTrigger("webhook", null, "test", [:], [testArtifact])
+      authentication = new Execution.AuthenticationDetails("parentUser", "acct1", "acct2")
+    }
+    def executionLauncher = Mock(ExecutionLauncher)
+    def templateLoader = Mock(TemplateLoader)
+    def applicationContext = new StaticApplicationContext()
+    def renderer = new JinjaRenderer(mapper, Mock(Front50Service), [])
+    def registry = new NoopRegistry()
+    def parameterProcessor = new ContextParameterProcessor()
+    def pipelineTemplatePreprocessor = new PipelineTemplatePreprocessor(
+      mapper,
+      new SchemaVersionHandler(
+        new V1SchemaHandlerGroup(
+          templateLoader,
+          renderer,
+          mapper,
+          registry),
+        Mock(V2SchemaHandlerGroup)),
+      new PipelineTemplateErrorHandler(),
+      registry)
+    applicationContext.beanFactory.registerSingleton("pipelineLauncher", executionLauncher)
+    dependentPipelineStarter = new DependentPipelineStarter(
+      applicationContext,
+      mapper,
+      parameterProcessor,
+      Optional.of([pipelineTemplatePreprocessor] as List<ExecutionPreprocessor>),
+      Optional.of(artifactResolver),
+      registry
+    )
+
+    and:
+    def execution
+    1 * executionLauncher.start(*_) >> {
+      execution = it[0]
+      execution = mapper.readValue(it[1], Map)
+      return pipeline {
+        JavaType type = mapper.getTypeFactory().constructCollectionType(List, Stage)
+        trigger = mapper.convertValue(execution.trigger, Trigger)
+        stages.addAll(mapper.convertValue(execution.stages, type))
+      }
+    }
+    1 * templateLoader.load(_ as TemplateConfiguration.TemplateSource) >> [triggeredPipelineTemplate]
+
+    artifactResolver.getArtifactsForPipelineId(*_) >> {
+      return new ArrayList<>()
+    }
+
+    when:
+    def result = dependentPipelineStarter.trigger(
+      triggeredPipelineConfig,
+      null /*user*/,
+      parentPipeline,
+      [:],
+      null,
+      buildAuthenticatedUser("user", [])
+    )
+
+    then:
+    execution.expectedArtifacts.size() == 1
+    execution.expectedArtifacts[0].id == "helm-chart"
+    result.trigger.resolvedExpectedArtifacts.size() == 1
+    result.trigger.resolvedExpectedArtifacts[0].id == "helm-chart"
+  }
 }
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/TemplatedPipelineRequest.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/TemplatedPipelineRequest.java
@@ -16,7 +16,6 @@
 package com.netflix.spinnaker.orca.pipelinetemplate;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact;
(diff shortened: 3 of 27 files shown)


xabierlaiseca:
@spinnaker/reviewers

orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGeneratorSpec.groovy
```diff
@@ -114,6 +115,79 @@ class V1SchemaExecutionGeneratorSpec extends Specification {
...
+    then:
+    result.expectedArtifacts*.id == expectedArtifactIds
+
+    where:
+    inherit               || expectedArtifactIds                                                                || artifactInRequest
+    ['expectedArtifacts'] || ['artifact-from-template', 'artifact-from-configuration']                          || false
+    []                    || ['artifact-from-configuration']                                                    || false
+    []                    || ['artifact-from-request', 'artifact-from-configuration']                           || true
+    ['expectedArtifacts'] || ['artifact-from-template', 'artifact-from-request', 'artifact-from-configuration'] || true
+  }
```
jervi: Nitpick: `||` is by convention used to separate inputs and expected outputs, so each unroll table should only have one column with these. Use `|` otherwise

orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGeneratorSpec.groovy
```diff
@@ -162,4 +236,17 @@ class V1SchemaExecutionGeneratorSpec extends Specification {
...
+    return new HashMap().with {
+      put('id', id)
+      put('displayName', effectiveName)
+      put('defaultArtifact', [customKind: true])
+      put('matchArtifact', [customKind: true, type: 'http/file', name: effectiveName])
+      put('useDefaultArtifact', false)
+      put('usePriorArtifact', false)
+      it
+    }
+  }
```
jervi: This being groovy, you can just
```groovy
    private Map<String, Object> createExpectedArtifact(String id, String name = null) {
	    def effectiveName = name ?: id
	    return [
	      id: id,
	      displayName: effectiveName,
	      defaultArtifact: [customKind: true],
	      matchArtifact: [customKind: true, type: 'http/file', name: effectiveName],
	      useDefaultArtifact: false,
	      usePriorArtifact: false
	    ]
	  }
```

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarterSpec.groovy
```diff
@@ -584,4 +584,145 @@ class DependentPipelineStarterSpec extends Specification {
...
+    then:
+    execution.expectedArtifacts.size() == 1
+    execution.expectedArtifacts[0].id == "helm-chart"
+    result.trigger.resolvedExpectedArtifacts.size() == 1
+    result.trigger.resolvedExpectedArtifacts[0].id == "helm-chart"
+
+//    result.stages.size() == 1
+//    result.stages[0].type == "wait"
+//    result.stages[0].refId == "wait1"
+//    result.stages[0].name == "Wait for 5 seconds"
```
jervi: Maybe remove these if not relevant anymore

orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGeneratorSpec.groovy
xabierlaiseca: fixed

orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGeneratorSpec.groovy
xabierlaiseca: replaced

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarterSpec.groovy
xabierlaiseca: removed

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -100,7 +100,14 @@ class DependentPipelineStarter implements ApplicationContextAware {
...
     //keep the trigger as the preprocessor removes it.
-    def expectedArtifacts = pipelineConfig.expectedArtifacts
+
+    for (ExecutionPreprocessor preprocessor : executionPreprocessors.findAll {
+      it.supports(pipelineConfig, ExecutionPreprocessor.Type.PIPELINE)
+    }) {
+      pipelineConfig = preprocessor.process(pipelineConfig)
+    }
+
+    pipelineConfig.trigger = trigger
```
jervi: Any reason you are moving this back? It breaks what I fixed in #2963.

xabierlaiseca: Apparently it will be required to run resolve artifacts twice. I added a comment in code explaining why

gardleopard:
@spinnaker/google-reviewers PTAL

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -100,14 +100,18 @@ class DependentPipelineStarter implements ApplicationContextAware {
 
     def trigger = pipelineConfig.trigger
     //keep the trigger as the preprocessor removes it.
-    def expectedArtifacts = pipelineConfig.expectedArtifacts
```
ezimanyi: It looks like this is reverting the fix that was made in spinnaker/orca#2430.  Have you tested that you haven't re-introduced the bug that was fixed there?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -100,14 +100,18 @@ class DependentPipelineStarter implements ApplicationContextAware {
...
 
     if (parentPipelineStageId != null) {
       pipelineConfig.receivedArtifacts = artifactResolver?.getArtifacts(parentPipeline.stageById(parentPipelineStageId))
     } else {
       pipelineConfig.receivedArtifacts = artifactResolver?.getAllArtifacts(parentPipeline)
     }
 
+    // For pipelines with dynamic source, it is required to perform the following steps:
+    //   1. Resolve source template artifact so the dynamic source expression can be evaluated.
```
ezimanyi: I'm not sure that I fully understand why we need to resolve artifacts twice here, and have some concerns about now always resolving artifacts twice.  I think partly it's that I don't fully understand what is a "pipeline with dynamic source" is.  Can you please give an example (or point to a test case) for what such a pipeline looks like and how it would get resolved?

xabierlaiseca: with pipeline with dynamic source I meant to have an expression to fetch source pipeline from artifacts such as 
```
"source": "{% for artifact in trigger.artifacts %}{% if artifact.name == 'pipeline-template' && artifact.type == 'http/file' %}http://{{artifact.reference}}{% endif %}{% endfor %}"
```
here a test with dynamic parent: https://github.com/spinnaker/orca/blob/2e07fcc7ecb5bf77e3a3a6a5df9e5d1be9c5eabd/orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarterSpec.groovy#L467

On why we need to resolve artifacts twice, the first time we need it in order to load parent pipeline, once we have loaded the parent pipeline, we need to resolve artifacts again as new artifacts could have been found in parent pipeline.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/TemplatedPipelineRequest.java
```diff
@@ -27,7 +26,7 @@
   String id;
   String schema;
   String type;
-  List<ExpectedArtifact> expectedArtifacts;
+  List<HashMap<String, Object>> expectedArtifacts;
```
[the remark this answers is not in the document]
xabierlaiseca: The only reason to move it to `HashMap<String, Object>` is that it's easier to merge artifacts from different sources given that the other sources are using `HashMap<String, Object>` (`TemplateConfiguration.PipelineConfiguration` and `PipelineTemplate.Configuration`).

I considered changing these 2 configuration classes to use `ExpectedArtifact` instead but many tests failed because of the lack of `Artifact boundArtifact` field in templates and configurations.

Given that orca is not using the expected artifacts in this class, I consider this change involves lower risk than doing a bigger refactoring in order to be able to use `ExpectedArtifact` in the other cases.

xabierlaiseca:
@ezimanyi I hope my comments give you some clarification on this. Let me know if more explanation is needed.

ezimanyi:
@xabierlaiseca : Thanks for the explanation, and the pointer to the test.  Looking at that example, it seems like the only effect of the first `resolveExpectedArtifacts` is just to copy the contents of `pipeline.recievedArtifacts` to `pipeline.trigger.artifacts`.

It seems like it might be less complex and less risky of a change to leave a single pass of artifact resolution (happening after running the pre-processors as you have here) and have the template refer to `{% for artifact in receivedArtifacts %}`.  If I understand how this works (which it's possible I don't fully yet) I *think* that should work.

Alternately I could also imagine just directly setting `trigger.artifacts` where we set `pipelineConfig.receivedArtifacts` a few lines above if for some reason we need to read these from the trigger.  That would avoid actually resolving twice (which has caused issues in the past).

ezimanyi:
The context behind my extreme caution around this is that there were a lot of very subtle bugs in artifact resolution for templated pipelines in the numerous comments on spinnaker/spinnaker#2654.  I spent a bunch of time fixing those, but the code is still more complex and not as well tested than I'd like so I want to be careful about introducing increased complexity.

I have less direct experience with how the V1 template processing works (which is why my comments are more around the artifact resolution in `DependentPipelineStarter`).  Would definitely appreciate a look from @robzienert or @marchello2000 who would be in a better position to comment on that part of it.

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGenerator.java
```diff
@@ -155,4 +153,39 @@ private void addTriggers(
...
+      PipelineTemplate template,
+      TemplateConfiguration configuration,
+      TemplatedPipelineRequest request) {
+
+    List<HashMap<String, Object>> mergedExpectedArtifacts =
+        mergeExpectedArtifact(
+            emptyListIfNull(configuration.getConfiguration().getExpectedArtifacts()),
+            emptyListIfNull(request.getExpectedArtifacts()));
+
+    if (configuration.getConfiguration().getInherit().contains("expectedArtifacts")) {
```
marchello2000: nit: i find this a bit cleaner/easier to parse:

```
	  if (configuration.getConfiguration().getInherit().contains("expectedArtifacts")) {
      mergedExpectedArtifacts = mergeExpectedArtifact(
              emptyListIfNull(template.getConfiguration().getExpectedArtifacts()),
              mergedExpectedArtifacts));
    }

    pipeline.put("expectedArtifacts", mergedExpectedArtifacts);
```

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGenerator.java
```diff
@@ -155,4 +153,39 @@ private void addTriggers(
...
+    } else {
+      pipeline.put("expectedArtifacts", mergedExpectedArtifacts);
+    }
+  }
+
+  private List<HashMap<String, Object>> emptyListIfNull(List<HashMap<String, Object>> eas) {
+    return Optional.ofNullable(eas).orElse(Collections.emptyList());
+  }
+
+  private List<HashMap<String, Object>> mergeExpectedArtifact(
```
marchello2000: just confirming this is the intent (which, looking at that usages, I think it is) if there are duplicates, the artifacts in the `eas2` will "win", right?

marchello2000:
Sorry, just saw this.
I agree with @ezimanyi comments. I don't have context on prior subtleties of failures other than they existed and were painful... @robzienert maybe you can look at this too?

marchello2000:
> @xabierlaiseca : Thanks for the explanation, and the pointer to the test. Looking at that example, it seems like the only effect of the first `resolveExpectedArtifacts` is just to copy the contents of `pipeline.recievedArtifacts` to `pipeline.trigger.artifacts`.
> 
> It seems like it might be less complex and less risky of a change to leave a single pass of artifact resolution (happening after running the pre-processors as you have here) and have the template refer to `{% for artifact in receivedArtifacts %}`. If I understand how this works (which it's possible I don't fully yet) I _think_ that should work.
> 
> Alternately I could also imagine just directly setting `trigger.artifacts` where we set `pipelineConfig.receivedArtifacts` a few lines above if for some reason we need to read these from the trigger. That would avoid actually resolving twice (which has caused issues in the past).

I agree, seems like this should work, @xabierlaiseca what do you think?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/TemplatedPipelineRequest.java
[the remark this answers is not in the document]
robzienert: Method and property types should use an interface type, so these should change to `Map<String, Object>`

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGenerator.java
xabierlaiseca: yes, that is my intent, this way am spinnaker configuration will be able to override an expected artifact in a template

xabierlaiseca:
@ezimanyi @marchello2000 I tried `{% for artifact in receivedArtifacts %}` but `receivedArtifacts` could not be resolved when rendering the template. 

I'll try the alternative way that was mentioned

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/v1schema/V1SchemaExecutionGenerator.java
xabierlaiseca: addressed

xabierlaiseca:
@ezimanyi @marchello2000 can you please have a look to my latest commit https://github.com/spinnaker/orca/pull/3023/commits/6f9408ff675da83e28d87776face4e2a1d4e9df0 , I believe this is what you suggested instead of resolving artifacts twice but I would like to get some feedback about it from you

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/TemplatedPipelineRequest.java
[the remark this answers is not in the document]
xabierlaiseca: @robzienert map type comment addressed

marchello2000:
Thanks, @xabierlaiseca, for reworking this. Unfortunately, I don't understand the subtleties of artifact resolution so I will let @ezimanyi review this

ezimanyi:
@xabierlaiseca : Yes, the latest commit (6f9408f) is indeed what I was suggesting to avoid resolving artifacts twice.  If that works for your use case then I think that would be more robust than always resolving twice.

I think the only remaining question from my perspective is still that it appears this reverts #2430 (which stored and re-set the expected artifacts).  Was that a deliberate change, or will this re-introduce the issue there?

PerGon:
@ezimanyi regarding #2430 , that fix is covered in line https://github.com/spinnaker/orca/pull/3023/files#diff-4f3d52b85acf51b2ed51955fdbbb78e3R165
so that is covered

xabierlaiseca:
@ezimanyi @marchello2000 given @PerGon's comment, I believe the PR is ready to be merged, I've just squashed my commits. Could you please re-review and merge it you consider is now ready?

ezimanyi:
OK, this final version looks good to me!  Thanks for addressing all of the comments!

PR #3024, opened 2019-07-04
feat(cloudformation): treat yaml as string instead of map

gregjones:
To allow the template to contain intrinsic functions (e.g. !Ref)
this changes the processing of the template to consider the parameter
as a string instead of parsing it as a yaml object.

For backwards compatibility, if the value is a map then it will
be converted to a string before it's passed to the task.


gregjones:
cc @xavileon

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/DeployCloudFormationTask.java
```diff
@@ -89,8 +91,11 @@ public TaskResult execute(@Nonnull Stage stage) {
       }
     }
 
-    Map templateBody = (Map) task.get("templateBody");
-    if (templateBody == null || templateBody.isEmpty()) {
+    Object templateBody = task.get("templateBody");
+    if (templateBody instanceof Map) {
+      templateBody = new Yaml().dump(templateBody);
```
xavileon: It is necessary to add the `templateBody` back into the task map after translating to string?

xavileon:
@spinnaker/reviewers PTAL @ethanfrogers I believe we discussed this some time ago, you may be interested.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/DeployCloudFormationTask.java
gregjones: Yes :) Done.

PR #3027, opened 2019-07-08
fix(queue): Propagate auth context on CancelStageHandler

evansb:
This fixes https://github.com/spinnaker/spinnaker/issues/4596

Authentication is not propagated in Cancellable stages such as `CIStage`, so cancellation will fail if using Fiat permissions in build masters.

The fix is to propagate authentication context to the cancellation handler.


marchello2000:
@robzienert I don't understand why Travis won't report status... the build succeeded:
https://travis-ci.org/spinnaker/orca/builds/555899012

PR #3030, opened 2019-07-11
fix(kotlin): spring config classes shouldn't use constructors

asher:


asher:

PR #3031, opened 2019-07-12
fix(RRB): Correctly determine source ASG

marchello2000:
Currently, if there are two ASGs:
(1) Newer one but disabled
(2) Older one but enabled

The RRB strategy will pick ASG (1) as the "source" which is incorrect for the deploy process.
This change takes advantage of the `determineSourceServerGroup` task and uses the ASG from it.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
@@ -17,12 +17,13 @@ package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
 
 import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.DisableClusterStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.ScaleDownClusterStage
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.CreateServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.DetermineTargetServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup
-import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroupResolver
 import com.netflix.spinnaker.orca.front50.pipeline.PipelineStage
 import com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategy
+import com.netflix.spinnaker.orca.kato.pipeline.support.StageData
 import com.netflix.spinnaker.orca.pipeline.WaitStage
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import com.netflix.spinnaker.orca.pipeline.model.SyntheticStageOwner
@@ -32,8 +33,6 @@ import org.springframework.context.ApplicationContext
 import org.springframework.context.ApplicationContextAware
 import org.springframework.stereotype.Component
 import static com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder.newStage
-
-import static com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategySupport.*
 
 @Slf4j
 @Component
@@ -54,9 +53,6 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
 
   @Autowired
   DetermineTargetServerGroupStage determineTargetServerGroupStage
-
-  @Autowired
-  TargetServerGroupResolver targetServerGroupResolver
 
   @Autowired
   ScaleDownClusterStage scaleDownClusterStage
@@ -99,8 +95,6 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
       desired: 0
     ]
 
-
-
     def targetPercentages = stageData.getTargetPercentages()
     if (targetPercentages.isEmpty() || targetPercentages[-1] != 100) {
       targetPercentages.add(100)
@@ -120,10 +114,29 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
       SyntheticStageOwner.STAGE_AFTER
     )
 
+    // Get source ASG from prior determineSourceServerGroupTask
     def source = null
+
     try {
-      source = getSource(targetServerGroupResolver, stageData, baseContext)
-    } catch (TargetServerGroup.NotFoundException e) {}
+      StageData.Source sourceServerGroup
+
+      Stage parentCreateServerGroupStage = stage.directAncestors().find() { it.type == CreateServerGroupStage.PIPELINE_CONFIG_TYPE }
+
+      StageData parentStageData = parentCreateServerGroupStage.mapTo(StageData)
+      sourceServerGroup = parentStageData.source
+
+      if (sourceServerGroup != null && sourceServerGroup.serverGroupName != null) {
+        source = new ResizeStrategy.Source(
+          region: sourceServerGroup.region,
+          serverGroupName: sourceServerGroup.serverGroupName,
+          credentials: stageData.credentials ?: stageData.account,
+          cloudProvider: stageData.cloudProvider
+        )
+      }
+    } catch (Exception e) {
+      // This probably means there was no parent CreateServerGroup stage - which should never happen
+      throw new IllegalStateException("Failed to determine source server group from parent stage while planning RRB flow", e)
+    }
 
     if (source == null) {
       log.warn("no source server group -- will perform RRB to exact fallback capacity $savedCapacity with no disableCluster or scaleDownCluster stages")
--- a/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategySpec.groovy
+++ b/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategySpec.groovy
@@ -5,22 +5,19 @@ import com.netflix.spinnaker.kork.dynamicconfig.SpringDynamicConfigService
 import com.netflix.spinnaker.moniker.Moniker
 import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.DisableClusterStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.ScaleDownClusterStage
-import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.ShrinkClusterStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.DetermineTargetServerGroupStage
-import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup
-import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroupResolver
 import com.netflix.spinnaker.orca.clouddriver.utils.TrafficGuard
 import com.netflix.spinnaker.orca.front50.pipeline.PipelineStage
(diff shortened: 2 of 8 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -120,10 +118,29 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
       SyntheticStageOwner.STAGE_AFTER
     )
 
+    // Get source ASG from prior determineSourceServerGroupTask
     def source = null
+
     try {
-      source = getSource(targetServerGroupResolver, stageData, baseContext)
-    } catch (TargetServerGroup.NotFoundException e) {}
+      StageData.Source sourceAsg
```
ajordens: `s/sourceAsg/sourceServerGroup`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -120,10 +118,29 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
...
+
+      Stage parentCreateAsgStage = stage.ancestors()
+                                        .find() { it.type == CreateServerGroupStage.PIPELINE_CONFIG_TYPE }
+
+      StageData parentStageData = parentCreateAsgStage.mapTo(StageData)
```
ajordens: `s/Asg/ServerGroup`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -120,10 +118,29 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
...
+    // Get source ASG from prior determineSourceServerGroupTask
     def source = null
+
     try {
-      source = getSource(targetServerGroupResolver, stageData, baseContext)
-    } catch (TargetServerGroup.NotFoundException e) {}
+      StageData.Source sourceAsg
+
+      Stage parentCreateAsgStage = stage.ancestors()
+                                        .find() { it.type == CreateServerGroupStage.PIPELINE_CONFIG_TYPE }
```
ajordens: is this what `spotlessApply` gave you ... seems like weird formatting but 🤷‍♀ 

more `s/Asg/ServerGroup`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -120,10 +118,29 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
...
+      if (sourceAsg != null && sourceAsg.serverGroupName != null) {
+        source = new ResizeStrategy.Source(
+          region: sourceAsg.region,
+          serverGroupName: sourceAsg.serverGroupName,
+          credentials: stageData.credentials ?: stageData.account,
+          cloudProvider: stageData.cloudProvider
+        )
+      }
+    } catch (Exception e) {
+      throw new IllegalStateException("Failed to locate source ASG while planning RRB flow", e)
```
ajordens: source server group while planning

ajordens: and we're consciously wanting to throw an exception? (previously it was just swallowed I guess?)

is there any context that you'd like to propagate in the exception message?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategySpec.groovy
```diff
@@ -52,15 +55,20 @@ class RollingRedBlackStrategySpec extends Specification {
       targetHealthyDeployPercentage: 90
     ]
 
-    def stage = new Stage(Execution.newPipeline("orca"), "whatever", ctx)
+    def pipeline = Execution.newPipeline("orca")
+
+    def createAsgStage = new Stage(pipeline, "createServerGroup", ctxCreateASG)
+    def stage = new Stage(pipeline, "whatever", ctx)
+
+    pipeline.stages << createAsgStage << stage
```
ajordens: for what it's worth the `ExecutionBuilder` makes building a pipeline graph slightly nicer.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -56,7 +56,7 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
   DetermineTargetServerGroupStage determineTargetServerGroupStage
 
   @Autowired
-  TargetServerGroupResolver targetServerGroupResolver
+  SourceResolver sourceResolver
```
ajordens: Doesn't look like this is actually used anywhere?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
ajordens: I'm always tentative with `ancestors()` looking only for stage type name ... are we confident this is only going to find what we want?

(vs. say looking at the parent of this stage and then it's synthetic children or something)

marchello2000: as in it could find a CreateServerGroup stage that's completely irrelevant.. that's fair... (it shouldn't but it's brittle, you're right). let me think about this a bit how to clean it up

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
marchello2000: i think an exception here is better - i'd rather it fail early than do something weird. Besides - this would truly be unexpected - nothing in here should throw an exception except for `parentCreateServerGroupStage` being null and then being deref'd which should not happen and if it does i'd rather it fail. we still allow an empty source server group (e.g. first deploy ever)

marchello2000: Let me clean up the error message though. it can be made clearer

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
marchello2000: thanks!

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategySpec.groovy
marchello2000: i was trying to do minimal changes (since the plan is to totally re-write this ut... one day)... going to fix it up since i tweaked some more code anyway

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
marchello2000: I made a helper that only traverses "real" parents, what do you think?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -405,9 +430,16 @@ public Task taskById(String taskId) {
   }
 
   /**
-   * Computes all ancestor stages, including those of parent pipelines
+   * Find the ancestor stage that satisfies the given predicate (including traversing the parent
+   * execution graphs)
+   *
+   * <p>Ancestor stages include: <br>
```
marchello2000: without these `<br>`'s spotless would reformat it all into one line...

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -120,10 +114,29 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
       SyntheticStageOwner.STAGE_AFTER
     )
 
+    // Get source ASG from prior determineSourceServerGroupTask
```
dreynaud: is this comment accurate? Looks like we're pulling the `source` from a `createServerGroup` stage, which has no direct reference to a `determineSourceServerGroupTask`

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -338,21 +338,44 @@ public Task taskById(String taskId) {
...
+  public List<Stage> directAncestors() {
+    if (execution != null) {
+      Set<String> visited = Sets.newHashSetWithExpectedSize(execution.getStages().size());
+      return ImmutableList.<Stage>builder().add(this).addAll(ancestorsOnly(visited, true)).build();
+    } else {
+      return emptyList();
+    }
+  }
+
+  private List<Stage> ancestorsOnly(Set<String> visited, boolean directParentOnly) {
```
dreynaud: I think I can make sense of the `ancestors` and `directAncestors` method names, but I'm not sure what `ancestorsOnly` is supposed to mean 🤔

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
marchello2000: it comes from the parent class: https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/AbstractDeployStrategyStage.groovy#L82

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: yeah, this was here before... but yeah, i don't like it... in my past life I would call a method like this `getAncestorsImpl` for being the actual implementation but I think that's the convention in Java?...

maybe just rename it to `ancestors` or `getAncestors` and have the fact that it's `private` be the indication that it's the "worker"?

dreynaud: up to you, I wouldn't be opposed to `getAncestorsImpl`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
dreynaud: ah I see, thanks

marchello2000:
thanks @dreynaud !

PR #3035, opened 2019-07-12
feat(queue): sql backed pending execution service

asher:
This is intended to become the default `PendingExecutionService` implementation when `executionRepository.sql.enabled` is true but for now must be explicitly enabled via the `queue.sql.pendingExecutionService.enabled` property. 

A forthcoming `DualPendingExecutionService` will allow for a migration from redis to sql without loss of pending queue state.


ajordens:
Any value in metrics / exception handling in the interface methods?

orca-queue-redis/orca-queue-redis.gradle
```diff
@@ -28,6 +28,7 @@ dependencies {
   implementation("com.netflix.spinnaker.keiko:keiko-redis-spring:$keikoVersion")
   implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
 
+  testCompile(project(":orca-queue"))
```
robzienert: I think we want this to be `testImplementation`, no?

PR #3037, opened 2019-07-15
fix(RRB): disable traffic on the source server group only 

marchello2000:
Make sure RRB disable step disabled the same number of instances as it brought online,
and only in the source server group.

Imagine there are multiple active server groups for RRB:

- v001 (capacity: 6, active)
- v002 (capacity: 10, active)

we are deploying v003, with the following steps: 50%, 100%

Currently, RRB will do the following:
1. Spin up 5 instances in v003 (50% of the source which is v002)
2. Disable 3 instances in v001 and 5 in v002 for a total of 8!
3. Spin up another 5 in v003 (100%)
4. Disable 3 and 5 respectively

Effectively we went from capacity 16 to 10 and are now under provisioned.
This is an edge case, but we want it to have predictable behavior that doesn't leave the cluster under provisioned

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
@@ -15,9 +15,9 @@
  */
 package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
 
-import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.DisableClusterStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.ScaleDownClusterStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.CreateServerGroupStage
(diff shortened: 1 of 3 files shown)


marchello2000:
Note: this is based on a branch for another PR (#3031) so you only need to review the last commit https://github.com/spinnaker/orca/pull/3037/commits/9c8ce49558f78bf767acd8884caac25cfccc8efd.

I will rebase when the source branch gets merged

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -33,15 +34,13 @@ import org.springframework.context.ApplicationContextAware
...
-import static com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategySupport.*
-
 @Slf4j
 @Component
 class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
   final String name = "rollingredblack"
 
   @Autowired
-  DisableClusterStage disableClusterStage
+  DisableServerGroupStage disableServerGroupStage
```
dreynaud: very elegant change, my mouth is agape in amazement

marchello2000: i thought it was clever :)

PR #3039, opened 2019-07-15
fix(core): Plan templated pipelines before triggering from start tasks

louisjimenez:
While fixing this [issue](https://github.com/spinnaker/spinnaker/issues/4623), I noticed that the pipeline run stage fails when a pipeline triggers a dependent templated pipeline. This was due to the plan step being omitted by the StartPipelineTask. These changes plan for v2 pipelines and allows them to trigger properly.


orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/tasks/StartPipelineTaskSpec.groovy
```diff
@@ -16,9 +16,11 @@
```
louisjimenez: @ezimanyi here are the changes for this test

ezimanyi: Looks good! :+1:

louisjimenez:

louisjimenez:

PR #3047, opened 2019-07-18
fix(preconfiguredJob): make cluster.application optional

emjburns:
With a titus preconfigured job you can override the cluster definition. If you override the cluster definition, you have to provide an application to run in. This means that all usages of the preconfigured job are in that application.

That's not ideal when you're using application based permissions. This PR allows you to optionally leave out `application` from a provided cluster override. If `application` is missing, we fill it in with the application the stage is running in.


ajordens:
LGTM

PR #3050, opened 2019-07-19
fix(RRB): fix NPE introduced in https://github.com/spinnaker/orca/pull/3031

marchello2000:
When there is no source server group and a validation pipeline has been specified
RRB flow composition will fail (erroneously) trying to construct parameters to be
passed into the validation pipeline (even though the source isn't even necessary!).

This fix does two things:
1. Correctly calculate parameters passed into the validation pipeline
2. Don't even bother composing the flow if we are being called during the `beforeStages`
   since all these stages will be discared anyway (none of the stages have `STAGE_BEFORE` synthetic)

Medium term: all of this should just go away - we shouldn't be using `aroundStages` which is a
source of lots of confusion. I will be refactoring this code to take advantage of newer
`beforeStages` and `afterStages`

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
@@ -15,6 +15,7 @@
  */
 package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
 
+import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.ScaleDownClusterStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.CreateServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.DisableServerGroupStage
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy
```diff
@@ -122,6 +114,14 @@ class RollingRedBlackStrategy implements Strategy, ApplicationContextAware {
 
       Stage parentCreateServerGroupStage = stage.directAncestors().find() { it.type == CreateServerGroupStage.PIPELINE_CONFIG_TYPE }
 
+      if (parentCreateServerGroupStage.status == ExecutionStatus.NOT_STARTED) {
+        // No point in composing the flow if we are called to plan "beforeStages" since we don't have any STAGE_BEFOREs.
```
dreynaud: I don't see an explicit connection between the condition being checked and being called to plan beforeStages, can you elaborate?

marchello2000: Remember this one: https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/StageDefinitionBuilder.java#L83

marchello2000: what happens is:
`DeployStrategyStage` implemented `aroundStages` which calls our `composeFlow`
the `DeployStrategyStage::aroundStages` is called from `StageDefinitionBuilder::beforeStages` (java) which is called from `StageDefinitionBuilder::buildBeforeStages` (kotlin), which is called from `RunStageHandler`

marchello2000:
@dreynaud i am going to merge this so we can potentially deploy it on monday. Let me know if you have any comments

PR #3051, opened 2019-07-19
chore(test): update test to reflect class change

asher:


orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/RestrictExecutionDuringTimeWindowSpec.groovy
```diff
@@ -203,10 +203,9 @@ class RestrictExecutionDuringTimeWindowSpec extends Specification {
     where:
     jitterEnabled | min  | max  | waitTaskCount
     true          | 60   | 600  | 1
-    true          | null | 600  | 1
-    true          | null | null | 0
```
asher: `JitterConfig.{minDelay, maxDelay}` are no longer nullable since `RestrictExecutionDuringTimeWindow` was converted from groovy to java.

PR #3057, opened 2019-07-24
fix(Orca):Bake of HelmBakeManifestRequest error converting YAML to JSON: yaml: control characters are not allowed #4570

neelimamanubolu:
Hi ,
Since UI is not able to load "HELM2" in template renderer, we are facing issue in creation on Bake manifest for Helm. For reference, please find the below link:

https://github.com/spinnaker/spinnaker/issues/4570


ethanfrogers:
@neelimamanubolu this has been fixed in spinnaker/deck#250 and will be available in the next patch release of 1.15 (1.15.1). 

Also, instead of hard coding the value here you can simply set the `templateRenderer` in the stage JSON manually by clicking "Edit Stage as JSON" and adding it there.

PR #3059, opened 2019-07-24
chore(intellij): Add IDEA project files.

plumpy:


ethanfrogers:
Do we need to commit these? Perhaps we could just promote the use of `./gradlew idea`?

PR #3061, opened 2019-07-24
fix(kayenta): plumb through siteLocal for kayenta email support

marchello2000:
This a temporary "hack" until https://github.com/spinnaker/kayenta/issues/600 is addressed.
This changes passes through the `siteLocal` stage context to the kayenta so that it can send
report emails. The `siteLocal` params are currently used at Netflix only but they exist in
OSS `deck`. Medium term, we will remove this from deck and orca


marchello2000:
Going to merge this in so i can get the train started for our testing, just let me know if you have comments and I can  address them after the fact

duftler:
These changes look fine to me.

marchello2000:
thanks @duftler !

PR #3064, opened 2019-07-29
feat(aws/findImage): if present use appversion to filter results

cfieber:
The image query will return a substring match on package name, which can result in an incorrect match if there are multiple packages with the same prefix (e.g. myapp and myappserver).

If there are appversion tags on the image (added by baking) then prefer using those as an indicator of the package name, and do an exact match on the appversion's package name as a filter of the image find results.

If no results pass the appversion filtering, fall back to existing behaviour


emjburns:
LGTM

PR #3069, opened 2019-07-30
feat(webhooks): add support for cancellation to webhooks

marchello2000:
for webhooks that have `waitForCompletion=true` (i.e. monitored webhook) it is
desirable to have an indication of cancellation (either user cancels execution
or the execution is terminated due to some failure)

NOTE: this also deleted fields marked deprecated from webhook response. These have been marked deprecated 1yr ago so it's time to let them go


marchello2000:
@ajordens reviewed offline

PR #3070, opened 2019-07-31
fix(pipelines): an execution only starts one dep pipeline once

emjburns:
If a pipeline that triggers a downstream pipeline is canceled, and that downstream pipeline doesn't limit concurrent pipelines, you can get two duplicate downstream pipeline executions.

This PR adds the `parent execution id` as a `correlationId` so that the downstream pipeline only run once.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
@@ -86,7 +86,7 @@ class DependentPipelineStarter implements ApplicationContextAware {
       parentPipelineStageId: parentPipelineStageId,
       parameters           : [:],
       strategy             : suppliedParameters.strategy == true,
-      correlationId        : parentPipeline.id
+      correlationId        : parentPipeline.id + "_" + pipelineConfig.id + "_" + parentPipeline.startTime
     ]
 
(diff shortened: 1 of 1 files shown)


ajordens:
There are a couple edge cases that this PR doesn't quite cover.

I _think_ creating a correlation id consisting of the parentId + childId + timestamp will be unique enough to support (a) multiple child pipelines and (b) restarts.

emjburns:
@ajordens updated to fix this other cases. Thanks for pointing them out!

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -85,7 +85,8 @@ class DependentPipelineStarter implements ApplicationContextAware {
       parentExecution      : parentPipeline,
       parentPipelineStageId: parentPipelineStageId,
       parameters           : [:],
-      strategy             : suppliedParameters.strategy == true
+      strategy             : suppliedParameters.strategy == true,
+      correlationId        : parentPipeline.id + "_" + pipelineConfig.id + "_" + parentPipeline.startTime
```
marchello2000: why `pipelineConfig.id`? would that change from execution to execution?

emjburns: A parent pipeline might have multiple child pipelines that trigger on completion.
Start time accounts for a restart, since we change the start time if the pipeline restarts.

PR #3073, opened 2019-07-31
feat(expressions): entpoint to eval an expression from a stage

emjburns:
People often try to fix expresions or test how they would evaluate by using the expressions testing endpoint. This runs in a different way than if you were to use an expression in a stage.

This PR adds a new endpoint that takes an execution id and a stage id from that execution. It evaluates the expression as if it was evaluated as that stage.

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
@@ -16,15 +16,10 @@
 
 package com.netflix.spinnaker.orca.controllers
 
-import java.nio.charset.Charset
-import java.time.Clock
-import java.time.ZoneOffset
-import java.time.temporal.ChronoUnit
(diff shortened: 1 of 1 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -516,6 +518,37 @@ class TaskController {
...
+      true
+    )
+    return [result: evaluated?.expression, detail: evaluated?.expressionEvaluationSummary]
+  }
+  
+  /**
+   * Adds trigger and execution to stage context so that expression evaluation can be tested.
+   * This is not great, because it's brittle, but it's very useful to be able to test expressions.
+   */
+  Map<String, Object> augmentContext(Stage stage) {
```
marchello2000:
```suggestion
  private Map<String, Object> augmentContext(Stage stage) {
```

PR #3074, opened 2019-08-01
fix(logging context header): add pipeline description details to MDC

aravindmd:
- Adding the pipeline description to `PIPELINE-DESCRIPTION` header , this will be consumed and set in `clouddriver` Eg  `X-Titus-CallReason` while making calls to titus.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
@@ -311,7 +311,9 @@ class RunTaskHandler(
     try {
       MDC.put("stageType", type)
       MDC.put("taskType", taskModel.implementingClass)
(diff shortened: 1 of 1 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -310,7 +311,7 @@ class RunTaskHandler(
     try {
       MDC.put("stageType", type)
       MDC.put("taskType", taskModel.implementingClass)
-
+      MDC.put(AuthenticatedRequest.Header.makeCustomHeader("pipeline-description"), execution.description)
```
marchello2000: you decided to use `.description` over `.name`?
In that case, since the `.description` is mostly empty, probably better to set it only when the description is nonempty

aravindmd: on second thought , i think it's better to use `.name` here , also added the null check for sanity.

PR #3077, opened 2019-08-02
fix(clouddriver): Deterministic request IDs for cloud operations

robzienert:
While working on idempotent cloud operations, we ran into an odd bug where the same operation was being processed twice but not being correlated to each other. It came down to the fact that each `requestOperation` creates its own UUID, rather than something more deterministic.

We want deterministic clientRequestIds for cloud operations because tasks & stages can be re-driven. Rather than restarting an operation from the beginning (which could be undesirable, like creating multiple server groups), we want to consistently signal that a particular operation is the same.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoService.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoService.groovy
@@ -16,8 +16,11 @@
 
 package com.netflix.spinnaker.orca.clouddriver
 
+
+import com.google.common.hash.Hashing
 import com.netflix.spinnaker.orca.clouddriver.model.Task
 import com.netflix.spinnaker.orca.clouddriver.model.TaskId
+import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 import rx.Observable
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoService.groovy
```diff
@@ -48,4 +55,22 @@ class KatoService {
...
+   * It is VERY IMPORTANT that callSiteIds are constant values between invocations, and are
+   * unique from other callSiteIds within the same Stage.
+   *
+   * A Stage's startTime is included to support the case where a Stage has been restarted. In
+   * this case, we would want to treat an operation request as different from a prior request.
+   */
+  @Nonnull
+  static String requestId(@Nonnull Stage stage, @Nonnull String callSiteId) {
+    return "${stage.getId()}:${stage.getStartTime()}:$callSiteId".toString()
+  }
```
robzienert: Would like scrutiny of this: Is my logic sound? Are there cases where this requestId generation strategy would cause bugs?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/snapshot/RestoreSnapshotTask.groovy
```diff
@@ -36,7 +36,7 @@ class RestoreSnapshotTask extends AbstractCloudProviderAwareTask implements Task
     String cloudProvider = getCloudProvider(stage)
     String account = getCredentials(stage)
     //TODO(nwwebb) emit events to echo for every resource that is being restored or destroyed
-    def taskId = kato.requestOperations(cloudProvider, [[restoreSnapshot: stage.context]])
+    def taskId = kato.requestOperations(stage.getId(), cloudProvider, [[restoreSnapshot: stage.context]])
```
cfieber: should this be a `KatoService.requestId`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoService.groovy
cfieber: I can 100% guarantee this will cause no bugs

cfieber: 1000% now

ajordens: Don’t we still want a `stageId` and `stageStartTime` in here?

robfletcher: I'm confused, why don't I see this change in the "Files changed" tab there?

ajordens: @robfletcher this PR was originally 107 files with an alternative implementation for generating the hash ... it's subsequently been trimmed down to 1 file which confused github.

robzienert: Ugh, derp. I didn't mean to merge. Will followup with the stage stuff. :(

robzienert: https://github.com/spinnaker/orca/pull/3080

PR #3081, opened 2019-08-05
fix(Orca): pipelines/templates : Pipleline template issue - Not able …

neelimamanubolu:
Hi ,
From UI , the  parameter "tag"is not able to load as "latest" in template renderer, which is default for every Pipeline template. For reference, please find the below link: https://github.com/spinnaker/spinnaker/issues/4670

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
@@ -169,6 +169,11 @@ public class Execution implements Serializable {
    */
   @JsonIgnoreProperties(value = "execution")
   public @Nonnull List<Stage> getStages() {
+    if (stages.size() == 0) {
+      setStartTime(new Date().getTime());
+      setEndTime(new Date().getTime());
+    }
+
     return stages;
   }
 


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/v2/DeleteV2PipelineTemplateTask.java
```diff
@@ -45,7 +45,7 @@ public TaskResult execute(Stage stage) {
     }
 
     String templateId = (String) stage.getContext().get("pipelineTemplateId");
-    String tag = (String) stage.getContext().get("tag");
+    String tag = "latest"; // (String) stage.getContext().get("tag");
```
marchello2000: Are you sure the intent here to always delete the `latest` no matter what the user specified in the payload?
If so, the commented-out line should be removed - it's confusing.

But this seems wrong, even if the UI doesn't allow the user to specify the template `tag` to delete, someone can do so by directly manipulating the `JSON` of the pipeline and then the behavior will be wrong. I assume what you really want here is to use `latest` only if nothing was specified

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/v2/CreateV2PipelineTemplateTask.java
```diff
@@ -66,6 +66,7 @@ public TaskResult execute(Stage stage) {
     validate(pipelineTemplate);
 
     String tag = (String) stage.getContext().get("tag");
+
```
marchello2000: no needed

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/tasks/v2/DeleteV2PipelineTemplateTask.java
louisjimenez: Agreed, we wouldn't want to hardcode this since we need flexibility to delete different tags (stable, experimental, etc.). 

The long term fix to the UI failing to delete templates is to display the underlying template versions and allow more fine grained deletion. For example, rendering the version hashes and allowing users to tag or delete them. 

We could also modify the delete endpoint to support a cleanup operation that would delete all versions of a template and update the UI to make use of that.

Finally, an existing workaround to this is to have the user use spin and pass tag arguments. I will update the issue with some more detail on that. 

@neelimamanubolu let me know if any of those options seem appealing and I can share some more detail on template versions.

neelimamanubolu: Then can  we send tag as "latest" and delete the pipeline template, and make the other template as latest..? Is this fine..?

louisjimenez: Is the proposal to modify deck so that it request for latest to be deleted? This could be a short term workaround, it would need to be tested out. It will leave some versions in front50 but a cleanup mechanism could be offered down the line.

PR #3082, opened 2019-08-05
fix(runjob): add account to destroy task context

ethanfrogers:
the kubernetes v2 provider requires the account to determine which
provider version to use for the task converter.


ethanfrogers:
Related to spinnaker/clouddriver#3919

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStageContext.java
```diff
@@ -82,4 +83,12 @@ public String getCredentials() {
   public void setCredentials(String credentials) {
     this.credentials = credentials;
   }
+
+  public String getAccount() {
```
ezimanyi: Maybe lombok to auto-generate all the getters/setters on this class?

PR #3084, opened 2019-08-06
fix(alibabacloud): modify creator to ensure it can get the imageId from the last pipeline

xiaozhu36:
This PR implements `DeploymentDetailsAware` and aims to fix the issue that `deploy stage` can not get the ImageId from `Bake` or `FindImageFromCluster`.


xiaozhu36:
Duplicate with #3091

PR #3085, opened 2019-08-07
fix(webhooks): correctly detect DNS failures

marchello2000:
It appears as though we can get `UnknownHostException` in several different ways.
Making the exception handler for webhook stages a bit more robust and less constrained
when looking for `UnknownHostException` in the cause to make sure we retry these failures


marchello2000:
> I approve this change

Thank you

PR #3086, opened 2019-08-08
feat(runJob): support artifacts in k8s run job

ethanfrogers:
adds support for artifacts in the v2 kubernetes run job implementation.
follows the same patterns as deploy manifest. the KubernetesJobRunner
class is responsible for fetching artifacts and setting up the operation
description.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/RunJobManifestContext.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/RunJobManifestContext.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2019 Armory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.tasks.manifest;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+import lombok.Builder;
+import lombok.Value;
+
+@Builder(builderClassName = "RunJobManifestContextBuilder", toBuilder = true)
+@JsonDeserialize(builder = RunJobManifestContext.RunJobManifestContextBuilder.class)
+@Value
+public class RunJobManifestContext implements ManifestContext {
+  @Nullable private Map<Object, Object> manifest;
+  @Nullable private List<Map<Object, Object>> manifests;
+
+  @Builder.Default private Source source = Source.Text;
+
+  private String manifestArtifactId;
+  private Artifact manifestArtifact;
+  private String manifestArtifactAccount;
+
+  private List<String> requiredArtifactIds;
+  private List<BindArtifact> requiredArtifacts;
+
+  @Builder.Default private boolean skipExpressionEvaluation = false;
+
+  @Override
+  public List<Map<Object, Object>> getManifests() {
+    if (manifest != null) {
+      return Arrays.asList(manifest);
+    }
+    return manifests;
+  }
+
+  @JsonPOJOBuilder(withPrefix = "")
+  public static class RunJobManifestContextBuilder {}
+}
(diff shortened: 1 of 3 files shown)


ethanfrogers:
Related to - https://github.com/spinnaker/spinnaker/issues/4605

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesJobRunnerSpec.groovy
```diff
@@ -0,0 +1,149 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
maggieneterval:
```suggestion
 * Copyright 2019 Armory
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/RunJobManifestContext.java
```diff
@@ -0,0 +1,64 @@
...
+    if (manifest != null) {
+      List<Map<Object, Object>> mfests = new ArrayList<>();
+      mfests.add(manifest);
+      return mfests;
+    }
+    return manifests;
+  }
+
+  @Override
+  public boolean isSkipExpressionEvaluation() {
```
maggieneterval: for a second I thought we were going to add the dreaded skip SpEL checkbox to another stage and I was sad and then very relieved :sweat_smile:

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/RunJobManifestContext.java
```diff
@@ -0,0 +1,64 @@
...
+  private Artifact manifestArtifact;
+  private String manifestArtifactAccount;
+
+  private List<String> requiredArtifactIds;
+  private List<BindArtifact> requiredArtifacts;
+
+  @Builder.Default private boolean skipExpressionEvaluation = false;
+
+  @Override
+  public List<Map<Object, Object>> getManifests() {
```
maggieneterval: might have missed some context (pun intended) but why might the manifest be under either the singular or plural key?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/RunJobManifestContext.java
ezimanyi: Should this return the value of `skipExpressionEvaluation`?  Or if we always want to return `false`, should we remove the `skipExpressionEvaluation` member variable?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesJobRunner.java
```diff
@@ -48,6 +56,23 @@ public KubernetesJobRunner(ArtifactResolver artifactResolver, ObjectMapper objec
       operation.putAll(stage.getContext());
     }
 
+    // previous versions of run job did not support artifacts so we'll default to text here
+    String source = (String) stage.getContext().getOrDefault("source", "text");
```
ezimanyi: Would it make sense to move the `stage.mapTo(RunJobManifestContext.class)` up here and then just use `runJobManifestContext.getSource()` to avoid the casting and string matching?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/kubernetes/KubernetesJobRunnerSpec.groovy
ezimanyi: s/Netflix/Armory/

ethanfrogers: I'll figure this out....eventually.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/RunJobManifestContext.java
ethanfrogers: Since this stage only supports single Job manifests the original implementation has the text manifest at `manifest`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/RunJobManifestContext.java
ethanfrogers: I think we can always return `false` here since the use case for `skipExpressionEvaluation` is to avoid parsing `${}` in bash scripts. Since this stage only supports single `Job` manifests we can avoid doing that.

ezimanyi: That makes sense, but in that case would we want to remove the member `skipExpressionEvaluation` from the class as it could be confusing that it can be set but is always ignored.

PR #3090, opened 2019-08-09
fix(runJob): inject manifest fcr for k8s runjob

ethanfrogers:
if the job is for kubernetes v2 (denoted by either `source` or
`manifest` keys), inject a couple of manifest refresh stages to prevent
cache misses when `WaitForJobCompletion` runs. this isn't the most ideal
way to implement this, however. we should eventually abstract this out
into the provider specific job runner classes.


ethanfrogers:
Working on figuring out why the build failed. Will merge once I fix it.

ethanfrogers:

ethanfrogers:

PR #3091, opened 2019-08-12
fix(alibabacloud): modify creator to ensure it can get the imageId from the last pipeline

xiaozhu36:
This PR implements `DeploymentDetailsAware` and aims to fix the issue that `deploy stage` can not get the ImageId from `Bake` or `FindImageFromCluster`.


xiaozhu36:
HI @robzienert Can you review and merge this PR? Please let me know if there is any issue. Thanks in advance.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/alicloud/AliCloudServerGroupCreator.groovy
```diff
@@ -0,0 +1,77 @@
...
+ */
+package com.netflix.spinnaker.orca.clouddriver.tasks.providers.alicloud
+
+import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.ServerGroupCreator
+import com.netflix.spinnaker.orca.kato.tasks.DeploymentDetailsAware
+import com.netflix.spinnaker.orca.pipeline.model.Stage
+import org.springframework.stereotype.Component
+
+@Component
+class AliCloudServerGroupCreator implements ServerGroupCreator, DeploymentDetailsAware {
```
robzienert: Please do not introduce new Groovy files into the `main` source set. Use Java instead.

xiaozhu36: HI @robertsigler I also want to use Java, but I found `DeploymentDetailsAware` is Groovy owned type and I can not implement it in the Java. Do you have any idea?

robzienert: That shouldn't be the case. We implement Groovy interfaces with Java all over the codebase.

xiahuai: Hi @robzienert , DeploymentDetailsAware is not a common Java interface but a Groovy trait, if I implement it , I must rewrite the default implement in DeploymentDetailsAware. Here is the usage of trait from Groovy docs:
"This means that even if a trait can be seen from a Java class as a regular interface, that interface will not have default methods, only abstract ones."

PR #3092, opened 2019-08-12
[DNMY] feat(authz): More robust handling of managed service accounts

jervi:
This commit partially reverts #2754, so that pipeline ID is always generated in front50, and change the order of save/update pipeline tasks: save pipeline first, then create and update managed service accounts. Also add some monitoring to check that the account has been created. This fixes an issue where you cannot reliably copy a pipeline with roles set if you are running multiple front50 instances.
Depends on https://github.com/spinnaker/front50/pull/567


jervi:
Depends on https://github.com/spinnaker/front50/pull/567

PR #3096, opened 2019-08-13
fix(runJob): inject manifest fcr for k8s runjob (#3090) (#3094)

duftler:
if the job is for kubernetes v2 (denoted by either `source` or
`manifest` keys), inject a couple of manifest refresh stages to prevent
cache misses when `WaitForJobCompletion` runs. this isn't the most ideal
way to implement this, however. we should eventually abstract this out
into the provider specific job runner classes.


duftler:
Manual cherry-pick of: https://github.com/spinnaker/orca/pull/3090

PR #3097, opened 2019-08-13
fix(authorization): Use `EXECUTE` privilege for restarting stages

AbdulRahmanAlHamali:
I am not sure if this is a breaking change. Were people counting on the fact that restarting a stage required a `WRITE` permission?

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
@@ -490,7 +490,7 @@ class TaskController {
     }
   }
 
-  @PreAuthorize("hasPermission(this.getPipeline(#id)?.application, 'APPLICATION', 'WRITE')")
+  @PreAuthorize("hasPermission(this.getPipeline(#id)?.application, 'APPLICATION', 'EXECUTE')")
   @RequestMapping(value = "/pipelines/{id}/stages/{stageId}/restart", method = RequestMethod.PUT)
   Execution retryPipelineStage(
     @PathVariable String id, @PathVariable String stageId) {


ajordens:
It looks like there are a handful of `TaskController` operations that still require `WRITE` vs `EXECUTE` (cancel, pause, resume, updatePipelineStage)

I think it'd be worth updating them all.

AbdulRahmanAlHamali:
Alright, I changed it in the places you mentioned, and other places where I felt it also made sense. Let me know what you think

robzienert:
I think we want to keep `deletePipeline` and `deleteTask` with a `WRITE` permission. As an operator / auditor, I would be surprised to learn people could start executions with `EXECUTE`, then cover their tracks with a follow-up delete of the execution.

evansb:
I think this should be a minor version bump (`feat` commit) since it potentially breaks some use cases.

For instance, we have one use case with manual judgment (`updatePipelineStage`) where we have a centralised pipeline that can be initiated self serve but only small set of engineers that can "approve" the execution somewhere in the middle.

AbdulRahmanAlHamali:
Alright I agree with the `deleting`. It is really not part of `executing`. I will set it back to `WRITE`.

As for `updatePipelineStage`, @evansb I totally agree with you. I am actually going to have a similar use-case soon where having a `WRITE` permission on that stage actually solves the issue. So, I will put it back as `WRITE`, but I think another PR should be submitted to provide special authorization on a more granular level to address this use-case. Relying on the fact that it just so happens to work for us is not a good long-term strategy

ajordens:
I believe `updatePipelineStage` was an oversight when `EXECUTE` came into the mix -- for consistency I still think it should be `EXECUTE` vs `WRITE`.

@evansb Is the small set of allowed engineers sufficiently enforceable with a `Check Preconditions` stage after the judgment. I do think there's some value in being able to more explicitly define who can perform a judgment.

evansb:
@ajordens Yes, I think we can use `Check Preconditions` as a stop-gap solution.

Meanwhile we will replace it with custom stage similar to manual judgment with minimum role requirement specific to our use case.

AbdulRahmanAlHamali:
As far as I can see, there is no pipeline expression function that can be used to find the current authorization. The only functions available are the ones defined in this directory:

https://github.com/spinnaker/orca/tree/498bc02a89fdbd4c1fcdc7f6764faa857f8d04d7/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions

But it seems rather simple to implement another one that fetches the authorization. I will submit another PR to do that.

AbdulRahmanAlHamali:
So if everyone agrees on those changes. Can we merge?

AbdulRahmanAlHamali:
To answer the question that I posed earlier:
> can you use Check Precondition to look at the authorization/authentication of the user?
The answer is yes, using this:
```
${#currentStage()['execution']['authentication']['user']}
```
So I think moving everything to `EXECUTE`, and using the expression above should be a good enough solution for now.

We should later figure out a way to provide granular authorization for the manual judgement stage though.

For now, can we merge this please?

PR #3105, opened 2019-08-16
feat(plugins): Added ability to create plugin stages

bpowell:
## Motivation

This adds the ability to add very simple Stages as plugins. These Stages are very limited in scope. This allows people to create custom Spinnaker stages without having to dig into and understand all of the Orca codebase. This also means that stages can be added without needing to create a custom version or wait till pull requests are merged.

 ## How to use

To build a plugin stage, one needs to implement the `Stage` interface. The `getName` method is used to determine the Stage's name. The `execute` method takes in a `StageInput<YourClassHere>`. `YourClassHere` is a class you create that has the definition of everything your `Stage` needs to run. At runtime, every `Stage` will be registered with the `StageResolver` after being turned into an `ApiStageDefinitionBuilder`.

The `ApiStageDefinitionBuilder` will build a `taskGraph` with a single `ApiTask` which introspects the type required by the `execute` method. The pipeline context will be converted to your `YourClassHere` and then will call the `execute` method, passing in the correct information.

A quick example on how to implement the Orca plugin stage API:
MyStageContext.java
```
    public class MyStageContext {
        public String artifactName;
        public String artifactLocation;
    }
```

MyStage.java
```
    @Component
    public class MyStage implements Stage<MyStageContext> {
        public String getName() { return "myStage"; }

        public <MyStageContext> StageOutput execute(StageInput<MyStageContext> stageInput) {
            StageOutput stageOutput = new StageOutput();
            // TODO add your stage execution stuff here!
            return stageOutput;
        }
    }
```

To see a more complete example check out: https://github.com/armory/stage-plugin

 ## Future

This is the start of the work needed to support plugin stages. As we get more feedback from users of plugins and plugin creators, we will make adjustments to what these stages can do. For example, a future enhancement maybe to add the ability to support `RetryableTask`s.

### DIFF
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStage.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStage.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2019 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+
+@Beta
+public interface SimpleStage<T> {
+  <T> SimpleStageOutput execute(SimpleStageInput<T> simpleStageInput);
+
+  String getName();
+}
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStageInput.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStageInput.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2019 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+
+@Beta
+public class SimpleStageInput<T> {
+  private T value;
+
+  public SimpleStageInput(T value) {
+    this.value = value;
+  }
+
+  public void setValue(T value) {
+    this.value = value;
+  }
+
+  public T getValue() {
+    return this.value;
+  }
+}
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStageOutput.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStageOutput.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2019 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+import java.util.Map;
+
+@Beta
+public class SimpleStageOutput {
+  private SimpleStageStatus status;
+
+  public void setStatus(SimpleStageStatus status) {
+    this.status = status;
+  }
+
+  public SimpleStageStatus getStatus() {
+    return this.status;
+  }
+
+  private Map outputs;
+
+  public void setOutputs(Map outputs) {
+    this.outputs = outputs;
+  }
+
+  public Map getOutputs() {
+    return this.outputs;
+  }
+}
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/Stage.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/Stage.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2019 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+
+@Beta
+public interface Stage<T> {
+  <T> StageOutput execute(StageInput<T> stageInput);
+
+  String getName();
+}
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/StageInput.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/StageInput.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2019 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+
+@Beta
+public class StageInput<T> {
+  private T value;
+
+  public StageInput(T value) {
+    this.value = value;
+  }
+
+  public void setValue(T value) {
+    this.value = value;
+  }
+
+  public T getValue() {
+    return this.value;
+  }
+}
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/StageOutput.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/StageOutput.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2019 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+import java.util.Map;
+
+@Beta
+public class StageOutput {
+  private StageStatus status;
+
+  public void setStatus(StageStatus status) {
+    this.status = status;
+  }
+
+  public StageStatus getStatus() {
+    return this.status;
+  }
+
+  private Map outputs;
+
+  public void setOutputs(Map outputs) {
+    this.outputs = outputs;
+  }
+
+  public Map getOutputs() {
+    return this.outputs;
+  }
+}
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/StageResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/StageResolver.java
@@ -18,10 +18,13 @@ package com.netflix.spinnaker.orca;
 
 import static java.lang.String.format;
 
+import com.netflix.spinnaker.orca.api.SimpleStage;
+import com.netflix.spinnaker.orca.pipeline.SimpleStageDefinitionBuilder;
 import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import javax.annotation.Nonnull;
 
 /**
@@ -32,7 +35,9 @@ import javax.annotation.Nonnull;
 public class StageResolver {
   private final Map<String, StageDefinitionBuilder> stageDefinitionBuilderByAlias = new HashMap<>();
 
-  public StageResolver(Collection<StageDefinitionBuilder> stageDefinitionBuilders) {
+  public StageResolver(
+      Collection<StageDefinitionBuilder> stageDefinitionBuilders,
+      Collection<SimpleStage> apiStages) {
     for (StageDefinitionBuilder stageDefinitionBuilder : stageDefinitionBuilders) {
       stageDefinitionBuilderByAlias.put(stageDefinitionBuilder.getType(), stageDefinitionBuilder);
       for (String alias : stageDefinitionBuilder.aliases()) {
@@ -46,6 +51,13 @@ public class StageResolver {
         }
 
         stageDefinitionBuilderByAlias.put(alias, stageDefinitionBuilder);
+      }
+    }
+
+    if (!Objects.equals(apiStages, null)) {
+      for (SimpleStage stage : apiStages) {
+        SimpleStageDefinitionBuilder builder = new SimpleStageDefinitionBuilder(stage);
+        stageDefinitionBuilderByAlias.put(stage.getName(), builder);
       }
     }
   }
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/config/OrcaConfiguration.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/config/OrcaConfiguration.java
@@ -25,6 +25,7 @@ import com.netflix.spinnaker.kork.expressions.ExpressionFunctionProvider;
 import com.netflix.spinnaker.orca.StageResolver;
 import com.netflix.spinnaker.orca.Task;
 import com.netflix.spinnaker.orca.TaskResolver;
+import com.netflix.spinnaker.orca.api.SimpleStage;
 import com.netflix.spinnaker.orca.commands.ForceExecutionCancellationCommand;
 import com.netflix.spinnaker.orca.events.ExecutionEvent;
 import com.netflix.spinnaker.orca.events.ExecutionListenerAdapter;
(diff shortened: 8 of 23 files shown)


orca-api/src/main/java/com/netflix/spinnaker/orca/api/Stage.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+
+@Beta
+public interface Stage<T> {
+  <T> StageOutput execute(StageInput<T> stageInput);
```
ncknt: The `<T>` can be removed.

orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStage.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+
+@Beta
+public interface SimpleStage<T> {
+  <T> SimpleStageOutput execute(SimpleStageInput<T> simpleStageInput);
```
ncknt: Remove the first `<T>`

orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStageOutput.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.google.common.annotations.Beta;
+import java.util.Map;
+
+@Beta
+public class SimpleStageOutput {
+  private SimpleStageStatus status;
+
+  public void setStatus(SimpleStageStatus status) {
+    this.status = status;
```
ncknt: Can you use `@Data` instead of adding getter/setter?

orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStageOutput.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  public SimpleStageStatus getStatus() {
+    return this.status;
+  }
+
+  private Map outputs;
```
ncknt: Can you explicit the type of the Map?

orca-api/src/main/java/com/netflix/spinnaker/orca/api/StageInput.java
```diff
@@ -0,0 +1,36 @@
...
+
+@Beta
+public class StageInput<T> {
+  private T value;
+
+  public StageInput(T value) {
+    this.value = value;
+  }
+
+  public void setValue(T value) {
```
ncknt: `@Data`

orca-api/src/main/java/com/netflix/spinnaker/orca/api/StageOutput.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+import java.util.Map;
+
+@Beta
+public class StageOutput {
+  private StageStatus status;
```
ncknt: Same comments as on `StateInput` (`@Data` and types for Map)

orca-core/src/main/java/com/netflix/spinnaker/orca/StageResolver.java
```diff
@@ -48,6 +53,13 @@ public StageResolver(Collection<StageDefinitionBuilder> stageDefinitionBuilders)
         stageDefinitionBuilderByAlias.put(alias, stageDefinitionBuilder);
       }
     }
+
+    if (!Objects.equals(apiStages, null)) {
```
ncknt: If you want to test if `apiStages` is null, just use `apiStages != null` or just filter out in the loop. 
If you want to test if any stage is null, use something like:
```
if (apiStages.stream().anyMatch(s -> s == null)) {
...
}
```

You could also just:
```
apiStages.stream().forEach(s -> stageDefinitionBuilderByAlias.put(s.getName(), new SimpleStageDefinitionBuilder(s));
```

orca-core/src/main/java/com/netflix/spinnaker/orca/config/OrcaConfiguration.java
```diff
@@ -187,8 +187,12 @@ public TaskResolver taskResolver(Collection<Task> tasks) {
   }
 
   @Bean
-  public StageResolver stageResolver(Collection<StageDefinitionBuilder> stageDefinitionBuilders) {
-    return new StageResolver(stageDefinitionBuilders);
+  public StageResolver stageResolver(
+      Collection<StageDefinitionBuilder> stageDefinitionBuilders,
+      Optional<Collection<SimpleStage>> simpleStages) {
+    Collection<SimpleStage> stages =
+        simpleStages.isPresent() ? simpleStages.get() : new ArrayList<SimpleStage>();
```
ncknt:
```
Collection<SimpleStage> stages = simpleStages.orElseGet(ArrayList::new)
```

orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStageInput.java
```diff
@@ -0,0 +1,36 @@
...
+
+@Beta
+public class SimpleStageInput<T> {
+  private T value;
+
+  public SimpleStageInput(T value) {
+    this.value = value;
+  }
+
+  public void setValue(T value) {
```
ncknt: `@Data`

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
```diff
@@ -0,0 +1,97 @@
...
+        case NOT_STARTED:
+          status = ExecutionStatus.NOT_STARTED;
+          break;
+        default:
+          status = ExecutionStatus.FAILED_CONTINUE;
+          break;
+      }
+    } catch (Exception e) {
+      log.error("Cannot execute stage " + simpleStage.getName());
+      log.error(e.getMessage());
```
ncknt: Move the exception in the log call to preserve the stack trace 
```log.error("Cannot ...", e)```

File: orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java [hunk 11]
```diff
@@ -0,0 +1,97 @@
...
+  SimpleTask(@Nullable SimpleStage simpleStage) {
+    this.simpleStage = simpleStage;
+  }
+
+  @Nonnull
+  public TaskResult execute(@Nonnull Stage stage) {
+    ObjectMapper objectMapper = OrcaObjectMapper.newInstance();
+
+    ExecutionStatus status;
+    SimpleStageOutput outputs = new SimpleStageOutput();
```
ncknt: just `output` not `outputs`

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
```diff
@@ -0,0 +1,97 @@
...
+
+    try {
+      List<Class<?>> cArg = Arrays.asList(SimpleStageInput.class);
+      Method method = simpleStage.getClass().getMethod("execute", cArg.toArray(new Class[0]));
```
ncknt: A few comments would help (getting the StageInputType generic class and using it to deserialize the context). Perhaps also moving the type logic into its own method.

orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStageOutput.java
bpowell: We are trying not to add more dependencies than what we need. Right now we only have one dependency, which is for the `@Beta` annotation . With how small the classes are, we didn't think it it was an issue to create the getter/setters

ncknt: It's just for code readability. I'm fine not adding lombok as a dependency if it's not there already.

robzienert: Lombok is not a runtime dependency, it will be fine to add.

orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStageOutput.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public void setStatus(SimpleStageStatus status) {
+    this.status = status;
+  }
+
+  public SimpleStageStatus getStatus() {
+    return this.status;
+  }
+
+  private Map<String, Object> outputs;
```
robzienert: Please keep properties up at the top, grouped together.

orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStage.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+
+@Beta
+public interface SimpleStage<T> {
```
robzienert: Interfaces must have class- and method-level docs.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
```diff
@@ -0,0 +1,99 @@
...
+      Type inputType = ResolvableType.forMethodParameter(method, 0).getGeneric().getType();
+      Map<TypeVariable, Type> typeVariableMap =
+          GenericTypeResolver.getTypeVariableMap(simpleStage.getClass());
+
+      return new SimpleStageInput(
+          objectMapper.convertValue(
+              stage.getContext(), GenericTypeResolver.resolveType(inputType, typeVariableMap)));
+    } catch (NoSuchMethodException exeception) {
+      log.error("Cannot get method for " + simpleStage.getName() + ": " + exeception.getMessage());
+    }
```
robzienert: Surely we don't want to just continue if the method doesn't exist? This strikes me as something that will lead to a lot of confusing bugs for integrators. We can always loosen up in the future, but making things more restrictive is much harder.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
```diff
@@ -0,0 +1,99 @@
...
+        status = ExecutionStatus.RUNNING;
+        break;
+      case COMPLETED:
+        status = ExecutionStatus.SUCCEEDED;
+        break;
+      case NOT_STARTED:
+        status = ExecutionStatus.NOT_STARTED;
+        break;
+      default:
+        status = ExecutionStatus.FAILED_CONTINUE;
```
robzienert: Why would the default be `FAILED_CONTINUE`?

orca-core/src/test/groovy/com/netflix/spinnaker/orca/StageResolverSpec.groovy
```diff
@@ -67,4 +72,16 @@ class StageResolverSpec extends Specification {
   @StageDefinitionBuilder.Aliases("notAliased")
   class AliasedStageDefinitionBuilder implements StageDefinitionBuilder {
   }
+
+  class SimpleStage implements SimpleStage {
```
robzienert: Let's uh, rename this? `TestSimpleStage` or something. Having a class the same name as an interface is mega confusing.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/SimpleTaskSpec.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    <Object> SimpleStageOutput execute(SimpleStageInput<Object> input) {
+      SimpleStageOutput output = new SimpleStageOutput()
+
+      Map<String, String> stageOutput = new HashMap<>()
+      stageOutput.put("hello", "world")
+
+      output.setStatus(SimpleStageStatus.COMPLETED)
+      output.setOutputs(stageOutput)
```
robzienert: 🤔 Wonder if we should add a generic output type as well?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
```diff
@@ -0,0 +1,99 @@
...
+
+    try {
+      List<Class<?>> cArg = Arrays.asList(SimpleStageInput.class);
+      Method method = simpleStage.getClass().getMethod("execute", cArg.toArray(new Class[0]));
+      Type inputType = ResolvableType.forMethodParameter(method, 0).getGeneric().getType();
+      Map<TypeVariable, Type> typeVariableMap =
+          GenericTypeResolver.getTypeVariableMap(simpleStage.getClass());
+
+      return new SimpleStageInput(
+          objectMapper.convertValue(
```
ajordens: Is this similar to `stage.mapTo(GenericTypeResolver.resolveType(..))`

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
```diff
@@ -0,0 +1,99 @@
...
+      case NOT_STARTED:
+        status = ExecutionStatus.NOT_STARTED;
+        break;
+      default:
+        status = ExecutionStatus.FAILED_CONTINUE;
+        break;
+    }
+
+    return TaskResult.builder(status)
+        .context(output.getOutputs())
```
ajordens: This doesn't seem quite what we want ... we need to be careful about what gets made available to downstream stages vs added back into the local stage context.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
ajordens: Should we just raise an exception on an unknown status ... `FAILED_CONTINUE` doesn't seem like the appropriate default.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/SimpleTaskSpec.groovy
ajordens: Probably.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
ajordens: Agreed, I'd suggest just raising one of our system exceptions.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  @Nonnull
+  public TaskResult execute(@Nonnull Stage stage) {
+    ExecutionStatus status;
+    SimpleStageInput simpleStageInput = getStageInput(stage);
+    SimpleStageOutput output = simpleStage.execute(simpleStageInput);
+
+    switch (output.getStatus()) {
+      case TERMINAL:
```
ajordens: Reviewer thoughts on using `FAILED` rather than `TERMINAL`? It looks like we've already flipped to `COMPLETED` from `SUCCEEDED`?

Or alternatively we keep things consistent and use `SUCCEEDED` for both (leaning this way).

orca-api/src/main/java/com/netflix/spinnaker/orca/api/Stage.java
ajordens: +1

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
robzienert: I might be misinterpreting what you're saying, but you're proposing setting the status to `SUCCEEDED` if the output status is `TERMINAL`?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
ethanfrogers: This was my fault. I didn't think that one through. You're right, I don't think we want to dump it back into the context. +1.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
bpowell: @robzienert would we want to throw an exception here as well then?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
bpowell: @ethanfrogers @ajordens do we want to git rid of `SimpleStageOutput` then and have the `execute` method return a `SimpleStageStatus` instead? That would also solve the issue above about making `SimpleStageOutput` use a generic

ethanfrogers: I don't think so. I think `SimpleStageOutput` is fine. I def want to be able to output things from a stage so that it lands back in the overall pipeline context. I think @ajordens point here is that we might not want the things meant for output to effect the context all the time. What if we added a `getContext` method alongside the `getOutput` method. That way the stage could supply both explicitly.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
bpowell: Kinda. This is used to convert the stage context to the generic `StageInput` class to pass into the `execute` method. This way everything is typed

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
ajordens: Sorry, that was unclear.

In `orca-api` we introduced `COMPLETED` in place of `SUCCEEDED`. My suggestion would be to just re-use the same status types and avoid a conversion (ie. just do `ExecutionStatus.valueOf()`).

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
ajordens: Yes to @ethanfrogers, we definitely want a task to be able to return values ... just have some control around how they're exposed (or not) to downstream stages.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
bpowell: @ajordens we can't use `ExecutionStatus` inside `orca-api`, that would lead to circular dependencies between `orca-core` and `orca-api`. But I can make the names of the statues be the same. Does that work?

bpowell: @ajordens never mind. I'm following you now. I was wrong on how `valueOf` works. Pushing up a change for that now. Is this more inline with what you were thinking?

bpowell:
@robzienert @ajordens @ethanfrogers I think I resolved all the issues that were brought up. Is there anything else that you may have noticed? I know I misunderstood some of the comments, but I think I addressed them correctly. Thanks everyone for all of the help on this PR!

orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStage.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+
+/**
+ * Used to create a more simplistic stage. This should be used by plugin developers to create plugin
+ * stages
+ *
+ * @param <T>
```
robzienert: What's `T`? (I know, but a reader won't)

orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStageOutput.java
```diff
@@ -0,0 +1,28 @@
...
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+import lombok.Data;
+
+@Beta
+@Data
+public class SimpleStageOutput<T, U> {
+  private SimpleStageStatus status;
+  private T outputs;
```
robzienert: `output`

orca-api/src/main/java/com/netflix/spinnaker/orca/api/SimpleStage.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.api;
+
+import com.google.common.annotations.Beta;
+
+/**
+ * Used to create a more simplistic stage. This should be used by plugin developers to create plugin
+ * stages
```
robzienert: `... more simplistic stage.` more than what? I'd just get rid of that sentence imo. These docs are a little more anemic than I'd prefer to see, but we can work with them as we go along and we figure more assumptions, etc.

bpowell:
@robzienert @ajordens I think I covered all the things. thanks again for all of the help on this PR! let me know if there is anything else that needs to be done on this please

PR #3115, opened 2019-08-21
fix(tasks): duplicate empty keys ignored

emjburns:
Currently when we fetch a task we pull all the stage context keys into a flatmap called `variables` so that the UI can look at them easily ([here](https://github.com/spinnaker/orca/blob/c11ba0c26e57b81db80e3ef738d914c843368981/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy#L715)). A key that has a null or empty value overwrites a previously set key with a good value. This is a problem in some deploy tasks because the UI uses the `deploy.server.groups` key to show the spiny wheel indicating that a task is working on a server group, and that value gets unset because the last task has the key `deploy.server.groups` with an empty value (default behavior of `ApplySourceServerGroupCapacityTask` I think). 

I discovered this while wondering why the `keel` tasks weren't showing the spiny wheel in the UI. 

This PR changes the behavior of the `variables` population so that if the map _already_ has the key in question, _and_ it would be set to a null or empty value, we disregard that update.

Can anyone think of a situation where we'd actually want to take the last state of a variable, even if it was empty?

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
@@ -712,11 +712,19 @@ class TaskController {
     return aStartTime <=> bStartTime ?: b.id <=> a.id
   }
 
+  static boolean shouldReplace(Map.Entry<String, Object> entry, Map variables) {
+    // a duplicate key with an empty value should
+    // not overwrite a previous key with a non empty value
+    return isNullOrEmpty(variables[entry.key]) || !isNullOrEmpty(entry.value)
+  }
+
   private OrchestrationViewModel convert(Execution orchestration) {
     def variables = [:]
     for (stage in orchestration.stages) {
       for (entry in stage.context.entrySet()) {
-        variables[entry.key] = entry.value
+        if (shouldReplace(entry, variables)) {
+          variables[entry.key] = entry.value
+        }
       }
     }
     new OrchestrationViewModel(
@@ -747,6 +755,19 @@ class TaskController {
       .collect(Collectors.toList())
 
     return pipelineConfigIds
+  }
+
+  private static boolean isNullOrEmpty(Object value) {
+    if (value == null) {
+      return true
+    } else if (value instanceof Collection) {
+      return value.isEmpty()
+    } else if (value instanceof String) {
+      return value.length() == 0
+    } else if (value == [:]) {
+      return true
+    }
+    return false
   }
 
   @VisibleForTesting
--- a/orca-web/src/test/groovy/com/netflix/spinnaker/orca/controllers/TaskControllerSpec.groovy
+++ b/orca-web/src/test/groovy/com/netflix/spinnaker/orca/controllers/TaskControllerSpec.groovy
@@ -32,6 +32,7 @@ import org.springframework.mock.web.MockHttpServletResponse
 import org.springframework.test.web.servlet.MockMvc
 import org.springframework.test.web.servlet.setup.MockMvcBuilders
 import spock.lang.Specification
+import spock.lang.Unroll
 
 import java.time.Clock
 import java.time.Instant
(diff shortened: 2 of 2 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -712,11 +712,24 @@ class TaskController {
...
+    def currentVal = variables[entry.key]
+    return !(!isNullOrEmpty(currentVal) && isNullOrEmpty(entry.value))
+  }
+
   private OrchestrationViewModel convert(Execution orchestration) {
     def variables = [:]
     for (stage in orchestration.stages) {
       for (entry in stage.context.entrySet()) {
-        variables[entry.key] = entry.value
+        if (shouldReplace(entry, variables)) {
```
marchello2000: Since `Map` returns `null` for nonexistent keys, I feel this would be easier to read (and no need for `shouldReplace` function)
```suggestion
        if (isNullOrEmpty(variables[entry.key]) || !isNullOrEmpty(entry.value)) {
```

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -749,6 +762,19 @@ class TaskController {
...
   }
 
+  private static Boolean isNullOrEmpty(Object value) {
+    if (value == null) {
+      return true
+    } else if (value instanceof Collection) {
+      return value.isEmpty()
+    } else if (value instanceof String) {
+      return value.length() == 0
+    } else if (value == [:] || value == []) {
```
marchello2000: Doesn't the `if instanceof Collection` cover `if value == []`?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
emjburns: 👍 that is better

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -712,11 +712,24 @@ class TaskController {
     return aStartTime <=> bStartTime ?: b.id <=> a.id
   }
 
+  static Boolean shouldReplace(Map.Entry entry, Map variables) {
```
robzienert: Why isn't this a primitive `boolean`?

Can we type the key & value of these maps & entries? At least to the point of a `String, Object`?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
emjburns: Used this logic but kept it in a separate function for easy testing.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
emjburns: good catch yes. It just doesn't work for an empty map, which was somewhat surprising to me.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -712,11 +712,19 @@ class TaskController {
...
+    // a duplicate key with an empty value should
+    // not overwrite a previous key with a non empty value
+    return isNullOrEmpty(variables[entry.key]) || !isNullOrEmpty(entry.value)
```
robzienert: This condition is so much better than what it was! 🙏

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
emjburns: Will fix both!

PR #3116, opened 2019-08-21
feat(pipeline): add stage status precondition task

maggieneterval:
Closes: https://github.com/spinnaker/spinnaker/issues/4768
Corresponding Deck PR: https://github.com/spinnaker/deck/pull/7348

Thoughts on surfacing useful error messages in Deck:
- Based on Deck's [expectations](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/orchestratedItem/orchestratedItem.transformer.ts#L171), non-thrown error messages should be set on a task's outputs `exception.details.error` or `exception.details.errors` to be surfaced in the `StageFailureMessage` component. I haven't encountered a standard way to do this in Orca but if you get this message and you know The Right Way please help me out :love_letter: 
- I'm also unsure of any conventions around when to throw an error vs. attach an error to the task context. I went with 1) throw an error when inputs are invalid, such that the task cannot be performed 2) attach an error message to the task context when the task completed but failed (in this case, because the precondition assertion was incorrect).

(Edit: should throw an error in both cases so it gets picked up by RunTaskHandler)

@robfletcher I feel like you might be an expert on error conventions -- any wisdom is appreciated :pray: :pray:

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/StageStatusPreconditionTask.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/StageStatusPreconditionTask.java
@@ -16,11 +16,9 @@
 
 package com.netflix.spinnaker.orca.pipeline.tasks;
 
-import com.google.common.collect.ImmutableMap;
 import com.netflix.spinnaker.orca.ExecutionStatus;
 import com.netflix.spinnaker.orca.TaskResult;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
-import java.util.Map;
 import javax.annotation.Nonnull;
 import lombok.Getter;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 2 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/StageStatusPreconditionTask.java
```diff
@@ -0,0 +1,84 @@
...
+            .findFirst()
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
+                        String.format(
+                            "Failed to find stage %s in execution. Please specify a valid stage name",
+                            stageName)));
+    String actualStatus = foundStage.getStatus().toString();
+    if (!actualStatus.equals(assertedStatus)) {
+      Map<String, Object> outputs =
```
ezimanyi: It might make sense to write a helper function to encapsulate the logic of where to put error messages.

maggieneterval: Agreed, will await the blessing of @robfletcher to make sure this is even the most up to date correct place to surface errors :ear:

marchello2000: I think the only "centralized"/"consistent" place where errors are handled is in the `RunTaskHander` which is basically what you're replicating here. I say, just throw and let https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt#L154 handle it

maggieneterval: @marchello2000 thank you so much, that makes sense! Updated the task and its tests accordingly :+1:

PR #3071, opened 2019-07-31
refactor(core): Use kork-expressions

scottfrederick:
This PR supersedes https://github.com/spinnaker/orca/pull/2649

* Rebase changes from the earlier PR on `master`
* Adapt to changes in `kork-expressions` based on https://github.com/spinnaker/orca/commit/a4815ce1b5f12e551d5d6dc05526a2158775a6a0
* Encapsulate all expression-handling functions in implementations of `ExpressionFunctionsProvider` and rationalize these implementations

depends on https://github.com/spinnaker/kork/pull/348

PR #3118, opened 2019-08-23
feat(core): Add SpEL Function Helpers for Resolved Trigger Artifacts

evansb:
Fixes https://github.com/spinnaker/spinnaker/issues/4783
This MR  adds the following helper functions to Spinnaker Pipeline Expressions

`#triggerResolvedArtifact(name)` gets *one* resolved expected artifact that matches the provided `name`. I used `name` instead of `type` here to make it consistent with `#stage(..)`.

`#triggerResolvedArtifactByType(type)` is similar to `#triggerResolvedArtifact` but uses `type` of the artifact.

### DIFF
--- a/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/expressions/functions/ArtifactExpressionFunctionProviderSpec.groovy
+++ b/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/expressions/functions/ArtifactExpressionFunctionProviderSpec.groovy
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.pipeline.expressions.functions
+
+import com.netflix.spinnaker.kork.artifacts.model.Artifact
+import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact
+import com.netflix.spinnaker.kork.expressions.SpelHelperFunctionException
+import com.netflix.spinnaker.orca.pipeline.model.DefaultTrigger
+import spock.lang.Shared
+import spock.lang.Specification
+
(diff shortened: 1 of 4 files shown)


gnarea:
Thanks for implementing this functionality @evansb! 🙇  Since the changes to the doc have already been merged, I'm wondering what's blocking the merge of this PR?

marchello2000:
I will review today

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/expressions/functions/ArtifactExpressionFunctionProviderSpec.groovy
```diff
@@ -0,0 +1,77 @@
...
+
+import static com.netflix.spinnaker.orca.pipeline.expressions.functions.ArtifactExpressionFunctionProvider.triggerResolvedArtifact
+import static com.netflix.spinnaker.orca.pipeline.expressions.functions.ArtifactExpressionFunctionProvider.triggerResolvedArtifactByType
+import static com.netflix.spinnaker.orca.test.model.ExecutionBuilder.pipeline
+
+
+class ArtifactExpressionFunctionProviderSpec extends Specification {
+  @Shared
+  def pipeline1 = pipeline {
+    def matchArtifact1 = new Artifact(type: "docker/image", "name": "covfefe")
```
marchello2000: nit: maybe not `covfefe` - don't want to (add) to already existing political references in the codebase + it's easier to read if you know what it is, e.g.
```suggestion
    def matchArtifact1 = new Artifact(type: "docker/image", "name": "artifactName1")
```

marchello2000: (and similar below)

marchello2000:
Thanks, @evansb : FYI, there is currently an issue with expressions that doesn't inject the `execution` in to the expressions - someone on our team is working on the fix

evansb:
Hi @marchello2000 , just to make sure understand that correctly, does it mean that these expressions I introduced only work if the evaluation context is the stage (because they have `execution` field?).

So we can interpolate inside stages but not in pipeline notifications, for instance?

marchello2000:
Right now there is an issue where calling the expression from the stage (or anywhere) will fail, if `execution` isn't passed in, e.g. normally you would:
`${#triggerResolvedArtifact('myArtifact')}` and under the covers, `orca` will interpolate it to:
`${#triggerResolvedArtifact(execution, 'myArtifact')}`

today, this interpolation is broken for new functions so if you want it to work in, you would have to write the expression as  `${#triggerResolvedArtifact(execution, 'myArtifact')}`
- @srekapalli is fixing this

(this is a regression from https://github.com/spinnaker/orca/pull/3071)

evansb:
Got it, thanks a lot for the explanation

gnarea:
FYI: This doesn't work for me: https://github.com/spinnaker/spinnaker/issues/5105

PR #3120, opened 2019-08-23
feat(bakeManifest/kustomize): add support for kustomize renderer

ethanfrogers:
Adds support for the kustomize as a template renderer. Abstracts some of the Bake Manifest code to support more than just Helm.


ethanfrogers:
@ezimanyi @maggieneterval the Rosco side of this is still going through internal review. I hope to have it submitted today.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/manifests/helm/HelmBakeManifestRequest.java
```diff
@@ -28,6 +32,24 @@
...
+  List<Artifact> inputArtifacts;
+
+  List<Artifact> values;
+
+  public HelmBakeManifestRequest(
+      BakeManifestContext bakeManifestContext,
+      List<Artifact> inputArtifacts,
+      String outputArtifactName,
+      Map<String, Object> overrides) {
+    this.setTemplateRenderer(bakeManifestContext.getTemplateRenderer());
```
ezimanyi: In general the constructor should just directly set properties of the object rather than call setters.  (I think the general reason is that we want to avoid calling methods on a not-yet-fully-initialized class.)

I also think that it would enforce more separation of concerns to call `super(...)` to use a constructor on the superclass for the properties that live there rather than directly setting them here.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/manifests/kustomize/KustomizeBakeManifestRequest.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
ezimanyi: s/Netflix/Armory/

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/manifests/kustomize/KustomizeBakeManifestRequest.java
```diff
@@ -0,0 +1,39 @@
...
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class KustomizeBakeManifestRequest extends BakeManifestRequest {
+  @JsonProperty("inputArtifact")
+  Artifact inputArtifact;
+
+  public KustomizeBakeManifestRequest(
+      BakeManifestContext bakeManifestContext, Artifact inputArtifact, String outputArtifactName) {
+    this.setTemplateRenderer(bakeManifestContext.getTemplateRenderer());
```
ezimanyi: Same comment as in the `HelmBakeManifestRequest` about not using setters and deferring to `super()`.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/manifests/BakeManifestRequest.java
```diff
@@ -18,24 +18,16 @@
...
-
-  List<Artifact> values;
-
   @JsonProperty("outputName")
   String outputName;
 
-  @JsonProperty("overrides")
-  Map<String, Object> overrides;
+  @JsonProperty("outputArtifactName")
+  String outputArtifactName;
```
ezimanyi: I realize this is pre-existing, but given the scope of changes to this class (and the subclasses) would be good to make these member variables `private` as there's no need for them to be `public` given that `@Data` will generate an accessor.

(In an ideal world it would be great for them to be immutable as well and make everything `private final` but that might require some more changes to work with deserialization.)

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/BakeManifestContext.java
```diff
@@ -36,19 +38,23 @@
   // Lombok (1.16.20) that
   // Jackson can use to deserialize.
   public BakeManifestContext(
-      @JsonProperty("inputArtifacts") List<CreateBakeManifestTask.InputArtifactPair> inputArtifacts,
+      @Nullable @JsonProperty("inputArtifacts")
```
ezimanyi: If the constructor can accept null for these two fields, it would make sense to add `@Nullable` to the fields themselves (as they'll end up null in some cases).

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/manifests/kustomize/KustomizeBakeManifestRequest.java
ethanfrogers: GETS ME EVERY TIME!

ezimanyi: Haha, me too...there's a way to set at least IntelliJ to do this correctly but I've never learned how.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/manifests/helm/HelmBakeManifestRequest.java
ethanfrogers: @ezimanyi I'm not sure what you mean in this case? Each of these classes extend the `BakeManifestRequest` but their inputs are different such that I don't want to have a matching constructor on the super class. It would require passing in more data than needed. Thoughts?

PR #3124, opened 2019-08-29
chore(sql): replace GPL licensed mysql-connector with mariadb-connector

Jammy-Louie:


asher:
- `mysql-connector-java` 8.x is licensed with a Universal FOSS Exception (https://oss.oracle.com/licenses/universal-foss-exception/) which should make it compatible with Spinnaker and unencumbered.
- we were using `mariadb-java-client` internally at Netflix and moved back to mysql-connector-java due to a number of issues. a move away from `mysql-connector-java` should be made with discussion and deliberation.

Jammy-Louie:
@robzienert @asher @ezimanyi 
Can we understand what these issue are and if we can resolve them?

Customers may want alternatives, could we build an alternative subproject for `mariadb`

ezimanyi:
@Jammy-Louie : Given @asher's above comment that `mysql-connector-java` is actually released under a license compatible with Spinnaker, why is there still a need to give customers the option of `mariadb`?

Even if we did the work to resolve any of the issues that arise from using `mariadb-java-client`, I'm not seeing the benefit of moving to it given that `mysql-connector-java` is actually released under a compatible license.

scottfrederick:
Disclaimer: IANAL and I don't speak for any company's legal department

> `mysql-connector-java` is actually released under a license compatible with Spinnaker, why is there still a need to give customers the option of `mariadb`?

The FOSS Exception to the `mysql-connector-java` GPL license apparently applies when Spinnaker is used or distributed in OSS form. It may not apply when Spinnaker is distributed under a commercial license, depending on the interpretation of the license terms by a distributor's license review team. If a license review determines that the FOSS Exception does not apply, then the GPL license becomes an issue.

asher:
I don't see why OSS Spinnaker needs to accommodate anyone looking to distribute Spinnaker under a commercial license. `mysql-connector-java` can be replaced with `mariadb-java-client` while building internally. See i.e. https://github.com/robzienert/orca-mariadb-extension

kdykeman:
Replacement probably isn't the right answer as different users have different needs, but it's definitely good for the community to have options (esp. as legal interpretations and circumstances can vary). I am interested in experimenting with `mariadb-java-client` not just here, but also in other places within spinnaker where it's pulled in - i.e. `front50/front50-sql-mysql`, `clouddriver/clouddriver-sql-mysql`, and `echo/echo-scheduler`.

Perhaps it's a better side conversation, but can you add some details about the issues Netflix ran into or things to watch out for as I experiment?

scottfrederick:
Thinking about alternatives to **_replacing_** `mysql-connector-java`. 

The relevant services (front50, orca, clouddriver) each have a `-sql-mysql` module that pulls in the `mysql-connector-java` dependency. A custom build could exclude this module to omit that jar file from the distribution. This would likely require a change to the `-web` module build file in each service to allow the `-sql-mysql` to be optionally excluded. As long as Spinnaker is configured with `sql.enabled=false`, excluding the connector jar file from the build should have no impact on deployed services. 

If a user of such a build did want to enable SQL, then they could add a JDBC driver of their choosing to the deployment of those services (somehow, not sure what might be required to do that). It seems to me that setting `sql.connection-pool.driver` to the class name of the selected JDBC driver would be sufficient to configure the `DataSource` used by the HikariCP setup without requiring custom code like shown in the [`orca-mariadb-extension`](https://github.com/robzienert/orca-mariadb-extension) sample project.

cfieber:
I believe this was addressed in #3175 please reopen if that is not the case

PR #3126, opened 2019-09-03
fix(error handling): add endpoint to mark executions failed

marchello2000:
When an execution fails in to start (e.g. before the execution is persisted) in `orca`,
`orca` will mark and store the execution as failed.
Since the artifact resolution is being moved to `echo`, `echo` needs the same ability
to mark executions as failed - otherwise any failures during artifact resolution in
`echo` will turn into silent failures for the user (if no execution is persisted then
the UI can't show any failures associated with that execution).

This change adds an endpoint that `echo` can call to mark the execution failed

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.fiat.model.UserPermission
 import com.netflix.spinnaker.fiat.model.resources.Role
 import com.netflix.spinnaker.fiat.shared.FiatService
 import com.netflix.spinnaker.fiat.shared.FiatStatus
+import com.netflix.spinnaker.kork.exceptions.SpinnakerException
 import com.netflix.spinnaker.kork.web.exceptions.InvalidRequestException
 import com.netflix.spinnaker.kork.web.exceptions.ValidationException
 import com.netflix.spinnaker.orca.clouddriver.service.JobService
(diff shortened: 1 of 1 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -123,6 +124,19 @@ class OperationsController {
...
 
+  /**
+   * Used by echo to mark an execution failure if it fails to materialize the pipeline
+   * (e.g. because the artifacts couldn't be resolved)
+   *
+   * @param pipeline pipeline json
+   */
+  @RequestMapping(value = '/fail', method = RequestMethod.POST)
+  void failPipeline(@RequestBody Map pipeline) {
+    String errorMessage = pipeline.remove("errorMessage")
```
emjburns: How come you're calling `remove` here?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -179,6 +193,30 @@ class OperationsController {
     }
   }
 
+  private void recordPipelineFailure(Map pipeline, String errorMessage) {
+    Exception pipelineError = null
+    try {
+      pipeline = parseAndValidatePipeline(pipeline)
```
ezimanyi: What's the purpose of calling `parseAndValidatePipeline` here if we already know that the pipeline is failing?  (I'd imagine we might not even expect the pipeline to be valid if we caught an exception while preparing it in `echo`.)

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
marchello2000: I didn't want to pollute the pipeline object since the error will be persisted as a `cancellationReason` by `markPipelineFailed` anyway

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
marchello2000: yeah, i was going back and forth on that... My thoughts were: if it does `parse` then it will be nicer from the UI perspective as the pipeline will be rendered "fully/correctly" (this especially applied to MPT pipelines)

If it fails, on the other hand, I figured we can capture both errors for the user to fix.

Do you think it's reasonable?

ezimanyi: Thanks for the explanation; yeah, that sounds reasonable to me

marchello2000: 👍 will add a comment to the code as well

PR #3133, opened 2019-09-08
refactor(StageDefinitionBuilder): Delete deprecated aroundStages

marchello2000:
Removing deprecated `aroundStages` (deprecated in early 2018)
Also, removing `DeployStage`/`DeployStrategyStage` (marked deprecated in 2015)

This logically simplifies the stage flow composition and gives us the opportunity for
having smarter deploy strategies (which can plan things on failure, etc) - this is already
added to the `Strategy` (e.g. `composeFlow` became `composeBeforeFlow`+`composeAfterFlow`)
but it's not really taken advantage of [yet]


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/QuickPatchStage.groovy
```diff
@@ -53,16 +56,21 @@ class QuickPatchStage implements StageDefinitionBuilder {
...
-    def stages = []
+  void beforeStages(@Nonnull Stage parent, @Nonnull StageGraphBuilder graph) {
+    // mark as SUCCEEDED otherwise a stage w/o child tasks will remain in NOT_STARTED
+    parent.status = ExecutionStatus.SUCCEEDED
+  }
+
+  @Override
+  void afterStages(@Nonnull Stage stage, @Nonnull StageGraphBuilder graph) {
+    ArrayList<Stage> stages = new ArrayList<>()
+    HashMap<String, Object> nextStageContext = new HashMap<>()
```
robzienert: Generally speaking, you shouldn't cast the variable with the implementation class, preferring the most common interface that you require. There are exceptions, but I don't believe this is one of them. There's a few spots that this is done in this PR that could be tidied a bit.

marchello2000: thanks!

PR #3134, opened 2019-09-09
feat(queue): add support for keiko-sql

asher:


orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/ExecutionMapper.kt
```diff
@@ -75,7 +77,24 @@ class ExecutionMapper(
...
+            execution.stages.filter { it.parentStageId == null }
+              .sortedBy { it.refId }
+          )
+
+          execution.stages.filter { it.parentStageId != null }
+            .sortedBy { it.refId }
+            .forEach {
+              when (it.syntheticStageOwner) {
+                STAGE_BEFORE -> stages.add(stages.indexOf(it.parent), it)
+                STAGE_AFTER -> stages.add(stages.indexOf(it.parent) + 1, it)
```
asher: Ensures `SqlExecutionRepository` returns executions with stages sorted in the same order as `RedisExecutionRepository`. Instead of sorting all stages by `refId`, synthetic `BEFORE` and `AFTER` stages are clustered around their parent.

orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueIntegrationTest.kt
```diff
@@ -309,10 +309,16 @@ abstract class QueueIntegrationTest {
...
+        refId = "2b1"
+        type = "wait"
+        context = mapOf("waitTime" to 2)
         requisiteStageRefIds = listOf("1")
       }
+      stage {
+        refId = "2b2"
+        type = "dummy"
+        requisiteStageRefIds = listOf("2b1")
+      }
```
asher: `keiko-sql` does not attempt to guarantee FIFO semantics; messages should only be delivered at or after their ready time and only once (within the `ackTimeout`), however ready messages are not necessarily read in insertion order. `RedisQueue` doesn't strictly guarantee ordered message executions either when running across multiple instances but does when running on a single instance such as during the execution of these tests.

Test changes here are made for tests with parallel execution branches, where assumptions were made that parallel branches without explicit ordering would still execute in a determinate order.

For this test, the original expectation:
```
       2a1 (TERMINAL) -> 2a2 (NOT_STARTED) ------\
1  <                                              3 (NOT_STARTED)
      2b (SUCCEEDED)  ----------------------------/
```
becomes:
```
       2a1 (TERMINAL) -> 2a2 (NOT_STARTED) -----------------\
1  <                                                         3 (NOT_STARTED)
      2b1 (SUCCEEDED or CANCELED) -> 2b2 (NOT_STARTED)  ----/
```
because while on most executions, execution of the A and B branches will be interleaved in exact message insertion order such that 2b/2b1 processes its `RunTask` message before 2a1 persists the execution with `isCanceled` set to true, it become possible for the A branch's `StartStage` and subsequent `RunTask` to complete before the B branch's `RunTask` is processed, so it can no longer be assumed to always `SUCCEED`. That should be ok. In a real HA Orca deployment using keiko-redis, the same thing could happen depending on the timing of GC pauses, network blips, or instance speed.

orca-queue-sql/src/test/kotlin/com/netflix/spinnaker/orca/q/sql/SqlQueueIntegrationTest.kt
```diff
@@ -0,0 +1,178 @@
...
+@Configuration
+class SqlTestConfig {
+  @Bean
+  fun jooq(): DSLContext {
+    val testDatabase = SqlTestUtil.initTcMysqlDatabase()
+    return testDatabase.context
+  }
+
+  @Autowired
+  fun sqlQueueObjectMapper(
```
robfletcher: I'm confused, is this configuring an existing bean or creating one?

asher: Configuring existing, the same thing is happening here for the redis equivalent via: https://github.com/spinnaker/orca/blob/master/orca-queue-redis/src/main/kotlin/com/netflix/spinnaker/config/RedisOrcaQueueConfiguration.kt#L51-L56

robfletcher: Hmm, but this is changing the mapper for everything then, right?

asher: Yeah, seems to be required for orca to function. Are you suggesting this should happen other than Queue specific configuration classes?

robfletcher: I'm just wondering if that configuration is Keiko-specific it might cause unexpected behavior elsewhere. I'm also surprised that `KotlinModule`, etc. aren't already configured.

I don't think it's a _big_ deal though.

asher: It might be nice wiring in `ExecutionTypeDeserializer` and `TaskTypeDeserializer` in `OrcaObjectMapper` to make it more obvious that they're globally applied but they aren't currently available from orca-core and I'm hesitant to change that.

PR #3138, opened 2019-09-11
feat(monitored deploy): add endpoint to get a list of deployment monitors

marchello2000:
Returns the names and ids of registered deployment monitors.
This will be used in `deck` to let the user pick a monitor (also need to plumb through `gate`)

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesController.java
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesController.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.controllers;
+
+import com.netflix.spinnaker.config.MonitoredDeployConfigurationProperties;
+import com.netflix.spinnaker.orca.model.DeploymentMonitorDefinition;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/** Controller intended for querying various capabilities of orca */
+@RestController
+public class CapabilitiesController {
+  private MonitoredDeployConfigurationProperties monitoredDeployConfigurationProperties;
+
+  public CapabilitiesController(
+      Optional<MonitoredDeployConfigurationProperties> monitoredDeployConfigurationProperties) {
+    this.monitoredDeployConfigurationProperties =
+        monitoredDeployConfigurationProperties.orElse(null);
+  }
+
+  @GetMapping("/capabilities/deploymentMonitors")
+  public List<DeploymentMonitorDefinition> getDeploymentMonitors() {
+    if (monitoredDeployConfigurationProperties == null) {
+      return Collections.emptyList();
+    }
+
+    return monitoredDeployConfigurationProperties.getDeploymentMonitors().stream()
+        .map(DeploymentMonitorDefinition::new)
+        .collect(Collectors.toList());
+  }
+}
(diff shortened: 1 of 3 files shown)


orca-web/src/test/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesControllerTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.security.test.context.support.WithMockUser;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
+import org.springframework.web.servlet.config.annotation.EnableWebMvc;
+
+@ExtendWith(SpringExtension.class)
```
marchello2000: @robzienert / @robfletcher I copied this pattern, but I question if we actually should test via the `WebMvc`? Perhaps it makes sense to validate that params are annotated correctly, but it seems like a big overhead (vs. testing the methods directly). It also makes the tests slower since a bunch of spring needs to be stood up for this style test
What are your thoughts?

orca-web/src/test/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesControllerTest.java
```diff
@@ -0,0 +1,78 @@
...
+    def1.setName("the name");
+    def1.setId("the id");
+    monitorDefinitions.add(def1);
+
+    when(configurationProperties.getDeploymentMonitors()).thenReturn(monitorDefinitions);
+
+    mvc.perform(request)
+        .andExpect(status().is2xxSuccessful())
+        .andExpect(
+            content().json(format("[{'id':'%s','name':'%s'}]", def1.getId(), def1.getName())));
```
marchello2000: there has got to be a better way than `format`, feedback appreciated!

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesController.java
```diff
@@ -0,0 +1,50 @@
...
+    this.monitoredDeployConfigurationProperties =
+        monitoredDeployConfigurationProperties.orElse(null);
+  }
+
+  @GetMapping("/capabilities/deploymentMonitors")
+  public List<DeploymentMonitorDefinition> getDeploymentMonitors() {
+    ArrayList<DeploymentMonitorDefinition> result = new ArrayList<>();
+
+    if (monitoredDeployConfigurationProperties != null) {
+      monitoredDeployConfigurationProperties
```
dreynaud: could _streamline_ this with streams:

```
if (monitoredDeployConfigurationProperties == null) {
    return Collections.emptyList();
}

return monitoredDeployConfigurationProperties
    .getDeploymentMonitors()
    .stream()
    .map(dm -> result.add(new DeploymentMonitorDefinition(dm)))
```

orca-web/src/main/groovy/com/netflix/spinnaker/orca/model/DeploymentMonitorDefinition.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.model;
+
+import lombok.Data;
+
+@Data
+public class DeploymentMonitorDefinition {
```
dreynaud: same class name but in a different package? This is triggering my anxiety

orca-web/src/test/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesControllerTest.java
dreynaud: it's a java file in `src/test/groovy` FWIW

orca-web/src/test/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesControllerTest.java
dreynaud: in pure Java I don't think so. But it seems like a job for `def1`'s `toString()`

orca-web/src/main/groovy/com/netflix/spinnaker/orca/model/DeploymentMonitorDefinition.java
marchello2000: this is the "model" for the API... should i call it something else - i don't want return all of the internal stuffs

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesController.java
marchello2000: thanks, i like it (even though i am not a fan of  java streams)

orca-web/src/test/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesControllerTest.java
marchello2000: going to convert it to a simple non-mvc test

PR #3128, opened 2019-09-05
feat(strategy): Monitored deploy strategy ALPHA

marchello2000:
This is the very beginning of the monitored deploy strategy.
There is still a lot of work to be done but that will be left for subsequent PRs as many of these features will require refactoring a bunch of stages under `orca-clouddriver`.
In order to use this feature one must have the config flag enabled as well as manually craft json for the deploy stage, so it's quite isolated.

Monitored deploy strategy operated like RollingRedBlack with the addition that it allows for 3rd party service (DeploymentMonitor) to provide input into the progress of the deploy.
After every step, Spinnaker will call into the DeploymentMonitor specified in the stage (and configured in `orca.yml`).
The DeploymentMonitor can then perform analysis on the newly deployed instances and provide input on whether the deployment should continue or not.
(detailed documentation will be added in subsequent PR).

The following features are still missing:
* Pinning and unpinning the source during deploy
* Shrink cluster at the end of deploy
* Ability to fast forward to 100% of the deployment
* Indicate success/failure in the DeploymentComplete call to DeploymentMonitor
* Dynamically set task timeout while evaluating deployment health
* Deck/UI changes to support this new strategy

PR #3139, opened 2019-09-11
fix(gce): revert a conditional that was accidentally flipped in #3128

plumpy:
cc @marchello2000


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.groovy
```diff
@@ -114,7 +114,7 @@ class GceDeployStagePreProcessor implements DeployStagePreProcessor  {
   }
 
   private static boolean shouldPinSourceServerGroup(String strategy) {
-    return (Strategy.fromStrategyKey(strategy) != Strategy.RED_BLACK)
```
marchello2000: oh, my gosh, sorry @plumpy and thank you!

PR #3150, opened 2019-09-13
fix(conditions): Only pause pipeline deploys

jeyrschabu:
- Pausing deployments should only apply to pipelines
- Orchestrations should be ignored


jeyrschabu:
This PR ensures that pausing deployments only applies to pipelines and ignore orchestrations.
When a deployment is paused as a result of a failover, the pipeline execution view provides the user the option to un-pause the deployment.
However, that option is not currently available in the task view, in the case where a user is manually cloning a server group.
This prevents the user from intervening to un-pause the deployment in case of an emergency.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/ConditionAwareDeployStagePreprocessor.java
```diff
@@ -48,7 +49,9 @@ public ConditionAwareDeployStagePreprocessor(
 
   @Override
   public boolean supports(Stage stage) {
-    return true;
+    // Restricting pausing deployments to Pipeline executions only
```
robzienert:
```suggestion
    // Restrict pausing deployments to Pipeline executions only
```

PR #3152, opened 2019-09-13
fix(test): Bumping execution latch to 10s

jeyrschabu:
- fixing test failures because wait time seems low
- Seeing the following failure occasionally in travis and locally:

``
com.netflix.spinnaker.orca.q.sql.SqlQueueIntegrationTest > can run a fork join pipeline FAILED
    java.lang.AssertionError: Pipeline did not complete
``

### DIFF
--- a/orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/ExecutionLatch.kt
+++ b/orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/ExecutionLatch.kt
@@ -43,7 +43,7 @@ class ExecutionLatch(private val predicate: Predicate<ExecutionComplete>) :
     }
   }
 
-  fun await() = latch.await(5, TimeUnit.SECONDS)
+  fun await() = latch.await(10, TimeUnit.SECONDS)
 }
 
 fun ConfigurableApplicationContext.runToCompletion(execution: Execution, launcher: (Execution) -> Unit, repository: ExecutionRepository) {


orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/ExecutionLatch.kt
```diff
@@ -43,7 +43,7 @@ class ExecutionLatch(private val predicate: Predicate<ExecutionComplete>) :
     }
   }
 
-  fun await() = latch.await(5, TimeUnit.SECONDS)
+  fun await() = latch.await(10, TimeUnit.SECONDS)
```
marchello2000: seems fine... 
but.. @asher : curious why would this be slower now that it's in SQL? is it because we spin up a docker for it? If that's the case maybe whatever spins it up can block so we don't play this cat/mouse game?

asher: This isn't related to docker startup times and test execution does block on container db availability. Wouldn't have even been able to persist the pipeline earlier in the test otherwise, and the you'd see the first `QueueIntegrationTest` fail if that was related. Not sure if this PR is the right answer though.

asher:
This is worth a shot. It would be nice if `runToCompletion` could retrieve the current execution to help troubleshoot this sort of issue. Something like this would prob work:

```
  assert(latch.await()) { "Pipeline did not complete: " +
    OrcaObjectMapper.getInstance().writeValueAsString(repository.retrieve(execution.type, execution.id))
  }
```

jeyrschabu:
@marchello2000 i haven’t run into this since.

asher:
@marchello2000 your idea sounds like a good balance between making the rare failure case rarer, vs. not overly increasing test times. I haven't seen this happen since originally reported or been able to repo across 1000 runs of `./gradlew :orca-queue-sql:test`, but there are plenty of reasons the mysql test container could experience stalls that makes a single queue poll take a second instead of 100ms under inhospitable circumstances. That makes any time sensitive tests built on top of it a bit dicey but not sure if there's a better way for this specific test scenario.

robzienert:
Reopen if still legit, but it looks like this died on the vine.

PR #3155, opened 2019-09-16
chore(monitored deploy): clean up naming broadcast -> notify

marchello2000:
left over cleanup from the previous PR


marchello2000:
merging as this is just syntactic change - no review needed

PR #3157, opened 2019-09-16
refactor(google): convert GceDeployStagePreProcessor to Java

maggieneterval:
- chore(google): add tests for GceDeployStagePreProcessor

  Adds tests that demonstrate that in a GCE red/black deployment, the `GceDeployStagePreProcessor` adds a `ResizeServerGroupStage`, which "pins" the source server group's minimum capacity as its desired capacity, _before_ a `CaptureSourceServerGroupCapacityTask`, which snapshots the source server group's minimum capacity (to later be re-applied to the target server group).

- refactor(google): convert GceDeployStagePreProcessor to Java

  For clarity, removes references to rolling red/black, because as far as I know we have no plans to implement this strategy in GCE.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
```diff
@@ -0,0 +1,152 @@
...
+  private static boolean shouldPinSourceServerGroup(String strategy) {
+    return Strategy.fromStrategyKey(strategy) == Strategy.RED_BLACK;
+  }
+
+  private Map<String, Object> getResizeContext(StageData stageData) {
+    AbstractDeployStrategyStage.CleanupConfig cleanupConfig =
+        AbstractDeployStrategyStage.CleanupConfig.fromStage(stageData);
+    Map<String, Object> resizeContext = new HashMap<>();
+
+    resizeContext.put(
```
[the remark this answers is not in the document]
maggieneterval: Great idea, I didn't even notice that! Since the next PR in this thrilling series is going to change even more about `getResizeContext`, I'll include the addition of a `toContext` method as a commit with that PR!

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessorTest.java
```diff
@@ -0,0 +1,132 @@
...
+    context.put("zone", "us-central1-f");
+    context.put(
+        "availabilityZones",
+        Collections.singletonMap("us-central1-f", Collections.singletonList("us-central1-f")));
+    context.put("application", "testapp");
+    stage.setContext(context);
+
+    // Before Stages
+    List<StageDefinition> beforeStages = preProcessor.beforeStageDefinitions(stage);
+    assertThat(beforeStages.stream().map(stageDefinition -> stageDefinition.stageDefinitionBuilder))
```
plumpy: I actually found a better (IMO) way to do this, but I didn't go back and fix the older tests I wrote:

```
assertThat(beforeStages)
    .extracting(stage -> stage.stageDefinitionBuilder)
    .containsExactly(resizeServerGroupStage)
```

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessorTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+  @Test
+  void redBlackStrategyTest() {
+    when(targetServerGroupResolver.resolve(any()))
+        .thenReturn(
+            ImmutableList.of(
+                new TargetServerGroup(
+                    ImmutableMap.of("name", "testapp-v000", "zone", "us-central1-f"))));
+    Stage stage = new Stage();
+    Map<String, Object> context = new HashMap<>();
```
plumpy: There's no hard and fast rule to this, and sometimes it's difficult/impossible, but I try to make my test methods contain all the _relevant_ details, but not additional ones. In this case, this test method is testing something about the context's strategy and the rest of the context is just boilerplate to get the thing to run. So I'd be inclined to hide these details in a helper method:

`Map<String, Object> context = createContextWithStrategy(redblack);`

or maybe this is better?
```
Map<String, Object> context = createDefaultContext();
context.put("strategy", "redblack");
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
```diff
@@ -0,0 +1,152 @@
...
+  public List<StepDefinition> additionalSteps(Stage stage) {
+    return ImmutableList.of(
+        new StepDefinition(
+            "snapshotSourceServerGroup", CaptureSourceServerGroupCapacityTask.class));
+  }
+
+  @Override
+  public List<StageDefinition> beforeStageDefinitions(Stage stage) {
+    StageData stageData = stage.mapTo(StageData.class);
+    if (shouldPinSourceServerGroup(stageData.getStrategy())) {
```
plumpy: Feel free to disagree and ignore, but I'd be inclined to flip this if statement. IMO it's usually easier to read if the bulk of the logic is at the bottom of the method instead of the middle.

```
if (!shouldPinSourceServerGroup(stageData.getStrategy())) {
  return ImmutableList.of();
}

// etc
```

Hot IntelliJ tip: If you agree, you can do this by hitting alt-enter on `if` and selecting `invert if condition`.

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessorTest.java
maggieneterval: Wow this is great, I now love `extracting`! :sweet_potato:

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
maggieneterval: I agreed and I hit alt-enter!

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessorTest.java
maggieneterval: love it, I went with option #2!

PR #3159, opened 2019-09-16
feat(orcestration): add support for dynamic task timeout

marchello2000:
Currently, task timeouts can't take anything about the execution environment into account (they are essentially a static method).
This change adds the ability for tasks to consider the current `stage` (`execution`) when evaluating timeout.

Additionally, the `onTimeout` method on the task can now return a `TaskResult`.
This allows the tasks to decide if they should be `TERMINAL` or `FAILED_CONTINUE` on timeout and return data to be added to the stage in a standard way

Both are needed for monitored deploy - as the timeout and what action to take depend on the monitor configuration.
This will also become useful for `clouddriver` tasks that are retryable (e.g. backed by sagas)

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/RunJobTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/RunJobTask.groovy
@@ -26,6 +26,9 @@ import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
+import javax.annotation.Nonnull
+import javax.annotation.Nullable
+
 @Slf4j
 @Component
 class RunJobTask extends AbstractCloudProviderAwareTask implements RetryableTask {
@@ -43,12 +46,15 @@ class RunJobTask extends AbstractCloudProviderAwareTask implements RetryableTask
   long timeout = 60000
 
   @Override
-  void onTimeout(Stage stage) {
+  @Nullable
+  TaskResult onTimeout(@Nonnull Stage stage) {
     jobUtils.cancelWait(stage)
+
+    return null;
   }
 
   @Override
-  void onCancel(Stage stage) {
+  void onCancel(@Nonnull Stage stage) {
     jobUtils.cancelWait(stage)
   }
 
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/RetryableTask.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/RetryableTask.java
@@ -27,6 +27,10 @@ public interface RetryableTask extends Task {
 
   long getTimeout();
 
+  default long getDynamicTimeout(Stage stage) {
+    return getTimeout();
+  }
+
   default long getDynamicBackoffPeriod(Duration taskDuration) {
     return getBackoffPeriod();
   }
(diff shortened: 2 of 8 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -104,15 +105,23 @@ class RunTaskHandler(
             registry
               .timeoutCounter(stage.execution.type, stage.execution.application, stage.type, taskModel.name)
               .increment()
-            task.onTimeout(stage)
-            throw e
+            taskResult = task.onTimeout(stage)
+
+            if ((taskResult?.status != FAILED_CONTINUE) &&
+              (taskResult?.status != TERMINAL)) {
```
robzienert: Don't need these wrapping parenthesis, do we?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/RunJobTask.groovy
```diff
@@ -43,12 +46,15 @@ class RunJobTask extends AbstractCloudProviderAwareTask implements RetryableTask
...
+  @Nullable
+  TaskResult onTimeout(@Nonnull Stage stage) {
     jobUtils.cancelWait(stage)
+
+    return null;
   }
 
   @Override
-  void onCancel(Stage stage) {
+  void onCancel(@Nonnull Stage stage) {
```
dreynaud: is adding `@Nonnull` actually buying us anything? I was operating under the assumption that if something isn't `@Nullable` then it has to be `@Nonnull`

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
dreynaud: how do you like this?
 `if (!setOf(TERMINAL, FAILED_CONTINUE).contains(taskResult?.status))`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/RunJobTask.groovy
marchello2000: but this is groovy, isn't everything `nullable`? Mostly, I added this because intellij complains when the base method has the annotation and the methods in derived classes don't

orca-core/src/main/java/com/netflix/spinnaker/orca/RetryableTask.java
```diff
@@ -27,6 +27,10 @@
 
   long getTimeout();
 
+  default long getTimeoutDynamic(Stage stage) {
```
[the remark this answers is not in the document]
marchello2000: duh... i don't know what I was thinking, thanks

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
marchello2000: the old me would be like... "i don't want to create a short-lived object on gen 0 gc heap" but that me is long gone

dreynaud: if you're this kind of person you can make `setOf(TERMINAL, FAILED_CONTINUE)` a static field so you only ever have one

PR #3160, opened 2019-09-17
fix(rollback): capture the source capacity before anything else

dreynaud:
Does not entirely address this issue but makes it less likely to happen:
https://github.com/spinnaker/spinnaker/issues/4895

The "danger zone" is when a resize of the source server group happens
between the start of the rollback and the moment we take the capacity
snapshot. This change should reduce that "danger zone" from minutes to
seconds.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
@@ -92,11 +92,14 @@ class ExplicitRollback implements Rollback {
       ]
     }
 
-    def stages = [enableServerGroupStage]
+    // https://github.com/spinnaker/spinnaker/issues/4895 - capture the source capacity as early as possible
+    def stages = []
     if (!parentStage.getContext().containsKey("sourceServerGroupCapacitySnapshot")) {
       // capacity has been previously captured (likely as part of a failed deploy), no need to do again!
       stages << buildCaptureSourceServerGroupCapacityStage(parentStage, parentStage.mapTo(ResizeStrategy.Source))
     }
+
+    stages << enableServerGroupStage
 
     Map resizeServerGroupContext = new HashMap(parentStage.context) + [
       action                       : ResizeStrategy.ResizeAction.scale_to_server_group.toString(),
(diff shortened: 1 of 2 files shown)


dreynaud:
FYI @ashutosh16

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
```diff
@@ -92,12 +92,15 @@ class ExplicitRollback implements Rollback {
       ]
     }
 
-    def stages = [enableServerGroupStage]
+    // https://github.com/spinnaker/spinnaker/issues/4895 - capture the source capacity as early as possible
+    def stages = []
     if (!parentStage.getContext().containsKey("sourceServerGroupCapacitySnapshot")) {
       // capacity has been previously captured (likely as part of a failed deploy), no need to do again!
```
marchello2000: this comment confuses me

ajordens: me too ... otherwise looks good

dreynaud: I mean... it comes from 

```commit ed90df775b9e5870c2c6779818293c90a34edb7d (tag: v6.4.0)
Author: Adam Jordens <adam@jordens.org>
Date:   Tue Feb 13 10:40:12 2018 -0800

    feat(rollback): Support for automatic rollback when a deployment fails (#1944)
    
    This PR injects a `RollbackCluster` stage if a particular cluster
    deployment has automatic rollback enabled.
```

:)

ajordens: My comment about the comment was more that it doesn’t really describe much, I find little value in linking out to issues that are also linked out to in the commit message itself. Perhaps others do tho.

dreynaud: I see, we're not talking about the same thing. I thought @marchello2000's comment was referring to this:

`// capacity has been previously captured (likely as part of a failed deploy), no need to do again!`

For me linking out from the commit and from the comment serve different purposes: linking from the commit makes it visible to github on the issue, and linking in the comment makes it readable from the source without having to dig through potentially layers of git blame.

PR #3162, opened 2019-09-17
feat(clouddriver): Basic support for retryable kato tasks

robzienert:
Depends on https://github.com/spinnaker/kork/pull/383

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
@@ -24,16 +24,18 @@ import com.netflix.spinnaker.orca.clouddriver.KatoService
 import com.netflix.spinnaker.orca.clouddriver.model.Task
 import com.netflix.spinnaker.orca.clouddriver.model.TaskId
 import com.netflix.spinnaker.orca.pipeline.model.Stage
+import com.netflix.spinnaker.orca.pipeline.model.SystemNotification
 import groovy.transform.CompileStatic
 import groovy.transform.TypeCheckingMode
+import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.stereotype.Component
 import retrofit.RetrofitError
 
 import java.time.Clock
-import java.util.concurrent.TimeUnit
 
+@Slf4j
 @Component
 @CompileStatic
 class MonitorKatoTask implements RetryableTask {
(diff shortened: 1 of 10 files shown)


cfieber:
Thanks I Love It

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -48,10 +50,12 @@ class MonitorKatoTask implements RetryableTask {
 
   private final Clock clock
   private final Registry registry
+  @Autowired private KatoService kato
```
cfieber: could you add this to the constructor instead

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -48,10 +50,12 @@ class MonitorKatoTask implements RetryableTask {
...
 
   @Autowired
-  public MonitorKatoTask(Registry registry) {
+  MonitorKatoTask(Registry registry) {
     this(registry, Clock.systemUTC())
+    this.kato = kato
```
cfieber: this seems wrong without KatoService as a constructor arg

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -23,9 +23,11 @@ import com.netflix.spinnaker.orca.TaskResult
 import com.netflix.spinnaker.orca.clouddriver.KatoService
 import com.netflix.spinnaker.orca.clouddriver.model.Task
 import com.netflix.spinnaker.orca.clouddriver.model.TaskId
+import com.netflix.spinnaker.orca.pipeline.model.Execution
 import com.netflix.spinnaker.orca.pipeline.model.Stage
 import groovy.transform.CompileStatic
 import groovy.transform.TypeCheckingMode
+import io.github.resilience4j.retry.RetryConfig
```
cfieber: I don't see this used anywhere in here, old?

orca-web/config/orca.yml
```diff
@@ -51,6 +51,26 @@ tasks:
 logging:
   config: classpath:logback-defaults.xml
 
+resilience4j.retry:
+  instances:
+    # TODO(rz): These defaults were just copied from the resilience4j website. They still need to be tuned.
+    default:
+      maxRetryAttempts: 3
+      waitDuration: 10s
+      enableExponentialBackoff: true
```
marchello2000: we talked in person but these defaults will blow orca's 60s task ack timeout which is going to cause issues/double executions, etc

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -149,7 +149,11 @@ class MonitorKatoTask implements RetryableTask {
       }
       katoTasks << m
       outputs["kato.tasks"] = katoTasks
+    }
 
+    if (status == ExecutionStatus.TERMINAL && katoTask.status.retryable) {
+      kato.resumeTask(katoTask.id)
+      status = ExecutionStatus.RUNNING
```
marchello2000: probably worth at least logging something here that we are "going to retry".
You are basically hiding an error from the user and it is possible the error goes away on retry or it becomes another error... and that makes it harder to debug issues. (this is the thing we talked about some standardization in orca about error messages... someday)

robzienert: You're right. I've updated the PR to show usage of the `SystemNotification` stuff I was talking about.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
robzienert: yeeeeeahh... I tried once and wound up reverting it. I'll try again.

PR #3169, opened 2019-09-18
chore(intellij): fixing first time builds in intellij

marchello2000:
for some reason, the latest intellij doesn't copy the sql migration (changelog-master) file.
this change fixes that so we don't have to `gradlew build` to copy the file over


marchello2000:
cc: @luispollo this fixes the first run not working in intelliJ

luispollo:
Thanks @marchello2000! I'm still having issues building in IntelliJ, but I'll check with you at work.

PR #3171, opened 2019-09-18
chore(kubernetes): upgrade kubernetes client-java from beta to 5.0.0

Jammy-Louie:
bump to use the same version as cloud driver, the version used before was a beta version


Jammy-Louie:

PR #3173, opened 2019-09-18
fix(google): fix deploying with red/black into an empty cluster

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4270

As a Spinnaker user deploying my first server group into a new cluster, I expect it to succeed if I have selected the red/black strategy.

Since Netflix has already added support for this user story in the AWS rolling red/black strategy, this PR takes the same approach: the GCE deploy stage pre-processor now short-circuits when attempting to add stages to pin and unpin the source server group's capacity if there is no source server group ([AWS code](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy#L173), for reference), and the red/black strategy `composeFlow` method now also short-circuits instead of adding stage to disable and shrink the non-existent source server group ([rolling red/black code](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy#L113) and [tests](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategySpec.groovy), for reference).

<details><summary>Before</summary>
<p>

3dtGMbet5xO

</p>
</details>

<details><summary>After</summary>
<p>

Yuy39tK63h7

</p>
</details>

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
@@ -24,14 +24,15 @@ import com.netflix.spinnaker.orca.clouddriver.pipeline.providers.aws.CaptureSour
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerGroupStage;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.AbstractDeployStrategyStage;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.DeployStagePreProcessor;
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroup;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.TargetServerGroupResolver;
 import com.netflix.spinnaker.orca.kato.pipeline.strategy.Strategy;
 import com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategy;
 import com.netflix.spinnaker.orca.kato.pipeline.support.StageData;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import javax.annotation.Nullable;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
@@ -66,7 +67,13 @@ public class GceDeployStagePreProcessor implements DeployStagePreProcessor {
       return ImmutableList.of();
     }
 
-    Map<String, Object> resizeContext = getResizeContext(stageData);
+    Optional<Map<String, Object>> optionalResizeContext = getResizeContext(stageData);
+    if (!optionalResizeContext.isPresent()) {
+      // no source server group, no need to resize
+      return ImmutableList.of();
+    }
+
+    Map<String, Object> resizeContext = optionalResizeContext.get();
     resizeContext.put("pinMinimumCapacity", true);
 
     return ImmutableList.of(
@@ -107,26 +114,25 @@ public class GceDeployStagePreProcessor implements DeployStagePreProcessor {
     return Strategy.fromStrategyKey(strategy) == Strategy.RED_BLACK;
   }
 
-  private Map<String, Object> getResizeContext(StageData stageData) {
-    AbstractDeployStrategyStage.CleanupConfig cleanupConfig =
-        AbstractDeployStrategyStage.CleanupConfig.fromStage(stageData);
-    Map<String, Object> resizeContext = new HashMap<>();
+  private Optional<Map<String, Object>> getResizeContext(StageData stageData) {
+    Map<String, Object> resizeContext =
+        AbstractDeployStrategyStage.CleanupConfig.toContext(stageData);
 
-    resizeContext.put(
-        cleanupConfig.getLocation().singularType(), cleanupConfig.getLocation().getValue());
-    resizeContext.put("cluster", cleanupConfig.getCluster());
-    resizeContext.put("moniker", cleanupConfig.getMoniker());
-    resizeContext.put("credentials", cleanupConfig.getAccount());
-    resizeContext.put("cloudProvider", cleanupConfig.getCloudProvider());
+    try {
+      ResizeStrategy.Source source = getSource(targetServerGroupResolver, stageData, resizeContext);
+      if (source == null) {
+        return Optional.empty();
+      }
 
-    ResizeStrategy.Source source = getSource(targetServerGroupResolver, stageData, resizeContext);
-    resizeContext.put("serverGroupName", source.getServerGroupName());
-    resizeContext.put("action", ResizeStrategy.ResizeAction.scale_to_server_group);
-    resizeContext.put("source", source);
-    resizeContext.put(
-        "useNameAsLabel", true); // hint to deck that it should _not_ override the name
-
-    return resizeContext;
+      resizeContext.put("serverGroupName", source.getServerGroupName());
+      resizeContext.put("action", ResizeStrategy.ResizeAction.scale_to_server_group);
+      resizeContext.put("source", source);
+      resizeContext.put(
+          "useNameAsLabel", true); // hint to deck that it should _not_ override the name
+      return Optional.of(resizeContext);
+    } catch (TargetServerGroup.NotFoundException e) {
+      return Optional.empty();
+    }
   }
 
   @Nullable
@@ -140,7 +146,13 @@ public class GceDeployStagePreProcessor implements DeployStagePreProcessor {
       return null;
     }
 
-    Map<String, Object> resizeContext = getResizeContext(stageData);
+    Optional<Map<String, Object>> optionalResizeContext = getResizeContext(stageData);
+    if (!optionalResizeContext.isPresent()) {
+      // no source server group, no need to unpin
+      return null;
+    }
+
+    Map<String, Object> resizeContext = optionalResizeContext.get();
     resizeContext.put("unpinMinimumCapacity", true);
 
     return new StageDefinition(
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
@@ -16,9 +16,11 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
 
+import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.DisableClusterStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.cluster.ScaleDownClusterStage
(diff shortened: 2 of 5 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
```diff
@@ -50,17 +52,32 @@ class RedBlackStrategy implements Strategy, ApplicationContextAware {
...
-    def cleanupConfig = AbstractDeployStrategyStage.CleanupConfig.fromStage(stage)
-
-    Map baseContext = [
-      (cleanupConfig.location.singularType()): cleanupConfig.location.value,
-      cluster                                : cleanupConfig.cluster,
-      credentials                            : cleanupConfig.account,
-      moniker                                : cleanupConfig.moniker,
-      cloudProvider                          : cleanupConfig.cloudProvider,
-    ]
+    List<Stage> stages = []
```
ezimanyi: :+1: to adding explicit types here

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
```diff
@@ -50,17 +52,32 @@ class RedBlackStrategy implements Strategy, ApplicationContextAware {
...
+    StageData stageData = stage.mapTo(StageData)
+    Map<String, Object> baseContext = AbstractDeployStrategyStage.CleanupConfig.toContext(stageData)
+
+    try {
+      Stage parentCreateServerGroupStage = stage.directAncestors().find() { it.type == CreateServerGroupStage.PIPELINE_CONFIG_TYPE }
+
+      if (parentCreateServerGroupStage.status == ExecutionStatus.NOT_STARTED) {
+        // No point in composing the flow if we are called to plan "beforeStages" since we don't have any STAGE_BEFOREs.
```
marchello2000: oh no, my crappy bandaid is spreading! :)

Btw, my pr #3133 will obviate the need for this (i will let you know when it's in and can help take advantage of it)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
```diff
@@ -50,17 +52,32 @@ class RedBlackStrategy implements Strategy, ApplicationContextAware {
...
+      StageData parentStageData = parentCreateServerGroupStage.mapTo(StageData)
+      StageData.Source sourceServerGroup = parentStageData.source
+
+      // Short-circuit if there is no source server group
+      if (sourceServerGroup == null || sourceServerGroup.serverGroupName == null) {
+        return []
+      }
+    } catch (Exception e) {
+      // This probably means there was no parent CreateServerGroup stage - which should never happen
+      throw new IllegalStateException("Failed to determine source server group from parent stage while planning red/black flow", e)
```
marchello2000: oh, also... there is a currently known issue with this that you might want to check if it is still an issue in gce:
if you select `clone server group` in the UI this fails in AWS today

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
maggieneterval: Sorry for spreading the crappy bandaid! Very excited for that PR to be merged, I can definitely update this flow once `aroundStages` is officially gone.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
```diff
@@ -107,26 +114,25 @@ private static boolean shouldPinSourceServerGroup(String strategy) {
...
-
-    ResizeStrategy.Source source = getSource(targetServerGroupResolver, stageData, resizeContext);
-    resizeContext.put("serverGroupName", source.getServerGroupName());
-    resizeContext.put("action", ResizeStrategy.ResizeAction.scale_to_server_group);
-    resizeContext.put("source", source);
-    resizeContext.put(
-        "useNameAsLabel", true); // hint to deck that it should _not_ override the name
-
-    return resizeContext;
+  private Optional<Map<String, Object>> getResizeContext(StageData stageData) {
```
marchello2000: FYI: I want to (and am planning on to do this this week) to remove this logic form the preprocessor. because it's a) confusing and not explicit and b) can cause issues...
the specific issue is, imagine you have:

```
v1 server group - disabled
v0 server group - enabled
```

now you run the deploy and expect this to pin `v0`, but in (at least AWS case today) this will pin `v1` because the source determination here runs before we run the determinesourceservergroup task.


my plan is to have a "PinSourceSupport" class that the strategy can call explicitly to add the pin/unpin stages - doesn't seem like this should belong to the preprocessor - and also no point for each cloud provider to duplicate this code

marchello2000:
@maggieneterval : this looks good to me. I added a few caveats that are known to us in the AWS space right now (we are in the process of re-writing RRB). Maybe these caveats aren't an issue with gce. Not promising anything, but i am in the processing of refactoring (and hopefully make this at least a little bit better). But up to you if you want to wait for that or not

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
```diff
@@ -50,17 +52,32 @@ class RedBlackStrategy implements Strategy, ApplicationContextAware {
...
+      }
+
+      StageData parentStageData = parentCreateServerGroupStage.mapTo(StageData)
+      StageData.Source sourceServerGroup = parentStageData.source
+
+      // Short-circuit if there is no source server group
+      if (sourceServerGroup == null || sourceServerGroup.serverGroupName == null) {
+        return []
+      }
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
maggieneterval: Good call, will replace the catch with a null check on `parentCreateServerGrouStage` (had originally lifted most of this code from the RollingRedBlackStrategy [here](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategy.groovy#L137) and assumed it was doing this for some reason I didn't understand but I think you're right that that is the only possible exception)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
maggieneterval: I so appreciate you reworking this logic and making it easier to understand!

Since for some reason a GCE red/black has a pin stage, but for AWS only a _rolling_ red/black has a pin stage, I think the GCE-specific pre-processor will continue need to add its own pin stage :cry: I have no context on the origin of that discrepancy but I'm hesitant to change it in case GCE users are relying on the current functionality.

maggieneterval:
@marchello2000 thank you for highlighting the AWS caveats and for refactoring all this strategy code: would it be easier for you if I didn't merge this until after your PRs are in? Otherwise I am happy to merge this now and update based on your refactors whenever they land!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RedBlackStrategy.groovy
maggieneterval: Wow good call, I had no idea you could select a strategy when doing an ad-hoc clone from the UI (at least in GCE, we disable strategies when doing an ad-hoc create server group from the UI so I had assumed this was the same for cloning). Do you think your refactor will resolve this?

maggieneterval: Just pushed a commit that addresses this for red/black by checking for an ancestor stage that is either a deploy stage or a clone stage, which at least fixes this for GCE!

marchello2000: i think we will still need something like this after refactor... to get the source. but seems like a reasonable fix for now (i will borrow it from you for AWS) :)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
marchello2000: I am impressed you pin for RB - it's a good idea, but pretty dangerous. We enabled it briefly on AWS and @dreynaud can tell you about the awesome incident it caused... anyway. most of the issues are around "guaranteeing" that the ASG gets unpinned which we don't do a very good job at today

marchello2000:
@maggieneterval - I don't want to block you from making progress, so feel free to merge if you are ok making changes to this code after the refactor. we can coordinate after i have a refactor on how you can take advantage of it

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
maggieneterval: Haha can't wait to hear about the incident... I only started looking into this because it recently came to our attention that in GCE, we are basically guaranteeing that the ASG will _never_ be unpinned correctly because we are snapshotting its capacity _after_ pinning it :face_with_head_bandage: (stay tuned for a fix for that in an upcoming PR)

maggieneterval:
@marchello2000 I am 100% okay with making changes to this code after the refactor, can't wait to clean it up! Thanks again for your help! :100: :100: :100:

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
dreynaud: I did a twitter webposting on it! It _will_ sound familiar to you 😅  https://twitter.com/FakeRyanGosling/status/1106714429247221761

maggieneterval: Amazing write-up! Love that it took y'all 24 hours to fix it and I think it will have taken us on the order of years to fix for GCE :see_no_evil: :see_no_evil: :see_no_evil: 

Excuse me while I LOL at your other tweets for the rest of the evening...

PR #3175, opened 2019-09-18
feat(sql): Conditionally allow MySQL Connector to be excluded from the build

scottfrederick:
This change allows a custom build of Spinnaker to exclude the `mysql-connector-java` driver from the generated distribution with a build command like:

`./gradlew orca-web:installDist -x test -PexcludeSqlDrivers`

The driver is included in the build by default, so the build results are not changed for existing open-source and custom builds. 

The Gradle dependency types were changed from `implementation` to `runtimeOnly` to reflect the fact that `mysql-connector-java` is not needed for compilation.


scottfrederick:
@robzienert @asher @ezimanyi

This is a follow-up / alternative to #3124.

PR #3177, opened 2019-09-18
fix(google): fix scale down cluster task in gce red/black

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4810

In response to the deprecation of GCE autoscaler's `ONLY_DOWN` mode, we set the mode to `OFF` in the `ScaleDownClusterTask` in this [PR](https://github.com/spinnaker/orca/pull/3087/files). However, when we scale down a server group with an autoscaler, we are actually enacting the scale-down by updating the capacity of the autoscaler, so if we set its mode to OFF, the scale-down will hang until the task times out. Since we are setting `min`, `max`, and `desired` capacity to 0, I don't think it will matter whether the mode is ONLY_DOWN or ON, since it will have to scale down in order to actuate the new capacity.

<details><summary>Before</summary>
<p>

6thykQUyAbX

</p>
</details>

<details><summary>After</summary>
<p>

NiUbZwBTMyh

</p>
</details>


maggieneterval:
@spinnakerbot :cherries: :pick: 1.15

maggieneterval:
@spinnakerbot cherries pick 1.16

maggieneterval:
@spinnakerbot :cherries: :pick: 1.16

PR #3180, opened 2019-09-19
fix(monitored deploy): log everything better

marchello2000:
This change cleans up and significantly improves logging for deployment monitors.
Additionally, it take advantage of the dynamic task timeout mechanism newly minted in orca (see #3159)

The following will be added to context for `deck` to parse and present to user:
* The current status the deployment monitor has provided (e.g. wait/continue/abort)
* In case of failures to communicate with the deployment monitor
* A better high level summary of response from deployment monitor

The following simple logging is improved:
* retrofit errors

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.monitoreddeploy;
 
+import com.google.common.io.CharStreams;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.config.DeploymentMonitorDefinition;
 import com.netflix.spinnaker.config.DeploymentMonitorServiceProvider;
@@ -25,12 +26,24 @@ import com.netflix.spinnaker.orca.TaskResult;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.MonitoredDeployStageData;
 import com.netflix.spinnaker.orca.deploymentmonitor.models.DeploymentStep;
 import com.netflix.spinnaker.orca.deploymentmonitor.models.EvaluateHealthResponse;
+import com.netflix.spinnaker.orca.deploymentmonitor.models.StatusExplanation;
+import com.netflix.spinnaker.orca.deploymentmonitor.models.StatusReason;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import retrofit.RetrofitError;
+import retrofit.client.Header;
+import retrofit.client.Response;
 
 public class MonitoredDeployBaseTask implements RetryableTask {
   private static final int MAX_RETRY_COUNT = 3;
@@ -54,8 +67,50 @@ public class MonitoredDeployBaseTask implements RetryableTask {
 
   @Override
   public long getTimeout() {
-    // TODO(mvulfson): Use DeploymentMonitorDefinition
-    return TimeUnit.MINUTES.toMillis(30);
+    // NOTE: This is not used since we override getDynamicTimeout
+    return 0;
+  }
+
+  @Override
+  public long getDynamicTimeout(Stage stage) {
+    ensureMonitorDefinition(stage);
+
+    final Duration defaultTimeout = Duration.ofMinutes(60);
+    long timeout;
+
+    try {
+      timeout = TimeUnit.MINUTES.toMillis(monitorDefinition.getMaxAnalysisMinutes());
+    } catch (Exception e) {
+      log.error(
+          "Failed to compute timeout for {}, returning {} min",
+          getClass().getSimpleName(),
+          defaultTimeout.toMinutes(),
+          e);
+
+      timeout = defaultTimeout.toMillis();
+    }
+
+    return timeout;
+  }
+
+  @Override
+  public @Nullable TaskResult onTimeout(@Nonnull Stage stage) {
+    ensureMonitorDefinition(stage);
+
+    ExecutionStatus taskStatus;
+    String message;
+
+    if (monitorDefinition.isFailOnError()) {
+      message =
+          "Deployment monitor failed to evaluate health in the allotted time, assuming failure because the monitor is configured to failOnError";
+      taskStatus = ExecutionStatus.TERMINAL;
+    } else {
+      message =
+          "Deployment monitor failed to evaluate health in the allotted time, proceeding anyway because the monitor is not configured to failOnError";
+      taskStatus = ExecutionStatus.FAILED_CONTINUE;
+    }
+
+    return buildTaskResult(TaskResult.builder(taskStatus), message);
   }
 
   @Override
@@ -63,13 +118,15 @@ public class MonitoredDeployBaseTask implements RetryableTask {
     MonitoredDeployStageData context = stage.mapTo(MonitoredDeployStageData.class);
 
     try {
-      this.stage = stage;
-      this.monitorDefinition =
-          deploymentMonitorServiceProvider.getDefinitionById(
-              context.getDeploymentMonitor().getId());
-
+      ensureMonitorDefinition(stage);
       return executeInternal();
     } catch (RetrofitError e) {
+      log.warn(
+          "HTTP Error encountered while talking to {}->{}, {}}",
+          monitorDefinition,
+          e.getUrl(),
+          getRetrofitLogMessage(e.getResponse()),
+          e);
 
       return handleError(context, e, true);
     } catch (DeploymentMonitorInvalidDataException e) {
@@ -120,10 +177,12 @@ public class MonitoredDeployBaseTask implements RetryableTask {
           monitorDefinition,
           e);
 
-      return TaskResult.builder(ExecutionStatus.TERMINAL)
-          // TODO(mvulfson)
-          // .context()
-          .build();
+      String userMessage =
+          String.format(
+              "Failed to get a valid response from deployment monitor %s, aborting because this deployment monitor is configured to failOnError",
+              monitorDefinition.getName());
+
+      return buildTaskResult(TaskResult.builder(ExecutionStatus.TERMINAL), userMessage);
     }
 
     log.warn(
@@ -132,10 +191,63 @@ public class MonitoredDeployBaseTask implements RetryableTask {
         monitorDefinition,
         e);
 
-    return TaskResult.builder(ExecutionStatus.SUCCEEDED)
-        // TODO(mvulfson)
-        // .context()
-        .build();
+    String userMessage =
+        String.format(
+            "Failed to get a valid response from deployment monitor %s, proceeding anyway because this deployment monitor is configured to not failOnError",
+            monitorDefinition.getName());
+
+    return buildTaskResult(TaskResult.builder(ExecutionStatus.SUCCEEDED), userMessage);
+  }
+
+  private void ensureMonitorDefinition(Stage stage) {
+    if (this.stage == null) {
+      MonitoredDeployStageData context = stage.mapTo(MonitoredDeployStageData.class);
+      this.stage = stage;
+      this.monitorDefinition =
+          deploymentMonitorServiceProvider.getDefinitionById(
+              context.getDeploymentMonitor().getId());
+    }
+  }
+
+  TaskResult buildTaskResult(
+      TaskResult.TaskResultBuilder taskResultBuilder, EvaluateHealthResponse response) {
+    List<StatusReason> statusReasons =
+        Optional.ofNullable(response.getStatusReasons()).orElse(Collections.emptyList());
+
+    String summary = "Deployment monitor requested to: " + response.getNextStep().getDirective();
+    StatusExplanation explanation = new StatusExplanation(summary, statusReasons);
+
+    return taskResultBuilder.context("deploymentMonitorReasons", explanation).build();
+  }
+
+  TaskResult buildTaskResult(TaskResult.TaskResultBuilder taskResultBuilder, String summary) {
+    StatusExplanation explanation = new StatusExplanation(summary);
+
+    return taskResultBuilder.context("deploymentMonitorReasons", explanation).build();
+  }
+
+  private String getRetrofitLogMessage(Response response) {
+    if (response == null) {
+      return "<NO RESPONSE>";
+    }
+
+    String body = "";
+    String status = "";
+    String headers = "";
+
+    try {
+      status = String.format("%d (%s)", response.getStatus(), response.getReason());
+      body =
+          CharStreams.toString(
+              new InputStreamReader(response.getBody().in(), StandardCharsets.UTF_8));
+      headers =
+          response.getHeaders().stream().map(Header::toString).collect(Collectors.joining("\n"));
+    } catch (Exception e) {
+      log.error(
+          "Failed to fully parse retrofit error while reading response from deployment monitor", e);
+    }
+
+    return String.format("status: %s\nheaders: %s\nresponse body: %s", status, headers, body);
   }
 
   void sanitizeAndLogResponse(EvaluateHealthResponse response) {
--- a/orca-deploymentmonitor/src/main/java/com/netflix/spinnaker/config/DeploymentMonitorServiceProvider.java
+++ b/orca-deploymentmonitor/src/main/java/com/netflix/spinnaker/config/DeploymentMonitorServiceProvider.java
@@ -33,15 +33,18 @@ public class DeploymentMonitorServiceProvider {
   private static final Logger log = LoggerFactory.getLogger(DeploymentMonitorServiceProvider.class);
 
   private Client retrofitClient;
+  private RestAdapter.LogLevel retrofitLogLevel;
   private RequestInterceptor spinnakerRequestInterceptor;
   private List<DeploymentMonitorDefinition> deploymentMonitors;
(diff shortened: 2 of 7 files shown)


marchello2000:
@dreynaud if you have time - would appreciate your eyes. But not critical

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -54,22 +66,66 @@ public long getBackoffPeriod() {
...
     try {
-      this.stage = stage;
-      this.monitorDefinition =
-          deploymentMonitorServiceProvider.getDefinitionById(
-              context.getDeploymentMonitor().getId());
-
+      ensureMonitorDefinition(stage);
       return executeInternal();
     } catch (RetrofitError e) {
+      log.error(
```
cfieber: wonder about logging this at `error` if this might be a something we retry on

orca-deploymentmonitor/src/main/java/com/netflix/spinnaker/config/DeploymentMonitorServiceProvider.java
```diff
@@ -75,8 +75,7 @@ private synchronized DeploymentMonitorService getServiceByDefinition(
               .setRequestInterceptor(spinnakerRequestInterceptor)
               .setEndpoint(definition.getBaseUrl())
               .setClient(retrofitClient)
-              // TODO(mvulfson): Change logging level
-              .setLogLevel(RestAdapter.LogLevel.HEADERS_AND_ARGS)
+              .setLogLevel(RestAdapter.LogLevel.BASIC)
```
cfieber: I think there is a `retrofitLogLevel` `@Bean` floating around the spring context that makes this value configurable via yaml

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -132,10 +190,60 @@ private TaskResult handleError(
...
+    String status = "";
+    String headers = "";
+
+    try {
+      status = String.format("%d (%s)", response.getStatus(), response.getReason());
+      body = CharStreams.toString(new InputStreamReader(response.getBody().in()));
+      headers =
+          response.getHeaders().stream().map(Header::toString).collect(Collectors.joining("\n"));
+    } catch (Exception unused) {
+      // Nothing to do
```
dreynaud: why nothing to do? What kind of exception do you expect in here, IO stuff? What if there's a sneaky NPE too?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -54,22 +66,66 @@ public long getBackoffPeriod() {
 
   @Override
   public long getTimeout() {
-    // TODO(mvulfson): Use DeploymentMonitorDefinition
-    return TimeUnit.MINUTES.toMillis(30);
+    // NOTE: This is not used since we override getDynamicTimeout
+    return 0;
```
luispollo: If it's not meant to be used, shouldn't it throw an exception instead? Others might miss your comment here...

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -54,22 +66,66 @@ public long getBackoffPeriod() {
...
+  }
+
+  @Override
+  public long getDynamicTimeout(Stage stage) {
+    ensureMonitorDefinition(stage);
+
+    final Duration defaultTimeout = Duration.ofMinutes(60);
```
luispollo: Do we have any conventions on whether defaults like this should be defined as constants?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -132,10 +190,60 @@ private TaskResult handleError(
...
+      return "<NO RESPONSE>";
+    }
+
+    String body = "";
+    String status = "";
+    String headers = "";
+
+    try {
+      status = String.format("%d (%s)", response.getStatus(), response.getReason());
+      body = CharStreams.toString(new InputStreamReader(response.getBody().in()));
```
dreynaud: it seems like we should specify an encoding somewhere in here

orca-deploymentmonitor/src/main/java/com/netflix/spinnaker/config/DeploymentMonitorServiceProvider.java
marchello2000: yeah, i was trying to find it but didn't look hard enough, let me look more harder... 
one thing i discovered... is that retrofit logging (outside of basic) kinda sorta sucks... hence i did it myway so i can actually find it/read it/use it

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -132,10 +190,60 @@ private TaskResult handleError(
...
+    List<StatusReason> statusReasons =
+        Optional.ofNullable(response.getStatusReasons()).orElse(Collections.emptyList());
+
+    String summary = "Deployment monitor requested to: " + response.getNextStep().getDirective();
+    StatusExplanation explanation = new StatusExplanation(summary, statusReasons);
+
+    return taskResultBuilder.context("deploymentMonitorReasons", explanation).build();
+  }
+
+  TaskResult buildTaskResult(TaskResult.TaskResultBuilder taskResultBuilder, String summary) {
```
luispollo: Wouldn't it be cleaner to add a `withSummary` method to `TaskResult.TaskResultBuilder` if it's such a common use case, and a usage we want to promote more? Same comment on the other version of the method above (maybe a `withResponseDetails` method?)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: hm.. that's fair, let me change to `warn`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -132,10 +190,60 @@ private TaskResult handleError(
...
+    String userMessage =
+        String.format(
+            "Failed to get a valid response from deployment monitor %s, proceeding anyway because this deployment monitor is configured to not failOnError",
+            monitorDefinition.getName());
+
+    return buildTaskResult(TaskResult.builder(ExecutionStatus.SUCCEEDED), userMessage);
+  }
+
+  private void ensureMonitorDefinition(Stage stage) {
+    if (this.stage == null) {
```
luispollo: This should be `!= null` right?

```suggestion
    if (this.stage != null) {
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -132,10 +190,60 @@ private TaskResult handleError(
...
+      MonitoredDeployStageData context = stage.mapTo(MonitoredDeployStageData.class);
+      this.stage = stage;
+      this.monitorDefinition =
+          deploymentMonitorServiceProvider.getDefinitionById(
+              context.getDeploymentMonitor().getId());
+    }
```
luispollo: What's the else case here? Especially since the method is called "ensure", I'd expect it to throw an exception?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: you think `utf-8`? i don't even know what the default for retrofit (accept header) is... let me dig

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -54,22 +66,66 @@ public long getBackoffPeriod() {
...
+    String message;
+
+    if (monitorDefinition.isFailOnError()) {
+      message =
+          "Deployment monitor failed to evaluate health in the allotted time, assuming failure because the monitor is configured to failOnError";
+      taskStatus = ExecutionStatus.TERMINAL;
+    } else {
+      message =
+          "Deployment monitor failed to evaluate health in the allotted time, proceeding anyway because the monitor is not configured to failOnError";
+      taskStatus = ExecutionStatus.FAILED_CONTINUE;
```
dreynaud: wouldn't `FAILED_CONTINUE` fail the deploy stage entirely but continue with the rest of the pipeline?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
luispollo: Log an error (or warning) at least?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -54,22 +66,66 @@ public long getBackoffPeriod() {
...
+
+  @Override
+  public long getDynamicTimeout(Stage stage) {
+    ensureMonitorDefinition(stage);
+
+    final Duration defaultTimeout = Duration.ofMinutes(60);
+    long timeout;
+
+    try {
+      timeout = TimeUnit.MINUTES.toMillis(monitorDefinition.getMaxAnalysisMinutes());
```
dreynaud: why would that throw an exception?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
dreynaud: I like

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: I don't really expect an exception (since I am not sure what retrofit may fail to include). but you are right I should at least log it - what i am doing is not acceptable.
and if there is an exception, we log best effort

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -54,22 +66,66 @@ public long getBackoffPeriod() {
...
-      this.stage = stage;
-      this.monitorDefinition =
-          deploymentMonitorServiceProvider.getDefinitionById(
-              context.getDeploymentMonitor().getId());
-
+      ensureMonitorDefinition(stage);
       return executeInternal();
     } catch (RetrofitError e) {
+      log.error(
+          "HTTP Error encountered while talking to {}->{}, {}}",
```
dreynaud: I like

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
dreynaud: I'm a bit ignorant about this but I'd say `utf-8` all the things is probably better than relying on some unspecified system default

marchello2000: yes, agree, updated.. apparently... google/android JVM defaults to UTF-8 (which of course can be changed). Others seem to have VERY convoluted logic to figure out the default encoding depending on your locale 😨

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: `TaskResultBuilder` is used with every task, so I don't want to add specifics related to my stage (deployment monitor stuff). But you are not wrong and we would benefit from more standardization of how to handle various errors, but that's a big task for another day :)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: this just sets instance vars (sort of lazily) so `==` is right. basically this is called from a bunch of spots to "init" the class instance vars

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: none really, see my above comment.. think of  this as an idempotent (and absolutely not thread safe - which is OK in this case since messages are handled serially on a single thread) init method of sorts

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: yeah, this is a bit weird. maybe we should just get rid of this method (but that's a bit more work that i don't want to lump here).
If you look at `RunTaskHandler` (https://github.com/spinnaker/orca/blob/725db2b0cadd6262ba6c8bafb4888f57fcf7f358/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt#L241) it always calls the `getDynamicTimeout`. and the only place where `getTimeout` is called by https://github.com/spinnaker/orca/blob/725db2b0cadd6262ba6c8bafb4888f57fcf7f358/orca-core/src/main/java/com/netflix/spinnaker/orca/RetryableTask.java#L30 but since i override it it will never be called

Could add an exception, but i would rather just delete this function in the medium term

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: Oh gross, and you're right... jeez, i demote myself from understanding orca
thanks!

marchello2000: wait... no, i don't think so... i am so confused now

dreynaud: 🙃

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
luispollo: Doh! 🤦‍♂ Of course. Please forget that I asked that.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: everything works as expected :) stop confusing me DANIEL! :)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: not that i know of

PR #3181, opened 2019-09-19
fix(google): do not pin source server group capacity in red/black

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4812

- When deploying to AWS, only the rolling red/black strategy pins the source server group's min capacity to its desired capacity (see [AwsDeployStagePreProcessor](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/AwsDeployStagePreProcessor.groovy#L155) and this [conversation](https://github.com/spinnaker/orca/pull/3173#discussion_r325905128) for details on how/why).
- This change brings the GCE implementation of red/black to parity with the AWS implementation: we will no longer pin the source server group's min capacity to its desired capacity. The potential drawback to this is that during the period when both the source and target server groups are taking traffic, the autoscaler may scale down the source server group as it will only be taking 50% of the traffic: in the case a rollback is necessary, it will need to scale back up. However, as Netflix has experienced, the potential downsides of pinning the source server group are much worse, as there are many unpredictable ways to get into a state where the server group is never unpinned (see Netflix post-mortem [here](https://twitter.com/FakeRyanGosling/status/1106714429247221761)).
- The other upside of bringing the GCE red/black to parity with the AWS red/black is the maintainability of the common strategy code, which @marchello2000 is currently refactoring.
- In the next Spinnaker release (1.17), I will summarize the rationale behind this change in the release notes so that any GCE users relying on the existing functionality are aware of the change.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
@@ -16,21 +16,14 @@
 
 package com.netflix.spinnaker.orca.clouddriver.pipeline.providers.gce;
 
-import static com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategySupport.getSource;
-
 import com.google.common.collect.ImmutableList;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.providers.aws.ApplySourceServerGroupCapacityStage;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.providers.aws.CaptureSourceServerGroupCapacityTask;
-import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.AbstractDeployStrategyStage;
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/gce/GceDeployStagePreProcessor.java
```diff
@@ -92,28 +63,12 @@ public GceDeployStagePreProcessor(
...
   public boolean supports(Stage stage) {
     StageData stageData = stage.mapTo(StageData.class);
     return stageData.getCloudProvider().equals("gce");
   }
 
-  private static boolean shouldPinSourceServerGroup(String strategy) {
-    return Strategy.fromStrategyKey(strategy) == Strategy.RED_BLACK;
-  }
-
   private Optional<Map<String, Object>> getResizeContext(StageData stageData) {
```
plumpy: This method can also be :hocho:

maggieneterval: good call, :hocho:  now...

PR #3184, opened 2019-09-20
fix(monitored deploy): don't store any context in the task

marchello2000:
Somehow... I forgot that tasks are singletons and shouldn't have any state stored...
Simple refactor to pass along the vars that are needed all the time


marchello2000:
since this is a simple refactor, i am going to merge this in

PR #3185, opened 2019-09-20
feat(rollback): avoid scaling down during rollback

dreynaud:
Should partly address spinnaker/spinnaker#4895 by avoiding resizing the restore server group down.
This does not address concurrent actions issues, so it is probably still possible to have race conditions result in bad things.

So fundamentally this changes:
* computes the max of the current restore server group size and the rollback server group size
* only generates a resize stage if that computed size is bigger than the current size

This also opens the possibility to skip the snapshot/restoreMin tasks if we are not going to resize (which would pin the min of the restore server group).


dreynaud:
cc @luispollo

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/ResizeStrategy.groovy
```diff
@@ -56,6 +56,13 @@ interface ResizeStrategy {
     Integer max
     Integer desired
     Integer min
+
+    // from https://codereview.stackexchange.com/questions/8801/returning-groovy-class-fields-as-a-map
+    public Map asMap() {
```
cfieber: I think you can just call `.properties()` to get the value as a map?

cfieber: or I guess that has some groovy magic-junk in it?

dreynaud:
cc @gal-yardeni

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/ResizeStrategy.groovy
dreynaud: I get this though

```
MissingMethodException: No signature of method: Capacity.properties() is applicable for argument types: () values: []
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
```diff
@@ -124,10 +127,73 @@ class ExplicitRollback implements Rollback {
...
+    }
+
+    Map resizeServerGroupContext = new HashMap(parentStage.context) + [
+      action                       : ResizeStrategy.ResizeAction.scale_exact.toString(),
+      capacity                     : newRestoreCapacity.asMap(),
+      asgName                      : restoreServerGroupName,
+      targetLocation               : restoreServerGroup.getLocation(),
+      account                      : restoreServerGroup.credentials,
+      cloudProvider                : restoreServerGroup.cloudProvider,
+      pinMinimumCapacity           : true,
```
marchello2000: wy do you need this if you already set min=desired

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
```diff
@@ -124,10 +127,73 @@ class ExplicitRollback implements Rollback {
...
+    if (currentCapacity == newRestoreCapacity) {
+      log.info('Skipping resize stage because the current capacity of the restore server group {} would be unchanged ({})',
+        restoreServerGroupName, newRestoreCapacity)
+      return null
+    }
+
+    Map resizeServerGroupContext = new HashMap(parentStage.context) + [
+      action                       : ResizeStrategy.ResizeAction.scale_exact.toString(),
+      capacity                     : newRestoreCapacity.asMap(),
+      asgName                      : restoreServerGroupName,
```
marchello2000: add `serverGroupName`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
```diff
@@ -124,10 +127,73 @@ class ExplicitRollback implements Rollback {
...
+      targetLocation               : restoreServerGroup.getLocation(),
+      account                      : restoreServerGroup.credentials,
+      cloudProvider                : restoreServerGroup.cloudProvider,
+      pinMinimumCapacity           : true,
+      targetHealthyDeployPercentage: targetHealthyRollbackPercentage
+    ]
+
+    return newStage(parentStage.execution, resizeServerGroupStage.type,
+      "Resize Restore Server Group to (min: ${newRestoreCapacity.min}, max: ${newRestoreCapacity.max}, desired: ${newRestoreCapacity.desired})",
```
marchello2000: nit: `Resize Restore` is pretty confusing wording.. maybe `Resize {serverGroupaName}`...?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
dreynaud: it's not needed, I thought I would just leave it like that to express the intent anyway

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
```diff
@@ -124,10 +127,73 @@ class ExplicitRollback implements Rollback {
...
+
+    try {
+      // TODO: throw some retries in there?
+      return oortHelper.getTargetServerGroup(
+        fromContext.credentials,
+        serverGroupName,
+        fromContext.location
+      ).orElse(null)
+    } catch(Exception e) {
+      log.error('Skipping resize stage because there was an error looking up {}', serverGroupName, e)
```
marchello2000: i think this error would be better in line 157/162.

that way you can differentiate between failure in asg lookup vs there legitimately not being an ASG

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
dreynaud: good point!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
dreynaud: I don't like it, but I will do it

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
```diff
@@ -124,10 +127,73 @@ class ExplicitRollback implements Rollback {
...
+    ResizeStrategy.Capacity newRestoreCapacity = [
+      max: Math.max(rollbackServerGroup.capacity.max, restoreServerGroup.capacity.max),
+      desired: Math.max(rollbackServerGroup.capacity.desired, restoreServerGroup.capacity.desired)
+    ]
+
+    // let's directly produce a capacity with a pinned min instead of relying on the resize stage
+    newRestoreCapacity.min = newRestoreCapacity.desired
+
+    ResizeStrategy.Capacity currentCapacity = restoreServerGroup.capacity
+    if (currentCapacity == newRestoreCapacity) {
```
luispollo: Shouldn't this be comparing with `.equals()` instead?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/support/ResizeStrategy.groovy
luispollo: If you're still interested, it's a property, not a method:
```
> import com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategy.Capacity
> 
> Capacity c = new Capacity(3,2,1)
> print(c.properties)
[class:class com.netflix.spinnaker.orca.kato.pipeline.support.ResizeStrategy$Capacity, max:3, min:1, desired:2]
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
dreynaud: it is Groovy so it does some magic behind the scenes that ultimately does call `.equals()`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
```diff
@@ -124,10 +127,73 @@ class ExplicitRollback implements Rollback {
...
-    stages << buildApplySourceServerGroupCapacityStage(parentStage, parentStage.mapTo(ResizeStrategy.Source))
+
+    // only restore the min if it was pinned, i.e. there was a resize
+    if (resizeStage != null) {
+      stages << buildApplySourceServerGroupCapacityStage(parentStage, parentStage.mapTo(ResizeStrategy.Source))
+    }
     return stages
   }
 
+  @Nullable TargetServerGroup lookupServerGroup(Stage parentStage, String serverGroupName) {
```
robzienert: nit: Wouldn't mind seeing `@Nonnull` etc added to the method args, too.

dreynaud: do we have an explicit convention around this or tooling that actually takes advantage of that?

Otherwise it seems overkill, for me the implicit contract is that things that are not explicitly `@Nullable` must be `@Nonnull`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
luispollo: Actually I think it's because of the way that [`@Canonical`](http://docs.groovy-lang.org/2.4.7/html/gapi/groovy/transform/Canonical.html) annotation is implemented (they use property values for `hashCode()` so `.equals()` ends up being the same as `==`. But that's TMI. 😄

luispollo:
@dreynaud Thanks a lot for tagging me for review. Still very early days for me so as you could see, not a lot of insightful comments, but for whatever it's worth, your changes made sense to me.

dreynaud:
Thanks for taking the time to look things over! 🙏

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/ExplicitRollback.groovy
dreynaud: chatted offline, will change the way we log when we get an empty optional

PR #3186, opened 2019-09-20
fix(pipelines): Copying of failed child pipeline outputs

luispollo:
This fixes an issue where the outputs of a failed stage that calls another pipeline would not be made available in the parent pipeline's subsequent stages.


luispollo:
@marchello2000 Could you please help me review this?

marchello2000:
thanks, @luispollo!

PR #3187, opened 2019-09-23
fix(webhooks): Construct Webhook status check URL honoring the origin…

srekapalli:
- Change here is to make the status check URL of a webhook to have the same protocol/scheme of the webhook.
- RestTemplate provider keys on the URL as a whole and it is important to keep the scheme as is from the original webhook so that the same RestTemplate provider as webhook will be used for making the status check call.

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
@@ -70,7 +70,7 @@ class WebhookService {
     )
     HttpHeaders headers = buildHttpHeaders(customHeaders)
     HttpEntity<Object> httpEntity = new HttpEntity<>(headers)
-    return restTemplateProvider.getRestTemplate().exchange(validatedUri, HttpMethod.GET, httpEntity, Object)
+    return restTemplateProvider.getRestTemplate(url).exchange(validatedUri, HttpMethod.GET, httpEntity, Object)
   }
 
   List<WebhookProperties.PreconfiguredWebhook> getPreconfiguredWebhooks() {
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
@@ -21,7 +21,6 @@ import com.fasterxml.jackson.core.JsonParseException
 import com.fasterxml.jackson.databind.JsonMappingException
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.jayway.jsonpath.JsonPath
-import com.jayway.jsonpath.PathNotFoundException
 import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.RetryableTask
 import com.netflix.spinnaker.orca.TaskResult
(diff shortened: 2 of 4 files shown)


orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
```diff
@@ -70,7 +70,7 @@ class WebhookService {
     )
     HttpHeaders headers = buildHttpHeaders(customHeaders)
     HttpEntity<Object> httpEntity = new HttpEntity<>(headers)
-    return restTemplateProvider.getRestTemplate().exchange(validatedUri, HttpMethod.GET, httpEntity, Object)
```
srekapalli: Method expects a URL to be provided so adding it though Groovy doesn't complain about it during compilation time

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
```diff
@@ -160,7 +160,6 @@ class MonitorWebhookTask implements OverridableTimeoutRetryableTask {
       return TaskResult.builder(statusMap[result.toString().toUpperCase()]).context(responsePayload).build()
     }
```
srekapalli: Removed unnecessary statement.

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -162,4 +157,41 @@ class CreateWebhookTask implements RetryableTask {
       return TaskResult.builder(ExecutionStatus.TERMINAL).context(outputs).build()
     }
   }
+
```
srekapalli: Below method adjusts the status check URL if the original webhook domain and status check URL domain given back matches.

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -119,23 +124,13 @@ class CreateWebhookTask implements RetryableTask {
...
-              statusUrl = JsonPath.compile(stageData.statusUrlJsonPath as String).read(response.body)
-            } catch (PathNotFoundException e) {
-              outputs.webhook << [error: e.message]
-              return TaskResult.builder(ExecutionStatus.TERMINAL).context(outputs).build()
-            }
+        String statusUrl
+        try {
+          statusUrl = determineWebHookStatusCheckUrl(response, stageData)
+        } catch (Exception e) {
+          log.error('Exception received while determining status change url', e)
```
marchello2000: We no longer bubble up the json parsing error (or other errors from within `determineWebHookStatusCheckUrl`) to the user as we used to here:
https://github.com/spinnaker/orca/pull/3187/files#diff-e737f2c3e6b9c31027d9d0670960c694L135

this means when something goes wrong we will have to dig in the logs, but it's better if the user can find the exceptions in the execution and self-service fix it (hopefully)

srekapalli: Good catch!.  Added the exception message to webhook output map.

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -32,13 +31,19 @@ import com.netflix.spinnaker.orca.webhook.service.WebhookService
...
 import org.springframework.web.client.HttpStatusCodeException
 
+import java.util.regex.Matcher
+import java.util.regex.Pattern
+
 @Slf4j
 @Component
 class CreateWebhookTask implements RetryableTask {
 
+  private static final Pattern URL_SCHEME = Pattern.compile("(.*)://(.*)")
```
marchello2000: was trying to see if there are ways to make the `URL` class do the "hard" work here but can't find anything.

srekapalli:
> @srekapalli I believe you mentioned there are some other URIs involved in the webhook flow besides status? How are those going to be handled

@cfieber : Actually they were all going through the same webHookService implementation and only status URL is derived from the webhook call response and that is the reason we needed special treatment for that.

PR #3193, opened 2019-09-24
fix(tasks): limit task maxBackoff time to 2min

asher:
- The `maxTaskBackoff` for `RetryableTask` `RunTask` messages was previously limited via `LockExtendingTaskExecutionInterceptor` which was removed by https://github.com/spinnaker/orca/pull/3136. This makes the prior backoff constraint the default as it assists with patching and cancellation of in-flight executions.


cfieber:
I think the limitation went in because we needed to be sure we refreshed the TTL on the lock - I'm not sure we need to constrain it any further. That said, it is probably better to revisit the task with some frequency - I suspect it might give us a better story around killing off things that have been cancelled etc

PR #3199, opened 2019-09-30
fix(monitored deploy): send complete notification on failure

marchello2000:
When deploy fails we need to sent completion notification to the deployment monitor

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
@@ -17,7 +17,6 @@ package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies
 
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.netflix.spinnaker.config.DeploymentMonitorServiceProvider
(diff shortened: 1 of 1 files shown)


marchello2000:
@dreynaud I also added a shrinky stage while i was at it

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
```diff
@@ -385,6 +396,28 @@ class MonitoredDeployStrategy implements Strategy {
       SyntheticStageOwner.STAGE_AFTER
     )
 
+    MonitoredDeployStageData stageData = parent.mapTo(MonitoredDeployStageData.class)
+    if (stageData.deploymentMonitor.id) {
+      def evalContext = stageData.getChildStageContext()
+      MonitoredDeployInternalStageData internalStageData = new MonitoredDeployInternalStageData()
```
dreynaud: is there anything there to indicate that the deployment failed?

marchello2000: not yet... there will be! for now, I just want to unblock a use case
https://github.com/spinnaker/orca/blob/master/orca-deploymentmonitor/src/main/java/com/netflix/spinnaker/orca/deploymentmonitor/models/DeploymentCompletedRequest.java#L35

PR #3200, opened 2019-10-01
fix(cfn): Add retries when force cache refreshing

xavileon:
For cache refreshing the CloudFormation cache can fail for
some reason (e.g. read timeout), but should generally succeed.
This patch makes this task more reliable by retrying if there was
a failure for up to 5 minutes.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/CloudFormationForceCacheRefreshTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/CloudFormationForceCacheRefreshTask.java
@@ -16,7 +16,7 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.providers.aws.cloudformation;
 
-import com.netflix.spinnaker.orca.Task;
+import com.netflix.spinnaker.orca.OverridableTimeoutRetryableTask;
 import com.netflix.spinnaker.orca.TaskResult;
 import com.netflix.spinnaker.orca.clouddriver.CloudDriverCacheService;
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/CloudFormationForceCacheRefreshTask.java
```diff
@@ -16,25 +16,29 @@
...
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 @Component
 public class CloudFormationForceCacheRefreshTask extends AbstractCloudProviderAwareTask
-    implements Task {
+    implements OverridableTimeoutRetryableTask {
```
marchello2000: Just making sure that you intended this to be `OverridableTimeoutRetryableTask` vs. `RetryableTask`

`OverridableTimeoutRetryableTask` allow the timeout to be overridden by the user in the stage config, is that what you want?

(https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt#L238)

xavileon:
Ah good point! Haven't thought about that. I don't think it need to be configurable but I see no harm on it in case the default I chose doesn't apply. Is there any hidden reason to not making it configurable? I think I'm fine either way.

marchello2000:
no real reason/harm in having it configurable, @xavileon  - just that it adds another point to think about when you're debugging a deploy that took way too long..? I will merge as is, and you can change it later if you want - unless you tell me otherwise - need to pull master and wait for travis anyway

xavileon:

PR #3202, opened 2019-10-01
fix(webhook): Don't try to deserialize fields we don't really need

marchello2000:
An assumption was made that body is a can be jackon parsed into a JSON object.
But that's not always the case, in fact, we actually account for it in the original parse:
https://github.com/spinnaker/orca/blob/master/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy#L71
But then disregard it in StageContext mapping

This change removes those fields from being parsed since we don't actually need them for anything
inside the stage (they were added in the attempt to be fully typed)

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/pipeline/WebhookStage.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/pipeline/WebhookStage.groovy
@@ -104,20 +104,26 @@ class WebhookStage implements StageDefinitionBuilder {
   }
 
   static class WebhookResponseStageData {
-    String statusEndpoint
-    Integer statusCodeValue
     String statusCode
-    Map body
     WebhookMonitorResponseStageData monitor
     String error
+
+    // NOTE: The fields below exist in the context because they are inserted by the CreateWebhookTask but they aren't
+    //       consumed by Spinnaker so they are commented out below - they are here for informational purposes only
+    //String statusEndpoint
+    //Integer statusCodeValue
+    //Map body
   }
 
   static class WebhookMonitorResponseStageData {
-    Integer statusCodeValue
-    String statusCode
-    Map body
     String error
-    String progressMessage
-    Number percentComplete
+
+    // NOTE: The fields below exist in the context because they are inserted by the MonitorWebhookTask but they aren't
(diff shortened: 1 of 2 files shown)


marchello2000:
Fixes https://github.com/spinnaker/spinnaker/issues/4881

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/pipeline/WebhookStage.groovy
```diff
@@ -104,20 +104,26 @@ class WebhookStage implements StageDefinitionBuilder {
...
 
   static class WebhookResponseStageData {
-    String statusEndpoint
-    Integer statusCodeValue
     String statusCode
-    Map body
     WebhookMonitorResponseStageData monitor
     String error
+
+    // NOTE: The fields below exist in the context because they are inserted by the CreateWebhookTask but they aren't
```
emjburns: 😆

emjburns:
What do you mean "they aren't consumed by spinnaker" - did they used to be?

marchello2000:
> What do you mean "they aren't consumed by spinnaker" - did they used to be?

We don't ever "read" them in orca (or anywhere) we store them for the user/deck to observe in the context. In my previous change, i wanted to be complete and define everything in the stagecontext for webhook, but `body` can be a string or a map so it's kinda annoying to parse out... (and causes exceptions) hence i just commented them out.

I think there is some marginal value in having them there... even if they are commented out... but i am happy to also remove them

emjburns:
I'd say leave them for now. I like having as much information as possible when you're in orca trying to fix a stage.

emjburns:
LGTM

ethanfrogers:
@marchello2000 is this okay to merge?

geojaz:
We'd love to get this one in and cut 1.16.3 asap :)

marchello2000:
@ethanfrogers yes, thank you! feel free to merge when tRavis does it's thing... or I can if I see it finish first

kevinawoo:

kevinawoo:
fyi, I merged/cherry-picked because @ethanfrogers told me to since he was ✈️ ing.

marchello2000:
Thanks, @kevinawoo !!!, CC: @louisjimenez as FYI that this should go to `1.16.3`

PR #3209, opened 2019-10-03
chore(core): Adding stage.toString()

jeyrschabu:
- added a toString method with the stage id & execution id


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -799,5 +799,10 @@ public boolean isJoin() {
         .collect(toList());
   }
 
+  @Override
+  public String toString() {
+    return "Stage {" + "id='" + id + '\'' + ", executionId=" + execution.getId() + '}';
```
dreynaud:
```suggestion
    return "Stage {id='" + id + "', executionId='" + execution.getId() + "'}";
```

PR #3210, opened 2019-10-03
fix(cfn): Return RUNNING if an error occurred

xavileon:
Force cache refreshing was a retryable task. However, it didn't return
RUNNING whenever the cache refresh failed, and thus the task failed
immediately. This patch makes the task return RUNNING whenever there's
an exception refreshing the cache so it actually retries.


xavileon:
Yes, sorry as well, I missed this too :-)

xavileon:

rampatnaik:
@spinnaker please reopen

rampatnaik:
@spinnakerbot I have similar issues. 
Please reopen

PR #3211, opened 2019-10-03
fix(cf): cf rolling red black did not resize old SG correctly

Jammy-Louie:
After some investigating it looks like the problem is that when attempting to restore the old number of instances for the old server group, the following task waits on the wrong number of server groups to be up. The old server group remains disabled (in cf terms - stopped) but the task waits for the old number of up instances.

fix for spinnaker/spinnaker#4786

Screen Shot 2019-10-03 at 11 25 54 AM


Jammy-Louie:

marchello2000:
@Jammy-Louie FYI: we've made a bunch of improvements in RRB for AWS - I saw that your strategy is quite different  so I didn't try to backport any of those changes, but just FYI

Jammy-Louie:
@marchello2000 Yup thats fine. We had to do something custom due certain limitations with cloud foundry

PR #3213, opened 2019-10-03
refactor(strategies): clean up some strategies from my past refactor

marchello2000:
also, remove a deprecated class


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/CustomStrategy.groovy
```diff
@@ -33,17 +33,11 @@ class CustomStrategy implements Strategy {
 
   @Override
   List<Stage> composeBeforeStages(Stage parent) {
-    return composeFlow(parent)
```
marchello2000: when i did the original composeflow refactor (https://github.com/spinnaker/orca/pull/3133/files) i took a shortcut and didn't fully use the `composeBefore/AfterStages` on strategy, this completes that work.

the thing to tripple check is that there were no "side-effects" in the compose calls since it used to be called during stage planning, but now it's not

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingPushStrategy.groovy
```diff
@@ -43,17 +43,17 @@ class RollingPushStrategy implements Strategy {
...
-      .findAll({ it.syntheticStageOwner == SyntheticStageOwner.STAGE_BEFORE })
-  }
+    def source = sourceResolver.getSource(parent)
 
-  @Override
-  List<Stage> composeAfterStages(Stage parent) {
-    return composeFlow(parent)
-      .findAll({ it.syntheticStageOwner == SyntheticStageOwner.STAGE_AFTER })
+    if (!source) {
+      throw new IllegalArgumentException("Could not find source server group for rolling push. Does the specified cluster exist?")
```
robzienert: Thoughts on using `SpinnakerException`?

marchello2000: yes, thank you!

marchello2000: actually, prob a userexception, this would happen if they try to rolling push without a prior asg

PR #3215, opened 2019-10-03
feat(Kayenta): Support canaryConfigName as canary identifier

nisanharamati:
- Allow canaries to be specified via canaryConfigName, in addition to canaryConfigId.
- canaryConfigId is still the default, and if it is provided, no additional work is necessary.
- if canaryConfigId is null, then if a canaryConfigName is provided, it can be used, together with the currentApplication name, to resolve a valid canaryConfigId using the Kayenta `/canaryConfig` end point.
- Once a list of candidate canaryConfigIds is produced, an additional check is performed to ensure only one candidate was returned. If no candidates were found, an error is thrown, and likewise if more than one candidate matches on both canaryConfigName and the current application name, an error is thrown.


orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/ResolveKayentaConfigIdTask.kt
```diff
@@ -0,0 +1,34 @@
...
+  override fun execute(stage: Stage): TaskResult {
+    val configName = stage.mapTo<String?>("/canaryConfigName")
+    val currentApplication = stage.execution.application
+    val canaryConfigList = kayentaService.getAllCanaryConfigs()
+    val candidates = canaryConfigList.asSequence()
+      .filter { it.name == configName && it.applications.contains(currentApplication) }
+      .toList()
+
+    if (candidates.size == 0) {
+      throw NoSuchElementException("Couldn't find a configId for configName $configName and application $currentApplication")
```
marchello2000: There is a `com.netflix.spinnaker.kork.web.exceptions.NotFoundException` that you should use instead
Also... 
```(suggestion)
throw NotFoundException("Couldn't find a canary configId for ...");
```

marchello2000: actually, use 
`com.netflix.spinnaker.kork.exceptions.UserException`

orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/ResolveKayentaConfigIdTask.kt
```diff
@@ -0,0 +1,34 @@
...
+    } else if (candidates.size > 1) {
+      throw IllegalArgumentException("Found more than one configId for configName $configName and application $currentApplication")
```
marchello2000: for new exceptions, we want to use the `com.netflix.spinnaker.kork.exceptions` an in this case `com.netflix.spinnaker.kork.exceptions.UserException` to make it clear that it was a user config error

orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/tasks/RunKayentaCanaryTask.kt
```diff
@@ -37,6 +37,10 @@ class RunKayentaCanaryTask(
   override fun execute(stage: Stage): TaskResult {
     val context = stage.mapTo<RunCanaryContext>()
 
+    if (context.canaryConfigId.isNullOrEmpty()) {
+      throw IllegalArgumentException("Invalid canaryConfigId. Was the correct configName or configId used?")
```
marchello2000: same here... either use `com.netflix.spinnaker.kork.exceptions.UserException` or (by old convention) `IllegalStateException`

orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/pipeline/RunCanaryPipelineStage.kt
```diff
@@ -37,6 +40,13 @@ class RunCanaryPipelineStage(
   private val log = LoggerFactory.getLogger(javaClass)
 
   override fun taskGraph(stage: Stage, builder: TaskNode.Builder) {
+      val context = stage.mapTo<RunCanaryContext>()
```
marchello2000: the one caveat we talked about here.. is that we will query the canary config every run which is suboptimal, but can be tracked separately.

nisanharamati: I was going to open an issue against Spinnaker/orca, but I can't find an `issues` tab for it. What's up with that?

nisanharamati: https://github.com/spinnaker/spinnaker/issues/5006

gregorymfoster:
@nisanharamati Love this pr. I was just writing something similar which leveraged the `pipelineDataMutator` extension point to patch in `canaryConfigId`s based on `canaryConfigName`s. This is obviously more ideal. 

Out of curiosity, what was your motivation for this change and what other changes do you see yourself making in this area?

nisanharamati:
@gregorymfoster Thanks!

The motivation for this is that we have some legacy use cases that use parametrized names as part of templated pipelines, so they only resolve configs at execution times, and do so by name. In order to be able to migrate those use cases off of our old legacy system, we needed to be able to support passing configuration Ids by name.

For future work, there's https://github.com/spinnaker/spinnaker/issues/5006 for starters, because the approach we're taking here is not terribly efficient.

I'm working on a more concrete roadmap for Kayenta work that I'm going to tackle over the next year, and I'll be happy to share it in a week or so once it's more polished.

gregorymfoster:
@nisanharamati sounds great! The more we can align on this, the more I can prevent Airbnb from duplicating work and help us upstream useful contributions :)

PR #3217, opened 2019-10-04
chore(mysql): add drop database to init script for convenience

marchello2000:
Just added this for convenience so it's easy to reset orca DB - this is only run manually


luispollo:
Thanks for adding this, @marchello2000!

PR #3219, opened 2019-10-04
fix(bake): Fail bake stage when image names don't match across regions

luispollo:
This adds some logic to `BakeStage` so that if the image names don't match across regions in a multi-region bake, then the stage optionally fails altogether. This is to prevent downstream problems with deploys that attempt to find existing images based on the name and will fail if they differ across regions for.

TO-DO:
- [x] Figure out how to enable feature toggles in tests
- [x] Add condition based on stage context (to enable user opting out on the UI later)


luispollo:
@marchello2000 @cfieber Please review.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -120,7 +127,7 @@ class BakeStage implements StageDefinitionBuilder {
       def globalContext = [
         deploymentDetails: relatedBakeStages.findAll{it.context.ami || it.context.imageId}.collect { Stage bakeStage ->
           def deploymentDetails = [:]
-          ["ami", "imageId", "amiSuffix", "baseLabel", "baseOs", "refId", "storeType", "vmType", "region", "package", "cloudProviderType", "cloudProvider"].each {
+          ["ami", "imageId", "imageName", "amiSuffix", "baseLabel", "baseOs", "refId", "storeType", "vmType", "region", "package", "cloudProviderType", "cloudProvider"].each {
```
robzienert: Let's make this a class constant instead.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -129,7 +136,28 @@ class BakeStage implements StageDefinitionBuilder {
...
+        def imageNames = globalContext.deploymentDetails.stream().map({details -> details['imageName']})
+          .distinct()
+          .collect(Collectors.toList())
+
+        if (imageNames.size() > 1) {
+          throw new IllegalStateException("Non-matching image names found in different regions: ${imageNames}. " +
+          "Failing bake stage as this would cause deployment failures. Please re-run the bake.")
+        }
+      }
+
```
robzienert: I'd maybe write this more like...

1. Use `Joiner` to create the `imageNames` variable: https://guava.dev/releases/16.0/api/docs/com/google/common/base/Joiner.html
2. Use `com.netflix.spinnaker.kork.exceptions` classes instead of root exceptions.
3. The exception message could be clearer. Perhaps something like this? `Image names found in different regions do not match: $imageNames. Re-run the bake to protect against deployment failures.`

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -129,7 +136,28 @@ class BakeStage implements StageDefinitionBuilder {
...
       TaskResult.builder(ExecutionStatus.SUCCEEDED).outputs(globalContext).build()
     }
+
+    def failOnImageNameMismatch() {
```
robzienert: Avoid using `def` at all times in Groovy's `main` source set, especially in method contracts.

luispollo: Any particular reason why?

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -129,7 +136,28 @@ class BakeStage implements StageDefinitionBuilder {
           return deploymentDetails
         }
       ]
+
+      // TODO: add condition for user-specified opt-out setting in addition to feature toggle
```
marchello2000: you don't have to wait for the UI to be implemented for this - you can implement it right away (users can always tweak the json before the UI is ready)

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
marchello2000: ideally, we want to be on Java/Kotlin instead of groovy. When we make changes to the groovy code there are two options: 
1. convert to Java
2. keep it in groovy but make it as java as possible (all valid java code is valid groovy code)

and... type inference like this is hard to read and can be error prone

marchello2000: (oh, and as Rob said, this applies to `main` not `test` source sets

emjburns: FYI backend conventions are here: https://www.spinnaker.io/community/contributing/back-end-code/

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
luispollo: I know. This was just a note-to-self... 😄  It's on the TO-DO list above.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
luispollo: Thanks for the context @marchello2000. In that case, shouldn't we apply that rule to any declaration (not just methods)? I see a bunch of `def variable` all over, and was declaring my variables in the same way. Should I always declare them with a type then?

marchello2000: Ideally, yes, I try not to use `def`s at all in any new/changed groovy code (other than tests)

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
luispollo: Done. For #2, used `ConstraintViolationException`, which was the closest thing I could find. Let me know if you're OK with that or if we should add a new exception type to kork.

luispollo:
@robzienert Please look this over one more time when you have a chance.
@marchello2000 @cfieber You too, pretty please. 😃

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -129,7 +157,31 @@ class BakeStage implements StageDefinitionBuilder {
...
         }
       ]
+
+      if (failOnImageNameMismatchEnabled() && !stage.context.dontFailOnImageNameMismatch) {
+         List<String> distinctImageNames = globalContext.deploymentDetails.stream()
+          .map({details -> details['imageName']})
+          .distinct()
+          .collect(Collectors.toList())
+
+        Joiner joiner = Joiner.on(", ")
```
robzienert: Why create this object here if it is only used within the following conditional?

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -129,7 +157,31 @@ class BakeStage implements StageDefinitionBuilder {
...
+        }
+      }
+
       TaskResult.builder(ExecutionStatus.SUCCEEDED).outputs(globalContext).build()
     }
+
+    private boolean failOnImageNameMismatchEnabled() {
+      try {
+        return dynamicConfigService.isEnabled("stages.bake.failOnImageNameMismatch", false)
+      } catch (Exception e) {
```
robzienert: Are you sure swallowing any exception without logging is a good idea?

luispollo: I would generally ask the same question, but the code @marchello2000 and I looked at as an example of using `DynamicConfigService` looked like this:
https://github.com/spinnaker/orca/blob/fba04ec8cfc5e9c8d5d0cb6219b2507822ae2358/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/StageDefinitionBuilder.java#L106-L120

...which is obviously not an excuse, but the method doesn't declare any exceptions, and I was told a common practice is to throw descendants of `RuntimeException`, so I made the incorrect assumption that returning the default when anything went wrong was OK.

I see now that most calls are actually not within a `try/catch` block. Should I just remove them?

robzienert: Definitely not the correct approach: Just because a bad pattern is done somewhere else doesn't mean we should spread it elsewhere! We don't do checked exceptions, so looking to see if a method has declared exceptions isn't going to get you very far. :)

It's up to you on if you catch an exception, but if you're going to catch and return a default value _the least_ you should do is `log.error` or `log.warn` the case. If this method failed for some reason (like NPE, or whatever), it would be phenomenally difficult to detect.

luispollo: > Definitely not the correct approach: Just because a bad pattern is done somewhere else doesn't mean we should spread it elsewhere! We don't do checked exceptions, so looking to see if a method has declared exceptions isn't going to get you very far. :)

[Agreed](https://github.com/spinnaker/orca/pull/3180#discussion_r326297016). I feel like I'm watching myself reviewing my own code. 😄

marchello2000:
Looks good, @robzienert i see your concerns are addressed.
Going to merge this when it goes green, unless you have other concerns

PR #3220, opened 2019-10-07
fix(pipeline): allow mid-task spel evaluation to reference prior stage outputs

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4932

- chore(pipeline): add buildExecutionContext test to ContextParameterProcessorSpec

  Adds test to demonstrate that currently, the context returned by `ContextParameterProcessor.buildExecutionContext`, when used as the context for `ContextParameterProcessor.process`, does not allow for the evaluation of variables that refer to the outputs of a prior stage.

- fix(pipeline): allow mid-task SpEL evaluation to reference prior stage outputs

  Changes `ContextParameterProcessor.buildExecutionContext` to return `StageContext`, which extends a `ForwardingMap` in order to override `get` to search the stage's ancestors' outputs if a given key does not exist on the delegate object. Uses the `Map` previously returned by this method as the delegate. Removes the `includeStageContext` argument since all existing consumers want stage context included. Updates `ExpressionAware` interface to use updated `buildExecutionContext` per todo.

To do: Investigate whether tasks like `ExpressionPreconditionTask` even need to explicitly take another pass at expression evaluation, since presumably any resolvable variables were evaluated in the `StartTaskHandler`. I think only if the task itself fetches an artifact that may contain SpEL should we need another evaluation pass?

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
@@ -17,12 +17,10 @@
 package com.netflix.spinnaker.orca.q.handler
 
 import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.module.kotlin.convertValue
 import com.netflix.spinnaker.kork.expressions.ExpressionEvaluationSummary
 import com.netflix.spinnaker.orca.exceptions.ExceptionHandler
 import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper
 import com.netflix.spinnaker.orca.pipeline.expressions.PipelineExpressionEvaluator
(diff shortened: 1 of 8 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
```diff
@@ -130,23 +129,12 @@ interface ExpressionAware {
...
     StageContext(stage, contextParameterProcessor.process(
       stage.context,
-      (stage.context as StageContext).augmentContext(stage.execution),
+      contextParameterProcessor.buildExecutionContext(stage),
       true,
       summary
     )
     )
 
-  // TODO (mvulfson): Ideally, we opt out of this method and use ContextParameterProcessor.buildExecutionContext
```
marchello2000: ❤️ You are brave soul, @maggieneterval, for tackling this :)

PR #3222, opened 2019-10-10
feat(monitored deploy): Support rollback on failure

marchello2000:
Add rollback support to monitored deploy strategy.
It's slightly different from the default you get with CreateServerGroup, hence it is removed from there for this strategy

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
@@ -26,6 +26,7 @@ import com.netflix.spinnaker.orca.clouddriver.pipeline.monitoreddeploy.EvaluateD
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.CloneServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.CreateServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.DestroyServerGroupStage
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.DisableServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.PinServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.ResizeServerGroupStage
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.DetermineTargetServerGroupStage
@@ -58,6 +59,14 @@ class MonitoredDeployStrategy implements Strategy {
 
   @Override
   List<Stage> composeBeforeStages(Stage stage) {
+    def stageData = stage.mapTo(MonitoredDeployStageData)
+
+    if (stageData.deploymentMonitor?.id) {
+      // Before we begin deploy, just validate that the given deploy monitor is registered
+      // Note: getDefinitionById will throw if no monitor is registered with the given ID
+      deploymentMonitorServiceProvider.getDefinitionById(stageData.deploymentMonitor.id)
+    }
+
     if (stage.context.useSourceCapacity) {
       stage.context.useSourceCapacity = false
     }
@@ -67,7 +76,6 @@ class MonitoredDeployStrategy implements Strategy {
     def savedCapacity = stage.context.savedCapacity ?: stage.context.capacity?.clone()
     stage.context.savedCapacity = savedCapacity
 
-    // FIXME: this clobbers the input capacity value (if any). Should find a better way to request a new asg of size 0
     stage.context.capacity = [
       min    : 0,
       max    : 0,
@@ -103,11 +111,6 @@ class MonitoredDeployStrategy implements Strategy {
       deploySteps.add(100)
     }
 
-    if (stageData.deploymentMonitor?.id) {
-      // Before we begin deploy, just validate that the given deploy monitor is registered
-      deploymentMonitorServiceProvider.getDefinitionById(stageData.deploymentMonitor.id)
-    }
-
     // Get source ASG from prior determineSourceServerGroupTask
     def source = null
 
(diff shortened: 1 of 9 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
```diff
@@ -89,13 +75,15 @@ class MonitoredDeployStrategy implements Strategy {
...
 
-    // FIXME: this clobbers the input capacity value (if any). Should find a better way to request a new asg of size 0
     stage.context.capacity = [
       min    : 0,
       max    : 0,
       desired: 0
     ]
 
+    // Don't allow old-school "rollback" key in the deploy stage, we handle our own rollback
+    stage.context.remove("rollback")
```
gal-yardeni: In which cases will we have "rollback" here as one of the stages?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
```diff
@@ -421,12 +410,74 @@ class MonitoredDeployStrategy implements Strategy {
...
 
+  List<Stage> composeRollbackStages(Stage parent) {
+    CreateServerGroupStage.StageData stageData = parent.mapTo(CreateServerGroupStage.StageData)
+    MonitoredDeployStageData monitoredDeployStageData = parent.mapTo(MonitoredDeployStageData)
+    String deployedServerGroupName = stageData.getServerGroup()
+
+    // Does the user want an automatic rollback?
+    if ((monitoredDeployStageData.failureActions.rollback != FailureActions.RollbackType.Automatic) &&
+      (monitoredDeployStageData.failureActions.rollback != FailureActions.RollbackType.Manual)) {
+      log.warn("Not performing automatic rollback on failed deploy of ${deployedServerGroupName ?: '<NO SERVER GROUP CREATED>'} because no rollback was requested by user in pipeline config")
```
gal-yardeni: Is this even a valid scenario? where the user can specify he doesn't want rollback?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
marchello2000: this is a context entry that is used by other strategies (and added by the UI if you use, say RedBlack)
I am removing it to reduce confusion when looking at execution JSON  since Monitored will not honor this setting - it will use `failureActions` instead. So it should never really be here, unless the user had a deploy stage using RB and then switched to Monitored strategy on that deploy stage

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
marchello2000: yes, that's the default for all strategies (except for Monitored - we we are opting users in by default and they can optout)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
```diff
@@ -421,12 +410,74 @@ class MonitoredDeployStrategy implements Strategy {
...
+    stages << newStage(
+      parent.execution,
+      RollbackClusterStage.PIPELINE_CONFIG_TYPE,
+      "Rollback ${stageData.cluster}",
+      [
+        credentials              : stageData.credentials,
+        cloudProvider            : stageData.cloudProvider,
+        regions                  : [stageData.region],
+        serverGroup              : stageData.serverGroup,
+        stageTimeoutMs           : MINUTES.toMillis(30), // timebox a rollback to 30 minutes
```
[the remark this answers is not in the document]
marchello2000: It's a bit arbitraty, but i reused/copy pasted from https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy#L133

The idea here is that we want to make sure that we don't stay in this "bad" state for long (bad being the deploy failed and we are rolling back, normally the rollback should take a few minutes at most). If the rollback times out the user will (hopefully) be notified because the pipeline will fail. (assuming they have notifications turned on)

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategySpec.groovy
```diff
@@ -0,0 +1,275 @@
...
+    when:
+    def failureStages = strategy.composeOnFailureStages(stage)
+
+    then: 'should unpin source'
+    failureStages.size() == 1
+    failureStages[0].type == PinServerGroupStage.TYPE
+    failureStages[0].context.serverGroupName == sourceServerGroupName
+    failureStages[0].name == "Unpin ${sourceServerGroupName}"
+  }
+
```
srekapalli: Tests are very helpful to read 💯

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
```diff
@@ -44,42 +46,26 @@ import java.util.concurrent.TimeUnit
...
-  ScaleDownClusterStage scaleDownClusterStage
-
   @Autowired
   DeploymentMonitorServiceProvider deploymentMonitorServiceProvider
 
   @Override
   List<Stage> composeBeforeStages(Stage stage) {
+    def stageData = stage.mapTo(MonitoredDeployStageData)
+
+    if (stageData.deploymentMonitor?.id) {
```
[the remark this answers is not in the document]
marchello2000: Yes. let me actually a comment about that - this has been confusing to at least two people

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategySpec.groovy
marchello2000: thanks!

gal-yardeni: +1!

PR #3232, opened 2019-10-17
feat(EvaluateVariables): Support complex objects in EvaluateVariables stage

christopherthielen:
You can create an adhoc spel expression that resolves to a complex object...

Screen Shot 2019-10-16 at 8 35 03 PM

... and we will happily expand the entire object.

Screen Shot 2019-10-16 at 8 33 54 PM

However, if you try to do the same thing in Evaluate Variables stage, you get an error:

```
java.lang.IllegalArgumentException: Unable to map context to class com.netflix.spinnaker.orca.pipeline.EvaluateVariablesStage$EvaluateVariablesStageContext
...
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `java.lang.String` out of START_ARRAY token
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: com.netflix.spinnaker.orca.pipeline.EvaluateVariablesStage$EvaluateVariablesStageContext[\"variables\"]
```

This PR changes `Variable` type from `[ key: string, value: string ]` to `[key: string, value: object]`

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/EvaluateVariablesStage.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/EvaluateVariablesStage.java
@@ -51,10 +51,10 @@ public class EvaluateVariablesStage implements StageDefinitionBuilder {
 
   public static class Variable {
     private String key;
-    private String value;
+    private Object value;
 
     @JsonCreator
-    public Variable(@JsonProperty("key") String key, @JsonProperty("value") String value) {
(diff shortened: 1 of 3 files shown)


cfieber:
This is a good change - I wonder how many people are working around the stringyness of the current implementation elsewhere  and if any of those spots might bust as a result

christopherthielen:
@cfieber If users _are_ currently working around this, they're coercing their expression to a string in Evaluate Expressions stage already.  I wouldn't expect this change to break since strings are still supported.  Unless I'm overlooking some other possibility...

emjburns:
Nice! Thanks!

PR #3233, opened 2019-10-17
feat(kubernetes): support rolling restart operation for deployments

maggieneterval:
Closes spinnaker/spinnaker#4757

As the many comments/emojis on this [Kubernetes feature request](https://github.com/kubernetes/kubernetes/issues/13488) indicate, there are several valid user stories around issuing a rolling restart of a Deployment. The `kubectl rollout restart` command was released with kubectl version 1.15, and we currently use kubectl version 1.16 in Clouddriver. Like the other kubectl rollout commands, it is valid for Deployments, StatefulSets, and DaemonSets. We currently only expose the ad-hoc rollout commands for Deployments in Deck, but could at a later date expose them for StatefulSets and DaemonSets if there is evidence from users that it would be useful.

Related PRs:
Clouddriver: https://github.com/spinnaker/clouddriver/pull/4100
Deck: https://github.com/spinnaker/deck/pull/7538

863ee1d0-ca4a-4d50-b8ef-870bcfa72908

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/RollingRestartManifestStage.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/RollingRestartManifestStage.java
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.orca.clouddriver.tasks.manifest.RollingRestartManif
 import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder;
 import com.netflix.spinnaker.orca.pipeline.TaskNode;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import javax.annotation.Nonnull;
 import org.springframework.stereotype.Component;
 
 @Component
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/RollingRestartManifestStage.java
```diff
@@ -0,0 +1,38 @@
...
+import com.netflix.spinnaker.orca.pipeline.TaskNode;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import org.springframework.stereotype.Component;
+
+@Component
+public class RollingRestartManifestStage implements StageDefinitionBuilder {
+  public static final String PIPELINE_CONFIG_TYPE = "rollingRestartManifest";
+
+  @Override
+  public void taskGraph(Stage stage, TaskNode.Builder builder) {
```
ezimanyi: I think `stage` should be `@Nonnull` to match the method it's overriding.

maggieneterval: :+1: thank you, good catch!

PR #3234, opened 2019-10-17
feat(cfn): request to delete CFN changeset if empty

xavileon:
This patch requests the deletion of the CFN change set if it's empty.
The rationale behind it is that whenever a change set does not contain
changes to the CFN template, the AWS API marks the changeset as FAILED.
However, in a CD world, an empty change should not be considered a
failure, and instead continue the normal execution flow.

In this patch we detect that the change set is empty by the status
message and request it's deletion accordingly. This is implemented as
a separate task so it can be reused in different scenarios planned for
the future (e.g. configuration option to delete all changesets that
have been executed).

Depends on https://github.com/spinnaker/clouddriver/pull/4101


xavileon:
@spinnaker/reviewers PTAL

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/DeleteCloudFormationChangeSetTask.java
```diff
@@ -0,0 +1,79 @@
...
+  @Autowired KatoService katoService;
+
+  public static final String TASK_NAME = "deleteCloudFormationChangeSet";
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull Stage stage) {
+    String cloudProvider = getCloudProvider(stage);
+
+    Map<String, Object> task = new HashMap<>(stage.getContext());
```
marchello2000: `stage.getContext()` already returns a `Map`, why do you need to copy it?
If you do need to do the copy, consider renaming `task` to `stageData` - that's what we use pretty much everywhere else and I think it's more representative of what this is

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/DeleteCloudFormationChangeSetTask.java
```diff
@@ -0,0 +1,79 @@
...
+      Map<String, Map> operation =
+          new ImmutableMap.Builder<String, Map>().put(TASK_NAME, task).build();
+
+      katoService
+          .requestOperations(cloudProvider, Collections.singletonList(operation))
+          .toBlocking()
+          .first();
+      return TaskResult.builder(ExecutionStatus.SUCCEEDED).build();
+    } else {
+      log.debug(
```
marchello2000: `debug` or `info`? Not sure how frequent/expected this logline would be (don't know much about cloudformation).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/DeleteCloudFormationChangeSetTask.java
```diff
@@ -0,0 +1,79 @@
...
+          "Not deleting CloudFormation ChangeSet {} from stack {}.", changeSetName, stackName);
+      return TaskResult.builder(ExecutionStatus.SKIPPED).build();
```
marchello2000: Currently, `DeleteCloudFormationChangeSetTask` is the last task in the stage, so returning `SKIPPED` is fine, but if you ever add other tasks after it, if you return `SKIPPED` subsequent tasks won't run (see https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteTaskHandler.kt#L71).

So I recommend either returning `SUCCEEDED` or `FAILED_CONTINUE`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/WaitForCloudFormationCompletionTask.java
```diff
@@ -72,8 +72,15 @@ public TaskResult execute(@Nonnull Stage stage) {
           isChangeSet
               ? getChangeSetInfo(stack, stage.getContext(), "status")
               : getStackInfo(stack, "stackStatus");
-      if (isComplete(status) || isEmptyChangeSet(stage, stack)) {
+      if (isComplete(status)) {
         return TaskResult.builder(ExecutionStatus.SUCCEEDED).outputs(stack).build();
+      } else if (isEmptyChangeSet(stage, stack)) {
```
marchello2000: Again, apologies - I don't know much about CF... I assume, but want to confirm: Will `isComplete` return false in case of an empty changeset? Or should the `if isEmptyChangeSet...` be before `if isCompelte`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/WaitForCloudFormationCompletionTask.java
```diff
@@ -72,8 +72,15 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+        String changeSetName = (String) result.get("changeSetName");
+        log.debug("CloudFormation ChangeSet {} empty. Requesting to be deleted.", changeSetName);
```
marchello2000: `debug` or `info`? again, just calling it out to make sure it's appropriate level - I think most installation have `debug` logging turned off

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/DeployCloudFormationStage.java
```diff
@@ -36,5 +38,8 @@ public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
         .withTask("monitorCloudFormation", MonitorKatoTask.class)
         .withTask("forceRefreshCache", CloudFormationForceCacheRefreshTask.class)
         .withTask("waitForCloudFormationCompletion", WaitForCloudFormationCompletionTask.class);
+    if ((boolean) Optional.ofNullable(stage.getContext().get("isChangeSet")).orElse(false)) {
+      builder.withTask("deleteCloudFormationChangeSet", DeleteCloudFormationChangeSetTask.class);
```
marchello2000: do you need to have some sort of `waitForCompletion` / `monitorKatoTask` after the deletion of the changeset?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/DeleteCloudFormationChangeSetTask.java
xavileon: Good point. We made a copy of the context map because we were using the contents of the context to create the contents of the kato task operation. In that sense, both the context and task are coupled. I will make it more explicit by creating the task object from scratch and so we don't rely on assumptions.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/DeleteCloudFormationChangeSetTask.java
xavileon: I'd leave it at `debug` for now. Right now it would be printed whenever there's a change in the cloudformation stack that is being deployed. I don't think it has operational importance (so no need to fill in the logs with this message), but it could be handy when debugging if the removal doesn't happen for some reason.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/DeleteCloudFormationChangeSetTask.java
xavileon: ah well spotted. We plan to add more tasks, but I don't think we want to impose the restriction of this one being the last. So I'll move it to `SUCCEEDED` even though it won't actually do any side effect.

Follow-up maybe as another feature/PR: Would it make sense to have a task result that is `NOOP` or the likes? (so we can show the user that that specific task didn't do any action).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/WaitForCloudFormationCompletionTask.java
xavileon: yes it will return false as an empty changeset shows as `FAILED` in cloud formation terms. It definitely needs to go before the `isFailed` though. Maybe this part is calling for a refactor in a later PR.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/WaitForCloudFormationCompletionTask.java
xavileon: I'll change this one to info as it is taking an action 👍

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/DeployCloudFormationStage.java
xavileon: Added the monitor kato task. I'd like to keep it simple for now (no retries). I was looking at the saga backed operations, but this looks like a bigger change and understanding better how it works.

xavileon:
Addressed comments in https://github.com/spinnaker/orca/pull/3234/commits/49e608a21c1d70288f0dc0d62e9091b95a8675ae Also adding a monitor kato task to react with a failed operation in clouddriver.

PR #3236, opened 2019-10-17
fix(bake): Fix the fix for inconsistent names in parallel bakes

luispollo:
This complements #3219 by ensuring that we only look at the image names for parallel bake stages associated with the top-level bake stage that the `CompleteParallelBakeTest` cares about when trying to decide whether to fail a parallel bake on inconsistent image names. This avoids failing the stage if there are unrelated bake stages in the pipeline (which would obviously generate different image names).

While troubleshooting this, I found that this task includes context information from the parallel bake stages of unrelated top-level bake stages in the pipeline, which is clearly visible in the execution JSON of such pipelines. I don't know if that is itself a bug or is by design, so I just left a note in a comment for now, but would be happy to correct that if it's not expected behavior.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
@@ -164,9 +164,9 @@ class BakeStage implements StageDefinitionBuilder {
       if (failOnImageNameMismatchEnabled()) {
         // find distinct image names in bake stages that are actually related to the stage passed into the task
         def distinctImageNames = relatedBakeStages
-          .findAll { it.parentStageId == stage.id && (it.context.ami || it.context.imageId) }
+          .findAll { childStage -> childStage.parentStageId == stage.id && childStage.context.imageName }
           .stream()
-          .map { childStage -> childStage.context['imageName'] }
+          .map { childStage -> childStage.context.imageName }
           .distinct()
           .collect(Collectors.toList())
 
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
@@ -141,6 +141,9 @@ class BakeStage implements StageDefinitionBuilder {
         it.parentStageId == stage.parentStageId && it.status == ExecutionStatus.RUNNING
       }
 
+      // FIXME? (lpollo): I don't know why we do this, but we include in relatedStages all parallel bake stages in the
+      //  same pipeline, not just the ones related to the stage passed to this task, which means this list actually
+      //  contains *unrelated* bake stages...
       def relatedBakeStages = stage.execution.stages.findAll {
         it.type == PIPELINE_CONFIG_TYPE && bakeInitializationStages*.id.contains(it.parentStageId)
       }
@@ -159,8 +162,11 @@ class BakeStage implements StageDefinitionBuilder {
       ]
 
       if (failOnImageNameMismatchEnabled()) {
-         List<String> distinctImageNames = globalContext.deploymentDetails.stream()
-          .map({details -> details['imageName']})
+        // find distinct image names in bake stages that are actually related to the stage passed into the task
+        def distinctImageNames = relatedBakeStages
+          .findAll { it.parentStageId == stage.id && (it.context.ami || it.context.imageId) }
+          .stream()
+          .map { childStage -> childStage.context['imageName'] }
           .distinct()
           .collect(Collectors.toList())
 
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
@@ -163,7 +163,7 @@ class BakeStage implements StageDefinitionBuilder {
 
       if (failOnImageNameMismatchEnabled()) {
         // find distinct image names in bake stages that are actually related to the stage passed into the task
-        def distinctImageNames = relatedBakeStages
+        List<String> distinctImageNames = relatedBakeStages
           .findAll { childStage -> childStage.parentStageId == stage.id && childStage.context.imageName }
           .stream()
           .map { childStage -> childStage.context.imageName }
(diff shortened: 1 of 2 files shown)


luispollo:
@marchello2000 @aravindmd Please review.

luispollo:
@cfieber FYI if you want to take a look a well.

orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStageSpec.groovy
```diff
@@ -154,6 +154,51 @@ class BakeStageSpec extends Specification {
...
+        it.context.imageName = "image-from-bake-stage-${stageId}"
+      }
+      pipeline.stages.addAll(childBakeStages)
+    }
+
+    dynamicConfigService.isEnabled("stages.bake.failOnImageNameMismatch", false) >> { true }
+
+    when:
+    new BakeStage.CompleteParallelBakeTask(dynamicConfigService).execute(pipeline.stageById("1"))
+
```
srekapalli: Minor... You expect some other exception to be thrown here ? If not we can say noExceptionThrown() here.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -159,8 +162,11 @@ class BakeStage implements StageDefinitionBuilder {
       ]
 
       if (failOnImageNameMismatchEnabled()) {
-         List<String> distinctImageNames = globalContext.deploymentDetails.stream()
-          .map({details -> details['imageName']})
+        // find distinct image names in bake stages that are actually related to the stage passed into the task
+        def distinctImageNames = relatedBakeStages
+          .findAll { it.parentStageId == stage.id && (it.context.ami || it.context.imageId) }
```
cfieber: is the `(it.context.ami || it.context.imageId)` to determine whether this is a bake stage?

would checking it.type == 'bake' accomplish this instead?

alternatively would `it.context.imageName` be more appropriate to check since we are mapping to that on L169?

orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStageSpec.groovy
```diff
@@ -154,6 +154,51 @@ class BakeStageSpec extends Specification {
...
+        ]
+        status = ExecutionStatus.RUNNING
+      }
+      // this is a sibling bake stage whose child bake contexts should not be included in stage 1's outputs, but are
+      stage {
+        id = "2"
+        type = "bake"
+        context = [
+          "region": "us-east-1",
+          "regions": ["eu-east-1"]
```
cfieber:
```suggestion
          "regions": ["us-east-1"]
```

seems odd that this wouldn't be a superset of the single region

luispollo:
@gal-yardeni FYI as well!

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
luispollo: I was following what was done here:
https://github.com/spinnaker/orca/blob/608099016ea9436a032a88487ffd296b7bdbd350/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy#L149
...which I guessed was filtering for stages that actually had bake output information in them (without `ami` or `imageId` in the context, the bake probably failed?).

Does that make sense? Maybe I could encapsulate that in a method like `hasValidBakeContext(stage)` to make that clearer and reusable?

orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStageSpec.groovy
luispollo: I know, right? But based on what I saw, only the `regions` field is actually used? I don't know what the singular `region` is used for, if anything...

orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStageSpec.groovy
luispollo: I was advised to use a specific exception from kork for this case, so here I'm looking for that specific exception. If any other type of exception is thrown, it's unexpected in this test and it should fail.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
marchello2000: then it's probably more appropriate to use:
```suggestion
          .findAll { it.parentStageId == stage.id && (it.context.imageName) }
```

since that's what you actually use for comparison on L169

luispollo: @marchello2000 Am I guaranteed to have `imageName` in the context if the bake succeeded?

luispollo: Changed it to use `imageName`. Better?

luispollo:
> While troubleshooting this, I found that this task includes context information from the parallel bake stages of unrelated top-level bake stages in the pipeline, which is clearly visible in the execution JSON of such pipelines. I don't know if that is itself a bug or is by design, so I just left a note in a comment for now, but would be happy to correct that if it's not expected behavior.

@cfieber BTW, was curious if you knew the history behind this?

orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStageSpec.groovy
luispollo: @cfieber Looks like that singular `region` field is not set by the UI, even when baking in a single region:
```json
    {
      "amiSuffix": "${new java.text.SimpleDateFormat(\"yyyyMMddHHmmss\").format(new java.util.Date())}-${baseLabel}",
      "baseLabel": "candidate",
      "baseOs": "xenial",
      "cloudProviderType": "aws",
      "completeOtherBranchesThenFail": true,
      "continuePipeline": false,
      "extendedAttributes": {},
      "failPipeline": false,
      "name": "Bake Xenial Candidate",
      "package": "jnett-dummyapp-server",
      "refId": "1",
      "regions": [
        "us-east-1"
      ],
      "requisiteStageRefIds": [],
      "storeType": "ebs",
      "type": "bake",
      "user": "jnett@netflix.com",
      "vmType": "hvm"
    }
```
Wonder if we should remove it from the tests...

PR #3239, opened 2019-10-21
fix(rrb): Store RRB before-cleanup pipeline ID in the right context field

luispollo:
This is an addendum for #3224, which was causing Rolling Red/Black deployments with a "before cleanup" pipeline to fail as `StartPipelineTask` was looking for `stage.context.pipeline` but the pipeline ID was stored in `stage.context.pipelineId` here.

@marchello2000 and I search for other similar uses of `pipelineId` in the context and it looks like it's safe to make this change here.


luispollo:
@marchello2000 @jeyrschabu Please review.

marchello2000:
I am not a fan of the test for RRB, but still worthwhile adding this validation to:
https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/RollingRedBlackStrategySpec.groovy#L159

jeyrschabu:
LGTM

PR #3242, opened 2019-10-21
feat(plugins): Add kork-plugins, wire up initial spel function extension point

robzienert:
Depends on https://github.com/spinnaker/kork/pull/398

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/config/OrcaConfiguration.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/config/OrcaConfiguration.java
@@ -20,6 +20,7 @@ import static org.springframework.context.annotation.AnnotationConfigUtils.EVENT
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.config.PluginsAutoConfiguration;
 import com.netflix.spinnaker.kork.core.RetrySupport;
 import com.netflix.spinnaker.kork.expressions.ExpressionFunctionProvider;
 import com.netflix.spinnaker.orca.StageResolver;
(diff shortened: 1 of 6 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/HelloExpressionFunctionProvider.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.pipeline.expressions.functions;
+
+import com.netflix.spinnaker.kork.expressions.ExpressionFunctionProvider;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.pf4j.Extension;
+
+@Extension
```
jonsie: So this would essentially be a "system extension"?

robzienert: That's right. Although I'm gonna delete this from the PR. I didn't mean for it to come along for the ride.

orca-core/src/main/java/com/netflix/spinnaker/orca/config/OrcaConfiguration.java
```diff
@@ -132,8 +137,8 @@ UserConfiguredUrlRestrictions userConfiguredUrlRestrictions(
 
   @Bean
   public ContextParameterProcessor contextParameterProcessor(
-      List<ExpressionFunctionProvider> expressionFunctionProviders) {
-    return new ContextParameterProcessor(expressionFunctionProviders);
+      List<ExpressionFunctionProvider> expressionFunctionProviders, PluginManager pluginManager) {
```
marchello2000: where is the bean for `PluginManager`, i am confused...

marchello2000: oh, does it come from pf4j?

jonsie: I think it comes from `SpinnakerPluginManager` which is configured in Kork

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/HelloExpressionFunctionProvider.java
robzienert: Yeah, I have no intention of changing how spel function customization is done, only in providing new ways to provde said functions.

PR #3244, opened 2019-10-22
refactor(kubernetes): clean up bake manifest error messaging

maggieneterval:
Related to https://github.com/spinnaker/spinnaker/issues/4752

While investigating Retrofit errors in the Bake Manifest stage, found a few small code improvements:

- Move conversion of singular `inputArtifact` (Kustomize only) to list from `CreateBakeManifestTask.execute` into `BakeManifestContext` constructor. Also handle defaulting of `inputArtifacts` to empty list so that it doesn't need to be `@Nullable`. I think this lets the task be a little dumber about the Kubernetes vs. Helm inputs.
- In `CreateBakeManifestTask`, replace field injection with constructor injection. Because the `BakeryService` bean is not required, make it `Optional` per Stack Overflow [advice](https://stackoverflow.com/questions/42135102/how-to-set-autowired-constructor-params-as-required-false-individually) (curious if you guys agree with SO here or if there's another pattern we prefer?).
- Improve error messaging around 1) Attempting to run a Bake (Manifest) task without a configured `BakeryService`, 2) Attempting to produce more or less than one expected artifact, 3) Supplying unresolvable input artifacts.
- Remove a couple unused imports.

Additional changes in Rosco will improve error handling when the bake fails, either because the request times out or the input artifact cannot be downloaded. ([PR](https://github.com/spinnaker/rosco/pull/453))

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/BakeManifestContext.java
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/BakeManifestContext.java
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import javax.annotation.Nullable;
 import lombok.Getter;
 
@@ -48,12 +49,12 @@ public class BakeManifestContext {
       @JsonProperty("namespace") String namespace,
       @Nullable @JsonProperty("inputArtifact") CreateBakeManifestTask.InputArtifact inputArtifact,
       @JsonProperty("rawOverrides") Boolean rawOverrides) {
-    this.inputArtifacts = inputArtifacts == null ? new ArrayList<>() : inputArtifacts;
+    this.inputArtifacts = Optional.of(inputArtifacts).orElse(new ArrayList<>());
     // Kustomize stage configs provide a single input artifact
     if (this.inputArtifacts.isEmpty() && inputArtifact != null) {
       this.inputArtifacts.add(inputArtifact);
     }
-    this.expectedArtifacts = expectedArtifacts;
+    this.expectedArtifacts = Optional.of(expectedArtifacts).orElse(new ArrayList<>());
     this.overrides = overrides;
     this.evaluateOverrideExpressions = evaluateOverrideExpressions;
     this.templateRenderer = templateRenderer;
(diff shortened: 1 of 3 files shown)


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/BakeManifestContext.java
```diff
@@ -39,24 +39,26 @@
...
       @JsonProperty("evaluateOverrideExpressions") Boolean evaluateOverrideExpressions,
       @JsonProperty("templateRenderer") String templateRenderer,
       @JsonProperty("outputName") String outputName,
       @JsonProperty("namespace") String namespace,
-      @Nullable @JsonProperty("inputArtifact")
-          CreateBakeManifestTask.InputArtifactPair inputArtifact,
+      @Nullable @JsonProperty("inputArtifact") CreateBakeManifestTask.InputArtifact inputArtifact,
       @JsonProperty("rawOverrides") Boolean rawOverrides) {
-    this.inputArtifacts = inputArtifacts;
+    this.inputArtifacts = inputArtifacts == null ? new ArrayList<>() : inputArtifacts;
```
ezimanyi: Maybe slightly more idiomatic as `this.inputArtifacts = Optional.of(inputArtifacts).orElse(new Arraylist<>())`.  But this is also completely fine too.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
```diff
@@ -101,13 +105,9 @@ public TaskResult execute(@Nonnull Stage stage) {
 
     List<ExpectedArtifact> expectedArtifacts = context.getExpectedArtifacts();
 
-    if (expectedArtifacts == null || expectedArtifacts.isEmpty()) {
+    if (expectedArtifacts == null || expectedArtifacts.size() != 1) {
```
ezimanyi: This `null` check could be removed if you also defaulted `expectedArtifacts` in the stage context...but now I'm taking an awesome refactor and increasing the scope so feel free to ignore!

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/BakeManifestContext.java
maggieneterval: That's way better, updated!

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
maggieneterval: :+1: :+1: done!

maggieneterval: (not done ignoring, done implementing your suggestion)

PR #3246, opened 2019-10-22
fix(gcb): process spel in gcb build definition artifact

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/5051


maggieneterval:
Don't look at this yet, I broke your lovely tests!

PR #3249, opened 2019-10-23
fix(conditions): invalid stageEnabled expressions disable stage

asher:


erikmunson:
love it 🙏 

separately, do you think that this expression evaluation should respect the 'fail stage on failed expressions' setting?

asher:
@erikmunson Failing (instead of skipping) on an invalid conditional expression if the option to do so is enabled is the right thing to do, but going to leave it as a follow-up. I initially attempted to rethrow an `InvalidExpression` exception from `OptionalStageSupport` if there's an invalid expression only to discover that doing so results in a zombie pipeline left in a "running" state every time. Treading lightly for now until I have time to fully go down the rabbit hole.

erikmunson:
ack, this is _far far_ better than the existing behavior either way.

PR #3252, opened 2019-10-23
feat(orchestration): Allow sibling stages to continue on FAILED_CONTINUE

marchello2000:
Normally, if a synthetic child stage fails with `FAILED_CONTINUE` the `CompleteStageHandler`
will propagate the `FAILED_CONTINUE` status to the parent. This will prevent all subsequent
sibling stages from executing.
This change allows for an option (similar to Tasks) to continue execution even when a
child stage has a `status=FAILED_CONTINUE`

This option can only be set on a stage that is a synthetic child.
This is needed for Monitored Deploy where we want to allow the deploy to proceed even if the
monitoring fails

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
@@ -64,6 +64,16 @@ public class StageContext extends ForwardingMap<String, Object> {
     }
   }
 
+  /**
+   * Get a value from the current context ONLY - never looking at the ancestors' outputs
+   *
+   * @param key The key to look
+   * @param defaultValue default value to return if key is not present
+   * @return value or null if not present
+   */
+  Object getCurrentOnly(@Nullable Object key, Object defaultValue) {
+    return super.getOrDefault(key, defaultValue);
+  }
   /*
    * Gets all objects matching 'key', sorted by proximity to the current stage.
    * If the key exists in the current context, it will be the first element returned
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
@@ -142,7 +142,7 @@ class CompleteStageHandler(
 
         // When a synthetic stage ends with FAILED_CONTINUE, propagate that status up to the stage's
         // parent so that no more of the parent's synthetic children will run.
-        if (stage.status == FAILED_CONTINUE && stage.syntheticStageOwner != null) {
+        if (stage.status == FAILED_CONTINUE && stage.syntheticStageOwner != null && !stage.allowSiblingStagesToContinueOnFailure) {
           queue.push(message.copy(stageId = stage.parentStageId!!))
         } else if (stage.status in listOf(SUCCEEDED, FAILED_CONTINUE, SKIPPED)) {
           stage.startNext()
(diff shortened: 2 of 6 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -142,7 +142,7 @@ class CompleteStageHandler(
 
         // When a synthetic stage ends with FAILED_CONTINUE, propagate that status up to the stage's
         // parent so that no more of the parent's synthetic children will run.
-        if (stage.status == FAILED_CONTINUE && stage.syntheticStageOwner != null) {
+        if (stage.status == FAILED_CONTINUE && stage.syntheticStageOwner != null && stage.allowSiblingStagesToContinueOnFailure) {
```
ajordens: Shouldn't this be `!stage.allow...`? (based on the comment of not wanting to propagate up to the parent if siblings are allowed to run)

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
```diff
@@ -64,6 +64,16 @@ public Object get(@Nullable Object key) {
...
   }
 
+  /**
+   * Get a value from the current context ONLY - never looking at the ancestors' outputs
+   *
+   * @param key The key to look
+   * @param defaultValue default value to return if key is not present
+   * @return value or null if not present
+   */
+  Object getCurrentOnly(@Nullable Object key, Object defaultValue) {
```
ajordens: Not sure how I feel about `getCurrentOnly` vs overloading `get()` -- will let someone else weigh in.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
marchello2000: Yes, thank you! change the name of the var last minute that inverted the meaning and forgot to update this. i guess i should find and update tests for this

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
marchello2000: you mean something like 
```
public Object get(@Nullable Object key, Object defaultValue, boolean localOnly)
```
?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
marchello2000: oh, look, the existing tests caught it... fixed up and added a new test

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
jeyrschabu: +1 on overloading `get(...)`

marchello2000: Just curious, why?

I think
```
stage.context.getCurrentOnly("someValue", false);
```

is more descriptive than:
```
stage.context.get("someValue", false, true);
```

jeyrschabu: get() because the context is a map. Matter of preference.

marchello2000: I hear you. This area has caused some confusion in the past (where something comes from in the context) so I am more inclined to make this very explicit

marchello2000:
Going to merge this because there are teams blocked on this. If you have feedback please give it to me and I will incorporate

PR #3255, opened 2019-10-25
fix(monitored deploy): add deploymentId parameter

marchello2000:
Monitors need an ability to correlate calls to their API.
This adds a `deploymentId` parameter to all requests (which is really the `stageId` for the `createServerGroup` stage)
This allows the monitor to correlate e.g. `evaluateHealth @ 20%` and `evaluateHealth @ 50%` to the same cluster
Especially if multiple shards are deployed in the same pipeline

### DIFF
--- a/orca-deploymentmonitor/src/main/java/com/netflix/spinnaker/orca/deploymentmonitor/models/RequestBase.java
+++ b/orca-deploymentmonitor/src/main/java/com/netflix/spinnaker/orca/deploymentmonitor/models/RequestBase.java
@@ -16,20 +16,59 @@
 
 package com.netflix.spinnaker.orca.deploymentmonitor.models;
 
+import com.netflix.spinnaker.orca.deploymentmonitor.DeploymentMonitorService;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
 import java.util.Map;
 import lombok.Data;
 
 @Data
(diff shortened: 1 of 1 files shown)


orca-deploymentmonitor/src/main/java/com/netflix/spinnaker/orca/deploymentmonitor/models/RequestBase.java
```diff
@@ -42,6 +81,7 @@ protected void fromStage(Stage stage) {
     application = stage.getExecution().getApplication();
     executionId = stage.getExecution().getId();
     stageId = stage.getId();
+    deploymentId = stage.getParentStageId();
```
dreynaud: does stage id actually identify a deployment? Why not execution id?

marchello2000: The `createServerGroup` (parent) stage does - `executionId` is already there but it doesn't work when the same pipeline has multiple deploy stages. `createServerGroup` works because even if a deploy stage deploys N clusters there will be N `createServerGroup` stages and all steps will be their children

gal-yardeni:
Thanks for adding the comments @marchello2000 , super helpful!

PR #3258, opened 2019-10-25
fix(dependent pipelines): failure to trigger a dependend pipeline shouldn't affect others

marchello2000:
Failure in planning a, say, a single v2 templated pipeline will cause no dependent pipelines to start.
This is due to exception not being localized to each individual pipeline

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy
@@ -74,37 +74,52 @@ class DependentPipelineExecutionListener implements ExecutionListener {
       // Resolve templated pipelines if enabled.
       allPipelines = allPipelines.collect { pipeline ->
        if (V2Util.isV2Pipeline(pipeline)) {
-         return V2Util.planPipeline(contextParameterProcessor, executionPreprocessors, pipeline)
+         try {
+           return V2Util.planPipeline(contextParameterProcessor, executionPreprocessors, pipeline)
+         } catch (Exception e) {
+           log.error("Failed to plan V2 templated pipeline {}", pipeline.getOrDefault("id", "<UNKNOWN ID>"), e)
+           return null
+         }
        } else {
          return pipeline
        }
-      }
+      }.findAll({ pipeline ->
+        pipeline != null
+      })
     }
 
-    allPipelines.findAll { !it.disabled }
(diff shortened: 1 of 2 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy
```diff
@@ -74,37 +74,52 @@ class DependentPipelineExecutionListener implements ExecutionListener {
...
+           return V2Util.planPipeline(contextParameterProcessor, executionPreprocessors, pipeline)
+         } catch (Exception e) {
+           log.error("Failed to plan V2 templated pipeline {}", pipeline.getOrDefault("id", "<UNKNOWN ID>"), e)
+           return null
+         }
        } else {
          return pipeline
        }
-      }
+      }.findAll({ pipeline ->
```
marchello2000: oh, i don't need this anymore...

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/spring/DependentPipelineExecutionListener.groovy
```diff
@@ -74,37 +74,50 @@ class DependentPipelineExecutionListener implements ExecutionListener {
       // Resolve templated pipelines if enabled.
       allPipelines = allPipelines.collect { pipeline ->
        if (V2Util.isV2Pipeline(pipeline)) {
-         return V2Util.planPipeline(contextParameterProcessor, executionPreprocessors, pipeline)
+         try {
+           return V2Util.planPipeline(contextParameterProcessor, executionPreprocessors, pipeline)
+         } catch (Exception e) {
+           log.error("Failed to plan V2 templated pipeline {}", pipeline.getOrDefault("id", "<UNKNOWN ID>"), e)
```
marchello2000: should prob add a metric for this? what do you think @dreynaud ?

marchello2000: thought unless it's on the dashboard noone will look at it.. 🤷‍♂

dreynaud: I don't think there's necessarily a lot of value in throwing that on a dashboard. Is it a page-able offense though? If yes it deserves a metric, otherwise let's keep it in logs

PR #3269, opened 2019-10-30
feat(triggers): add action field to git triggers context

ivanphdz:
When dealing with webhook git triggers the event of the webhook does not reach the pipeline contexts. This PR makes this event available in the contexts of the pipeline execution.

This should no affect the current conditional for matching pipelines (source, project, slug, branch), it is just informative like the hash field.

For example: allow only github events where someone open/close a pull request using check preconditions, or if there is github push event follow a different branch in the pipeline, instead of one for pull request events.

For bitbucket server I used the header [ X-Event-Key ](https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html)

For bitbucket cloud I used the header [X-Event-Key](https://confluence.atlassian.com/bitbucket/event-payloads-740262817.html#EventPayloads-Push)

For gitlab I used the field in the payload named [object_kind](https://docs.gitlab.com/ee/user/project/integrations/webhooks.html)

For github at the moment only supports push and PR, so I used the field named [action](https://developer.github.com/webhooks/#example-delivery) of the payload.

related to https://github.com/spinnaker/echo/pull/693

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
@@ -138,7 +138,8 @@ internal class TriggerDeserializer :
           get("source").textValue(),
           get("project").textValue(),
           get("branch").textValue(),
-          get("slug").textValue()
+          get("slug").textValue(),
+          get("action").textValue()
         )
         looksLikeCustom() -> {
           // chooses the first custom deserializer to keep behavior consistent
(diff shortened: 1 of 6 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
```diff
@@ -138,7 +138,8 @@ internal class TriggerDeserializer :
           get("source").textValue(),
           get("project").textValue(),
           get("branch").textValue(),
-          get("slug").textValue()
+          get("slug").textValue(),
+          get("action").textValue()
```
cfieber: think we'll need to supply a default value here or make action on GitTrigger nullable

would also be good to ensure there is a test case that attempts deserialization with no supplied action

ivanphdz: thanks for feedback!

PR #3270, opened 2019-10-31
fix(cloudformation): Support Cloudformation templates as lists

jervi:
Something has changed between Spinnaker 1.15.6 and 1.16.0 with regard to how Cloudformation templates entered into the inline editor in Deck are parsed. Updating Orca to account for this change.


jervi:

jervi:

PR #3271, opened 2019-10-31
fix(plugins): Remove unnecessary code

robzienert:
* I bashed my head against the wall a little too much on the `SimpleStage` implementation: There's a lot of assumptions made that certain properties are non-null or off a certain type, but the generic type constraints do not enforce anything. I added `Optional` around the use of the problem properties, but didn't bother to harden the generic types or the handling of them - I'm not yet sure what I want to do with this.
* Removed unnecessary creation of `ObjectMapper` in `SimpleTask`.
* Removed `PluginManager` from the `StageResolver`.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
@@ -23,15 +23,13 @@ import com.netflix.spinnaker.orca.TaskResult;
 import com.netflix.spinnaker.orca.api.SimpleStage;
 import com.netflix.spinnaker.orca.api.SimpleStageInput;
 import com.netflix.spinnaker.orca.api.SimpleStageOutput;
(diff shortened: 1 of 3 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SimpleTask.java
```diff
@@ -23,15 +23,13 @@
...
 import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
srekapalli: Minor: wildcard imports above.

PR #3273, opened 2019-11-02
feat(expressions): add ability to query expression capabilities

marchello2000:
This will be used by `deck` to provide nice auto-complete for end-users typing out SpEL expressions

The endpoint (`/capabilities/expression`) will return
* a list of supported expression spel evaluators (currently v3 only)
* a list of functions and their descriptions&parameters so that `deck` can consume in autocomplete lists
  Note: all functions (including those from the plugins will be returned)
  Also note that implied `Execution` parameters will be stripped as they are not needed for the end-user to understand

Here is an example response:
```
{
  "functions": [
    {
      "name": "stage",
      "description": "Locates a stage by name",
      "parameters": [
        {
          "name": "idOrName",
          "description": "The name or id of the stage to find",
          "type": "String"
        }
      ],
      "documentation": {
        "documentation": "The most common use of this function is to check if a particular stage has succeeded or failed, but it can also be used to retrieve any information about a stage.",
        "examples": [
          {
            "usage": "#stage('bake').hasSucceeded",
            "description": "Returns true if the stage with the name `bake` has succeeded"
          },
          {
            "usage": "#stage('bake').hasFailed",
            "description": "Returns true if the stage with the name `bake` has failed"
          }
        ]
      }
    },
    ...
  ]
  "spelEvaluators": [
    {
      "versionKey": "v3",
      "description": "some description"
    }
  ]
}
```

Depends on https://github.com/spinnaker/kork/pull/410 and co-exists with `gate` PR which is TBD


marchello2000:
will need to wait for kork bump to merge

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/PipelineExpressionEvaluator.java
```diff
@@ -37,6 +37,31 @@
   public static final String SUMMARY = "expressionEvaluationSummary";
   public static final String ERROR = "Failed Expression Evaluation";
 
+  public enum SpelEvaluatorVersion {
+    V4("v4", false),
```
marchello2000: support for this is coming in a subsequent PR

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesController.java
```diff
@@ -46,4 +55,30 @@ public CapabilitiesController(
...
+
+    for (PipelineExpressionEvaluator.SpelEvaluatorVersion spelVersion :
+        PipelineExpressionEvaluator.SpelEvaluatorVersion.values()) {
+      if (spelVersion.getIsSupported()) {
+        result.getSpelEvaluators().add(spelVersion.getKey());
+      }
+    }
+
+    return result;
+  }
```
robzienert: Should pull this logic out into a service class and write some tests for it.

emjburns:
🎩 🐇 🎆 I'm so excited about this!

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/CapabilitiesController.java
marchello2000: donzo

orca-deploymentmonitor/src/main/java/com/netflix/spinnaker/orca/deploymentmonitor/models/DeploymentMonitorDefinition.java
```diff
@@ -23,6 +23,7 @@
   private String id;
   private String name;
   private String supportContact;
+  private boolean isStable;
```
robzienert: Do we feel like these data classes should have docs? What are the side effects of a deployment monitor being stable or not?

marchello2000: that's a good point, it's documented here https://github.com/spinnaker/orca/blob/master/orca-deploymentmonitor/src/main/java/com/netflix/spinnaker/config/DeploymentMonitorDefinition.java but i will replicate it, or at least link to the other file

PR #3274, opened 2019-11-04
feat(terminal-retry): Retry terminal monitor kato tasks with dynamic back off

jonsie:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -178,12 +180,36 @@ class MonitorKatoTask implements RetryableTask, CloudProviderAware {
...
 
     TaskResult.builder(status).context(outputs).build()
   }
 
+  private boolean terminalRetryCondition(Stage stage, Task katoTask, ExecutionStatus status) {
+    return (
+      status == ExecutionStatus.TERMINAL
+      && katoTask.status.retryable
+      && dynamicConfigService.isEnabled("tasks.monitor-kato-task.terminal-retries", true)
+      && (stage.context."kato.task.terminalRetryCount" as Integer) < dynamicConfigService.getConfig(
```
jonsie: Instead of `tasks.monitor-kato-task.saga-retries` I change this to `tasks.monitor-kato-task.terminal-retries` as I didn't want Orca referring to anything implementation specific on the CloudDriver side.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -115,15 +125,17 @@ class MonitorKatoTask implements RetryableTask, CloudProviderAware {
...
       'kato.task.firstNotFoundRetry': -1L,
-      'kato.task.notFoundRetryCount': 0
+      'kato.task.notFoundRetryCount': 0,
+      'kato.task.lastStatus': status
     ] as Map<String, ?>
 
+    if (status != ExecutionStatus.TERMINAL && status != ExecutionStatus.SUCCEEDED) {
+      status = ExecutionStatus.RUNNING
+    }
+
```
marchello2000: looking at `katoStatusToTaskStatus` i don't even understand why we need this `if`... anyway, either way, you can move it back where it came from

jonsie: Yah, good point. 🤷‍♂

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -136,6 +148,9 @@ class MonitorKatoTask implements RetryableTask, CloudProviderAware {
       }
 
       if (stage.context."kato.task.retriedOperation" == true) {
+        Integer totalRetries = stage.context."kato.task.terminalRetryCount" as Integer
+        log.info("Completed kato task ${katoTask.id} after ${totalRetries} retries.")
```
marchello2000: i think it would be useful to spit the exception that necessitated the retries (if there was one...) here - will make our log parsing easier

marchello2000:
Can you add some 'words' to the commit description for you and me 6 months from now to remember what this was about?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -71,6 +73,14 @@ class MonitorKatoTask implements RetryableTask, CloudProviderAware {
 
   long getTimeout() { 3600000L }
 
+  @Override
+  long getDynamicBackoffPeriod(Stage stage, Duration taskDuration) {
+    if ((stage.context."kato.task.lastStatus" as ExecutionStatus) == ExecutionStatus.TERMINAL) {
+      return Math.max(backoffPeriod, TimeUnit.MINUTES.toMillis(2))
+    }
+    return backoffPeriod
```
marchello2000: this is extra defensive programming Mark talking: do you think worth wrapping in `try/catch` so we never fail during task planning?
e.g.
```java
try {
    if ((stage.context."kato.task.lastStatus" as ExecutionStatus) == ExecutionStatus.TERMINAL) {
      return Math.max(backoffPeriod, TimeUnit.MINUTES.toMillis(2))
    }
} catch(e) {
  log(e)
}

return backoffPeriod
```

jonsie: I don't think that'll be necessary (famous last words).  The `if` statement should be safe.  If `stage.context."kato.task.lastStatus"` is null that'll just return false.

PR #3283, opened 2019-11-06
fix(kubernetes): fix patchBody typing depending on strategy type

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/5093

Corresponding Deck PR: https://github.com/spinnaker/deck/pull/7600

Clouddriver's patch operation expects Orca to send it a `patchBody` that is a list when the merge strategy is of type `json`, and a map otherwise. Prior to Spinnaker 1.15, Deck also saved the `patchBody` as either a list or map depending on merge strategy. In Spinnaker 1.15, the artifacts rewrite introduced a bug that caused the `patchBody` to be saved as a `Map` regardless of merge strategy, which broke the JSON merge strategy and Deck's YAML editor. This change:

1. Deserializes `patchBody` as an `Object`, and coerces it to a `List` to conform to the shared `ManifestContext` interface.
2. Validates that the list of patches is non-empty and that only JSON-strategy patches have more than one item in the list.
3. Sends Clouddriver the `patchBody` as a `List` when the merge strategy is of type `json`, and a `Map` otherwise.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/PatchManifestContext.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/PatchManifestContext.java
@@ -16,10 +16,11 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.manifest;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
-import java.util.Collections;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/PatchManifestContext.java
```diff
@@ -43,10 +44,51 @@
...
+     * regardless of merge strategy, but we should handle receiving a Map from pipelines configured
+     * in 1.15 and 1.16.
+     */
+    if (patchBody == null) {
+      return null;
+    }
+    if (patchBody instanceof List) {
+      return (List) patchBody;
+    }
+    List<Map<Object, Object>> manifests = new ArrayList<>();
```
ezimanyi: This could be simplified to `ImmutableList.of((Map<Object, Object>) patchBody)`; this would of course also make the return value immutable, which might be a good side-effect.

(I tend to avoid having a function sometimes return a mutable list and sometimes an immutable one, so if you take this suggestion it might also be good to change the above return to `ImmutableList.copyOf((List) patchBody)`).

And similarly in the null case, we might want to return `ImmutableList.of()` and annotate the function `@Nonnull` to make it easier on callers, as I don't think there would ever be a real difference in meaning when this returns `null` vs. an empty list.

maggieneterval: Loooove it, thank you, updated on all counts!

maggieneterval:
@spinnakerbot :cherries: :pick: 1.17

PR #3285, opened 2019-11-07
fix(core): Replace Unchecked Cast in TriggerDeserializer

evansb:
This PR replaces two unchecked casts in `listValue` and `mapValue` helper function in `TriggerDeserializer` with strongly typed deserializer.

Due to the unchecked cast and type erasure in the collection item type, some fields in trigger can have unexpected type in runtime.

For instance, in this bug: https://github.com/spinnaker/spinnaker/issues/5105, the type of `resolvedExpectedArtifacts` is a `List<LinkedHashMap<String, Any>>` instead of `List<ExpectedArtifact>`.

This is okay for use in SpEL because things like field access still work but the moment the object is used in code it will throw `ClassCastException`

e.g https://github.com/spinnaker/orca/blob/3d040623dda7d3b7c2c6bf0fa8dd4f7691a48945/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ArtifactExpressionFunctionProvider.java#L73

**Expected Behaviour**
Fields inside `Trigger` deserialize to the correct type in runtime.

**Actual Behaviour**
Inside Trigger, the actual run-time type for collection fields like `notifications` and `artifacts` is incorrect due to unsafe casts. For instance, the field `resolvedExpectedArtifacts`, the runtime type is `List<LinkedHashMap<String, Any>>` instead of `List<ExpectedArtifact>`.

**Test Case To Demonstrate**
```
 def "correctly map collection item to its type"() {
    given:
    def trigger = mapper.readValue(triggerJson, Trigger)

    expect:
    with(trigger) {
      artifacts[0] instanceof Artifact // Before the fix this is false
    }

    where:
    triggerJson = '''
    {
      "type": "manual",
      "artifacts": [
        {
          "type": "test/artifact",
          "reference": "testreference"
        }
      ]
    }'''
  }
```


evansb:
The previous CI that failed seems to be a flaky test:

https://travis-ci.org/spinnaker/orca/builds/612571445?utm_medium=notification

PR #3286, opened 2019-11-07
feat(expressions): Enhance expression evaluation order

marchello2000:
Today, expressions are evaluated indiscriminately.
This change allows for a few useful changes:
1. Only evaluate SpEL expressions in a stage when that stage runs.
  This allows for the SpEL expressions to such as `Instant.now()` to be used and produce correct value during stage execution
2. Allow evaluating SpEL expressions in the `EvaluateVariables` stage in order.
  This allows for defining variables that depend on ones defined earlier.
3. We can now capture expressions before they are evaluated during stage execution and perform useful actions.
  For example, prior to this change, it was impossible to know what was the "source" expression in the `CheckPrecondition` stage (this is especially annoying when the expression evaluates to `false`)
  Now, it is possible to capture the expression and make it clear to the end-user what went wrong.

All of this is, for now, an opt-in behavior by setting `spelEvaluator: "v4"` on the pipeline

Things still to do:

- [x] Add EP to evaluate expressions like EvaluateVariables stage
- [ ] Fixup CheckPreconditions stage to capture expressions
- [x] deal with v2 spelevaluator

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/EvaluateVariablesStage.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/EvaluateVariablesStage.java
@@ -18,21 +18,82 @@ package com.netflix.spinnaker.orca.pipeline;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.kork.expressions.ExpressionEvaluationSummary;
+import com.netflix.spinnaker.orca.pipeline.expressions.PipelineExpressionEvaluator;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import com.netflix.spinnaker.orca.pipeline.model.StageContext;
 import com.netflix.spinnaker.orca.pipeline.tasks.EvaluateVariablesTask;
-import java.util.List;
+import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor;
+import java.util.*;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 @Component
 public class EvaluateVariablesStage implements StageDefinitionBuilder {
-
   public static String STAGE_TYPE = "evaluateVariables";
+
+  private ObjectMapper mapper;
+
+  @Autowired
+  public EvaluateVariablesStage(ObjectMapper objectMapper) {
+    mapper = objectMapper;
+  }
 
   @Override
   public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
     builder.withTask("evaluateVariables", EvaluateVariablesTask.class);
+  }
+
+  @Override
+  public boolean processExpressions(
+      @Nonnull Stage stage,
+      @Nonnull ContextParameterProcessor contextParameterProcessor,
+      @Nonnull ExpressionEvaluationSummary summary) {
+
+    EvaluateVariablesStageContext context = stage.mapTo(EvaluateVariablesStageContext.class);
+    StageContext augmentedContext = contextParameterProcessor.buildExecutionContext(stage);
+    Map<String, Object> varSourceToEval = new HashMap<>();
+    int lastFailedCount = 0;
+
+    List<Variable> variables =
+        Optional.ofNullable(context.getVariables()).orElse(Collections.emptyList());
+
+    for (Variable var : variables) {
+      if (var.getValue() instanceof String) {
+        var.saveSourceExpression();
+        varSourceToEval.put("var", var.getValue());
+
+        Map<String, Object> evaluatedVar =
+            contextParameterProcessor.process(varSourceToEval, augmentedContext, true, summary);
+
+        // Since we process one variable at a time, the way we know if the current variable was
+        // evaluated properly is by
+        // checking if the total number of failures has changed since last evaluation. We can make
+        // this nicer, but that
+        // will involve a decent refactor of ExpressionEvaluationSummary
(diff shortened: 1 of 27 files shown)


marchello2000:
There are still a few things to wrap up, but wanted to get early feedback

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/EvaluateVariablesStage.java
```diff
@@ -67,12 +103,26 @@ public void setValue(Object value) {
...
     public Object getValue() {
       return value;
     }
+
+    public Object getSourceValue() {
+      return sourceValue;
+    }
+
+    public void saveSourceValue() {
+      if (sourceValue == null && value instanceof String) {
```
srekapalli: toString is not necessary i think, since this only execute if it is a string.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/EvaluateVariablesStage.java
```diff
@@ -35,6 +39,36 @@ public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
...
+        varSourceToEval.put("var", var.getValue());
+
+        Map<String, Object> evaluatedVar =
+            contextParameterProcessor.process(varSourceToEval, augmentedContext, false, summary);
+
+        var.setValue(evaluatedVar.get("var"));
+        augmentedContext.put(var.key, var.value);
+      }
+    }
+
```
srekapalli: Question ? Don't we have a ObjectMapper wired up to use as a Spring bean ?  May be we can inject that and use that instead of newing up ObjectMapper here.

robzienert: +1

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
```diff
@@ -73,6 +73,31 @@ public ContextParameterProcessor(
...
+  /**
+   * Process pipeline to evaluate spel expressions. Note that 'stages' key is not processed if we
+   * are using spel v4
+   */
+  public Map<String, Object> processPipeline(
+      Map<String, Object> pipeline, Map<String, Object> context, boolean allowUnknownKeys) {
+
+    final String spelEvaluatorKey = "spelEvaluator";
+    ExpressionEvaluationSummary summary = new ExpressionEvaluationSummary();
+    String spelEvaluatorVersion = (String) pipeline.getOrDefault(spelEvaluatorKey, "v3");
```
robzienert: I'd kinda like to see this `"v3"` string pulled into a constant, like `SpelEvaluatorVersion.DEFAULT_VERSION` or something?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt
```diff
@@ -126,14 +128,27 @@ interface ExpressionAware {
...
+    }
+
+    if (shouldContinueProcessing) {
+      return StageContext(stage, contextParameterProcessor.process(
+        stage.context,
+        contextParameterProcessor.buildExecutionContext(stage),
+        true,
+        summary
+      )
+      )
```
robzienert: nit: Wrap this to the previous line.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/util/ExpressionUtils.java
```diff
@@ -0,0 +1,128 @@
...
+      StageDefinitionBuilderFactory stageDefinitionBuilderFactory) {
+    this.contextParameterProcessor = contextParameterProcessor;
+    this.stageDefinitionBuilderFactory = stageDefinitionBuilderFactory;
+  }
+
+  public Map<String, Object> evaluateVariables(
+      Execution execution,
+      List<String> requisiteStageRefIds,
+      String spelVersionOverride,
+      List<Map<String, String>> expressions) {
```
robzienert: nit: How do you feel about `@Nonnull` annotations on public methods args & return types?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/util/ExpressionUtils.java
```diff
@@ -0,0 +1,128 @@
...
+    // form the specified stage are included in the evaluation
+    Map<String, Object> stageContext = new HashMap<>();
+    stageContext.put("refId", "_AD_HOC_EVALUATE_VARIABLES_STAGE_");
+    stageContext.put("requisiteStageRefIds", requisiteStageRefIds);
+    stageContext.put("variables", expressions);
+
+    if (!Strings.isNullOrEmpty(spelVersionOverride)) {
+      if (!PipelineExpressionEvaluator.SpelEvaluatorVersion.isSupported(spelVersionOverride)) {
+        throw new IllegalArgumentException(
+            "SpEL evaluator version " + spelVersionOverride + " is not supported");
```
robzienert: Maybe a `IllegalArgumentException : UserException`?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/EvaluateVariablesStage.java
marchello2000: yup, good point, thanks!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
marchello2000: yup, totally forgot to do that, thanks!

orca-web/src/main/groovy/com/netflix/spinnaker/orca/util/ExpressionUtils.java
marchello2000: yes, thanks!

PR #3288, opened 2019-11-08
feat(gcb): Ability to invoke existing GCB triggers

afcastano:
Added ability to call existing GCB triggers.
This is part of: spinnaker/spinnaker#5076

1. Enhanced `GoogleCloudBuildStageDefinition.java` to include the necessary fields for `igor` cloud build trigger endpoint:
   - triggerId
   - repoSource

2. Modified `StartGoogleCloudBuildTask.java` to check for the new build definition source and call the corresponding methods in `igorService`

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/GoogleCloudBuildStageDefinition.java
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/GoogleCloudBuildStageDefinition.java
@@ -29,6 +29,8 @@ public class GoogleCloudBuildStageDefinition implements RetryableStageDefinition
   private final Map<String, Object> buildDefinition;
   private final String buildDefinitionSource;
   private final GoogleCloudBuildDefinitionArtifact buildDefinitionArtifact;
+  private final String triggerId;
+  private final GoogleCloudBuildRepoSource repoSource;
   private final int consecutiveErrors;
 
   // There does not seem to be a way to auto-generate a constructor using our current version of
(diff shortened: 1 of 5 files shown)


afcastano:
@ezimanyi These are the changes in orca. I tested locally with previous stage definition and they seem to be working ok.
Please let me know what do you think.
Thanks.
=)

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartGoogleCloudBuildTask.java
```diff
@@ -55,15 +56,28 @@ public TaskResult execute(@Nonnull Stage stage) {
...
 
-    Map<String, Object> buildDefinition;
-    if (stageDefinition.getBuildDefinitionSource() != null
-        && stageDefinition.getBuildDefinitionSource().equals("artifact")) {
-      buildDefinition = getBuildDefinitionFromArtifact(stage, stageDefinition);
-    } else {
-      buildDefinition = stageDefinition.getBuildDefinition();
+    GoogleCloudBuild result;
+    String source = Optional.ofNullable(stageDefinition.getBuildDefinitionSource()).orElse("");
+    switch (source) {
```
ezimanyi: One thought---now that we're adding to the values that can be in `source`, I wonder if it would make sense to change it to an `enum` in `GoogleCloudBuildStageDefinition` (taking `STAGE`, `ARTIFACT`, or `TRIGGER`).  Then the deserialization logic can handle setting it to the enum value (defaulting to `STAGE` if `null` or if we get a `IllegalArgumentException`).  Then consumers of the class can deal only with the enum value and we avoid switches on Strings.

afcastano: Make sense... will implement.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/GoogleCloudBuildStageDefinition.java
```diff
@@ -67,4 +74,18 @@ public GoogleCloudBuildDefinitionArtifact(
...
   }
+
+  public enum BuildDefinitionSource {
+    STAGE,
+    ARTIFACT,
+    TRIGGER;
+
+    public static BuildDefinitionSource fromString(String buildDefinitionSource) {
+      try {
+        return valueOf(buildDefinitionSource.toUpperCase());
```
afcastano: Transformed to uppercase for backwards compatibility

afcastano:
@ezimanyi Changes ready. Added a new commit to make it easier to review.
=)

PR #3291, opened 2019-11-11

sanopsmx:
…stage #4792

1) Get the application roles, stage roles and the user roles.
2) Check each of the user roles whether they are contained in the stage and application roles.
3) If yes/no, set a isAuthorized flag to true/false in each of the stage. By default, all the stages except Manual Judgment are true.
4) Check for the flag in ManualJudgmentStage.groovy whether to execute to the next stage or not.
5) If yes/no, continue with the next stages/continues running the same stage.


sanopsmx:
I am closing this Pull Request. I am raising a new one.

PR #3294, opened 2019-11-13
fix(monitoreddeploy): unset pinMinimumCapacity when scaling to 100%

asher:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
```diff
@@ -194,6 +194,7 @@ class MonitoredDeployStrategy implements Strategy {
         scalePct                     : p,
         pinCapacity                  : p < 100,  // if p < 100, capacity should be pinned (min == max == desired)
         unpinMinimumCapacity         : p == 100, // if p == 100, min capacity should be restored to the original unpinned value from source
+        pinMinimumCapacity           : p < 100,  // pinMinimumCapacity should be false when unpinMinimumCapacity is true
```
marchello2000: oh, god, thank you, @asher .

but i am confused why this gets needs an explicit `false` when that's the default.
i think the real culprit is https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy#L155-156

where we(i) inadvertently modify the `basecontext` instead of `resizeContext`... ugh, this is embarrassing.

anyway, if you don't get to it, i can clean this up next week when I am back from vaca/spinnaker summit

asher: I (mis)interpreted 155-156 as deliberately modifying `baseContext` :) Agree that some further cleanup is warranted since that also unintentionally pollutes `disableContext`. That doesn't appear to have negative implications but uncertain if that's always the case. Not sure if I'll give this another pass before the summit but at least this PR fixes the immediate production impact.

PR #3295, opened 2019-11-13

sanopsmx:
…#4792

Modified the testcases.


sanopsmx:
Closing this PR and raising a new one.

PR #3298, opened 2019-11-13

sanopsmx:
…#4792

Modified the testcases.

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.orca.controllers
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.fiat.model.UserPermission
 import com.netflix.spinnaker.fiat.model.resources.Role
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator
 import com.netflix.spinnaker.fiat.shared.FiatService
 import com.netflix.spinnaker.fiat.shared.FiatStatus
 import com.netflix.spinnaker.kork.exceptions.SpinnakerException
@@ -29,6 +30,7 @@ import com.netflix.spinnaker.orca.exceptions.OperationFailedException
 import com.netflix.spinnaker.orca.extensionpoint.pipeline.ExecutionPreprocessor
 import com.netflix.spinnaker.orca.front50.Front50Service
 import com.netflix.spinnaker.orca.front50.PipelineModelMutator
+import com.netflix.spinnaker.orca.front50.model.Application
 import com.netflix.spinnaker.orca.igor.BuildService
 import com.netflix.spinnaker.orca.pipeline.ExecutionLauncher
 import com.netflix.spinnaker.orca.pipeline.model.Execution
@@ -40,6 +42,7 @@ import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor
 import com.netflix.spinnaker.orca.pipelinetemplate.PipelineTemplateService
 import com.netflix.spinnaker.orca.webhook.service.WebhookService
 import com.netflix.spinnaker.security.AuthenticatedRequest
+import groovy.json.JsonSlurper
 import groovy.util.logging.Slf4j
 import javassist.NotFoundException
 import org.springframework.beans.factory.annotation.Autowired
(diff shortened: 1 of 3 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -102,6 +108,8 @@ class OperationsController {
   @Autowired(required = false)
   Front50Service front50Service
 
+  def jsonSlurper = new JsonSlurper()
```
robzienert: While this is a groovy class, I'd prefer to not increase reliance on Groovy-only classes, as it'll make things just that much more difficult to refactor this out of Groovy in the future. Use Jackson instead!

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -470,4 +478,58 @@ class OperationsController {
       pipeline.origin = AuthenticatedRequest.spinnakerUserOrigin.orElse('unknown')
     }
   }
+
+  private void addStageAuthorizedRoles(def request, Map pipeline) {
+
+    def jsonObject = jsonSlurper.parseText(request)
```
robzienert: I'm kinda confused why you need to do any JSON slurping here to begin with. On L177, we do `def request = objectMapper.writeValueAsString(pipeline)`, and then on this line we're deserializing it again. It's wasteful, as you're passing the exact same original payload as `Map pipeline` into this method.

We can do without the json hoops entirely.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -470,4 +478,58 @@ class OperationsController {
...
+    if (jsonObject.application) {
+      Application application = front50Service.get(jsonObject.application)
+      if (application) {
+        def username = AuthenticatedRequest.getSpinnakerUser().orElse("")
+        def permissions = objectMapper.writeValueAsString(application.getPermission().permissions)
+        def permJsonObject = jsonSlurper.parseText(permissions)
```
robzienert: More serde hoops that I don't think are necessary. If you want `permissions` as a simple map, you could do `objectMapper.convertValue(permissions, new TypeReference<Map<String, Object>>() {})` or similar and skip the serde hoops entirely.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -470,4 +478,58 @@ class OperationsController {
...
+        roleView.each { it -> userRoles.add(it.getName().trim()) }
+        def stageList = pipeline.stages
+        def stageRoles = []
+        stageList.each { item ->
+          stageRoles = item.selectedStageRoles
+          item.isAuthorized = checkAuthorizedGroups(userRoles, stageRoles, applicationRoles)
+        }
+      }
+    }
+  }
```
robzienert: This entire function feels kind of brittle to me. The fact that we're converting strongly-typed objects into a map and iterating over the keys is quite concerning.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
sanopsmx: Incorporated the change.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
sanopsmx: Incorporated the change.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
sanopsmx: Incorporated the change.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
sanopsmx: I need only the roles, which are of List<String> and it needs to check with stageRoles and applicationRoles. This is the best, we can do.

sanopsmx:
Raising a new PR.

PR #3299, opened 2019-11-13
feat(cf): add SpEL support for cf manifests

Jammy-Louie:
add SpEL support for CF Manifests


Jammy-Louie:
relies on https://github.com/spinnaker/clouddriver/pull/4167

PR #3301, opened 2019-11-14
Kork sqlize

dreynaud:
Some of the early SQL implementation in orca has been migrated and modernized independently in `kork-sql`. This PR makes orca's sql configuration a lot more like clouddriver's -- it should not change its behavior directly.


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
```diff
@@ -61,15 +64,18 @@ void resume(
   void updateStatus(ExecutionType type, @Nonnull String id, @Nonnull ExecutionStatus status);
 
   @Nonnull
+  @Instrumented(metricName = "retrieveById")
```
dreynaud: small change here, this used to be named something not-very-descriptive like `retrieve2`. I favored more descriptive names

asher: :thankyou:

orca-queue-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlPendingExecutionConfiguration.kt
```diff
@@ -13,7 +14,7 @@ import org.springframework.context.annotation.Configuration
 import java.time.Clock
 
 @Configuration
-@EnableConfigurationProperties(SqlProperties::class)
+@EnableConfigurationProperties(com.netflix.spinnaker.kork.sql.config.SqlProperties::class)
```
asher: It's imported so shouldn't need the package name here?

dreynaud: indeed, will remove

PR #3302, opened 2019-11-15
feat(sql): add support for a secondary sql execution repo

dreynaud:
This makes it possible to conditionally create a secondary SqlExecutionRepository,
pointing at a specific named connection pool.

This also makes it possible to look up beans in DualExecutionRepository by name,
not just by class, which comes in handy when we want 2 SqlExecutionRepository beans.


dreynaud:
Example config that makes this work:

```sql:
  connection-pools:
    executions:
      jdbcUrl: beep
      default: true
    foreign-executions:
      jdbcUrl: boop

executionRepository:
  dual:
    enabled: true
    primaryName: sqlExecutionRepository
    previousName: secondarySqlExecutionRepository

  sql:
    enabled: true
    secondary:
      enabled: true
      pool-name: foreign-executions```

PR #3306, opened 2019-11-19
fix(perf): Make `getAncestors` call faster

marchello2000:
This change unrolls the recursion of the `getAncestorImpl` call into an imperative call.
getAncestors is actually called A TON! (especially since all `context.get` evaluations depend on it).
If the execution depth is large (e.g. canary with many stages) this call can take a while and, in some cases, throw a StackOverflow excetion.
Additionally, I added some logic to cache the `getAncesors` calls in the expression evaluation as it can't change

For context: `getAncestors` is executed 64 times for a simple wait stage execution

Results:
Time for `getAncestors` was reduced by a factor of 7x (1000 executions on 100 stage pipeline went from 7.2s to 0.9s)
Furthermore, due to caching the number of calls to `getAncestors` is reduces by ~3 (from 64 to 19)

So a total improvement time about 21x

As another example, planning a kayenta stage with 300 intervals went from 23.6s to 1.2s

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
@@ -38,7 +38,6 @@ import com.fasterxml.jackson.databind.node.TreeTraversingParser;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Sets;
 import com.netflix.spinnaker.kork.exceptions.SpinnakerException;
 import com.netflix.spinnaker.orca.ExecutionStatus;
 import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper;
@@ -46,17 +45,7 @@ import com.netflix.spinnaker.orca.pipeline.model.support.RequisiteStageRefIdDese
 import de.huxhorn.sulky.ulid.ULID;
 import java.io.IOException;
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Base64;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 import javax.annotation.Nonnull;
@@ -366,14 +355,10 @@ public class Stage implements Serializable {
    */
   public List<Stage> ancestors() {
     if (execution != null) {
-      Set<String> visited = Sets.newHashSetWithExpectedSize(execution.getStages().size());
-      return ImmutableList.<Stage>builder()
-          .add(this)
-          .addAll(getAncestorsImpl(visited, false))
-          .build();
-    } else {
-      return emptyList();
+      return getAncestorsImpl(false);
     }
+
+    return emptyList();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


marchello2000:
also see semi-related change: https://github.com/spinnaker/orca/pull/3307

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -384,81 +369,76 @@ public Task taskById(String taskId) {
...
+                              && it.getSyntheticStageOwner() == SyntheticStageOwner.STAGE_BEFORE)
+                  .collect(toList()));
+        }
+
+        // 2.2 Add our direct parent to the queue
+        try {
+          ancestors.add(curStage.getParent());
+        } catch (IllegalArgumentException e) {
+          // It's not really possible for the parent not to exist.. But unittests beg to differ
+          // Not logging anything here since not having a parent would have failed and loffed in
```
dreynaud: `loffed`... logged?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -384,81 +369,76 @@ public Task taskById(String taskId) {
...
    * of the current stage.
    *
-   * @param visited list of visited nodes
    * @param directParentOnly true to only include direct parents of the stage, false to also include
    *     stages this stage depends on (via requisiteRefIds)
    * @return list of ancestor stages
    */
-  private List<Stage> getAncestorsImpl(Set<String> visited, boolean directParentOnly) {
-    visited.add(this.refId);
+  public ImmutableList<Stage> getAncestorsImpl(boolean directParentOnly) {
```
ezimanyi: I believe this can be `private`

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -384,81 +369,76 @@ public Task taskById(String taskId) {
...
-                          .addAll(parent.getAncestorsImpl(visited, directParentOnly))
-                          .build())
-              .orElse(emptyList()));
-
-      return ancestors;
-    } else {
-      return emptyList();
+      // 2. If we have a parent stage
+      if (curStage.parentStageId != null && !visited.contains(curStage.parentStageId)) {
+        List<Stage> ancestors = new ArrayList<>();
```
ezimanyi: Any reason to have this temporary variable vs. just directly doing `toVisit.addAll()` and `toVisit.add()` below?  (I'm asking primarily for clarity/readability, though I suppose it would probably also be a bit faster to remove it as well.)

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -384,81 +369,76 @@ public Task taskById(String taskId) {
...
+        } catch (IllegalArgumentException e) {
+          // It's not really possible for the parent not to exist.. But unittests beg to differ
+          // Not logging anything here since not having a parent would have failed and loffed in
+          // other places
+        }
+        toVisit.addAll(ancestors);
+      }
     }
+
+    return ImmutableList.<Stage>builder().addAll(allAncestorStages).build();
```
ezimanyi: This would be clearer as `ImmutableList.copyOf(allAncestorStages)`.  Or, alternately (and perhaps better) `allAncestorStages` could be an `ImmutableList.Builder` that we add to during the function, then the return statement would just be `return allAncestorStages.build()`.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -384,81 +369,76 @@ public Task taskById(String taskId) {
...
+      if (visited.contains(curStage.id)) {
+        continue;
+      }
+
+      visited.add(curStage.id);
+      allAncestorStages.add(curStage);
+
+      // 1. Find all stages we depend on through requisiteStageRefIds and push them on the queue
+      if (!curStage.requisiteStageRefIds.isEmpty() && !directParentOnly) {
+        List<Stage> refIdPriorStages =
```
ezimanyi: I suspect the performance can be further improved by removing this (and the following) iteration over all stages from the loop...this is still an `O(N^2)` algorithm because we iterate (potentially twice) over all stages for every visited stage.  (I presume the issues are mostly for executions with a lot of stages, where changing this to an `O(N)` algorithm would have a lot of benefit.)

Instead you could build :
```
ImmutableMap<String, Stage> stageByRefId = execution.getStages().stream()
  .collect(toImmutableMap(...))`
ImmutableSetMultiMap<String, Stage> stageByParentId = execution.getStages().stream()
  .filter(it -> it.getSyntheticStageOwner() == SyntheticStageOwner.STAGE_BEFORE)
  .collect(toImmutableSetMultimap(...))`
```
And then look up the required values in these pre-computed structures from within the `while` loop.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: ah, yes, changed it for testing and forgot to undo, thank you!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: thanks!

marchello2000:
there are decent tests in `StageSpec` and `StageNavigatorSpec`. They seemed reasonably adequate to me, but I will look over them some more

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: i was trying to be clearer, but you're right that it would probably be faster and generate less GC cruft. let me try it with the immutablelist builder

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
```diff
@@ -56,7 +57,10 @@ public Object get(@Nullable Object key) {
     if (delegate().containsKey(key)) {
       return super.get(key);
     } else {
-      return stage.ancestors().stream()
+      if (ancestors == null) {
```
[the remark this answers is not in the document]
marchello2000: That's true. I think caching in the stage is just a fraught since the stage wouldn't know when a new stage is inserted into an execution upstream which could happen during stage planning.

The interesting bit is that, technically (today), no mutation will alter the behavior of this `get`ter, the reason is that all this "lookup" in prior stage "outpus" is really only used in bake and evalvars stages today, and it's not like an upstream bake stage will complete while a stage's task is being evaluated (which is the lifetime of stagecontext).

This is all WAY too convoluted. But in the end, I agree with you that even though (I believe) this is safe today, it might not be tomorrow and the code has a considerable "surprise" factor. So I will take it out for now and if I think of a better way of doing this will put up another PR (the perf improvement from recursion removal is pretty considerable already)

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: that's a good point, thanks!

marchello2000: well... it turns out this is not that easy... there is an expectation that the stages are returned in the order they appear in the execution and using a map doesn't preserve that order. I am tempted to say - noone should care but, then again, this is orca and I am sure I will break something if i change it. So going to leave as is for now. Ugh, this makes me upset...

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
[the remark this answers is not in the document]
marchello2000: Oh, i have an idea... @ezimanyi what do you think about caching the total number of stages in the execution and invalidating the cache when that changed? That actually seems safe. It's not immune to stages updating their `requisiteStageRefIds` but that doesn't happen AFAIK

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
ezimanyi: Ah, yes I see what you're saying...if some stage has `requisiteStageRefIds = ["1", "2"]` we are expecting that we'll add the two stages to `toVisit` in the order that the stages appear in `getStages()` rather than in the order they `refId`s appear in `requisiteStageRefIds` (which is actually using the `Collection` interface and so may be unordered anyway).

I guess one solution might be to make a wrapper class `OrderedStage` that contains a `Stage` and its order in the result of `getStages()` and store that in the map; then the call to build `refIdPriorStages` could order on the order field, then map to just get out the `Stage`.  That would replace an `O(N)` operation (where `N` is the total number of stages) with one that is `M log(M)` (where `M` is the average size of `requisiteStageRefIds`).  So the overall algorithm would go from `O(N^2)` to `O(N M log (M))`.  I think that would be better as the average size of dependencies generally be much smaller than the number of stages (ie, most stages would in general only depend on one or two other stages).

Obviously that does add some complexity, though might have a reasonable impact for large enough `N` (though definitely would want to profile to be sure).  Either way, completely up to you to either leave this as is (which is better than it was) or try this here.  If the change already committed fixes the issue so it's not really a problem anymore, it might not be worth adding this complexity.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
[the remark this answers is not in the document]
ezimanyi: I think that sounds reasonable...

I was just looking that how stages get added to an `Execution` and if I'm understanding correctly, there's no `addStage` directly, but rather callers use `getStages` and mutate the resulting array.  Based on that, we really don't have a lot of control over how/when to invalidate the cache (from anywhere...either from stage or from `Execution`).  So I still think this would be a bit risky to cache the value...but given that the current usage pattern appears to be to only ever add stages to the array, caching based on number of stages is probably a reasonable solution if the performance benefits warrant it.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: That makes sense. Seeing as I haven't returned to this PR in a few months... I am going to merge it as is and then noodle on these additional improvements

PR #3310, opened 2019-11-21
feat(executions): Adding new ADMIN endpoint to import execution

srekapalli:
Adding new Admin endpoint to import execution from another spinnaker instance.

Useful to get a execution into other envs for troubleshooting.
Only imports executions with TERMINAL, SUCCEEDED or CANCELLED states.

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
@@ -19,9 +19,11 @@ package com.netflix.spinnaker.orca.controllers
 
 import com.netflix.spinnaker.kork.exceptions.HasAdditionalAttributes
 import com.netflix.spinnaker.kork.web.exceptions.ValidationException
+import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.commands.ForceExecutionCancellationCommand
 import com.netflix.spinnaker.orca.eureka.NoDiscoveryApplicationStatusPublisher
 import com.netflix.spinnaker.orca.pipeline.model.Execution
+import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.beans.factory.annotation.Qualifier
@@ -42,6 +44,9 @@ class AdminController {
   @Autowired(required = false)
   @Qualifier("discoveryStatusPoller")
   ApplicationListener<ContextRefreshedEvent> discoveryStatusPoller
+
+  @Autowired
+  ExecutionRepository executionRepository
 
   @Autowired
   ForceExecutionCancellationCommand forceExecutionCancellationCommand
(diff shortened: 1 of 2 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
```diff
@@ -74,6 +79,23 @@ class AdminController {
     forceExecutionCancellationCommand.forceCancel(executionType, executionId, canceledBy)
   }
 
+
+  @RequestMapping(value = "/executions/", method = RequestMethod.POST)
+  @ResponseStatus(HttpStatus.OK)
+  void createExecution(@RequestBody Execution execution) {
+
+    if (execution.status in [ExecutionStatus.CANCELED, ExecutionStatus.SUCCEEDED, ExecutionStatus.TERMINAL]) {
```
marchello2000: do you think it's worth validating that the app & pipeline exist. Maybe at least spitting our an http response with some text:
```
Imported execution with ID: {} for app: {}, pipeline: {} with {n} stages
Note: the app doesn't exist in this environment
```

I think it would be valuable otherwise folks might import it and not see it and be confused?
Also, would be useful to know the ID of newly imported execution (though it's the same as the payload)

marchello2000:
Thanks for doing this, I think it would be useful and LGTM, but would like another opinion as to any side effects this could cause, @cfieber / @robzienert ?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
```diff
@@ -74,6 +87,35 @@ class AdminController {
     forceExecutionCancellationCommand.forceCancel(executionType, executionId, canceledBy)
   }
 
+
+  @PostMapping(value = "/executions/")
+  @ResponseStatus(HttpStatus.CREATED)
+  Map<String, String> createExecution(@RequestBody Execution execution) {
+
+    if (front50Service && !front50Service.get(execution.application)) {
```
ajordens: I believe `front50Service.get()` will raise an exception if the application doesn't exist (vs. returning null).

I'm not sure we want to import executions for applications that don't exist?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
```diff
@@ -74,6 +87,35 @@ class AdminController {
...
+      log.warn('No application exists with name: ' + execution.application)
+    }
+
+    try {
+      executionRepository.retrieve(execution.type, execution.id)
+      log.warn('Execution found with id: []', execution.id)
```
ajordens: `warn` seems like the wrong logging level? `debug` maybe? (or perhaps it's not even relevant given it's going to raise an exception immediately)

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
```diff
@@ -74,6 +87,35 @@ class AdminController {
...
+      throw new InvalidRequestException('Execution already exists with id: ' + execution.id)
+    } catch(ExecutionNotFoundException e) {
+      log.info('Execution not found .. can import it..')
```
ajordens: This log message seems irrelevant and would get lost in the noise? There's no identifier included in the output.

PR #3311, opened 2019-11-21
fix(notifications/expressions): properly construct spel context

marchello2000:
When I fixed https://github.com/spinnaker/orca/pull/3280 I expanded the `execution` as top level map.
But this is not what we do everywhere else, normally, top level is reserved for stage context while
`execution` is placed under the `execution` key, e.g.:
```
[
	execution: [...]
	trigger: [...]
	..
	stage context
	..
]
```

This change fixes this and adds some UTs, I will also refactor this so that it's harder to make this mistake in the future in my upcoming SpEL PR


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ContextParameterProcessor.java
```diff
@@ -101,20 +101,43 @@ public ContextParameterProcessor(
...
     return new StageContext(stage, augmentedContext);
   }
 
+  /**
+   * Builds a context for the SpEL evaluator to use while processing an execution This involves
+   * merging the following into a map - execution object (if PIPELINE) - trigger object (if
+   * PIPELINE)
+   *
+   * @param execution Execution to build context for
+   * @return Map for of the merged context
```
maggieneterval:
```suggestion
   * @return Map of the merged context
```

marchello2000:
thank you both!

PR #3314, opened 2019-11-22
perf(clouddriver): config to pass scoped manifests to clouddriver when checking pending force cache updates

fkyc:
`ManifestForceCacheRefreshTask` calls Clouddriver when it tries to `checkPendingRefreshes`. This call is inefficient because it doesn't pass any information about which manifests it cares about. As a result, Clouddriver has to look up everything for a particular `cloudProvider` and `OnDemandType`. 

We discovered this issue when trying to kick off 100 concurrent Kubernetes deploys on Spinnaker, with Clouddriver being backed by Redis. Clouddriver will fire off an excessive amount of `sscan` and `mget` calls to Redis while check pending refreshes, and Redis' CPU utilization will get pegged at 100%. 

With this fix in Clouddriver https://github.com/spinnaker/clouddriver/pull/4180 and this fix in this PR, we dramatically improve Spinnaker's ability to handle concurrent manifest deploys when using Redis. We were able to go from 100 concurrent deploys to over 700 concurrent deploys. 

Furthermore, since these optimizations are data store agnostic, we should be able to see performance gains from the SQL side of things as well.

This PR is separated into 2 commits for easier review:
- https://github.com/spinnaker/orca/commit/c3baf209dfc66b5619d3ffbba740773b6d25bb46
  This adds query params to pendingForceCacheUpdates in CloudDriverCacheStatusService.
  This allows us to use information from scoped manifests to help Clouddriver look up cache data more efficiently.
- https://github.com/spinnaker/orca/commit/027758a047a0114c60210206b0c04b8ac2c060b9
  This adds a configuration to have ManifestForceCacheRefreshTask pass scoped manifests to pendingForceCacheUpdates.
  Clouddriver can use information in scoped manifests to look up cache data more efficiently.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/CloudDriverCacheStatusService.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/CloudDriverCacheStatusService.groovy
@@ -18,9 +18,11 @@ package com.netflix.spinnaker.orca.clouddriver
 
 import retrofit.http.GET
 import retrofit.http.Path
+import retrofit.http.QueryMap
 
 interface CloudDriverCacheStatusService {
   @GET("/cache/{cloudProvider}/{type}")
   Collection<Map> pendingForceCacheUpdates(@Path("cloudProvider") String cloudProvider,
-                                           @Path("type") String type)
+                                           @Path("type") String type,
+                                           @QueryMap Map<String, String> queryParams)
 }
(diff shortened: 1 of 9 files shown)


ezimanyi:
Thanks for the PR @fkyc and for all the work that went into uncovering this performance issue! I've added some comments to the linked clouddriver PR and will wait until we resolve those before looking at this in detail.

maggieneterval:
Going to close this PR since it's against the 1.15 release branch, and the corresponding [Clouddriver PR](https://github.com/spinnaker/clouddriver/pull/4180) was also closed.

PR #3316, opened 2019-11-25
fix(executions): Only import executions if application exists in Front50

srekapalli:
- Log stmts clean up
- Only import execution if app can be retrieved from front50.

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
@@ -41,6 +41,7 @@ import org.springframework.web.bind.annotation.RequestMethod
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
+import retrofit.RetrofitError
 
 @RestController
 @RequestMapping("/admin")
@@ -92,16 +93,18 @@ class AdminController {
   @ResponseStatus(HttpStatus.CREATED)
   Map<String, String> createExecution(@RequestBody Execution execution) {
 
-    if (front50Service && !front50Service.get(execution.application)) {
-      log.warn('No application exists with name: ' + execution.application)
+    // Check if app exists before importing execution.
+    try {
+      front50Service.get(execution.application)
+    } catch(RetrofitError e) {
+      throw new InvalidRequestException("Error received retrieving application: " + execution.application, e)
     }
 
     try {
       executionRepository.retrieve(execution.type, execution.id)
-      log.warn('Execution found with id: []', execution.id)
       throw new InvalidRequestException('Execution already exists with id: ' + execution.id)
     } catch(ExecutionNotFoundException e) {
-      log.info('Execution not found .. can import it..')
+      log.info('Execution not found: {}, Will continue with importing..', execution.id)
     }
 
     if (execution.status in [ExecutionStatus.CANCELED, ExecutionStatus.SUCCEEDED, ExecutionStatus.TERMINAL]) {
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
@@ -18,15 +18,11 @@
 package com.netflix.spinnaker.orca.controllers
 
 import com.netflix.spinnaker.kork.exceptions.HasAdditionalAttributes
-import com.netflix.spinnaker.kork.web.exceptions.InvalidRequestException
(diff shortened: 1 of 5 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
```diff
@@ -92,16 +93,18 @@ class AdminController {
...
+      throw new InvalidRequestException("Error received retrieving application: " + execution.application, e)
     }
 
     try {
       executionRepository.retrieve(execution.type, execution.id)
-      log.warn('Execution found with id: []', execution.id)
       throw new InvalidRequestException('Execution already exists with id: ' + execution.id)
     } catch(ExecutionNotFoundException e) {
-      log.info('Execution not found .. can import it..')
+      log.info('Execution not found: {}, Will continue with importing..', execution.id)
```
marchello2000: don't think we need this since we have a log on line 111

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/AdminController.groovy
```diff
@@ -92,16 +93,18 @@ class AdminController {
   @ResponseStatus(HttpStatus.CREATED)
   Map<String, String> createExecution(@RequestBody Execution execution) {
 
-    if (front50Service && !front50Service.get(execution.application)) {
-      log.warn('No application exists with name: ' + execution.application)
+    // Check if app exists before importing execution.
+    try {
+      front50Service.get(execution.application)
+    } catch(RetrofitError e) {
+      throw new InvalidRequestException("Error received retrieving application: " + execution.application, e)
```
marchello2000: Talking to the team, i think it would be better to:
1. Ignore the failures here (e.g. allow importing into an app that doesn't exist - otherwise this EP is not super useful)
2. Add spring config prop to explicitly enable this EP (we would do this on prestaging/test/local only)

orca-web/src/main/java/com/netflix/spinnaker/orca/controllers/ExecutionsImportController.java
```diff
@@ -0,0 +1,106 @@
...
+
+  @PostMapping(value = "")
+  @ResponseStatus(HttpStatus.CREATED)
+  Map<String, Object> createExecution(@RequestBody Execution execution) {
+
+    // Check if app exists before importing execution.
+    try {
+      front50Service.get(execution.getApplication());
+    } catch (RetrofitError e) {
+      log.warn("Exception received while retrieving application from font50", e);
```
marchello2000:
```suggestion
      log.warn("Exception received while retrieving application from front50", e);
```

orca-web/src/main/java/com/netflix/spinnaker/orca/controllers/ExecutionsImportController.java
```diff
@@ -0,0 +1,106 @@
...
+
+  @PostMapping(value = "")
+  @ResponseStatus(HttpStatus.CREATED)
+  Map<String, Object> createExecution(@RequestBody Execution execution) {
+
+    // Check if app exists before importing execution.
+    try {
+      front50Service.get(execution.getApplication());
+    } catch (RetrofitError e) {
+      log.warn("Exception received while retrieving application from front50", e);
```
cfieber: for the sake of logging noise it might be worth differentiating this message for the potentially expected case (404 no existing app) vs some other failure
can examine e.getResponse to determine if it was an HTTP failure and if it was a 404

orca-web/src/main/java/com/netflix/spinnaker/orca/controllers/ExecutionsImportController.java
```diff
@@ -0,0 +1,106 @@
...
+  @Autowired
+  ExecutionsImportController(
+      ExecutionRepository executionRepository, Front50Service front50Service) {
+    this.front50Service = front50Service;
+    this.executionRepository = executionRepository;
+  }
+
+  @PostMapping(value = "")
+  @ResponseStatus(HttpStatus.CREATED)
+  Map<String, Object> createExecution(@RequestBody Execution execution) {
```
jonsie: I think for new endpoints we should strive to return a POJO rather than `Map<String, Object>`.

orca-web/src/main/java/com/netflix/spinnaker/orca/controllers/ExecutionsImportController.java
srekapalli: Fixed now @cfieber

orca-web/src/main/java/com/netflix/spinnaker/orca/controllers/ExecutionsImportController.java
srekapalli: Fixed now @jonsie

PR #3317, opened 2019-11-25
fix(execution): Resume parent pipeline when a failed stage in a child pipeline restarts

AbdulRahmanAlHamali:
If we have a pipeline, where one of the stages starts another pipeline, and then a stage fails in the inner pipeline. We can restart that inner stage, but the parent pipeline will not be notified of the restart, and thus will stay stuck. The only way to be able to actually move forward is to restart the whole failed inner pipeline, which is expensive and dangerous.

This PR provides a method to solve this problem. The main concept of the fix is to find and restart the stage in the parent pipeline, but to inform it that the child has already been restarted, and that it just needs to monitor it


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt
```diff
@@ -58,13 +59,38 @@ class RestartStageHandler(
...
           repository.updateStatus(topStage.execution.type, topStage.execution.id, RUNNING)
           queue.push(StartStage(startMessage))
         }
       }
     }
   }
 
+  private fun restartParentPipelineIfNeeded(message: RestartStage, topStage: Stage) {
+    val trigger = topStage.execution.trigger
+    if (!topStage.execution.trigger.type.equals("pipeline")) {
```
marchello2000: you are probably better of checking the type of the trigger in a type-safe way
```suggestion
    if (!topStage.execution.trigger instanceof PipelineTrigger) {
```

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt
```diff
@@ -58,13 +59,38 @@ class RestartStageHandler(
...
+      return
+    }
+
+    val parentExecution = trigger.other["parentExecution"] as Execution
```
marchello2000: then you can do `(trigger as PipelineTrigger).parentExecution`

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt
```diff
@@ -58,13 +59,38 @@ class RestartStageHandler(
...
+    parentStage.addAlreadyRunning(message)
+
+    queue.push(RestartStage(parentExecution, parentStage.id, message.user))
+  }
+
+  /**
+   * Inform the parent stage when it restarts that the child is already running
+   */
+  private fun Stage.addAlreadyRunning(message: RestartStage) {
+    message.withExecution { execution ->
```
marchello2000: this seems weird, why add these props on the execution it should be on the stage... but also... let me think about a better way of doing this

marchello2000: ok, so i think it would be better to do the following:
1. add something to the stage context like `context["_skipPipelineRestart"] = true`
2. No need to add `executionId` or `executionName` since those are already there
3. You will need to store the execution to persist this change to the stage in the executionRepository
4. `PipelineStage.prepareStageForRestart` should check the `_skipPipelineRestart` field, and if `true` not delete the `executionId/executionName`, from the context. It should *also clear* the `_skipPipelineRestart` flag so that if the user wanted to actually restart it explicitly afterwards it would work as expected (e.g. restart the child)
5. the `PipelineStage.taskGraph` should only create the start pipelinetask if there is no `executionId` present in the *current* stage context.

AbdulRahmanAlHamali:
Hi @marchello2000 I really appreciate the feedback. I will address your points and submit a fix.

Regarding your concerns:

> what happens if the user specified they don't want to wait for the child pipeline to complete?
I think we can pass down in the trigger a flag that says: I'm not waiting for you, and based on that we simply ignore the whole process.

> what happens if the parent pipeline is still running from the previous run?
If the parent pipeline is waiting for the child pipeline, and it failed because of the child pipeline's failure, then this scenario wouldn't happen. Unless there is some specific case where it could happen?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt
AbdulRahmanAlHamali: Alright I have addressed your points, thanks for the feedback!

I did not really understand point 3 though. I think this is already being done elsewhere in the code, but I might be wrong.

Also, regarding point 5, the stage context passed to `taskGraph` is the *current* one, right?

Let me know what you think, and thanks again!

marchello2000:
> 
> > what happens if the parent pipeline is still running from the previous run?
> 
> If the parent pipeline is waiting for the child pipeline, and it failed because of the child pipeline's failure, then this scenario wouldn't happen. Unless there is some specific case where it could happen?

Oh, this can absolutely happen.
1. Imagine a pipeline where the user has specified "continue on failure" and the "continuation" part is LONG (like hours - which is pretty common) - this is similar to my first concern
2. Imagine a pipeline that has two branches: one is the where this child pipeline lives and another, parallel branch. Now the pipeline stage can say "failPipeline=true" and "continuePipeline=true" which means it would wait for the parallel branch to complete before completing the parent pipeline.

I am not sure what the best approach here.
I will take a look at your commits later - gotta run

AbdulRahmanAlHamali:
Hey @marchello2000, sorry for the late reply.

OK, so we can say we have three principal cases:
1. The parent pipeline is halted, either because of the failure of the child pipeline, or at a later stage. In that case, a restart of the child pipeline should simply restart the pipeline stage in the parent pipeline, as I have already implemented.
2. The parent pipeline is running a stage that is dependent directly or indirectly on the pipeline stage: I think we can detect that easily in the code, and in that case, we do not restart the pipeline stage.
3. The parent pipeline is running a stage that is independent from the pipeline stage: In that case the restart should work normally. I think we already allow restarting stages while others are still running, right?

So, as a solution, we can simply check before restarting the parent pipeline that it is either halted, or running a stage that is not dependent on that one. Thoughts?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt
```diff
@@ -58,13 +59,34 @@ class RestartStageHandler(
...
+  private fun restartParentPipelineIfNeeded(message: RestartStage, topStage: Stage) {
+    if (topStage.execution.trigger !is PipelineTrigger) {
+      return
+    }
+
+    val trigger = topStage.execution.trigger as PipelineTrigger
+
+    val parentStage = trigger.parentExecution.stageById(trigger.parentPipelineStageId)
+    parentStage.addSkipRestart()
+
```
marchello2000: I believe you will need a `repository.storeStage(parentStage)` here to make sure your changes are persisted

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt
```diff
@@ -58,13 +59,34 @@ class RestartStageHandler(
...
+    queue.push(RestartStage(trigger.parentExecution, parentStage.id, message.user))
+  }
+
+  /**
+   * Inform the parent stage when it restarts that the child is already running
+   */
+  private fun Stage.addSkipRestart() {
+    context["_skipPipelineRestart"] = "true"
```
marchello2000: why not
```suggestion
    context["_skipPipelineRestart"] = true
```

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
```diff
@@ -65,8 +68,18 @@ public void taskGraph(Stage stage, TaskNode.Builder builder) {
   @Override
   public void prepareStageForRestart(Stage stage) {
     stage.getContext().remove("status");
-    stage.getContext().remove("executionName");
-    stage.getContext().remove("executionId");
+    if (!stage
```
marchello2000: nit: i, personally, find this really hard to read. I would maybe rewrite as:

```java
StageContext context = (StageContext)stage.getContext();
boolean restartChild = (boolean)(context.getCurrentOnly("_skipPipelineRestart", false));

if (restartChild) {
  ..
 } else {
  ..
}
```

Note the use of `getCurrentOnly` - that way you are only inspecting the context of the current stage not any of the outputs of its parents (a bug that we've run into a bunch)

marchello2000:
@AbdulRahmanAlHamali : thanks for pushing on this.
At a high level, you are correct. I think there are subtleties though (e.g. a direct or indirect descendant can have a conditional on the success of the stage in question and hence will fall into your case #2 but a restart will probably be a better bet.

I am leaning towards "only restart parent pipeline stage if the parent pipeline is halted" - it's simpler, more predictable, and is much easier to reason about (both in code and from user's perspective). I also think it will cover the large majority of the use cases. How do you feel about that?

Also, would love input from @robfletcher or @ajordens - as they have a ton more operational experience in this area

AbdulRahmanAlHamali:
Hey Mark, thanks for the review. I will submit changes to address your comment in a bit.

Regarding restarting only in the halted pipeline case. I don't mind that. We might have some twisted use cases for our users that are not covered by this. But for now I think this covers almost everything

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt
AbdulRahmanAlHamali: Shouldn't I use this instead?
```
repository.updateStageContext(parentStage)
```

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageContext.java
```diff
@@ -71,7 +71,7 @@ public Object get(@Nullable Object key) {
    * @param defaultValue default value to return if key is not present
    * @return value or null if not present
    */
-  Object getCurrentOnly(@Nullable Object key, Object defaultValue) {
+  public Object getCurrentOnly(@Nullable Object key, Object defaultValue) {
```
AbdulRahmanAlHamali: made this public to be able to access it from `PipelineStage`

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RestartStageHandler.kt
```diff
@@ -58,13 +60,40 @@ class RestartStageHandler(
...
   }
 
+  private fun restartParentPipelineIfNeeded(message: RestartStage, topStage: Stage) {
+    if (topStage.execution.trigger !is PipelineTrigger) {
+      return
+    }
+
+    val trigger = topStage.execution.trigger as PipelineTrigger
+
+    if (trigger.parentExecution.status != TERMINAL) {
```
AbdulRahmanAlHamali: I used `TERMINAL` here, but I could have used something else like `STOPPED`. Not sure which is best

marchello2000: Why only restart in the `TERMINAL` case? I think it's ok to "restart" so long as the parent is not RUNNING, no?
In that case, you can
```suggestion
    if (trigger.parentExecution.status.isComplete()) {
```

AbdulRahmanAlHamali:
From what I see, `isComplete` is set to true for success, cancellation and other cases. Do we want to restart even in those cases?

AbdulRahmanAlHamali:
Regarding the testing, I agree; unit tests don't add much here, since the change is mostly about the interaction with the different components. But I will run some end-to-end tests on that before we merge it.

marchello2000:
> From what I see, `isComplete` is set to true for success, cancellation and other cases. Do we want to restart even in those cases?

I think so... I guess your team is running into these use cases - curious what your perspective is. For me, I just wanted to make sure that we don't "restart" a stage while the pipeline is running because that could lead to some weird situations that I am having a hard time visualizing...

I was thinking that perhaps you have a child pipeline stage that is marked as "ignore failure"so the parent pipeline will succeed but you might still want to rerun the child if it failed

AbdulRahmanAlHamali:
Hey mark. 

My exact use case is this: I have multiple environments: dev, qa, prd, etc. each represented by its own pipeline.

Then, I have a parent pipeline, which triggers those pipelines one after the other, and does some other extra logic.

If the 10th stage in dev fails for example, the parent pipeline stops. And the only reason to make the parent pipeline resume is by restarting all of dev environment, which is not optimal, since 10 steps have already been executed in `dev`. So what I really want is for the user to simply restart the failed stage in `dev`, which will awaken the parent pipeline, so that it goes back to waiting for dev, and when dev is finished, it continues deploying in the other environments.

In case the parent pipeline ignores the failure of the child pipeline, then it really is not a problem for me. The only problem is when the parent pipeline fails because of the failure of `dev`, which could either be immediate (`failPipeline=true`), or not (`completeOtherBranchesThenFail`=true).

AbdulRahmanAlHamali:
I think the above are the only two cases where a restart is necessary, and maybe we can add a more specific check just for them.

What do you think?

marchello2000:
Got it. I think we can keep the check as is for now (e.g. if the pipeline is not running restart/awaken it). I just think the more complex the check the harder it is to get it right and, more importantly, reason about it for the end users

AbdulRahmanAlHamali:
Hey mark, did you mean to keep it as `TERMINAL`?

marchello2000:
Just to capture our offline convo: I think the change here look good, thank you for pushing them through. I will help test it (today/tomorrow) and then mark it approved/merge

AbdulRahmanAlHamali:
@spinnakerbot cherrypick 1.17

AbdulRahmanAlHamali:

AbdulRahmanAlHamali:

PR #3318, opened 2019-11-25
feature(customJobStage): Allow users to provide parameters to custom job stage without providing a mapping

AbdulRahmanAlHamali:
Addresses issue #5147:

https://github.com/spinnaker/spinnaker/issues/5147

We already have `parameters` in the context, but it would throw exceptions before if we didn't have a mapping defined. This PR should solve the problem, and allow us to include expressions like `${parameters['my_param']}` in the manifest


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.groovy
```diff
@@ -72,17 +72,20 @@ class PreconfiguredJobStage extends RunJobStage {
         context[it] = preconfiguredMap[it]
       }
     }
```
AbdulRahmanAlHamali: the code structure was a bit uncomfortable, so I restructured it a bit

AbdulRahmanAlHamali: But the main idea is basically confirming that `mapping` is defined on the parameter before we use it

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStageSpec.groovy
```diff
@@ -53,6 +53,7 @@ class PreconfiguredJobStageSpec extends Specification {
     expectedField   | expectedValue   | stageName | stageContext                                                              | preconfiguredJobProperties
     "cloudProvider" | "kubernetes"    | "testJob" | [account: "test-account"]                                                 | new KubernetesPreconfiguredJobProperties(enabled: true, label: "testJob", type: "testJob", parameters: [], cloudProvider: "kubernetes")
     "cloudProvider" | "titus"         | "testJob" | [account: "test-account"]                                                 | new KubernetesPreconfiguredJobProperties(enabled: true, label: "testJob", type: "testJob", parameters: [new PreconfiguredJobStageParameter(mapping: "cloudProvider", defaultValue: "titus")], cloudProvider: "kubernetes")
+    "cloudProvider" | "titus"         | "testJob" | [account: "test-account"]                                                 | new KubernetesPreconfiguredJobProperties(enabled: true, label: "testJob", type: "testJob", parameters: [new PreconfiguredJobStageParameter(mapping: "cloudProvider", defaultValue: "titus", "name": "cloudProvider")], cloudProvider: "kubernetes")
```
robzienert: Doesn't look like this row changes anything? Are we sure this test is actually asserting what you're changing?

robzienert:
Ping.

AbdulRahmanAlHamali:
Hi @robzienert 
I'm actually not sure if that change is going to work. And we have dropped using the CustomJobStage in favour of simply using the normal JobStage.

Shall I close this PR? I'll re-open, and finish properly if we end up actually needing this feature.

PR #3322, opened 2019-11-27
fix(spel): make contextParameterProcess.process side-effect free

marchello2000:
With SpEL v4 Eval Vars calls process once for each variable, that doesn't work because .process has sideeffects and mutates context in a way that prevents it from running twice (bad, bad).
The culprit is `precomputeValues` which expands and does naughty things with the trigger. Making a copy of the context is a simple way to prevent the sideeffects

Also, as part of this i noticed that if `StartStageHandler` fails the error handling is not great and we are stuck with a zombie pipeline

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
@@ -74,54 +74,69 @@ class StartStageHandler(
 
   override fun handle(message: StartStage) {
     message.withStage { stage ->
-      stage.withAuth {
-        if (stage.anyUpstreamStagesFailed()) {
-          // this only happens in restart scenarios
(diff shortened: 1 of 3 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
```diff
@@ -74,54 +74,69 @@ class StartStageHandler(
...
-
-                message.setAttribute(MaxAttemptsAttribute(40))
-                queue.push(message, retryDelay)
-              } else {
-                log.error("Error running ${stage.type} stage for ${message.executionType}[${message.executionId}]", e)
-                stage.apply {
-                  context["exception"] = exceptionDetails
-                  context["beforeStagePlanningFailed"] = true
-                }
+      try {
```
marchello2000: review with ignore white space

PR #3323, opened 2019-12-02
fix(cfn): include credentials on the task sent to clouddriver

xavileon:
Clouddriver expects the credentials to be on the kato task context we
sent from Orca. That was not the case for the DeleteCloudFormationTask
so it was failing in Clouddriver.

This patch adds the credentials field so clouddriver can do the
task conversion properly.


xavileon:

PR #3326, opened 2019-12-03
feat(aws/cfn): Execute cloudformation ChangeSets

sergiopena:
Add cloudformation execution tasks, there are two new tasks for this stage:

* EvaluateCloudFormationChangeSetExecution

This class holds the logic for the execution when changeset contains a replacement, the main problem is that a changeset can have **replacements** which means some of the infra will be replaced losing data, (db changes, etc). We cannot know if a changeset contains a replacement until it is created in the previous tasks. 

So this class do next when there is a replacement:
- If there is a pre-configured behavior set in deck to skip the execution, run it anyway or fail the stage, it set the context with this flag that will be used in the next task.
- If the preconfigured behavior is to "ask"  it does pretty much the same than manualJudgment stage, marks the context for deck and enters in a loop, waiting for user judgment.

* ExecuteCloudFormationChangeSet

This class runs the changeset, skips it or fails the stage based on user decision.

Issue link:
https://github.com/spinnaker/spinnaker/issues/5091


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/EvaluateCloudFormationChangeSetExecutionTask.java
```diff
@@ -0,0 +1,116 @@
...
+  public @Nonnull TaskResult execute(Stage stage) {
+    String actionOnReplacement =
+        (String) Optional.ofNullable(stage.getContext().get("actionOnReplacement")).orElse("ask");
+
+    if (!actionOnReplacement.equals("ask")) {
+      return TaskResult.SUCCEEDED;
+    }
+
+    Optional<Map> currentChangeSet = getCurrentChangeSet(stage);
+    Optional<Boolean> changeSetIsReplacement =
```
xavileon: nit: maybe better `changeSetContainsReplacement` and all others isReplacement to containsReplacement. It better describes the actual flag.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/EvaluateCloudFormationChangeSetExecutionTask.java
```diff
@@ -0,0 +1,116 @@
...
+        Optional.ofNullable(getChangeSetIsReplacement(stage));
+
+    if (!changeSetIsReplacement.isPresent()) {
+      Boolean isReplacement;
+      if (isAnyChangeSetReplacement(currentChangeSet.get())) {
+        isReplacement = true;
+      } else {
+        isReplacement = false;
+      }
```
xavileon: you can get the same with
Boolean isReplacement = isAnyChangeSetReplacement(currentChangeSet.get());

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/EvaluateCloudFormationChangeSetExecutionTask.java
```diff
@@ -0,0 +1,116 @@
...
+  public long getTimeout() {
+    return Integer.MAX_VALUE;
+  }
+
+  private String getChangeSetExecutionChoice(Stage stage) {
+    return (String)
+        Optional.ofNullable(stage.getContext().get("changeSetExecutionChoice")).orElse("none");
+  }
+
+  private Boolean getChangeSetIsReplacement(Stage stage) {
```
xavileon: You are returning a null here if the key is not there, and then transforming it back to optional on the caller. Maybe return Optionals on all these helper methods and react accordingly on the main function? We could get rid of the "none" string with is just signaling a non-existing value.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/ExecuteCloudFormationChangeSetTask.java
```diff
@@ -0,0 +1,129 @@
...
+  public TaskResult execute(@Nonnull Stage stage) {
+
+    String actionOnReplacement =
+        (String) Optional.ofNullable(stage.getContext().get("actionOnReplacement")).orElse("fail");
+
+    Optional<Map> currentChangeSet = getCurrentChangeSet(stage);
+    if (currentChangeSet.isPresent()) {
+      if (isAnyChangeSetReplacement(currentChangeSet.get())) {
+        switch (actionOnReplacement) {
+          case "skip":
```
xavileon: These actions should probably be constants somewhere, so the constant can be reused in other tasks and we don't risk breaking the workflow if we decide to change the value of the replacement string.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/WaitForCloudFormationCompletionTask.java
```diff
@@ -67,9 +67,12 @@ public TaskResult execute(@Nonnull Stage stage) {
               + stack.get("stackStatus"));
       boolean isChangeSet =
           (boolean) Optional.ofNullable(stage.getContext().get("isChangeSet")).orElse(false);
+      boolean isChangeSetDeploy =
```
xavileon: What is the meaning of this flag?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/DeployCloudFormationStage.java
```diff
@@ -38,9 +40,23 @@ public void taskGraph(@Nonnull Stage stage, @Nonnull TaskNode.Builder builder) {
...
         .withTask("waitForCloudFormationCompletion", WaitForCloudFormationCompletionTask.class);
+
     if ((boolean) Optional.ofNullable(stage.getContext().get("isChangeSet")).orElse(false)) {
       builder.withTask("deleteCloudFormationChangeSet", DeleteCloudFormationChangeSetTask.class);
       builder.withTask("monitorDeleteCloudFormationChangeSet", MonitorKatoTask.class);
     }
+
+    if ((boolean) Optional.ofNullable(stage.getContext().get("isChangeSet")).orElse(false)
+        && (boolean)
+            Optional.ofNullable(stage.getContext().get("executeChangeSet")).orElse(false)) {
```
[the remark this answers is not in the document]
sergiopena: You are right, much more readable. Changing it!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/EvaluateCloudFormationChangeSetExecutionTask.java
sergiopena: 😅 I've changed this! Thanks!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/WaitForCloudFormationCompletionTask.java
sergiopena: That flag is set by  [ExecuteCloudFormationChangeSetTask:101](https://github.com/sergiopena/orca/blob/execute_changesets/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/ExecuteCloudFormationChangeSetTask.java#L101) is used to distinguish when we are creating the ChangeSet from when it's being deployed, so we can reuse the logic and wait for changeSets execution too. Shall I change the name to isChangeSetExecution to make it clearer?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/DeployCloudFormationTask.java
```diff
@@ -115,6 +116,12 @@ public TaskResult execute(@Nonnull Stage stage) {
               .stream().map(part -> new Yaml().dump(part)).collect(Collectors.joining("\n---\n"));
       task.put("templateBody", templateBody);
     }
+
+    if (templateBody instanceof List && !((List) templateBody).isEmpty()) {
```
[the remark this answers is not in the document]
sergiopena: Agree, I've removed the duplicated code.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/WaitForCloudFormationCompletionTask.java
sergiopena: Renamed to isChangeSetExecution

xavileon:
@spinnaker/oss-approvers PTAL

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/EvaluateCloudFormationChangeSetExecutionTask.java
sergiopena: Done!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/EvaluateCloudFormationChangeSetExecutionTask.java
sergiopena: I've changed the logic of this. WDYT?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/ExecuteCloudFormationChangeSetTask.java
sergiopena: Thank you, good point. I've moved those values to constants.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/EvaluateCloudFormationChangeSetExecutionTask.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  @Override
+  public long getBackoffPeriod() {
+    return 1_000;
+  }
+
+  @Override
+  public long getTimeout() {
+    return Integer.MAX_VALUE;
```
robzienert: Ha, whoa there. Is this actually what you want?

sergiopena: A lifetime timeout is a bit too much 😅 I took that from the [WaitUntilTask.java](https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/WaitUntilTask.java#L87)
Same timeout than manualJudgement task will be better, 3 days. I will change it

Thanks a lot for the review @robzienert 🙇

xavileon:
@spinnakerbot add-label ready-to-merge

xavileon:
@spinnakerbot add-label "ready to merge"

xavileon:
@spinnakerbot add-label ready%20to%20merge

xavileon:
Thanks! ❤️

ezimanyi:
Opened spinnaker/spinnaker#5260 because it seems there's actually no way to get spinnakerbot to add a label containing a space.

PR #3328, opened 2019-12-04
refactor(artifacts): Use builder to create artifacts

ezimanyi:
The all-arg and no-arg constructors for Artifact and ExpectedArtifact were deprecated in spinnaker/kork#429. Replace their usages with a builder.

### DIFF
--- a/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolverSpec.groovy
+++ b/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolverSpec.groovy
@@ -100,14 +100,14 @@ class ArtifactResolverSpec extends Specification {
         name = "upstream stage"
         type = "stage1"
         refId = "1"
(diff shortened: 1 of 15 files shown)


orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolverSpec.groovy
```diff
@@ -394,22 +400,66 @@ class ArtifactResolverSpec extends Specification {
     artifacts.size == 2
     artifacts*.type == ["1", "trigger"]
   }
+
   @Unroll
-  def "should resolve expected artifacts from pipeline for #expectedArtifacts using #available and prior #prior"() {
+  def "should resolve expected artifacts from pipeline for default #defaultArtifact, available #available, and prior #prior"() {
```
ezimanyi: I minorly refactored this and the next test, as the enormous table in the `where` clause was getting hard to read.  I pulled the one case with multiple expected artifacts into its own test case, and simplified this test given that it only had to consider one artifact.

PR #3329, opened 2019-12-04
fix(monitored deploy): Give more leniency to scale down operations

marchello2000:
During a rolling deploy, if `scaleDown`, `notifyDeployCompleted`, or `shrinkCluster` fail and auto rollback was requested, a rollback will be performed.
This is suboptimal because you can easily argue that these operations are non-critical. In fact at this point the deploy is completed (and the old ASG is disabled).
So these are purely cleanup actions and having a perfectly good deploy fail and rollback due to a cleanup action failing is silly.

This change makes the three stages return `FAILED_CONTINUE` instead `TERMINAL` on failure thus not triggering a rollback.
The pipeline will still be marked as `FAILED_CONTINUE`

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.config.DeploymentMonitorDefinition;
 import com.netflix.spinnaker.orca.ExecutionStatus;
 import com.netflix.spinnaker.orca.RetryableTask;
 import com.netflix.spinnaker.orca.TaskResult;
+import com.netflix.spinnaker.orca.clouddriver.pipeline.monitoreddeploy.NotifyDeployCompletedStage;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.strategies.MonitoredDeployStageData;
 import com.netflix.spinnaker.orca.deploymentmonitor.DeploymentMonitorServiceProvider;
 import com.netflix.spinnaker.orca.deploymentmonitor.models.DeploymentStep;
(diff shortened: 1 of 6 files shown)


orca-kayenta/src/main/kotlin/com/netflix/spinnaker/orca/kayenta/pipeline/CleanupCanaryClustersStage.kt
```diff
@@ -47,15 +46,15 @@ class CleanupCanaryClustersStage : StageDefinitionBuilder {
           it.context.putAll(pair.control.toContext)
           it.context["remainingEnabledServerGroups"] = 0
           // Even if disabling fails, move on to shrink below which will have another go at destroying the server group
-          it.setContinueOnFailure(true)
```
gal-yardeni: Did you rename `setContinueOnFailure` to `continuePipelineOnFailure`? It looks like both methods are there.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -188,6 +189,22 @@ private TaskResult handleError(
       }
     }
 
+    // Don't let failures in NotifyDeployComplete cause rollback
+    if (stage.getType().equalsIgnoreCase(NotifyDeployCompletedStage.PIPELINE_CONFIG_TYPE)) {
```
gal-yardeni: Can you elaborate on how this part is preventing rollback in case one of the 3 stages fails?

gal-yardeni:
Thanks for adding this @marchello2000! Is it a part of a reported bug? or just enhancement?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -803,6 +803,17 @@ public void setAllowSiblingStagesToContinueOnFailure(boolean propagateFailuresTo
...
   }
 
+  @JsonIgnore
+  public void setContinuePipelineOnFailure(boolean continuePipeline) {
+    context.put("continuePipeline", continuePipeline);
+  }
+
+  @JsonIgnore
+  public boolean getContinuePipelineOnFailure() {
+    StageContext context = (StageContext) getContext();
```
dreynaud: slightly off topic but what's with the cast here? Can we add a `getStageContext` method that returns a `StageContext`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
marchello2000: this is part of it, this will cause, e.g. http errors to monitor service to bubble as `FAILED_CONTINUE` instead of `TERMINAL` then the changes https://github.com/spinnaker/orca/pull/3329/files#diff-cc03fa80e1a96b9c0443a8fe1ef22a6fR349-R351 will instruct orca to continue with other stages

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
marchello2000: yeah, that would be better, i will add to a list of things for me to fix, thanks!

PR #3335, opened 2019-12-09
fix(expression): various fixes related to SpEL v4

marchello2000:
* Fix MPTv2 with SpEL v4.
  The issue here was that `templateVariables.*` would need to be expanded in stages at pipeline creation.
  Since this is not possible anymore, `templateVariables.*` are now added to the execution and used in the evaluation context during stage execution.
  Thus the template variables are evaluated at runtime instead of creation time.
* Made SpEL v3 not deprecated (in the capabilities)
* Added description to variable in the eval vars stage

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
@@ -318,6 +318,16 @@ public class Execution implements Serializable {
     this.spelEvaluator = spelEvaluatorVersion;
   }
 
+  private Map<String, Object> templateVariables = null;
+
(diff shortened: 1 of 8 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Execution.java
```diff
@@ -318,6 +318,16 @@ public void setSpelEvaluator(@Nullable String spelEvaluatorVersion) {
     this.spelEvaluator = spelEvaluatorVersion;
   }
 
+  private Map<String, Object> templateVariables = null;
```
marchello2000: this will need an echo change as well

PR #3341, opened 2019-12-11
fix(rollback): Ignore disabled server groups in automatic rollbacks

luispollo:
We observed that, when a `RollbackClusterStage` was executed to rollback a failed deployment, the `DetermineRollbackCandidatesTask` was including disabled server groups when determining the target server group to rollback to, despite explicit class comments to the contrary. It did filter out disabled server groups when determining the newest server group, but not when actually determining the one to rollback to.

This was easily reproducible by creating 2 server groups `v000` and `v001` in a cluster, then disabling `v001`, then using a pipeline to deploy a new server group `v002` to the cluster. When this deployment failed, instead of rolling back to `v000`, we would rollback to the disabled group `v001`.

The fix should be self-explanatory given the context above.

Update: upon further consideration, we realized that changing this behavior might affect existing pipelines with a `RollbackClusterStage`, and so a condition was added to only exclude disabled server groups if the stage was initiated as part of an automatic rollback from a failed deployment stage.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
@@ -121,6 +121,11 @@ class CreateServerGroupStage extends AbstractDeployStrategyStage {
       additionalRollbackContext.disableOnly = true
     }
 
+    // When initiating a rollback automatically as part of deployment failure handling, only rollback to a server
+    // group that's enabled, as any disabled ones, even if newer, were likely manually marked so for being "bad"
+    // (e.g. as part of a manual rollback).
+    additionalRollbackContext.disregardDisabledCandidates = true
+
     if (strategySupportsRollback) {
       graph.add {
         it.type = rollbackClusterStage.type
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
@@ -124,7 +124,7 @@ class CreateServerGroupStage extends AbstractDeployStrategyStage {
     // When initiating a rollback automatically as part of deployment failure handling, only rollback to a server
     // group that's enabled, as any disabled ones, even if newer, were likely manually marked so for being "bad"
     // (e.g. as part of a manual rollback).
-    additionalRollbackContext.disregardDisabledCandidates = true
+    additionalRollbackContext.onlyEnabledServerGroups = true
 
     if (strategySupportsRollback) {
       graph.add {
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
@@ -211,12 +211,12 @@ class MonitoredDeployStrategy implements Strategy {
         resizeContext.capacity = savedCapacity // will scale to a percentage of that static capacity
       }
 
-      log.info("Adding `Grow to $p% of Desired Size` stage with context $resizeContext [executionId=${stage.execution.id}]")
+      log.info("Adding `Grow new server group to $p% of Desired Size` stage with context $resizeContext [executionId=${stage.execution.id}]")
 
       def resizeStage = newStage(
         stage.execution,
         ResizeServerGroupStage.TYPE,
-        "Grow to $p% of Desired Size",
+        "Grow new server group to $p% of Desired Size",
         resizeContext,
         stage,
         SyntheticStageOwner.STAGE_AFTER
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
@@ -459,7 +459,11 @@ class MonitoredDeployStrategy implements Strategy {
         serverGroup              : stageData.serverGroup,
         stageTimeoutMs           : MINUTES.toMillis(30), // timebox a rollback to 30 minutes
         additionalRollbackContext: [
-          enableAndDisableOnly: true
+          enableAndDisableOnly: true,
+          // When initiating a rollback automatically as part of deployment failure handling, only rollback to a server
+          // group that's enabled, as any disabled ones, even if newer, were likely manually marked so for being "bad"
+          // (e.g. as part of a manual rollback).
+          disregardDisabledCandidates: true
         ]
       ],
       parent,
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
@@ -463,7 +463,7 @@ class MonitoredDeployStrategy implements Strategy {
           // When initiating a rollback automatically as part of deployment failure handling, only rollback to a server
           // group that's enabled, as any disabled ones, even if newer, were likely manually marked so for being "bad"
           // (e.g. as part of a manual rollback).
-          disregardDisabledCandidates: true
+          onlyEnabledServerGroups: true
         ]
       ],
       parent,
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
@@ -211,12 +211,12 @@ class MonitoredDeployStrategy implements Strategy {
         resizeContext.capacity = savedCapacity // will scale to a percentage of that static capacity
       }
 
-      log.info("Adding `Grow new server group to $p% of Desired Size` stage with context $resizeContext [executionId=${stage.execution.id}]")
+      log.info("Adding `Grow ${internalStageData.newServerGroup} to $p% of Desired Size` stage with context $resizeContext [executionId=${stage.execution.id}]")
 
       def resizeStage = newStage(
         stage.execution,
         ResizeServerGroupStage.TYPE,
-        "Grow new server group to $p% of Desired Size",
+        "Grow ${internalStageData.newServerGroup} to $p% of Desired Size",
         resizeContext,
         stage,
         SyntheticStageOwner.STAGE_AFTER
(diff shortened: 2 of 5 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
```diff
@@ -211,12 +211,12 @@ class MonitoredDeployStrategy implements Strategy {
...
       }
 
-      log.info("Adding `Grow to $p% of Desired Size` stage with context $resizeContext [executionId=${stage.execution.id}]")
+      log.info("Adding `Grow new server group to $p% of Desired Size` stage with context $resizeContext [executionId=${stage.execution.id}]")
 
       def resizeStage = newStage(
         stage.execution,
         ResizeServerGroupStage.TYPE,
-        "Grow to $p% of Desired Size",
+        "Grow new server group to $p% of Desired Size",
```
marchello2000: if you are doing this, then you should put the actual ASG name in here, you can get it from
`internalStageData.newServerGroup` (same with the log statement above)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
```diff
@@ -211,12 +211,12 @@ class MonitoredDeployStrategy implements Strategy {
         resizeContext.capacity = savedCapacity // will scale to a percentage of that static capacity
       }
 
-      log.info("Adding `Grow to $p% of Desired Size` stage with context $resizeContext [executionId=${stage.execution.id}]")
+      log.info("Adding `Grow new server group to $p% of Desired Size` stage with context $resizeContext [executionId=${stage.execution.id}]")
```
luispollo: This is unrelated. I was just confused with the name of this task/log entry and thought I'd make it clearer.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -163,28 +163,28 @@ public TaskResult execute(@Nonnull Stage stage) {
             cluster.get("serverGroups"), new TypeReference<List<ServerGroup>>() {});
     serverGroups.sort(Comparator.comparing((ServerGroup o) -> o.createdTime).reversed());
 
+    List<ServerGroup> enabledServerGroups =
```
gal-yardeni: LGTM - so by filtering disabled server groups in `serverGroupsInRegion` , we are making sure to rollback only to an enable server group?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -163,28 +163,28 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+        serverGroups.stream()
+            .filter(s -> s.disabled == null || !s.disabled)
+            .collect(Collectors.toList());
+
     List<Map> imagesToRestore = new ArrayList<>();
     for (String region : stageData.regions) {
-      List<ServerGroup> serverGroupsInRegion =
-          serverGroups.stream()
+      List<ServerGroup> enabledServerGroupsInRegion =
+          enabledServerGroups.stream()
```
emjburns: I think you could also filter for not null here, and keep the rename to  `enabledServerGroupsInRegion`. It doesn't seem like you use the `enabledServerGroups` list anywhere else.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
luispollo: Yep, that's it!

luispollo:
> it seems good to me. but please double-check with red/black

Tested locally with monitored and red/black strategies.

luispollo:
@marchello2000 Before we merge this, I wanted to double check on potential side-effects. I realized the `RollbackClusterStage` is available for users to compose their pipelines, which means there are potentially lots of users who expected/relied on the existing behavior... How do we make sure we're not going to break a ton of pipelines for users inside and outside of Netflix? Is there a way to tell that the stage was added by the user versus the pipeline itself like in the case of deployment failure with auto-rollback?

luispollo:
> @marchello2000 Before we merge this, I wanted to double check on potential side-effects. I realized the `RollbackClusterStage` is available for users to compose their pipelines, which means there are potentially lots of users who expected/relied on the existing behavior... How do we make sure we're not going to break a ton of pipelines for users inside and outside of Netflix? Is there a way to tell that the stage was added by the user versus the pipeline itself like in the case of deployment failure with auto-rollback?

After discussing this with @marchello2000 offline, we agreed the right thing to do would be to keep the old behavior for a rollback stage defined in a user pipeline, and use new behavior only for automatic rollbacks initiated by deployment strategies, to avoid making any assumptions about how users might be using the stage externally.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
luispollo: I'm not sure I follow... Filter on not-null on which property, exactly?

emjburns: Ah sorry, filter out the disabled server groups!

luispollo: This is a different list, though -- the filter here is to take only the enabled server groups in one region.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
luispollo: Done!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
emjburns: never mind!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
```diff
@@ -121,6 +121,11 @@ class CreateServerGroupStage extends AbstractDeployStrategyStage {
       additionalRollbackContext.disableOnly = true
     }
 
+    // When initiating a rollback automatically as part of deployment failure handling, only rollback to a server
+    // group that's enabled, as any disabled ones, even if newer, were likely manually marked so for being "bad"
+    // (e.g. as part of a manual rollback).
+    additionalRollbackContext.disregardDisabledCandidates = true
```
luispollo: This is to account for simpler deployment strategies like `RedBlackStrategy`, which don't plan their own rollback stage.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -165,27 +165,41 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+            moniker.get().getCluster(),
+            region);
         continue;
       }
 
+      ServerGroup newestEnabledServerGroupInRegion = enabledServerGroupsInRegion.get(0);
+      boolean disregardDisabledCandidates =
+          stageData.additionalRollbackContext != null
+              && stageData.additionalRollbackContext.containsKey("disregardDisabledCandidates")
+              && ((Boolean) stageData.additionalRollbackContext.get("disregardDisabledCandidates"))
```
marchello2000: i think you can probably do:
`&& ((Boolean) stageData.additionalRollbackContext.getOrDefault("disregardDisabledCandidates", false))`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
marchello2000: nit: Personally, I prefer something like this:
```suggestion
    additionalRollbackContext.onlyEnabledServerGroups = true
```

(or `excludeDisabledServerGroups`)

marchello2000: findImageFromCluster already uses `onlyEnabled` but maybe that's too ambiguous in this context

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -165,27 +165,41 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+        // no enabled server groups in this region, nothing to rollback from!
+        logger.warn(
+            "No enabled server groups in cluster {} and region {} to rollback from. Skipping this region.",
+            moniker.get().getCluster(),
+            region);
         continue;
       }
 
+      ServerGroup newestEnabledServerGroupInRegion = enabledServerGroupsInRegion.get(0);
+      boolean disregardDisabledCandidates =
```
marchello2000: Another nit here: I find it easier to use positive booleans, e.g. `includeDisabled` or `onlyEnabled`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
luispollo: Yeah, this context switching between Java and Kotlin is getting to me! 😄

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
luispollo: I went with "disabled" because that's what clouddriver uses in the response:
https://github.com/spinnaker/orca/blob/688bd3d1af8d366f425059c9be89839ffe6a71f3/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java#L184

Will change to `excludeDisabledServerGroups`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
luispollo: Me too, except "include" should be the default in this case when nothing is added to the context...

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/CreateServerGroupStage.groovy
luispollo: Never mind, will take your suggestion.

PR #3344, opened 2019-12-14
feat(managed-delivery): Introduce stage to publish Delivery Config manifests to keel

luispollo:
In conjunction with https://github.com/spinnaker/igor/pull/573, this PR introduces a new pipeline stage which pulls a Managed Delivery config manifest from a source control repository specified by the user (via `igor`), and publishes that manifest to the `keel` service.

The high-level flow for the stage would be:
1. Pipeline is triggered upon changes to SCM repo. The stage will have information about where to find the Delivery Config manifest in the repo.
2. `orca` calls `igor` to retrieve the Delivery Config manifest and publishes it to `keel`

This flow is our solution at the moment for _GitOps_ workflows with Managed Delivery. At Netflix, this is intended to replace the existing "pre-configured job" stage which achieves more or less the same thing, but was written to publish individual resources (which is no longer supported), as opposed to delivery config manifests.

Another step towards closing https://github.com/spinnaker/keel/issues/608 (deck is the last piece)


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/BuildService.java
```diff
@@ -56,12 +56,4 @@ public Map queuedBuild(String master, String item) {
       Integer buildNumber, String fileName, String master, String job) {
     return igorService.getArtifacts(buildNumber, fileName, master, encode(job));
   }
-
-  public List compareCommits(
```
luispollo: I moved this to the new `ScmController` because, well, it doesn't have anything to do with builds. 😃

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/ScmService.java
```diff
@@ -0,0 +1,25 @@
...
+
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+import java.util.Map;
+
+@Component
+@RequiredArgsConstructor
+public class ScmService {
```
luispollo: Turns out this ended up being just a thin wrapper around `igorService`. Not sure it's worth it, but I kept it to follow the pattern of `BuildService`.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
```diff
@@ -0,0 +1,131 @@
...
+@Autowired
+constructor(
+  private val keelService: KeelService,
+  private val keelObjectMapper: ObjectMapper,
+  private val scmService: ScmService
+) : RetryableTask {
+  private val log = LoggerFactory.getLogger(javaClass)
+
+  override fun execute(stage: Stage): TaskResult {
+//      parameters:
```
luispollo: These are just here temporarily as a reference for what we currently have in the pre-configured job stage. If it's not obvious, I've favored convention in terms of where to look for the manifests, and have also replaced the repo URL with discrete parameters for the SCM system type, project key and repo, which seems to be the pattern used elsewhere.

The `ref` parameter is intended to contain a git reference, which would allow to specify any branch or commit, but as noted below I believe that information should come from the trigger, not the user.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
```diff
@@ -0,0 +1,131 @@
...
+//        defaultValue: ${trigger.hash}
+//        label: SHA
+//        order: 5
+
+    val context = keelObjectMapper.convertValue<PublishKeelSpecsContext>(stage.context)
+    val manifests = scmService.listKeelManifests(
+      context.scmType, context.projectKey, context.repositorySlug, context.ref)
+
+    var failedManifest: String? = null
+    for (manifest in manifests) {
```
luispollo: TODO: this block could be run in parallel for each manifest.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/IgorService.java
```diff
@@ -85,4 +85,19 @@ GoogleCloudBuild runGoogleCloudBuildTrigger(
       @Path("account") String account,
       @Path("triggerId") String triggerId,
       @Body GoogleCloudBuildRepoSource repoSource);
+
+  @GET("/keel/{scmType}/{projectKey}/{repositorySlug}/manifests")
```
emjburns: commented on the igor pr, but also commenting here: I don't think these endpoints should be prefixed with `keel`. They should be descriptive of what they do not who they're for

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/ScmService.java
emjburns: Hmm. I'd say, probably not worth it if you're not changing the api calls at all. I'd just wire in the `igorService` instead of creating this file.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
```diff
@@ -0,0 +1,131 @@
...
+//        label: Branch
+//        order: 4
+//      - name: gitSha
+//        mapping: cluster.env.GIT_SHA
+//        defaultValue: ${trigger.hash}
+//        label: SHA
+//        order: 5
+
+    val context = keelObjectMapper.convertValue<PublishKeelSpecsContext>(stage.context)
+    val manifests = scmService.listKeelManifests(
```
emjburns: We should have some retries around this

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
emjburns: I think that you're right, the commit should come from the trigger. But, if there is no trigger, we should default to master/the latest commit. The hash can be supplied by the normal git trigger (not the rocket trigger, which is internal to netflix). In this case it makes sense to consciously support both at first!

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
```diff
@@ -0,0 +1,131 @@
...
+
+  private fun Response.successful() =
+    this.status in 200..299
+
+  data class PublishKeelSpecsContext(
+    val scmType: String,
+    val projectKey: String,
+    val repositorySlug: String,
+    val ref: String? = "refs/heads/master"
+    // TODO: forceRepublish -- what does it do?
```
emjburns: For the current stage we only publish manifests if the last commit contains changes to the directory they're in. If force republish is true, we always re-publish manifests.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
emjburns: We could delete the resources if some fail?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/IgorService.java
luispollo: I struggled with naming for this. I thought of using `managed-delivery` but it occurred to me that the external feature name we use might change, whereas the service name was likely to stay forever, hence the current name.

Any suggestions for a better name? Perhaps remove `/keel` and just rename `/manifests` to `md-manifests` at the end?

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
luispollo: Absolutely. This was just a draft to see if the overall flow of things made sense. Will add that to the TO-DO list.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
luispollo: Got it! Thanks for explaining.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
luispollo: We could, but then we need to add extra logic to figure out if the resources existed in keel prior to publishing from here (in which case I don't think we can delete them). Do you feel like that next level of error handling could be part of a separate PR, perhaps?

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
luispollo: Absolutely. This was just a draft to see if the overall flow of things made sense. Will add that to the TO-DO list.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/IgorService.java
gal-yardeni: I agree with Emily - why we should limit the scope only for md/keel?

luispollo: To summarize our discussion in person:
- My concern was to avoid exposing an API that allows any caller on any Spinnaker instance the ability to  retrieve any file from any source control repo using igor's credentials. This was noted [here](https://github.com/spinnaker/igor/blob/6660b4b081ead1cdb69d5f76e00926a42eb8a82b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy#L58-L59).
- @emjburns's comments were more around exposing the name `keel` in the API, not so much restricting what source files the API exposes
- I will make changes to the API names as discussed but retain some level of restriction as to what kinds of directories and files can be accessed for the reason above

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
emjburns: Yes definitely

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
```diff
@@ -0,0 +1,131 @@
...
+      try {
+        resourceSpec = keelObjectMapper.convertValue(manifestContents)
+      } catch (e: java.lang.Exception) {
+        throw IllegalArgumentException("Unable to parse resource spec at $uri")
+      }
+
+      // TODO: confirm this is the right thing to do
+      val user = stage.execution.trigger.user ?: "anonymous"
+      log.info("Publishing manifest $manifest to keel on behalf of $user")
+      val response = keelService.publishSpec(resourceSpec, user)
```
emjburns: let's assume for now that the only flow is this:
- you have one file
- it must be called `spinnaker.yml`
- it it is published to the delivery config endpoint, not the resource endpoint

This is a change based on https://github.com/spinnaker/keel/issues/660

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishKeelSpecsTask.kt
luispollo: Not applicable anymore since we're refactoring this to publish delivery configs, not resources.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
```diff
@@ -0,0 +1,157 @@
...
+
+    if (context.attempt > context.maxRetries!!) {
+      val error = "Maximum number of retries exceeded (${context.maxRetries})"
+      log.error("$error. Aborting.")
+      return TaskResult.builder(ExecutionStatus.TERMINAL).context(mapOf("error" to error)).build()
+    }
+
+    // if the pipeline has a git trigger, infer what context we can from the trigger
+    if (trigger is GitTrigger) {
+      // FIXME: GitTrigger props are non-null, but IIUC branch and hash would be mutually exclusive, so what to do here?
```
luispollo: Since the hash trumps the branch, for now I'm prioritizing the hash. Please let me know if you understand this differently.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
```diff
@@ -0,0 +1,157 @@
...
+      }
+    } catch (e: Exception) {
+      log.error("Unexpected exception while executing {}, aborting.", javaClass.simpleName, e)
+      buildError(e.message)
+    }
+  }
+
+  override fun getBackoffPeriod() = TimeUnit.SECONDS.toMillis(30)
+
+  override fun getTimeout() = TimeUnit.SECONDS.toMillis(180)
```
luispollo: I didn't think these (or `MAX_RETRIES`) needed to be configurable for now, but let me know.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/ScmService.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class ScmService {
```
emjburns: If this class is just a wrapper around the igor service (i.e. function calls are just passthroughs to igor), do you think it's worth it to have this class? Are you planning on adding more complex logic here in the future?

luispollo: Like I mentioned earlier, this started as I looked into the existing implementation of `BuildService` (which is also just a pass-through), but seemed to exist to abstract the fact that it's igor under the covers into a more human-friendly, purpose-oriented class name. My thinking was that, if I was to just use `IgorService`, then to be consistent we should change everywhere that's using `BuildService`, and I felt that was way beyond the scope of this PR and didn't actually buy us anything.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
```diff
@@ -0,0 +1,157 @@
...
+    } else {
+      if (context.ref == null) {
+        context.ref = "refs/heads/master"
+      }
+      if (context.scmType == null || context.project == null || context.repository == null) {
+        throw IllegalArgumentException("scmType, project and repository are required fields in the stage if there's no git trigger.")
+      }
+    }
+
+    val manifestLocation = "${context.scmType}://${context.project}/${context.repository}/.netflix/${context.directory
```
emjburns: `.netflix` shouldn't be hardcoded here, it should be a property value that defaults to `.spinnaker` that we can override internally

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
```diff
@@ -0,0 +1,157 @@
...
+      ?: ""}/${context.manifest}@${context.ref}"
+
+    return try {
+      log.info("Retrieving keel manifest at $manifestLocation")
```
emjburns: debug logs are probably better

emjburns: Also, do you know if these logs also log execution and stage id along with this message? if not, it should, so that we can easily follow all these logs when troubleshooting

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
```diff
@@ -0,0 +1,157 @@
...
+      val deliveryConfig = scmService.getDeliveryConfigManifest(
+        context.scmType, context.project, context.repository, context.directory, context.manifest, context.ref)
+
+      log.info("Publishing manifest ${context.manifest} to keel on behalf of $user")
+      keelService.publishDeliveryConfig(deliveryConfig, user)
```
emjburns: There should probably be retries here

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
```diff
@@ -0,0 +1,157 @@
...
+  data class PublishDeliveryConfigContext(
+    var scmType: String? = null,
+    var project: String? = null,
+    var repository: String? = null,
+    var directory: String? = null, // as in, the directory *under* whatever manifest base path is configured in igor (e.g. ".netflix")
+    var manifest: String? = "spinnaker.yml",
+    var ref: String? = null,
+    var attempt: Int = 1,
+    val maxRetries: Int? = MAX_RETRIES
+//    val forceRepublish: Boolean? = false // TODO
```
emjburns: I think we probably don't need this, it's not a big deal to always republish ( i have found in practice i always want to republish)

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
luispollo: AFAIK, yes:
https://github.com/spinnaker/orca/blob/cd6b6cad218757a735e4b16d983127c2bc345a58/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/audit/ExecutionTrackingMessageHandlerPostProcessor.kt#L48-L52

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/IgorService.java
```diff
@@ -85,4 +86,13 @@ GoogleCloudBuild runGoogleCloudBuildTrigger(
       @Path("account") String account,
       @Path("triggerId") String triggerId,
       @Body GoogleCloudBuildRepoSource repoSource);
+
+  @GET("/delivery-config/manifest")
+  Map<String, Object> getDeliveryConfigManifest(
+      @Query("scmType") String scmType,
+      @Query("project") String project,
+      @Query("repository") String repository,
```
gal-yardeni: Not sure if we want to be consistent here, but there are some naming conventions that other services are using them already, like repoSlug instead of repository. You can view it here https://boost.test.netflix.net/swagger-ui.html#/builds-controller/findBuildsUsingGET

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/ScmService.java
```diff
@@ -0,0 +1,32 @@
...
+  public List compareCommits(
+      String repoType,
+      String projectKey,
+      String repositorySlug,
```
gal-yardeni: Inconsistent param names - here its repositorySlug, above its repository

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
luispollo: I can change the logs to debug, although I'm surprised that's turned on in prod in our deployment...

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/KeelService.kt
```diff
@@ -16,22 +16,13 @@
...
 interface KeelService {
-  @GET("/v1/intents/")
-  fun getIntents(): Response
-
-  @POST("/v1/intents/")
-  fun upsertIntents(@Body upsertIntentRequest: UpsertIntentRequest): Response
-
-  @DELETE("/v1/intents/{id}")
-  fun deleteIntents(@Path("id") id: String, @Query("status") status: String?): Response
+  @POST("/delivery-configs/")
```
gal-yardeni: I have a general question - what happens when there's an existing delivery config for an application? will it be replaced/ignored?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/IgorService.java
luispollo: That's a good question. I noticed that and, in fact, had originally named everything following that convention. Then I realized that "slug" is an [Atlassian term](https://confluence.atlassian.com/bitbucket/what-is-a-slug-224395839.html), apparently, whereas this API is supposed to be SCM-type agnostic. I don't have a super strong opinion, although at this point igor has been merged with the different names as well (except in the `StashClient`, which is Stash-specific).

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/ScmService.java
luispollo: This is an example of what I was saying in my previous comment. This API was moved from `BuildService`, I didn't actually write it, and it used the Atlassian-ish names. Let's see what we agree on, then I'll change one or the other accordingly.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/KeelService.kt
luispollo: It gets updated:
https://github.com/spinnaker/keel/blob/cb6c5f8c977a61e67ecdccc627d154b92c4cbb6d/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/DeliveryConfigController.kt#L39-L44

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/IgorService.java
gal-yardeni: Hmm interesting. I vote to keep it consistent with boost, as we might use your code in the future if we'll want to add a ci stage for example. any thoughts @emjburns ?

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
luispollo: @emjburns I was thinking about that and I think the right way to do this would be for `igor` to expose this config via an API, instead of duplicating it here, since it's truly only used for logging here.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
```diff
@@ -0,0 +1,156 @@
...
+
+    return try {
+      log.debug("Retrieving keel manifest at $manifestLocation")
+      val deliveryConfig = scmService.getDeliveryConfigManifest(
+        context.scmType, context.project, context.repository, context.directory, context.manifest, context.ref)
+
+      log.debug("Publishing manifest ${context.manifest} to keel on behalf of $user")
+      keelService.publishDeliveryConfig(deliveryConfig, user)
+
+      TaskResult.builder(ExecutionStatus.SUCCEEDED).context(emptyMap<String, Any?>()).build()
```
luispollo: This is really the meat of the task... It's fairly simple: pull the delivery config from igor, push to keel.

luispollo:
Thanks @gal-yardeni. I've renamed the variables/arguments as suggested.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/PublishDeliveryConfigTask.kt
```diff
@@ -0,0 +1,172 @@
...
+    } catch (e: RetrofitError) {
+      handleFailure(e, context)
+    } catch (e: Exception) {
+      log.error("Unexpected exception while executing {}, aborting.", javaClass.simpleName, e)
+      buildError(e.message)
+    }
+  }
+
+  private fun processDeliveryConfigLocation(trigger: Trigger, context: PublishDeliveryConfigContext): String {
+    if (trigger is GitTrigger) {
```
luispollo: This section basically just looks at the git trigger, if present, and attempts to fill in any blanks in the stage context not filled by the user and that can be inferred from the trigger (namely the `ref`, `repoType`, `projectKey` and `repositorySlug`, all of which are found in the trigger). If a corresponding field is present in the context, the context data wins (hence the `context.field == null` checks).

PR #3349, opened 2019-12-20
feat(queue): use keiko 3.1.0 with eventless metrics

dreynaud:
The queue event listener and queue state poller are now native in keiko, and the events no longer go through Spring.

AtlasQueueMonitor used to do listen to queue events, poll the queue state and check for zombies. It is now only responsible for the zombie check, so let's rename it accordingly to ZombiePipelineCheckingAgent.

Since it looks more like an actual agent now, I think it makes sense to make it extend AbstractPollingNotificationAgent like the other agents, so it doesn't need to do its own scheduling and locking anymore. Which is why we can remove that test case from ZombiePipelineCheckingAgentTest


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/ZombiePipelineCheckingAgent.kt
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Optional
+
+/**
+ * Monitors a queue and generates Atlas metrics.
+ */
+@Component
+@ConditionalOnExpression(
+  "\${queue.zombie-check.enabled:false}")
+@ConditionalOnBean(MonitorableQueue::class)
+class ZombiePipelineCheckingAgent
```
robzienert: I like this change.

nit: Since we are checking both `PIPELINE` and `ORCHESTRATION`, what about renaming the agent to something like `ZombieExecutionCheckingAgent`?

PR #3352, opened 2020-01-06
refactor(managed-delivery): Rename publishDeliveryConfig to importDeliveryConfig

luispollo:
As discussed in https://github.com/spinnaker/deck/pull/7733, this renames the `publishDeliveryConfig` stage and associated classes to `importDeliveryConfig` for clarity, since this concept is exposed to users and from their perspective we're not publishing the delivery config anywhere, but rather pulling (i.e. importing) it into Spinnaker.

### DIFF
--- a/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/pipeline/ImportDeliveryConfigStage.kt
+++ b/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/pipeline/ImportDeliveryConfigStage.kt
@@ -17,8 +17,9 @@
 package com.netflix.spinnaker.orca.keel.pipeline
 
 import com.netflix.spinnaker.orca.Task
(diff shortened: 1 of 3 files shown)


orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/pipeline/ImportDeliveryConfigStage.kt
```diff
@@ -29,9 +30,10 @@ import kotlin.reflect.KClass
  * Generally this will be added to a single-stage pipeline with a git trigger to support GitOps flows.
  */
 @Component
-class PublishDeliveryConfigStage : StageDefinitionBuilder {
+@Aliases("publishDeliveryConfig")
```
luispollo: Leaving this in here as an alias in case deck gets merged first...

PR #3356, opened 2020-01-08
feat(plugins): Add upsert and delete plugin artifact tasks

jonsie:
Closes https://github.com/spinnaker/spinnaker/issues/5288


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/model/PluginArtifact.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.front50.model;
+
+import java.util.List;
+import java.util.regex.Pattern;
+import javax.annotation.Nonnull;
+import lombok.Data;
+
```
srekapalli: Can we move these model/DTO objects to kork-plugins-api ?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/UpsertPluginArtifactStage.java
```diff
@@ -0,0 +1,32 @@
...
+
+package com.netflix.spinnaker.orca.front50.pipeline;
+
+import com.netflix.spinnaker.orca.front50.tasks.UpsertPluginArtifactTask;
+import com.netflix.spinnaker.orca.pipeline.StageDefinitionBuilder;
+import com.netflix.spinnaker.orca.pipeline.TaskNode;
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import javax.annotation.Nonnull;
+import org.springframework.stereotype.Component;
+
```
srekapalli: Tests around this new stage will be nice.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/model/PluginArtifact.java
jonsie: I think that's a broader issue - we do this type of model duplication across many services in Spinnaker.  I don't think we have a great story around sharing these.  I don't think I want `kork-plugins-api` to be a dependency in `orca-front50` since `kork-plugins-api` is really the library for writing plugins/extension points, and I'm not sure I want to break away from current patterns at the moment.

robzienert: `kork-plugins-api` is for the API contract between actual plugins and Spinnaker. Actual plugins won't need these models.

robzienert: If we're thinking about shared models, I think that needs a separate discussion, as there's a lot of paths forward here and we'd want consensus on direction before doing anything.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/model/PluginArtifact.java
```diff
@@ -0,0 +1,59 @@
...
+  private String lastModifiedBy;
+
+  @Data
+  public static class Release {
+    public static final Pattern SUPPORTS_PATTERN =
+        Pattern.compile(
+            "^(?<service>[\\w\\-]+)(?<operator>[><=]{1,2})(?<version>[0-9]+\\.[0-9]+\\.[0-9]+)$");
+    public static final String SUPPORTS_PATTERN_SERVICE_GROUP = "service";
+    public static final String SUPPORTS_PATTERN_OPERATOR_GROUP = "operator";
+    public static final String SUPPORTS_PATTERN_VERSION_GROUP = "version";
```
robzienert: I don't think we need this. Orca shouldn't be doing validation of models that are owned by other services, so it won't need to do any of this parsing.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/model/PluginArtifact.java
```diff
@@ -0,0 +1,59 @@
...
+@Data
+@Builder
+public class PluginArtifact {
+  @Nonnull private String id;
+  private String description;
+  private String provider;
+  @Nonnull private List<Release> releases;
+  private Long createTs;
+  private Long lastModified;
+  private String lastModifiedBy;
```
robzienert: 1. `createTs` - Is this intended to be when the artifact was published? If that's the case, then it should be something like `publishedAt`.
1. All 3 of these fields are private as far as front50 is concerned and the values cannot be affected by outside data - I don't think we need them.
1. The codebase is a huge mixed bag on this because legacy, but our modern preference is to represent all timestamps as `Instant`.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeletePluginArtifactTask.java
```diff
@@ -0,0 +1,77 @@
...
+
+    log.debug("Deleting front50 plugin artifact `{}`", pluginArtifactId);
+    Response response = front50Service.deletePluginArtifact(pluginArtifactId);
+
+    if (response.getStatus() != HttpStatus.NO_CONTENT.value()) {
+      log.warn(
+          "Failed to delete `{}`, received unexpected response status `{}`",
+          pluginArtifactId,
+          response.getStatus());
+      return TaskResult.ofStatus(ExecutionStatus.TERMINAL);
```
robzienert: This task should succeed if the status is 404.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/model/PluginArtifact.java
jonsie: I wanted these fields so that when we get the PluginArtifact back from front50 after the upsert we can add the artifact to the task outputs and see the exact PluginArtifact that was upserted.  Of course, I'm not expecting us to pass these fields to front50 when we do the upsert.  Maybe not necessary?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/DeletePluginArtifactTask.java
jonsie: Fair dues.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/model/PluginArtifact.java
jonsie: Oh yea, I think I probably deleted this right as you left that message - it's gone now.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/model/PluginArtifact.java
srekapalli: Not sure if all the audit type fields that are private to Front50 are required to be in Orca. Wouldn't hurt but may be not necessary.

jonsie: Alrighty, I'll remove them - we can always add them back if we want to see this data in the task outputs.

PR #3357, opened 2020-01-08
feat(managed-delivery): Add support for other types of code triggers

luispollo:
This adds support for other types of source code triggers to the delivery config import stage, beyond just git, by introducing a common interface for those types of triggers.

Note: while testing this, I actually found out that, because for some reason we deserialize, then re-serialize, then deserialize the trigger and/or pipeline JSON when triggering pipelines, during that second deserialization pass my new code trigger type, which is private to Netflix and looks almost the same as a `GitTrigger`, ends up getting converted to a `GitTrigger` by `TriggerDeserializer`. This doesn't have any undesired side-effects, but I thought I'd mention it because I wasn't sure this was intended behavior (the back-and-forth serialization, that is).

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
@@ -31,11 +31,11 @@ import com.netflix.spinnaker.orca.pipeline.model.NexusTrigger
 import com.netflix.spinnaker.orca.pipeline.model.PipelineTrigger
 import com.netflix.spinnaker.orca.pipeline.model.Trigger
 
-internal class TriggerDeserializer :
+class TriggerDeserializer :
   StdDeserializer<Trigger>(Trigger::class.java) {
 
   companion object {
-    val customTriggerSuppliers: MutableList<CustomTriggerDeserializerSupplier> = mutableListOf()
+    val customTriggerSuppliers: MutableSet<CustomTriggerDeserializerSupplier> = mutableSetOf()
   }
 
   override fun deserialize(parser: JsonParser, context: DeserializationContext): Trigger =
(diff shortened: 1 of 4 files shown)


orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -152,7 +148,7 @@ constructor(
       "$message: ${cause?.message ?: ""}"
     }
 
-  data class PublishDeliveryConfigContext(
+  data class ImportDeliveryConfigContext(
```
luispollo: This is unrelated but was a left-over from renaming the classes before.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -61,23 +61,19 @@ constructor(
...
     }
   }
 
-  private fun processDeliveryConfigLocation(trigger: Trigger, context: PublishDeliveryConfigContext): String {
-    if (trigger is GitTrigger) {
-      // if the pipeline has a git trigger, infer what context we can from the trigger
-      // FIXME: GitTrigger props are non-null, but IIUC branch and hash would be mutually exclusive, so what to do here?
-      if (trigger.hash != null && context.ref == null) {
+  private fun processDeliveryConfigLocation(trigger: Trigger, context: ImportDeliveryConfigContext): String {
+    if (trigger is SourceCodeTrigger) {
```
luispollo: Because of the serialization back-and-forth behavior I mentioned in the description, in the end this wasn't strictly necessary, but I thought it conveyed more clearly that it doesn't _have_ to be a `GitTrigger`, specifically.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
```diff
@@ -31,11 +31,11 @@ import com.netflix.spinnaker.orca.pipeline.model.NexusTrigger
 import com.netflix.spinnaker.orca.pipeline.model.PipelineTrigger
 import com.netflix.spinnaker.orca.pipeline.model.Trigger
 
-internal class TriggerDeserializer :
+class TriggerDeserializer :
```
luispollo: This was needed so we could register custom deserializers into `customTriggerSuppliers` below.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
emjburns: I like

gal-yardeni:
LGTM!

PR #3359, opened 2020-01-10
fix(core): Remove useless entries from `Stage.context`

pdelagrave:
Fixes spinnaker/spinnaker#5257

The `Stage` constructor was keeping `refId`, `startTimeExpiry` and `requisiteStageRefIds` from the context map argument in its internal context map. This caused issues when a parent stage's context was used to create a child stage, as these entries aren't expected to be inherited. This wasn't a problem for a long time because these values were accidentally removed by the side effect of `StartStageHandler#shouldSkip`. `shouldSkip()` uses `objectMapper.convertValue()` on a stage context expecting to have a cloned map returned. A long standing design/documentation issue with Jackson made `convertValue()` return the input parameter as is instead of converting/cloning it. This made every `Stage` passing through `StartStageHandler.handle()` having their context map indirectly 'corrected' and persisted.
An upcoming upgrade to Spring Boot 2.2 comes with Jackson 2.10 where `objectMapper.convertValue()` has a new behaviour and would make `shouldSkip()` fail to silently 'correct' the stages context map.
https://github.com/FasterXML/jackson-databind/issues/2220

`BakeStageSpec` was validating `context['refId']` on the generated contexts but the contexts aren't Stages yet; once a context is made into a Stage and added to the graph (`graph.add`) it will have its `refId` generated based on the parent refId; which isn't under test in that spec.


pdelagrave:
@robzienert @robfletcher

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Stage.java
```diff
@@ -114,6 +113,8 @@ public Stage(Execution execution, String type, String name, Map<String, Object>
     this.requisiteStageRefIds =
         Optional.ofNullable((Collection<String>) context.remove("requisiteStageRefIds"))
             .orElse(emptySet());
+
```
marchello2000: this whole function seems suspect to me (i've looked into this before but ran into some issues trying to fix it - don't recall what they were)
the `context` this function receives as the param is passed by ref so we actually modify the context from the caller which is a bit messed up.
but AFAIR there were no calls to this that mattered so i sort of put it aside...
anyway, the change LGTM

robzienert: Since it is passed by reference, shouldn't a stage copy the context? Would we need to do a deep copy or would a copy of the map only be sufficient?

PR #3363, opened 2020-01-14
feat(core): stage to depend on an uncorrelated execution

asher:


asher:
I'm interested in using this with keel (managed delivery) in the context of staggering the regional deployment of multi-region cluster resources. In this scenario, the straight forward path is for keel to emit an execution or executions covering all regions when the initial resource diff has been detected.

This can be handled via a single orchestration chaining deploy and wait stages, but I currently think it would be advantageous to preserve keel's current behavior of one orchestration per resource and region. Launching all at once (but serially, as opposed to doing more sophisticated orchestration within keel) helps with treating the collection of tasks as an atomic unit where actuation remains in progress until all have completed.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/DependsOnExecutionStage.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.netflix.spinnaker.orca.pipeline;
+
+import com.netflix.spinnaker.orca.pipeline.model.Stage;
+import com.netflix.spinnaker.orca.pipeline.tasks.DependsOnExecutionTask;
+import javax.annotation.Nonnull;
+
+public class DependsOnExecutionStage implements StageDefinitionBuilder {
+
+  public static String STAGE_TYPE = "dependsOnExecution";
```
marchello2000: are we doing `STAGE_TYPE` or `PIPELINE_CONFIG_TYPE`?

Either way, probably needs to be `final`

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/DependsOnExecutionTask.java
```diff
@@ -0,0 +1,86 @@
...
+import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionNotFoundException;
+import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+import javax.validation.constraints.NotNull;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class DependsOnExecutionTask implements RetryableTask {
```
marchello2000: do you want to be able to override timeout in stage config?
```suggestion
public class DependsOnExecutionTask implements OverridableTimeoutRetryableTask {
```

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/DependsOnExecutionStage.java
asher: @marchello2000 which one should be used? I see both all over.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/DependsOnExecutionTask.java
asher: Yes, thanks!

PR #3365, opened 2020-01-15
chore(retrofit): Expose OK3Client using api dependency

evansb:
`orca-retrofit` configuration provides a bean of type `Ok3Client` which is useful to create new Retrofit client when extending Orca and retaining metrics, user agent, etc.

However, the class is defined `com.jakewharton.retrofit:retrofit1-okhttp3-client` and is not exposed as `api` dependency, so we'll have to add it ourselves (which is not ideal as this package is not tracked `orca-bom`). This PR fixes that.

### DIFF
--- a/orca-retrofit/orca-retrofit.gradle
+++ b/orca-retrofit/orca-retrofit.gradle
@@ -19,6 +19,7 @@ apply from: "$rootDir/gradle/groovy.gradle"
 dependencies {
   api("com.squareup.retrofit:retrofit")
   api("com.squareup.retrofit:converter-jackson")
+  api("com.jakewharton.retrofit:retrofit1-okhttp3-client:1.1.0")
 
   implementation(project(":orca-core"))
   implementation("com.squareup.okhttp:okhttp")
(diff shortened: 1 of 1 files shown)


orca-retrofit/orca-retrofit.gradle
```diff
@@ -19,14 +19,14 @@ apply from: "$rootDir/gradle/groovy.gradle"
 dependencies {
   api("com.squareup.retrofit:retrofit")
   api("com.squareup.retrofit:converter-jackson")
+  api("com.jakewharton.retrofit:retrofit1-okhttp3-client:1.1.0")
```
cfieber: orca-retrofit is depended on by a bunch of modules so I'd rather avoid adding things to its api classpath

You could add this version declaration to the orca-bom:
https://github.com/spinnaker/orca/blob/d123a1480c4922fa56734254c653c2b6b763a215/orca-bom/orca-bom.gradle#L41

keep  `implementation("com.jakewharton.retrofit:retrofit1-okhttp3-client")`  in here as an implementation dependency here, and in any extension module add an implementation dependency there as well

evansb:
apologise for the brief hiatus, I found out that this package is depended by many services https://github.com/search?q=org%3Aspinnaker+com.jakewharton.retrofit%3Aretrofit1-okhttp3-client%3A1.1.0&type=Code so I think adding in `kork-bom` makes more sense, will work towards that. Closing this PR.

PR #3370, opened 2020-01-15
fix(queue): Allow execution without a queue + polling agent cleanup

marchello2000:
There must've been a regression where the queue can no longer really be disabled. If it is,
orca will fail to start due to missing bean.
Adding a `NoopQueue` that is created when the queue bean is missing.

Also, removing Rx from `PollingNotificationAgent` and making it so that the first invocation starts
without a delay and all subsequent invocations are delayed from the prior run's completion time

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/notifications/AbstractPollingNotificationAgent.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/notifications/AbstractPollingNotificationAgent.java
@@ -17,27 +17,23 @@ package com.netflix.spinnaker.orca.notifications;
 
 import static com.netflix.appinfo.InstanceInfo.InstanceStatus.UP;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.netflix.spinnaker.kork.eureka.RemoteStatusChangedEvent;
-import java.util.concurrent.TimeUnit;
+import com.netflix.spinnaker.orca.NamedThreadFactory;
+import java.util.concurrent.*;
 import javax.annotation.PreDestroy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 import org.springframework.context.ApplicationListener;
-import rx.Observable;
-import rx.Scheduler;
-import rx.Subscription;
-import rx.schedulers.Schedulers;
 
 public abstract class AbstractPollingNotificationAgent
     implements ApplicationListener<RemoteStatusChangedEvent> {
 
   private final Logger log = LoggerFactory.getLogger(AbstractPollingNotificationAgent.class);
+  private ScheduledExecutorService executorService = null;
+  private ScheduledFuture agentRunFuture = null;
 
   protected final NotificationClusterLock clusterLock;
-  protected Scheduler scheduler = Schedulers.io();
-  protected Subscription subscription;
 
   public static final String AGENT_MDC_KEY = "agentClass";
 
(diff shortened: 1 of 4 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/NoopQueue.kt
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  override fun poll(callback: QueueCallback) {
+  }
+
+  override fun poll(maxMessages: Int, callback: QueueCallback) {
+  }
+
+  override fun push(message: Message, delay: TemporalAmount) {
+  }
```
marchello2000: i think i am going to  log an error here, what do you think?

orca-core/src/main/java/com/netflix/spinnaker/orca/notifications/AbstractPollingNotificationAgent.java
```diff
@@ -17,27 +17,23 @@
 
 import static com.netflix.appinfo.InstanceInfo.InstanceStatus.UP;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.netflix.spinnaker.kork.eureka.RemoteStatusChangedEvent;
-import java.util.concurrent.TimeUnit;
+import io.micrometer.core.instrument.util.NamedThreadFactory;
```
dreynaud: sounds like the wrong place to import this

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/NoopQueue.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.netflix.spinnaker.q.metrics.EventPublisher
+import com.netflix.spinnaker.q.metrics.MonitorableQueue
+import com.netflix.spinnaker.q.metrics.NoopEventPublisher
+import com.netflix.spinnaker.q.metrics.QueueState
+import org.slf4j.LoggerFactory
+import org.threeten.extra.Minutes
+import java.time.temporal.TemporalAmount
+
+/**
+ * A Noop queue do be used when no Queue bean is found (e.g. when Queue is disabled)
```
dreynaud:
```suggestion
 * A Noop queue to be used when no Queue bean is found (e.g. when Queue is disabled)
```

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/NoopQueue.kt
```diff
@@ -0,0 +1,65 @@
...
+ */
+class NoopQueue : MonitorableQueue {
+  override val publisher: EventPublisher = NoopEventPublisher()
+  private val log = LoggerFactory.getLogger(this.javaClass)
+
+  init {
+    log.error("${this.javaClass.simpleName} was created - all queue operations will be NOOP'd. This is OK if the queue was intended to be disabled")
```
dreynaud: it's not an error though, maybe warn instead?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/NoopQueue.kt
dreynaud: I'd be fine with doing nothing (as a noop queue arguably should), otherwise maybe log a warning

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/NoopQueue.kt
marchello2000: yeah, was debating back and forth with @robzienert im on the fence. The idea was to make is super obvious... what you think?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/NoopQueue.kt
marchello2000: i think i would feel better with a warn here... otherwise that original error/warn in .ctor is easy to miss

orca-core/src/main/java/com/netflix/spinnaker/orca/notifications/AbstractPollingNotificationAgent.java
marchello2000: what do you mean?

dreynaud: it seems weird to pull a util from micrometer

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/NoopQueue.kt
dreynaud: yeah I get it, I just don't like hijacking the error log level for visibility, I'd like to keep it for actual errors. This is a legit use case for log.warn, still plenty visible IMO

orca-core/src/main/java/com/netflix/spinnaker/orca/notifications/AbstractPollingNotificationAgent.java
marchello2000: in orca core... yeah. true. do you think it's better to duplicate that code (it's ~15 lines)

marchello2000: it is clearly gradle imported already though...

marchello2000: i copied it from river/cats

marchello2000:
> I like the changes but wouldn't it make sense for the NoopQueue to live in keiko instead?

yes, i guess i was lazy putting it into keiko. I will do that after this merges (since i  need soon)

PR #3375, opened 2020-01-17
refactor(artifacts): Rename ArtifactResolver to ArtifactUtils

ezimanyi:
* refactor(artifacts): Rename ArtifactResolver to ArtifactUtils 

  This class does a lot more than resolve artifacts, and an upcoming commit will pull out the resolving-specific functionality into its own, smaller class.  In order to leave the ArtifactResolver class name available for that new class, rename this to ArtifactUtils.

  In general, *Utils classes tend to get a lot of functionality that should live in better purpose-built classes so are often not a great idea. In this case, the class already has a bunch of functionality so the name accurately describes it; the upcoming commit will pull out one part of the functionality and hopefully future refactors will further pare down this class.

* refactor(artifacts): Convert find artifacts test to java 

  An upcoming commit will require some changes to this test; as it's easy to convert it to java (and makes the changes easier) just convert it to java here.


orca-web/src/test/groovy/com/netflix/spinnaker/orca/controllers/OperationsControllerSpec.groovy
```diff
@@ -482,9 +482,9 @@ class OperationsControllerSpec extends Specification {
       startedPipeline
     }
     executionRepository.retrievePipelinesForPipelineConfigId(*_) >> Observable.empty()
-    ArtifactResolver realArtifactResolver = new ArtifactResolver(mapper, executionRepository, new ContextParameterProcessor())
+    ArtifactUtils realArtifactUtils = new ArtifactUtils(mapper, executionRepository, new ContextParameterProcessor())
 
-    // can't use @subject, since we need to test the behavior of otherwise mocked-out 'artifactResolver'
+    // can't use @subject, since we need to test the behavior of otherwise mocked-out 'artifactUtils'
```
maggieneterval: A+ for attention to detail and finding this comment #responsibly

PR #3376, opened 2020-01-17
fix(core): fix issue planning non-linear orchestration executions

asher:


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartStageHandler.kt
```diff
@@ -203,7 +203,7 @@ class StartStageHandler(
       return false
     }
 
-    val clonedContext = objectMapper.convertValue(this.context, Map::class.java) as Map<String, Any>
+    val clonedContext: MutableMap<String, Any> = this.context.toMutableMap()
     val clonedStage = Stage(this.execution, this.type, clonedContext).also {
```
asher: Because `clonedContext` was really `this.context`, this had the effect of mutating the current stage context via the `Stage()` constructor call in a way that's critical for some functionality, notably stripping `requisiteStageRefids` and `refId`.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
```diff
@@ -64,6 +64,10 @@ internal interface OrcaMessageHandler<M : Message> : MessageHandler<M> {
       try {
         execution
           .stageById(stageId)
+          .also {
+            // Mutates it.context in a required way for some non-linear stage features
```
luispollo: I wasn't sure what the "required way" meant here. Maybe worth expanding the comment a little bit?

PR #3377, opened 2020-01-18
refactor(artifacts): Extract artifact resolution into ArtifactResolver class

ezimanyi:
* refactor(artifacts): Add ArtifactResolver class 

  The ArtifactUtils class handles a lot of functionality and is difficult to understand. Simplify this by creating an ArtifactResolver class to handle resolving artifacts, so that functionality can be removed from ArtifactUtils.

  One important change in this class is that it does not mutate any of the Artifact or ExpectedArtifact instances it receives; it returns the results of its work (containing new instances) which the caller should then use as appropriate. This will help reduce bugs where the caller is depending on a mutation happening several calls below the call stack, and brings us closer to making Artifact and ExpectedArtifact immutable.

* refactor(artifacts): Update callers to use ArtifactResolver 

  This commit updates all artifact resolving code to use the new ArtifactResolver class.

  A few things of note in the changes:
  * FindArtifactFromExecutionTaskTest was mocking resolveExpectedArtifacts to return resolvedArtifacts; this was subtly different than the actual function because the actual resolvedArtifacts function mutated the input expectedArtifacts to add a boundArtifact to each. As the resolveExpectedArtifacts method has been moved to a simpler class that we don't need to mock, the test is calling the actual method now and so we need to update the expected result to account for the fact that the resolvedExpectedArtifacts now have a bound artifact.
  * All of the tests deleted from ArtifactUtilsSpec have equivalent new tests in ArtifactResolverTest which test the new implementation of artifact resolving.
  * The one changed test in ArtifactUtilsSpec (...sets the bound artifact...) is a result of fixing what was a bug in the prior code. As the prior code was mutating each ExpectedArtifact, we were mutating the expected artifacts on the pipeline itself (and testing that). resolveArtifacts is storing the resolved artifacts in the trigger, under resolvedArtifacts; this is where downstream stages look to find bound artifacts and where we should be testing that the bound artifact is set. The test now checks that the bound artifacts are set in the trigger.

* refactor(artifacts): Clean up some stream code 

  The code to map input artifacts in resolveArtifacts() is more complex and verbose than necessary.  Simplify it by moving the .orElse before the map operation (which means we don't need to have a nested stream) and also inline a two streams that are used only once.

* perf(artifacts): Lazily compute past pipeline artifacts 

  The resolveArtifacts function always passes the prior artifacts to the ArtifactResolver (except when it short-circuits in the event of no expected artifacts). Looking up artifacts from prior executions can be expensive, so ideally we'd defer looking these up until they are actually needed.

  This change updates the constructor for ArtifactResolver to accept a Supplier of prior artifacts; it will memoize this supplier and only invoke it if the resolution requires looking up prior artifacts
  (that is, the first time it encounters an expected artifact with usePriorAritfact set to true that does not match one of the current artifacts).

  This will also allow us in the next commit to remove the short-circuiting for the case where there are no expected artifacts without worrying about negatively affecting performance.

* refactor(artifacts): A few minor refactors of resolveArtifacts 

  Replace the List<ExpectedArtifact> with an ImmutableList<> as we never modify it. (And there's a tiny performance benefit that ArtifactResolver.getInstance should short-circuit the copy.)

  Replace .distinct() collecting to a List<> of receivedArtifacts with directly collecting to toImmutableSet() as it will remove duplicates and return a set whose iteration order is the encounter order in the stream.

  Remove the short-circuit that returns early if there are no expected artifacts; the ArtifactResolver is a no-op that returns empty lists for resolvedArtifacts and resolvedExpectedArtifacts in that case, so there is no need to special-case it.

  Replace the LinkedHashSet code that creates a deduplicated list of artifacts in the supplied order with an ImmutableSet that does the same.


ezimanyi:
After this PR, `ExpectedArtifact` will be effectively immutable; that is, none of its setters are actually called, so we can remove them and make all of its fields final.  (To be precise, it is shallowly-immutable; it still has a couple of fields that are `Artifact` objects which are themselves mutable.)

This should also eliminate the class of bug that was worked around in 688e1f9612f9f52ab3c38d85f4feae889804c7a9; in that bug we were mutating an `ExpectedArtifact` and expecting the changes to be visible several levels in the call stack above the change.  This behavior was actually depending on Jackson short-circuiting a call to `objectMapper.convertValue` to return the input value by reference; a Jackson upgrade changed that so we were operating on a copy and the top-level function no longer saw the mutation.  Now artifact resolution does not make any mutating changes to artifacts or expected artifacts and communicates all output via its return value.  (CC @pdelagrave who figured out this initial bug and had the original fix for it before the upgrade was rolled back for other reasons.)

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -0,0 +1,184 @@
...
+   * @param expectedArtifacts The expected artifacts to resolve
+   * @return The result of the artifact resolution
+   */
+  public ResolveResult resolveExpectedArtifacts(
+      @Nullable Iterable<ExpectedArtifact> expectedArtifacts) {
+    if (expectedArtifacts == null) {
+      expectedArtifacts = ImmutableList.of();
+    }
+    // We keep track of resolved artifacts in an ImmutableSet.Builder so that duplicates are not
+    // added (in the case that an artifact matches more than on expected artifact). An ImmutableSet
```
maggieneterval:
```suggestion
    // added (in the case that an artifact matches more than one expected artifact). An ImmutableSet
```

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -0,0 +1,184 @@
...
+  private final boolean requireUniqueMatches;
+
+  private ArtifactResolver(
+      @Nullable Iterable<Artifact> currentArtifacts,
+      @Nullable Supplier<? extends Iterable<Artifact>> priorArtifacts,
+      boolean requireUniqueMatches) {
+    this.currentArtifacts =
+        currentArtifacts == null ? ImmutableList.of() : ImmutableList.copyOf(currentArtifacts);
+    this.priorArtifacts =
+        Suppliers.memoize(
```
maggieneterval: :man_cartwheeling: This is so cool!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -0,0 +1,175 @@
...
+   * artifacts, and prior artifacts.
+   *
+   * @param currentArtifacts The current artifacts to consider when resolving expected artifacts
+   * @param priorArtifacts The prior artifacts to consider when resolving expected artifacts
+   * @param requireUniqueMatches Whether the resolver should require that each expected artifact
+   *     matches at most one artifact.
+   * @return An instance of {@link ArtifactResolver}
+   */
+  public static ArtifactResolver getInstance(
+      @Nullable Iterable<Artifact> currentArtifacts,
```
plumpy: This is personal preference, so feel free to ignore it, but I think `@Nullable` parameters should be used sparingly, especially for collection parameters. `currentArtifacts` is never null (in production code anyway) and clients could just pass `ImmutableSet.of()` instead of `null` to `priorArtifacts`...  (Or just add an overload that removes the `priorArtifacts` parameter entirely, which is arguably a better move anyway)

To me, nullable types have pretty serious downsides, so I try to avoid passing across API boundaries unless there's a compelling reason, but I don't really see one here...

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -0,0 +1,175 @@
...
+   * <p>If an expected artifact does not match any artifacts, an {@link InvalidRequestException} is
+   * thrown.
+   *
+   * <p>If {@link #requireUniqueMatches} is true, and an expected artifact matches more than one
+   * artifact in any of the above steps, an {@link InvalidRequestException} is thrown.
+   *
+   * @param expectedArtifacts The expected artifacts to resolve
+   * @return The result of the artifact resolution
+   */
+  public ResolveResult resolveExpectedArtifacts(
```
plumpy: very nitpicky, and let me know if this isn't how Spinnaker does it, but the _most common_ way to order Java methods is to put your public methods first, and your private methods close to the public method that uses them, like this:
* public void foo()
* private void calledByFoo()
* private void calledByCalledByFoo()
* public void bar()
* private void usedByBar()
* private void usedBySeveralPublicMethods()

This is completely backwards of idiomatic C++ style, I think.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
```diff
@@ -64,11 +64,12 @@ public TaskResult execute(@Nonnull Stage stage) {
           artifactUtils.getArtifactsForPipelineId(pipeline, executionOptions.toCriteria());
     }
 
-    Set<Artifact> matchingArtifacts =
-        artifactUtils.resolveExpectedArtifacts(expectedArtifacts, priorArtifacts, null, false);
+    ArtifactResolver.ResolveResult resolveResult =
+        ArtifactResolver.getInstance(priorArtifacts, null, false)
```
plumpy: more nitpicky stuff, but I think it's helpful to add a parameter comment to constant parameters... go/java-practices/comments#arguments

```
ArtifactResolver.getInstance(
    priorArtifacts, /* priorArtifacts= */ null, /* requireUniqueMatches= */ false);
```

That way readers can see what's going on without having to click through to `#getInstance`. You can add these comments in IntelliJ with `ctrl-alt-/` (this is an internal extension, doesn't work outside Google)

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
maggieneterval: This :1st_place_medal: is going straight into my Java bible

maggieneterval: (Which is a google doc of stuff you and Eric have commented on PRs)

ezimanyi: Yeah, I don't think that Spinnaker has a particular well-defined order so definitely will follow this order instead of the random order I've been using.

Does that order include constructors?  So should I move above `public static ArtifactResolver getInstance` above the private constructor?  (I presume yes from the order above, but just checking there's no exception for constructors.)

plumpy: Hm, that's a good question. I would say I generally do see an exception for constructors and would expect a static factory method below them, like:

constants
fields
constructors
static factory methods (like `getInstance`)
API methods in the order described above

That's how I most often see them ordered, both inside and outside of Google. go/java-practices/class-structure has some more info. They actually do list "visible static methods" right below the constants (i.e. all the static stuff listed together), which doesn't feel particularly normal to me, but could definitely make sense! I don't know. I'd still be inclined to put `#getInstance` after the constructor still, but I'm not sure I'd feel super confident about it anymore 😁

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
```diff
@@ -228,65 +225,44 @@ public ArtifactUtils(
...
+    ImmutableSet<Artifact> allArtifacts =
+        ImmutableSet.<Artifact>builder()
+            .addAll(receivedArtifacts)
+            .addAll(resolveResult.getResolvedArtifacts())
+            .build();
 
     try {
       trigger.put(
           "artifacts",
           objectMapper.readValue(objectMapper.writeValueAsString(allArtifacts), List.class));
```
dbyron0: This is a bit of a tangent to the exciting change you're making here, so feel free to ignore.

I was in this neighborhood a couple of weeks ago trying to debug a situation where spinnaker was picking up an artifact from a prior execution, even though I thought I specified enough via manual execution.  I'm still not quite sure what was happening, though I was able to disable use prior artifact in those pipelines, so if a similar thing happens again I figure I'll get a failed pipeline execution instead of using the 'wrong' artifact.

Anyway, the struggle I had was that by changing trigger.artifacts here, it's hard to see what actually came via the trigger vs. what happened during artifact resolution.  This would help me figure out whether my struggle was in e.g. deck or gate vs. here.

A log message would help, but if there's a way to reflect this in the pipeline execution json, it'd facilitate my troubleshooting even more.

Thanks for reading :)

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
ezimanyi: My thought here was that it would be more flexible for API consumers if the API accepted and did something sensible with null values (but never returned null values)...but I can see how it might just encourage consumers to avoid thinking about what nulls mean.  I've updated the methods in the class to no longer accept null parameters, and added an overload to handle the case of `priorArtifacts` being null.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
ezimanyi: Makes sense!  Pushed a commit to update this.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/FindArtifactFromExecutionTask.java
ezimanyi: Interesting...I've seen these around before but haven't been in the habit of adding them.  The way IntelliJ is configured for me, the parameter name shows up in the UI:
image
But I can see the advantage of having this directly in the text and not depend on a particular IDE.  I can't seem to get that shortcut to work, but maybe I don't have the right extension installed...I'll see if I can get it to work.

ezimanyi: Ok, couldn't get the extension to work but did use search/replace to add these!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
ezimanyi: I agree this is confusing...in general I think it's confusing that sometimes we store artifacts in `pipeline.receivedArtifacts` and sometimes in `trigger.artifacts` (and in this case concatenate the two and store them on the trigger).  One day it would be great to refactor this and have a consistent contract for where artifacts should go to be resolved...but I'm worried about making any changes to that here as I'm sure there are a lot of things depending on it.

dbyron0: Yeah, had a feeling changes here would ripple out a bunch.  Would it help if I create a separate issue?

ezimanyi: Yeah, that would be helpful---I can't promise any changes to how this works in the near future as it might be a big project, but I'd definitely like to understand better the confusing case you ran into.

dbyron0: Here goes: https://github.com/spinnaker/spinnaker/issues/5344

PR #3379, opened 2020-01-19
feat(admin): Add queue admin EPs

marchello2000:
Adds two endpoints:
1. Generic endpoint to post any message into the queue
2. A simplified EP to specifically post `StartWaitingExecutions` into the queue

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/admin/web/QueueAdminController.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/admin/web/QueueAdminController.kt
@@ -15,11 +15,15 @@
  */
 package com.netflix.spinnaker.orca.q.admin.web
 
+import com.netflix.spinnaker.orca.q.StartWaitingExecutions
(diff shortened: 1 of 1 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/admin/web/QueueAdminController.kt
```diff
@@ -15,22 +15,27 @@
...
 @RequestMapping("/admin/queue")
 class QueueAdminController(
-  private val hydrateCommand: HydrateQueueCommand
+  private val hydrateCommand: HydrateQueueCommand,
+  private val queue: Queue
 ) {
 
-  @RequestMapping(value = ["/hydrate"], method = [(RequestMethod.POST)])
+  @PostMapping(value = ["/hydrate"])
   fun hydrateQueue(
```
robzienert: I should just delete this endpoint... it's not possible without making executions immutable (Mark, do you uh, wanna get crazy sometime this year? 😆 )

PR #3392, opened 2020-01-24
fix(execution): Fix START_TIME_OR_ID comparator

ezimanyi:
* test(execution): Add tests to START_TIME_OR_ID 

  The next commit will fix a bug in the START_TIME_OR_ID comparator; add tests to document the current behavior and ensure that the bug fix only affects the intended case.

* fix(execution): Fix START_TIME_OR_ID comparator 

  I noticed that the START_TIME_OR_ID comparator violates the general contract of Comparator in the case where one of hte executions has a null start time. I haven't noticed this causing actual bugs, but I think the failure mode would be intermittent cases of
  "Comparison method violates its general contract" being thrown when trying to sort based on this comparator, which would be annoying and hard to debug, so fixing this now.


ezimanyi:
@spinnakerbot cherry-pick 1.18

PR #3399, opened 2020-01-29
fix(wait stage): Make wait stage restartable

marchello2000:
Noticed that when I restarted a WaitStage it didn't actually wait.
That's because it caches the stage `startTime`.
Not sure why it does since the stage should have actual start time the time the task runs, so making the wait task use that instead of caching it's own time.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/WaitStage.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/WaitStage.java
@@ -22,9 +22,9 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
 import com.netflix.spinnaker.orca.pipeline.tasks.WaitTask;
 import java.time.Duration;
-import java.time.Instant;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import org.jetbrains.annotations.NotNull;
 import org.springframework.stereotype.Component;
 
 @Component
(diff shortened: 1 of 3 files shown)


marchello2000:
I don't know if there was some special historical reason why this was necessary so if there was let me know and I can workaround it/abandon/reword this change - it's clearly not VERY frequent use case seeing as restarts are an admin EP

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/WaitStage.java
```diff
@@ -33,44 +33,33 @@
   public static String STAGE_TYPE = "wait";
 
   @Override
-  public void taskGraph(Stage stage, TaskNode.Builder builder) {
+  public void taskGraph(@NotNull Stage stage, TaskNode.Builder builder) {
```
ezimanyi: The super method uses `@Nonnull` from `javax.annotation`, which is in general the non-null annotation we use more commonly (including elsewhere in this class) so I'd suggest using that here.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/WaitTask.java
```diff
@@ -44,17 +41,14 @@ public WaitTask(Clock clock) {
...
     WaitStage.WaitStageContext context = stage.mapTo(WaitStage.WaitStageContext.class);
 
-    if (context.getWaitTime() == null) {
-      return TaskResult.SUCCEEDED;
-    }
-
     Instant now = clock.instant();
 
-    if (context.isSkipRemainingWait()) {
+    if (context.isSkipRemainingWait() || context.getWaitDuration().isZero()) {
```
ezimanyi: This is maybe a silly minor comment, but do we need to special case the zero wait time here?  Wouldn't the start time plus zero always be before now in the next block which would simplify the number of cases to consider?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/WaitStage.java
marchello2000: silly me for blindly accepting intelliJ offer.. Thanks for catching it!

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/WaitTask.java
marchello2000: yeah, good call - thank you!

marchello2000: Also realized the `context.getWaitDuration()` check for `null` on line 72 wasn't necessary and removed it

PR #3400, opened 2020-01-29
fix(rrb): workaround for rrb eureka deregistration issue

asher:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForRequiredInstancesDownTask.groovy
```diff
@@ -37,7 +37,28 @@ class WaitForRequiredInstancesDownTask extends AbstractWaitingForInstancesTask {
...
+      List<String> skippedIds = getSkippedInstances(stage)
+      List<Map> skippedInstances = instances.findAll { skippedIds.contains(it.name) }
+      List<Map> instancesToDisable = getInstancesToDisable(stage, instances)
+
+      if (!skippedInstances.isEmpty()) {
+        List<Map> skippedButNotDown = skippedInstances.findAll {
+          !HealthHelper.someAreDownAndNoneAreUp(it, interestingHealthProviderNames)
+        }
+        Map<String, List<Map>> skippedInstanceHealths = skippedButNotDown.collectEntries { instance ->
+          [(instance.name): HealthHelper.filterHealths(instance, interestingHealthProviderNames)]
```
asher: For instances that clouddriver skipped deregistering from eureka but that still appear to be up, collect and log their health information. This should help identify if further action can be taken from orca, if skip decisions are being made in error, or if there are other issues with handling eureka status.

PR #3401, opened 2020-01-30
fix(monitor kato): Retry MonitorKatoTask more aggressively

marchello2000:
In the past, we would retry the `MonitorKatoTask` with fairly complicated logic of checking if it was a replica or a master we got the 404 from.
This change simplifies that by retrying a whole bunch (always from read replica).

Other than basic resilience, this is needed if/when one performs a clouddriver database migration.
The situation is as follows:
 - the old clouddriver cluster is pointing to the old db
 - the new clouddriver cluster is pointing to the new and old db so it can serve getTask requests in either DB
 - during the deploy (with 0 downtime) both old and new will be in the load balancer taking traffic from orca.
 - if an orca request creates a task in new clouddriver but then hits monitor on the old clouddriver it will fail with a 404

Instead of making the old cluster aware of the new DB, we just add a bunch of retries.
Note: they are not made configurable for the sake of not adding unnecessary complexity (not sure when someone would tweak the number of retries), but if you feel strongly it should be configurable i can change it

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,3 +7,4 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = space
 indent_size = 2
+continuation_indent_size = 4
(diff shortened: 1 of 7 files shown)


.editorconfig
```diff
@@ -7,3 +7,4 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = space
 indent_size = 2
+continuation_indent_size = 4
```
dreynaud: I see it's not just me 😅

marchello2000: this was driving me bonkers!

marchello2000:
going to merge this because we need it urgently, if you have feedback, please let me know and I will be happy to address it after

PR #3402, opened 2020-01-30
refactor(kubernetes): clean up source manifest resolution

maggieneterval:
The goal of this refactor is to make Orca's input manifest resolution easier to reason about so that we can enhance it with [label filtration](https://github.com/spinnaker/spinnaker/issues/3695).

- **refactor(kubernetes): add tests for ManifestEvaluator**

  Adds tests documenting the functionality of `ManifestEvaluator.evaluate` to ensure we don't break anything in the next commit.

- **refactor(kubernetes): clean up ManifestEvaluator**

  Divides the functionality of `ManifestEvaluator.evaluate` into private methods responsible for a single bit of work. Adds immutability and cleans up streams where possible.

- **refactor(kubernetes): move task-specific ManifestEvaluator logic to tasks**

  Removes `ManifestEvaluator.buildTaskResult`, which makes a call to Clouddriver to initiate a task and composes the task outputs. Since this logic is only consumed by the Deploy and Patch (Manifest) tasks, push that logic to those tasks. Divide Deploy and Patch (Manifest) tasks, which currently house all of their logic in the `execute` method, into three methods each: `getOperation`, `executeOperation`, and `getOutputs`. Since something that doesn't really fit into any of those three categories doesn't seem like it belongs in these tasks, hopefully that explicit naming will help us not overload these tasks in the future. In the next commit, we'll pull out manifest resolution, which belongs in a task of its own.

- **refactor(kubernetes): extract source manifest resolution to own task**

  Adds a new tasks responsible only for resolving a source manifest and its associated artifacts for Deploy and Patch (Manifest) stages. Ideally these stages could share one task, but since tasks cannot be parameterized and we need to deserialize the Patch and Deploy (Manifest) stage's context with different implementations of `ManifestContext` (and the Patch and Deploy operations expect the `optionalArtifacts` to be on different context keys), delegate to deploy- and patch-specific resolution tasks.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
@@ -16,12 +16,14 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.manifest;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.Collections.emptyList;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Streams;
 import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
@@ -29,6 +31,7 @@ import com.netflix.spinnaker.kork.core.RetrySupport;
 import com.netflix.spinnaker.orca.clouddriver.OortService;
 import com.netflix.spinnaker.orca.clouddriver.tasks.manifest.ManifestContext.BindArtifact;
 import com.netflix.spinnaker.orca.clouddriver.utils.CloudProviderAware;
+import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper;
 import com.netflix.spinnaker.orca.pipeline.expressions.PipelineExpressionEvaluator;
 import com.netflix.spinnaker.orca.pipeline.model.Stage;
 import com.netflix.spinnaker.orca.pipeline.util.ArtifactUtils;
@@ -36,7 +39,6 @@ import com.netflix.spinnaker.orca.pipeline.util.ContextParameterProcessor;
 import java.io.InputStream;
 import java.time.Duration;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -57,10 +59,10 @@ import retrofit.client.Response;
 public class ManifestEvaluator implements CloudProviderAware {
   private static final ThreadLocal<Yaml> yamlParser =
       ThreadLocal.withInitial(() -> new Yaml(new SafeConstructor()));
+  private static final ObjectMapper objectMapper = OrcaObjectMapper.getInstance();
 
   private final ArtifactUtils artifactUtils;
   private final ContextParameterProcessor contextParameterProcessor;
-  private final ObjectMapper objectMapper;
   private final OortService oortService;
   private final RetrySupport retrySupport;
 
@@ -68,12 +70,10 @@ public class ManifestEvaluator implements CloudProviderAware {
   public ManifestEvaluator(
       ArtifactUtils artifactUtils,
       ContextParameterProcessor contextParameterProcessor,
-      ObjectMapper objectMapper,
       OortService oortService,
       RetrySupport retrySupport) {
     this.artifactUtils = artifactUtils;
     this.contextParameterProcessor = contextParameterProcessor;
-    this.objectMapper = objectMapper;
     this.oortService = oortService;
     this.retrySupport = retrySupport;
   }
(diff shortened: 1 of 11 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
```diff
@@ -16,172 +16,245 @@
...
+  }
+
+  private ImmutableList<Map<Object, Object>> getSpelEvaluatedManifests(
+      ImmutableList<Map<Object, Object>> unevaluatedManifests, Stage stage) {
+    Map<String, Object> manifestWrapper = new HashMap<>();
+    manifestWrapper.put("manifests", unevaluatedManifests);
+
+    manifestWrapper =
+        contextParameterProcessor.process(
+            manifestWrapper, contextParameterProcessor.buildExecutionContext(stage), true);
```
plumpy: What is `true`? Not in a philosophical sense, but more like what is that parameter doing? go/java-practices/comments#arguments suggests annotating it for readers with `/* allowUnknownKeys= */ true`. (If you put your cursor on `true` and hit `ctrl+alt+/` it should automatically add that comment.)

This is super subjective, but I also don't know about reusing `manifestWrapper`. It feels like you have two different variables that coincidentally have the same type, but are conceptually different things? I'd be inclined to call one `processorInput` and `processorResult`? I know nothing about the `ContextParameterProcessor` so I definitely may be wrong here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
```diff
@@ -16,172 +16,245 @@
...
+  }
+
+  private ImmutableList<Map<Object, Object>> getManifests(Stage stage, ManifestContext context) {
     if (ManifestContext.Source.Artifact.equals(context.getSource())) {
-      Artifact manifestArtifact =
-          artifactUtils.getBoundArtifactForStage(
-              stage, context.getManifestArtifactId(), context.getManifestArtifact());
+      return getManifestsFromArtifact(stage, context);
+    }
+    return getManifestsFromText(context);
```
plumpy: Personally I'd lean towards a switch for `enum` types since it makes it a little more clear to readers what types you're handling and makes it less likely to have strange things happen if you ever add a third type:
```
    switch (context.getSource()) {
      case Artifact:
        return getManifestsFromArtifact(stage, context);
      case Text:
        return getManifestsFromText(context);
    }
    throw new IllegalStateException("Unknown ManifestContext.Source " + context.getSource());
```

Another bonus of this is static analysis could catch at compile time that you're not covering all the cases... internally this code would no longer compile if someone added a third enum value, but we don't have that set up in Spinnaker. Maybe some day, though!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
```diff
@@ -16,172 +16,245 @@
...
+    // Once the legacy artifacts feature is removed, all trigger expected artifacts will be
+    // required to define an account up front.
+    if (context.getManifestArtifactAccount() != null) {
+      manifestArtifact.setArtifactAccount(context.getManifestArtifactAccount());
     }
 
-    log.info("Artifacts {} are bound to the manifest", requiredArtifacts);
+    if (manifestArtifact.getArtifactAccount() == null) {
+      throw new IllegalArgumentException("No manifest artifact account was specified.");
+    }
```
plumpy: equivalent code:
`checkArgument(manifestArtifact.getArtifactAccount() != null, "No manifest artifact account was specified.");`

Internally we use the [`Preconditions`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/base/Preconditions.html) methods in practically every file, but maybe it would be confusing inside Spinnaker so isn't worth using, I don't know...

maggieneterval: Love it, adding to the plumpstuff java bible.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
maggieneterval: :1234: :tada: :+1:

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
maggieneterval: Still don't know what is philosophically `true`, but added the arguments annotation so folks can at least try to understand what `allowUnknownKeys: true` means (comes from the [expression logic in Kork](https://github.com/spinnaker/kork/blob/master/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/whitelisting/MapPropertyAccessor.java#L28) which I don't really understand the purpose of to be honest).

Good call on distinguishing input vs. result, updated that as well!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
```diff
@@ -16,172 +16,245 @@
...
+
+    return manifestArtifact;
+  }
+
+  private ImmutableList<Map<Object, Object>> getSpelEvaluatedManifests(
+      ImmutableList<Map<Object, Object>> unevaluatedManifests, Stage stage) {
+    Map<String, Object> manifestWrapper = new HashMap<>();
+    manifestWrapper.put("manifests", unevaluatedManifests);
```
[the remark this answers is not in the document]
maggieneterval: Up to me?! Updated, always up for saving a line.

PR #3403, opened 2020-01-31
fix(stage): entity tags task succeeds

emjburns:
We don't actually handle returning `SKIPPED` for status here. It makes it so the stage always fails because `SKIPPED` isn't a valid status. I choose to return `SUCCEEDED` because there were 0 tags, so it's a success. I don't think that tags are required for all spinnakers so i don't think it's correct to fail the stage if there are no tags.


marchello2000:
I think it makes sense

emjburns:
@jonsie I'm curious about that too. With the specific example I was looking into it seemed like there were some errors from titus (connection closed etc) which could have resulted in no server groups. Not sure though, I couldn't find a good log trace or enough logs.

PR #3404, opened 2020-01-31
chore(InvalidTaskId): adding logs for when we can't find a task

emjburns:
We're seeing this error case but it's not clear why this is happening or what message we're failing on. Adding a log to hopefully help find the cause.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
@@ -36,9 +36,15 @@ import com.netflix.spinnaker.orca.q.StartStage
 import com.netflix.spinnaker.orca.q.TaskLevel
 import com.netflix.spinnaker.q.Message
 import com.netflix.spinnaker.q.MessageHandler
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 import java.time.Duration
 
 internal interface OrcaMessageHandler<M : Message> : MessageHandler<M> {
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
   val repository: ExecutionRepository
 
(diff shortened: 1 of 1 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
```diff
@@ -52,6 +58,7 @@ internal interface OrcaMessageHandler<M : Message> : MessageHandler<M> {
         .taskById(taskId)
         .let { task ->
           if (task == null) {
+            log.error("InvalidTaskId: Unable to find task {} in existing tasks {} while processing message {}", taskId, stage.tasks.map { it.id }, this)
```
asher: Will `this` actually print a human readable representation of the message vs the object type?

emjburns: Good call, I'm not sure and I can't get orca to start to test it out. Changing to print the class name

marchello2000: yes, since `this` is of type `Message` and since every message is a `data class` they should get a sensible `toString()` from the compiler

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
```diff
@@ -52,6 +58,7 @@ internal interface OrcaMessageHandler<M : Message> : MessageHandler<M> {
         .taskById(taskId)
         .let { task ->
           if (task == null) {
+            log.error("InvalidTaskId: Unable to find task {} in existing tasks {} while processing message {}", taskId, stage.tasks.map { it.id }, this.javaClass)
```
marchello2000: you should probably switch this back to printing `this` instead of `this.javaClass`. While I have a hard time believing the message can be corrupt, it would be useful to ascertain whether it is or is not

PR #3406, opened 2020-02-03
chore(plugins): adding plugins dir and application name

link108:
This creates the default plugins directory so plugins can be downloaded and placed locally. I needed to add the spring.application.name in gradle.properties as well, otherwise plugins would fail to start when running in a container.

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -3,5 +3,6 @@ MAINTAINER delivery-engineering@netflix.com
 COPY --from=compile /compiled_sources/orca-web/build/install/orca /opt/orca
 RUN apk --no-cache add --update bash openjdk8-jre
 RUN adduser -D -S spinnaker
+RUN mkdir plugins && chown -R 100:100 plugins
 USER spinnaker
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -4,3 +4,4 @@ korkVersion=7.16.1
 spinnakerGradleVersion=7.0.1
 enablePublishing=false
 org.gradle.parallel=true
+spring.application.name=orca
```
robzienert: Hmm, why is this necessary? [It should already set by the application itself](https://github.com/spinnaker/orca/blob/master/orca-web/src/main/groovy/com/netflix/spinnaker/orca/Main.groovy#L97).

link108: I need to do some digging on this one, not sure why its needed, especially since its in the code already. When running in a container, plugins would fail to be found unless I added this.

marchello2000: i also added this to tests here: https://github.com/spinnaker/orca/pull/3405

link108: not needed, thank you both!

PR #3408, opened 2020-02-04
fix(ci): Properly set artifacts on CI stage output

jervi:
The presence of a property file should not affect whether or not artifacts should be added to the output. The Travis and Wercker integrations in Igor don't even care about this property, and the Jenkins integration also works even if the property file is null or blank.

This will also copy artifacts already present in the `buildInfo` to the stage context, to be more in line with CI triggers (where this already happens).

This fix makes it possible to use the "Find Artifacts from Execution" stage with Travis and Wercker, as well as increasing the usability for use with Jenkins.

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/CIStageDefinition.java
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/CIStageDefinition.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import lombok.Getter;
 
@@ -29,6 +30,7 @@ public class CIStageDefinition implements RetryableStageDefinition {
   private final String job;
   private final String propertyFile;
   private final Integer buildNumber;
+  private final Map<String, Object> buildInfo;
   private final boolean waitForCompletion;
   private final List<ExpectedArtifact> expectedArtifacts;
   private final int consecutiveErrors;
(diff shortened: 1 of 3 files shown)


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/model/CIStageDefinition.java
```diff
@@ -41,13 +43,15 @@ public CIStageDefinition(
       @JsonProperty("job") String job,
       @JsonProperty("property") String propertyFile,
       @JsonProperty("buildNumber") Integer buildNumber,
+      @JsonProperty("buildInfo") Map<String, Object> buildInfo,
```
ezimanyi: I'd like to avoid having to handle ad-hoc `convertValue` operations inside of tasks (as we do for the artifacts in `GetBuildArtifactsTask`), as well as having untyped maps in the stage context.

I think this would be cleaner if we strongly-type the `buildInfo` by adding an inner class below `CIStageDefinition`:
```
  @Getter
  public static class BuildInfo {
    private final ImmutableList<Artifact> artifacts;
    
    public BuildInfo(@JsonProperty("artifacts") List<Artifact> artifacts) {
      this.artifacts =
          Optional.ofNullable(artifacts).map(ImmutableList::copyOf).orElse(ImmutableList.of());
    }
  }
```
and then also in the `CIStageDefinition` constructor just default `buildInfo` to an empty `BuildInfo` if the input is null.

Then we'd only do a single mapping at the beginning of the stage and all the rest of the logic would get type safety and null safety.  (You'd be able to remove the `convertValue` and the null-checks on buildInfo later.)

jervi: Thanks @ezimanyi! I agree, that is _a lot_ cleaner. Can you please take another look?

PR #3410, opened 2020-02-04
perf(clouddriver): Remove down instance check in disable server group…

mhumeSF:
… stage.

When a server group is disabled in AWS, you don't need to check for down instances. That's not the point of the stage, as it does not scale down the instances. It simply detaches them from the load balancer so they no longer receive traffic. This step alone is vastly slowing down our release cadence.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DisableServerGroupStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DisableServerGroupStage.groovy
@@ -47,7 +47,13 @@ class DisableServerGroupStage extends TargetServerGroupLinearStageSupport {
       .withTask("determineHealthProviders", DetermineHealthProvidersTask)
       .withTask("disableServerGroup", DisableServerGroupTask)
       .withTask("monitorServerGroup", MonitorKatoTask)
-      .withTask("waitForDownInstances", WaitForRequiredInstancesDownTask)
+
+    // When using AWS, disabling a server group doesn't cause the instances to go
+    // down or scale in. So there's no point in waiting for down instances - it just
+    // slows things down.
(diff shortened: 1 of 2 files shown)


cfieber:
If there is a source of instance health in AWS (the instances are in a load balancer for example) then this wait is necessary... depending on the load balancer settings the instances will not be marked as down until connection draining completes, and this stage needs to wait until that happens.

goodspark:
The point is that 'down' is not 'disabled' or 'detached'. 'Disabling' a server group is meant to detach the instances from the load balancer but keep them around for whatever reason (eg. the connection draining you referenced). The point of this stage (at least for AWS + LBs) is to stop sending traffic to the instances so that a new server group can take the full share of traffic.

There may be cases where we want a pipeline that looks like Deploy -> Disable -> Run some E2E tests -> Rollback on error.

By using disable instead of a the red/black strategy in Spinnaker, we can keep the old server group alive with approximately the right amount of instances (the 'active' server group might scale up between the deploy and rollback decision). Then we can simply and quickly reenable the old group, wait for old instances to get healthy (generally fast), and disable the new group to do a deploy. We skip the whole 'scale up' part of a rollback in Spinnaker. In a follow up we can clean up the new server group without blocking the pipeline.

cfieber:
there is some poor terminology here, but really this is checking for instances to be out of service

are you saying in our example you would want to 

deploy -> disable (but not wait for deregistration delays) -> run some tests -> re-enable if things are bad

you could accomplish that by parallelizing your pipeline:

deploy ---> disable
             \---> run some tests ---> re-enable

maybe with a sleep in there before the tests to allow the instances to be deregistered from load balancers before your tests start...

regardless, there isn't a way this change can land as is, there are too many assumptions built in to the system that this stage is complete once your instances are no longer taking traffic (not that they have been requested to stop taking traffic)

I could potentially see a parameter to the stage that would conditionally skip waiting for instance health to change

robzienert:
Good discussion here - ping @mhumeSF, wdyt of @cfieber's suggestion?

PR #3413, opened 2020-02-05
fix(kubernetes): Fix and simplify manifest stability logic

ezimanyi:
The primary goal here is to remove the distinction between a manifest's stable/failed status being `null` and its being `false` so that we can stop returning these as `null` from clouddriver, simplifying the code both here and in `clouddriver` and making it more null-safe.

* test(kubernetes): Add tests to WaitForManifestStable and Manifest 

  There are some existing tests in groovy on WaitForManifestStableTask but they are missing some edge cases that will prove important in the next commit.  Add a new suite of tests in Java; these duplicate the Groovy tests and also add more test cases. We'll wait to delete the Groovy tests until after the refactor in an upcoming commit just to get a bit more validation that it doesn't break anything.

  Also to prepare for some upcoming changes, add tests to the Manifest class to ensure we can correctly deserialize manifests.

* fix(kubernetes): Fix and simplify manifest stability logic 

  The logic in WaitForManifestStable is quite complex; in particular it separately distinguishes true, false, and null on the manifest's status field and has different behavior in each case. Likely due to this complexity, there are a number of bugs that were surface when writing tests in the last commit.

  Fix these bugs, which also allows us to stop treating a null status as different from false; the cases where we did were just what was leading to the bugs noted earlier.

  The tests should clearly show the changes, but in particular:
  * The status is fixed in a couple of cases where we either failed too early when one manifest was failed, or eventually returned success incorrectly even though a manifest had failed on an earlier poll.
  * Duplicate log messages are removed; if a manifest has failed, we now just log that it failed instead of also logging that we are waiting for it to stabilize.

* refactor(kubernetes): Delete unused ObjectMapper and groovy test 

  The ObjectMapper is unused. The groovy test was superseded by the Java tests added a few commits ago, but was kept around temporarily to give additional validation to the refactor in the last commit.

* refactor(kubernetes): Make Manifest classes final and immutable 

  Let's make the Manifest class and contained classes immutable.

  The changes to WaitForManifestStableTaskTest are just to take advantage of the more convenient constructors/builders that were created on those classes; none of the actual test results change.

* refactor(kubernetes): Default null entries in manifest 

  Add nullity annotations to the Manifest class (and contained classes) and default any null entries on deserialization.

  The fact that none of the tests on WaitForManifestStableTask broke means that switching from having a null status to a false status has not changed any observable behavior.


orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTaskTest.java
```diff
@@ -0,0 +1,383 @@
...
+    WaitForManifestStableTask task = new WaitForManifestStableTask(oortService, new ObjectMapper());
+
+    Stage myStage =
+        createStageWithManifests(ImmutableMap.of(NAMESPACE, ImmutableList.of(MANIFEST_1)));
+
+    when(oortService.getManifest(ACCOUNT, NAMESPACE, MANIFEST_1, false))
+        .thenReturn(manifestBuilder().stable(true).failed(true).build());
+
+    TaskResult result = task.execute(myStage);
+    AssertionsForClassTypes.assertThat(result.getStatus()).isEqualTo(ExecutionStatus.TERMINAL);
```
maggieneterval: I see we're using an `AssertionsForClassTypes.assertThat` in one other test to create an assertion for an Optional...when should we use this vs. vanilla `assertThat`?

ezimanyi: Great question!  I meant to add a comment on the PR for this, thanks for reminding me!  This is effectively a case study of Effective Java item 52 "Use Overloading Judiciously".

The reason is that `ExecutionStatus` implements `CharSequence` and there are two methods in  `Assertions.java` (where I've replaced the return with `U` in both cases as it's not important):
```
static <T extends Comparable<? super T>> U assertThat(T actual)
```
and a:
```
static U assertThat(CharSequence actual)
```

The compiler is tasked with finding the most specific invocation (using the compile-time types) when passed an `ExecutionStatus` but determines that neither of these methods is "most specific".  Per the JLS 15.12.2.5 "The informal intuition is that one method is more specific than another if any invocation handled by the first method could be passed on to the other one without a compile-time error" (which then goes on to get into a bunch of details I don't fully understand).  But I think here it's that:
* We can't replace an arbitrary invocation of the `CharSequence` one with the generic one because a `CharSequence` isn't guaranteed to implement `Comparable`.
* We can't replace invocations of the generic one with the `CharSequence` one because we might have called it with something that's not a `CharSequence`.
Since neither is "more specific" than the other, the compiler can't figure out which method to call and fails.  I think that if the generic one didn't have the bound `extends Comparable<? super T>` this *would* as it would view the `CharSequence` one as more specific and would call that one.

To get around this, AssertJ [suggests using AssertionsForClass](http://joel-costigliola.github.io/assertj/assertj-core.html#ambiguous-compilation-error) which supports the same generic assertions but does not include overloads accepting interfaces and sidesteps this bug.

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTaskTest.java
```diff
@@ -0,0 +1,383 @@
...
+
+    result =
+        task.execute(
+            createStageWithContext(
+                ImmutableMap.<String, Object>builder()
+                    .putAll(myStage.getContext())
+                    .putAll(result.getContext())
+                    .build()));
+    // TODO(ezimanyi): This is a bug; we should fail the task because my-manifest-1 failed on the
+    // first poll
```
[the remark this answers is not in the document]
ezimanyi: I found that one by reading the code and trying to understand where we were keeping track of whether anything failed on a prior poll...then realizing we just weren't at all!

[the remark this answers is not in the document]
maggieneterval: :man_scientist: :hammer: :bug:

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTaskTest.java
maggieneterval: Wow, that's so confusing, thank you for explaining that. My next question was "why does `ExecutionStatus` implement `CharSequence`?" but the answer is, perhaps unsurprisingly, SpEL. :cry:

ezimanyi: Ha, I hadn't actually dug that far as to ask why it implemented `CharSequence` :cry: :sob:

PR #3417, opened 2020-02-07
fix(artifacts): don't output buildArtifacts as korkArtifacts

asher:


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildArtifactsTask.java
```diff
@@ -45,10 +44,6 @@
             stageDefinition.getPropertyFile(),
             stageDefinition.getMaster(),
             stageDefinition.getJob());
-    if (artifacts == null) {
-      artifacts = new ArrayList<>();
-    }
-    artifacts.addAll(stageDefinition.getBuildInfo().getArtifacts());
```
asher: @jervi This essentially reverts a small part of your recent PR, #3408. It appears to have triggered a latent regression in `PackageInfo. orca-1.2358.0-h333.bb7d1a2-x86_64()` that only surfaces when `CreateBakeTask` is passed a fully qualified package name with suffix (i.e. "orca-1.2358.0-h333.bb7d1a2-x86_64.deb" instead of just "orca"). This is a pattern in use by a few teams at Netflix, hence the revert here. I haven't succeeded at unwinding the behavior in `findTargetPacakge` that this impacts.

jervi: Hmm, I'll try to figure out a fix that let's us keep this without breaking your workflows. I'll assign the PR to you once it is ready :)

jervi: I guess you haven't enabled the artifact decorator in your Spinnaker installation? Because the deb decorator in Igor should have renamed that package to orca if it came from Jenkins

jervi:
@asher Can you please have a look at #3422? I think this fix will let us keep the new functionality without breaking anyone's workflow.

PR #3420, opened 2020-02-09
fix(kubernetes): skip over empty manifests in artifacts

PerGon:
When deploying to K8s and using helm to bake the artifacts, Helm has the habit of adding empty yaml documents to the manifest. So, a baked helm chart can look like this:
```
---
# Source: k8s_namespaces/templates/00-something/01-file.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: other_namespace
---
# Source: k8s_namespaces/templates/00-namespaces.yaml

---
apiVersion: v1
kind: Namespace
metadata:
  name: namespace-dev
---
```
(notice the empty yaml manifest in the middle)

This PR allows to jump over those empty documents instead of failing with a NPE.


PerGon:
@spinnaker/reviewers PTAL

PerGon:
@spinnaker/oss-approvers PTAL

PR #3422, opened 2020-02-10
fix(artifacts): Output decorated ci stage artifacts as kork artifacts

jervi:
This PR makes it possible to use the "Produces Artifacts" feature for Travis and Jenkins stages (if the decorator is enabled in Igor).
image
It also makes it possible to bind artifacts (e.g. Docker images) from a Jenkins or Travis stage to a Deploy Manifest stage, for instance.

This PR reverts #3417, but does so in a way that should not affect Netflix' pipelines.


ezimanyi:
Looking at both this and #3417, I'm still not sure I fully understand what the bug was and how this change is safe.  (I might need also a quick reminder on how artifact decoration in igor works.)

Would an alternative be to have the `/builds/artifacts` endpoint in `igor` return decorated artifacts directly so we don't need to separately add them to the map?  (And it also would remove what seems like a fragile contract between orca and igor that orca will only add artifacts with a particular boolean in the metadata field.)

jervi:
I don't really understand the bug either 🙄 
The `/builds/artifacts` endpoint is actually a bit misleading. What it actually returns are artifacts that are found using the method described here: https://www.spinnaker.io/reference/artifacts/from-build-triggers/
However, both the Jenkins and the Travis integration in Spinnaker already adds artifacts to `buildInfo.artifacts`. If the artifact decorator is enabled, the artifacts will also be decorated. This means there in most cases is no need to call a igor endpoint to get artifacts, they are already present in `trigger.buildInfo` or in `<stageContext>.buildInfo.artifacts`.

### Decorator primer:
The decorator will convert this artifact:
```
fileName: "openmotif22-libs-2.2.4-192.1.3.x86_64.rpm"
displayPath: "openmotif22-libs-2.2.4-192.1.3.x86_64.rpm"
relativePath: "openmotif22-libs-2.2.4-192.1.3.x86_64.rpm"
```
into
```
fileName: "openmotif22-libs-2.2.4-192.1.3.x86_64.rpm"
displayPath: "openmotif22-libs-2.2.4-192.1.3.x86_64.rpm"
relativePath: "openmotif22-libs-2.2.4-192.1.3.x86_64.rpm"
reference: "openmotif22-libs-2.2.4-192.1.3.x86_64.rpm"
name: "openmotif22-libs"
type: "rpm"
version: "2.2.4-192.1.3.x86_64"
decorated: "true"
```
The built in decorator suppors `deb` and `rpm`, but it can be extended using config like this:
```yaml
artifact:
#  This is a feature toggle for decoration of artifacts.
  decorator:
    enabled: true
    fileDecorators:
       - type: docker/image
         decoratorRegex: '(.+):(.+)'
         identifierRegex: '(.+\/.+:.+)'
```

asher:
@jervi here's some more data around the related failures at nflx. This deployment flow is atypical but preferred by some teams responsible for stateful services.

1) Pipeline initiates a dpkg build via a Jenkins stage, reading back properties.
2) Outputs from the jenkins stage includes:
```
"buildInfo": {
  "artifacts": [
    {
			"reference": "build/distributions/my-app_0.1.0-h114.4d6cd75_all.deb",
			"fileName": “my-app_0.1.0-h114.4d6cd75_all.deb",
			"relativePath": "build/distributions/my-app_0.1.0-h114.4d6cd75_all.deb",
			"name": “my-jenkins-job”,
			"displayPath": “my-app_0.1.0-h114.4d6cd75_all.deb",
			"type": "jenkins/file",
			"version": "114"
    }.
    {
			"reference": "merged_ami_params.json",
			"fileName": "merged_ami_params.json",
			“relativePath": "merged_ami_params.json",
			"name": “my-jenkins-job“,
			"displayPath": "merged_ami_params.json",
			"type": "jenkins/file",
			"version": "114"
    }
  ]
}
```
Before #3408, the outputs from this jenkins stage also contains `artifacts` as an empty list. After #3408 but before #3417,`outputs.buildInfo.artifacts` was unchanged, but `outputs.artifacts` contained:
```
"artifacts": [
  {
    	"type": "jenkins/file",
	"customKind": false,
	"name": "my-jenkins-job",
	"version": "114",
	"reference": "build/distributions/my-app_0.1.0-h114.4d6cd75_all.deb",
	"metadata": {
		"fileName": "my-app_0.1.0-h114.4d6cd75_all.deb",
		"relativePath": "build/distributions/my-app_0.1.0-h114.4d6cd75_all.deb",
		"displayPath": "my-app_0.1.0-h114.4d6cd75_all.deb"
	}
  },
  {
	"type": "jenkins/file",
	"customKind": false,
	"name": "my-jenkins-job",
	"version": "114",
	"reference": "merged_ami_params.json",
	"metadata": {
		"fileName": "merged_ami_params.json",
		"relativePath": "merged_ami_params.json",
		"displayPath": "merged_ami_params.json"
	}
  }
]
```
3) Immediately following the Jenkins stage is a Bake stage, where package is populated via SPeL to a fully versioned w/suffix package. In this example, the back stage context contains:
```
"package": "my-app_0.1.0-h114.4d6cd75_all.deb"
```

When the back stage encounters the above the jenkins stage `output.buildInfo.artifacts` but an empty list in `outputs.artifacts`, it runs fine. When `outputs.artifacts` includes these artifacts as produced by #3408, `CreateBakeTask` fails with the following:
```
"java.lang.IllegalStateException: Unable to find deployable artifact starting with [my-app_0.1.0-h114.4d6cd75_all.deb_] and ending with .deb in [] and []. Make sure your deb package file name complies with the naming convention: name_version-release_arch.
  at com.netflix.spinnaker.orca.pipeline.util.PackageInfo.createAugmentedRequest(PackageInfo.java:274)
  at com.netflix.spinnaker.orca.pipeline.util.PackageInfo.findTargetPackage(PackageInfo.java:112)
```

jervi:
Ah. The correct fix would be to update the SPeL expression to use the name field of the artifact instead of fully qualified name then I guess. The version will be autoinjected by Orca/Rosco. In other words, the package field of the bakery should only contain `my-app`.

jervi:
Don't merge this yet. I'm doing some investigation after chatting with @asher on Slack.

jervi:
I'm closing this, seems it isn't necessary after all. Artifacts are still found when baking even if they reside inside buildInfo. Sorry about the fuss, @asher 😊

jervi:
I really need to revive this PR, @asher and @ezimanyi 😃 
As my last comment above says: this PR is not needed for bakes. But it is however needed to use the "Produces Artifacts" feature of the Travis stage, for K8S manifest deploys etc. It has become more pressing for us after [this change in Clouddriver](https://github.com/spinnaker/clouddriver/pull/4874) broke a bunch of our pipelines.
I think this should be safe for Netflix as well, as you don't have the decorator in Igor enabled AFAIK.

ezimanyi:
@jervi : I'm not working on Spinnaker anymore so will leave it to others to review this...maybe someone from Netflix who worked on the initial rollback would be a good choice for reviewing this.  Thanks!

jervi:
Sorry to bother you Eric, I wasn't aware!
@spinnaker/oss-reviewers PTAL

jervi:
@asher 🙏 ?

jervi:
@ajordens Any chance you could have a look at this? 🙏

jervi:
@Mergifyio update

dbyron-sf:
@mergifyio update

PR #3424, opened 2020-02-10
feat(codebuild): Add more options to AWS CodeBuild stage

Kaixiang-AWS:
1. Add source override. Users could override their source by specifying
Spinnaker artifacts. Only S3 objects, GitHub repo and BitBucket repo are
accepted as valid input. Note that multiple sources are not yet supported,
this will be added as secondarySourceOverride in a later version.

2. Add source version. Users could specify the source version of their
sources. This source version takes precedence of the version specified
in Spinnaker artifact.

3. Add environment variables. Users could specify env vars that would be
used during the build.


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartAwsCodeBuildTask.java
```diff
@@ -51,4 +69,74 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+        break;
+      case "git/repo":
+        requestInput.put(SOURCE_LOCATION, artifact.getReference());
+        if (artifact.getReference().matches("^(http(s)?://)github.com/(.*)$")) {
+          requestInput.put(SOURCE_TYPE, "GITHUB");
+        } else if (artifact.getReference().matches("^(http(s)?://)(.*@)?bitbucket.org/(.*)$")) {
+          requestInput.put(SOURCE_TYPE, "BITBUCKET");
+        } else {
+          throw new IllegalStateException("Only GitHub and BitBucket repositories are supported");
+        }
```
clareliguori: No support for GitHub Enterprise or CodeCommit?  Could the user specify what provider the artifact is for?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartAwsCodeBuildTask.java
```diff
@@ -51,4 +69,74 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+  private void appendSource(Map<String, Object> requestInput, Artifact artifact) {
+    switch (artifact.getType()) {
+      case "s3/object":
+        requestInput.put(SOURCE_TYPE, "S3");
+        String s3Location = artifact.getReference();
+        requestInput.put(
+            SOURCE_LOCATION, s3Location.startsWith("s3://") ? s3Location.substring(5) : s3Location);
+        break;
+      case "git/repo":
+        requestInput.put(SOURCE_LOCATION, artifact.getReference());
```
clareliguori: I notice that the widget in deck has options for selecting branch and checkout sub-path.  Are those artifact properties ignored here?  Should they be used, or should this throw an error if they are filled out?

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartAwsCodeBuildTask.java
Kaixiang-AWS: Yes CodeCommit should be added here. That was a miss. For GHE, there's no proper regex match that we can do. Unless we categorize all non GitHub/BitBucket/CodeCommit repos to GHE, there's no easy way to enable this. 

Reference for git/repo type artifact https://www.spinnaker.io/reference/artifacts/types/git-repo/

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartAwsCodeBuildTask.java
Kaixiang-AWS: Branch goes to version field and subpath is ignored as CodeBuild doesn't support subpath checkout.

clareliguori: If it's not used, I think validation should be done to ensure it's not set.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartAwsCodeBuildTask.java
clareliguori: I think GHE will be important for the typical Spinnaker user.  What about a separate field for users to specify the "CodeBuild source type"?  It can be non-required so you can do nice inferring logic here, but allow customers to specify when it's not clear.

Kaixiang-AWS: That's a good idea. I will add a source type field to let users specify type explicitly.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartAwsCodeBuildTask.java
Kaixiang-AWS: will add

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartAwsCodeBuildTask.java
```diff
@@ -51,4 +72,82 @@ public TaskResult execute(@Nonnull Stage stage) {
...
+          requestInput.putIfAbsent(SOURCE_TYPE, "GITHUB");
+        } else if (artifact.getReference().matches("^(http(s)?://)(.*@)?bitbucket.org/(.*)$")) {
+          requestInput.putIfAbsent(SOURCE_TYPE, "BITBUCKET");
+        } else if (artifact
+            .getReference()
+            .matches("^https://git-codecommit.(.*).amazonaws.com/(.*)$")) {
+          requestInput.putIfAbsent(SOURCE_TYPE, "CODECOMMIT");
+        } else {
+          throw new IllegalStateException("Source type could not be inferred from location");
+        }
```
clareliguori: This will still throw, even if I provided sourceType explicitly.  This should check if source type is already set (and unit test should test this)

Kaixiang-AWS: Fixed

PR #3426, opened 2020-02-11
removed sketchy thread stuff from DependentPipelineStarter

robfletcher:
This was originally required for Spring Batch (back in the day). I saw an issue while on call that I suspect may have been called by the callables in this not working correctly. Specifically assigning the `pipeline` variable _inside_ the callable rather than returning it seems weird.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
@@ -155,37 +155,26 @@ class DependentPipelineStarter implements ApplicationContextAware {
     log.debug("Source thread: MDC user: " + AuthenticatedRequest.getAuthenticationHeaders() +
       ", principal: " + principal?.toString())
 
-    def pipeline
-    def callable
+    Callable<Execution> callable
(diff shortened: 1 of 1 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -155,37 +155,26 @@ class DependentPipelineStarter implements ApplicationContextAware {
...
+      } as Callable<Execution>, true, principal)
     } else {
       callable = AuthenticatedRequest.propagate({
         log.debug("Destination thread user: " + AuthenticatedRequest.getAuthenticationHeaders())
-        pipeline = executionLauncher().fail(PIPELINE, json, artifactError)
-      } as Callable<Void>, true, principal)
-    }
-
-    //This needs to run in a separate thread to not bork the batch TransactionManager
-    //TODO(rfletcher) - should be safe to kill this off once nu-orca merges down
```
emjburns: Finally! Today is the day!! 😄 😄

robfletcher: I dread to think how long that comment has been there

PR #3430, opened 2020-02-12
feat(peering): Peering of executions cross orca DBs

marchello2000:
Adds (semi-experimental) peering support which allows us to run two orcas (say, in two different regions) and each region would see the other regions' executions.
This is accomplished by a peering agent that essentially performs a migration (copy) of all executions from the other DB.

There are a few parts to this work:
1. Copy executions
2. Add the ability to act on foreign executions
3. Take ownership and start/resume executing of foreign executions

This PR addresses #1 above but there are still a few caveats to be completed, biggest are:
- Deletes are currently not replicated
- Ability to peer multiple DBs is not supported
- Ability to run multiple peering agents (peering the same DB) is not supported

See orca-peering/readme.md for a bit more information


marchello2000:
I am still working on the tests, but would love your feedback in the mean time

orca-sql/src/test/groovy/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepositorySpec.groovy
```diff
@@ -25,6 +25,7 @@ import com.netflix.spinnaker.orca.pipeline.model.Execution
 import com.netflix.spinnaker.orca.pipeline.model.Stage
```
marchello2000: these are still work in progress

.editorconfig
```diff
@@ -8,3 +8,7 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 2
 continuation_indent_size = 4
+
+[{*.kt, *.kts}]
+continuation_indent_size = 2
```
asher: ❤️

orca-peering/README.md
```diff
@@ -0,0 +1,113 @@
+# Orca Peering
+
+This is an semi-experimental approach to solve the problem of having multiple `orca` installations (each with their own database) communicate changes with each other, for instance in a multi-region Spinnaker installation or during a database migration.
+
+The peering mechanism accomplishes a few things:
+1. (complete) Peer executions (pipelines and orchestrations) from one database (region) to another database (region)
+2. (in-progress) Allows for an `orca` operating with one database to perform actions on an execution being executed by an `orca` operating with a different database
```
asher:
```suggestion
2. (in-progress) Allow for an `orca` instance to perform actions on an execution running on a foreign partition
```

orca-peering/README.md
```diff
@@ -0,0 +1,113 @@
...
+3. (still to come) Take ownership and resume an execution previously operated on by an `orca` with a different database
```
asher:
```suggestion
3. (still to come) The ability to migrate incomplete executions across partitions
```

asher: Not sure if these are good suggestions due to some terminology confusion. Might help to define Peer, define Partition, and explain how the concepts relate.

orca-peering/src/main/kotlin/com/netflix/spinnaker/config/PeeringAgentConfiguration.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import java.util.concurrent.Executors
+
+@Configuration
+/**
+ * TODO(mvulfson): this needs to support multiple (arbitrary number of) beans / peers defined in config
+ * TODO(mvulfson): prob also worth creating ConfigurationProperties class for the props
```
cfieber: yes please!

marchello2000:
> This looks pretty good, though I have a general concern about the low level of operation. Orca's DB persistence is basically an MVP that has lots of room for improvement. At times I've thought it would be nice to revisit in ways, including changing how completed executions are stored (i.e. in a single row). It's only a matter of time until it evolves. That this code is tightly coupled to `SqlExecutionRepository` implementation details without code reuse seems a recipe for future regret. I'm guessing this approach was taken for pure performance reasons but can batch operations with minimal deserialization become a feature of ExecutionRepository?

This is a good point. I would also love to revisit how we persist executions and I like your idea about single row for completed ones. The mapping of executions/etc was a huge overhead when I first tried the naive approach of getting stuff from execution repository. But let me think of ways to do better re-use and less coupling - totally valid concern!

orca-peering/src/main/kotlin/com/netflix/spinnaker/config/PeeringAgentConfiguration.kt
```diff
@@ -0,0 +1,71 @@
...
+    jooq: DSLContext,
+    clusterLock: NotificationClusterLock,
+    dynamicConfigService: DynamicConfigService,
+    registry: Registry,
+    @Value("\${pollers.peering.pool-name}") peeredPoolName: String,
+    @Value("\${pollers.peering.id}") peeredId: String,
+    @Value("\${pollers.peering.interval-ms:5000}") pollIntervalMs: Long,
+    @Value("\${pollers.peering.thread-count:200}") threadCount: Int,
+    @Value("\${pollers.peering.chunk-size:100}") chunkSize: Int,
+    @Value("\${pollers.peering.clock-drift-ms:5000}") clockDriftMs: Long
```
cfieber: are we using instance timestamps rather than database timestamps for execution timestamps? I'm not totally sure why we need this

cfieber: (I guess maybe we are instance time-stamping from pre SQL repository days)

cfieber:
I agree this definitely seems like a tradeoff on performance vs maintenance, but I'd be ok starting out this way and considering how we align execution repository and peering in follow up iterations

orca-peering/src/main/kotlin/com/netflix/spinnaker/config/PeeringAgentConfiguration.kt
asher: "MySQL converts TIMESTAMP values from the current time zone to UTC for storage, and back from UTC to the current time zone for retrieval." 

I think you're right as to why millisecond epoch timestamps are still used, but I think the above is a good enough reason to continue down that path. Mostly personal preference (I don't like thinking about or converting TZ's) but bugs can result from the difference in behavior on a locally zoned laptop vs. servers in UTC.

orca-peering/README.md
marchello2000: 👍

orca-peering/README.md
marchello2000: good point, doing now!

orca-peering/src/main/kotlin/com/netflix/spinnaker/config/PeeringAgentConfiguration.kt
marchello2000: We (annoyingly) use instance timestamps for updates (https://github.com/spinnaker/orca/blob/master/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt#L652).
I suppose this is something we can change to have MySQL put in the timestamp on insert/update. But until we have that I added this drift thing (which I don't love)
So if there is consensus to use `UNIX_TIMESTAMP()` in our SQL statements, i am all for changing it (in a future PR)

orca-peering/src/main/kotlin/com/netflix/spinnaker/config/PeeringAgentConfiguration.kt
marchello2000: done

orca-peering/README.md
```diff
@@ -0,0 +1,139 @@
...
+* `pollers.peering.lag > 60` for some period of time (~3 minutes)
+
+
+### Dynamic properties
+The following dynamic properties are exposed and can be controlled at runtime via `DynamicConfigService`.
+
+| Property | Default | Notes |
+|----------|---------|-------|
+|`pollers.peering.enabled`          | `true` | if set to `false` turns of all peering |
+|`pollers.peering.<PEERID>.enabled` | `true` | if set to `false` turns of all peering for peer with give ID |
```
dreynaud:
```suggestion
|`pollers.peering.<PEERID>.enabled` | `true` | if set to `false` turns off all peering for peer with given ID | 
```

orca-peering/README.md
```diff
@@ -0,0 +1,139 @@
+# Orca Peering
+
+This is an semi-experimental approach to solve the problem of having multiple `orca` installations (each with its own database) communicate changes with each other, for instance in a multi-region Spinnaker installation or during a database migration.
+
+Note: the word *instance* is used to refer to `orca` below but most likely this is actually a cluster of `orca` instances that all talk to the same database
+ 
+**Definitions:**
+* `peer`
+    An `orca` instance whose database (can be a read replica) we copy data from. Each `orca` instance has an ID, for example `us-east-1` or `us-west-2`.
```
dreynaud: the term "instance" is maybe a bit overloaded. In this case, we mean something like "installation", right?

orca-peering/README.md
```diff
@@ -0,0 +1,139 @@
...
+  partitionName: LOCAL_PARTITION_NAME
+
+  connectionPools:
+    foreign:
+      jdbcUrl: jdbc:mysql://${sql.foreignBaseUrl}/orca?ADD_YOUR_PREFFERED_CONNECTION_STRING_PARAMS_HERE
+      user: orca_service
+      password: ${sql.passwords.orca_service}
+      connectionTimeoutMs: 5000
+      validationTimeoutMs: 5000
+      maxPoolSize: ${pollers.peering.clockDriftMs}
```
dreynaud: delicious copy-pasta 
```suggestion
      maxPoolSize: ${pollers.peering.threadCount}
```

orca-peering/README.md
```diff
@@ -0,0 +1,139 @@
+# Orca Peering
```
dreynaud: very cool doc IMO 👍

jonsie: Yeah, great overview - 💯

orca-peering/README.md
marchello2000: That's why i added the note above... you think it's not enough?

jonsie: Any reason not to just call it a cluster?  In practice, it is essentially a cluster for peering (even if only running a single node).

orca-sql/src/test/groovy/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepositorySpec.groovy
```diff
@@ -71,17 +72,21 @@ class SqlExecutionRepositorySpec extends ExecutionRepositoryTck<ExecutionReposit
...
-      "namespace")
+    return createExecutionRepository("test")
   }
 
   @Override
   ExecutionRepository createExecutionRepositoryPrevious() {
     new SqlExecutionRepository("test", previousDatabase.context, mapper, new RetryProperties(), 10, 100, "poolName")
   }
 
+  ExecutionRepository createExecutionRepository(String partition) {
```
dreynaud: nit, I think a default argument would work nicely here 
```suggestion
  ExecutionRepository createExecutionRepository(String partition = "test") {
```

orca-peering/README.md
marchello2000: good point, updated

orca-sql/src/test/groovy/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepositorySpec.groovy
marchello2000: can't because there is one already in the baseclass :(

PR #3431, opened 2020-02-12
feat(codebuild): Bind artifacts produced by CodeBuild

Kaixiang-AWS:
AWS CodeBuild is able to produce multiple artifacts to S3 location
as defined by users in their project configuration. However, these
artifacts are not visible to downstream stages, since they are not
Spinnaker artifacts. The two tasks added in this PR will fetch S3
location from CodeBuild and then bind it to the stage output, which
allows downstream stages to consume it as needed.


clareliguori:
Dependent on https://github.com/spinnaker/igor/pull/624

PR #3432, opened 2020-02-12
fix(old pipe cleanup): chunk up results of old pipeline cleanup agent

marchello2000:
Today, the query `OldPipelineCleanupAgent` performs is quite expensive:
```sql
select application, config_id, count(id) as `count` from pipelines where (build_time <= ? and (status IN ('CANCELED','SUCCEEDED','STOPPED','SKIPPED','TERMINAL','FAILED_CONTINUE'))) group by application, config_id having count(id) > ?;
```

Most of the time this query times out on our installation.
This change breaks it up into two queries:
1. first, get apps that have at least the min number of executions required
2. then, use that list, chunk it up and perform a query as above with additional restriction where `application in (.. chunk ..)`

This makes a lot more queries but each query is quicker

### DIFF
--- a/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/cleanup/OldPipelineCleanupPollingNotificationAgent.kt
+++ b/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/cleanup/OldPipelineCleanupPollingNotificationAgent.kt
@@ -98,37 +98,47 @@ class OldPipelineCleanupPollingNotificationAgent(
   private fun performCleanup() {
     val thresholdMillis = Instant.ofEpochMilli(clock.millis()).minus(thresholdDays, ChronoUnit.DAYS).toEpochMilli()
 
-    val pipelineConfigsWithOldExecutions = jooq
-      .select(field("application"), field("config_id"), count(field("id")).`as`("count"))
+    val candidateApplications = jooq
+      .select(field("application"))
       .from(table("pipelines"))
-      .where(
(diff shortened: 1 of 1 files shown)


marchello2000:
note: review with whitespace ignore

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/cleanup/OldPipelineCleanupPollingNotificationAgent.kt
```diff
@@ -98,37 +98,47 @@ class OldPipelineCleanupPollingNotificationAgent(
...
+    val candidateApplications = jooq
+      .select(field("application"))
       .from(table("pipelines"))
-      .where(
-        field("build_time").le(thresholdMillis).and(
-          "status IN (${completedStatuses.joinToString(",") { "'$it'" }})"
-        )
-      )
-      .groupBy(field("application"), field("config_id"))
+      .groupBy(field("application"))
```
asher: No longer need the `groupBy` at all here.

marchello2000: but how would i get the aggregate count?

testing these two queries:
`select application from pipelines group by application having count(id) > 200;`
and
`select application from pipelines having count(id) > 200;`

they don't produce the same result. Or am I misunderstanding what you mean?

PR #3433, opened 2020-02-13
feat(codebuild): Add support for overriding image and buildspec

Kaixiang-AWS:


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartAwsCodeBuildTask.java
```diff
@@ -119,6 +124,19 @@ private void appendSourceVersion(Map<String, Object> requestInput, String source
...
+  private void appendBuildSpec(Map<String, Object> requestInput, String buildspec) {
+    if (buildspec != null && !buildspec.equals("")) {
+      requestInput.put(BUILDSPEC, buildspec);
+    }
+  }
+
+  private void appendImage(Map<String, Object> requestInput, String image) {
+    if (image != null && !image.equals("")) {
+      requestInput.put(IMAGE_LOCATION, image);
+      requestInput.put(IMAGE_CREDENTIALS_TYPE, "SERVICE_ROLE");
```
clareliguori: Does this mean ECR only, and not other registries?

Kaixiang-AWS: This means ALL registries except CodeBuild curated images
https://docs.aws.amazon.com/codebuild/latest/APIReference/API_StartBuild.html#CodeBuild-StartBuild-request-imagePullCredentialsTypeOverride

If it's a private registry, user needs to specify a Secrets Manager ARN that contains credentials.
https://docs.aws.amazon.com/codebuild/latest/APIReference/API_StartBuild.html#CodeBuild-StartBuild-request-registryCredentialOverride

clareliguori: Where would they specify the registry credentials though, does that need to be added here?

Kaixiang-AWS: I assume that user has specified that in the project configuration already. I think fields that could be configured through Spinnaker interface should be those that could be parameterized and changed in different pipeline executions (such as image tag, execution-specific env vars, etc.). I couldn't find a scenario where their registry credentials need to be parameterized so I choose not to add it in the stage definition. What do you think?

clareliguori: Ok makes sense

PR #3436, opened 2020-02-14
fix(aws/cloudformation): Stage won't try to execute deleted change sets.

sergiopena:
By default empty cloud formation changeSets are deleted to keep things tidy. 
There is an unexpected behavior when the user has the executionChangeSet flag enabled and uploads an empty changeSet, the changeset gets deleted and then tries to be executed, causing the stage to fail as the changeSet cannot be found anymore.

This change skips the execution when s a previous changeSet deletion (empty changeSet) is found.

It covers the most used case as we allow to apply empty cloud formation changesets from git and if there are no infra changes proceed with the application deployment without any failure.


sergiopena:
It would be great if this fix it would make it to 1.18.x brach

sergiopena:
@spinnakerbot cherry-pick 1.18

xavileon:
@spinnaker/oss-reviewers PTAL

duftler:
@spinnakerbot cherry-pick 1.18

PR #3441, opened 2020-02-14
fix(clusterCalc): changing ceiling calculation to round for healthyDeployPercentage

gal-yardeni:
Changing the calculation for determent the number of healthy clusters from ceiling to round.


gal-yardeni:
@marchello2000 Let's hope for the best :)

PR #3442, opened 2020-02-14
fix(clouddriver): Adding deleteLaunchConfigurationStage

jeyrschabu:
- added deleteLaunchConfigurationStage to support launch config deletion


jeyrschabu:
I don't see a lot of value in a unit test for this. No real logic to test.

PR #3444, opened 2020-02-15
fix(alicloud): Deployment cannot find cloud vendor

956081712:
When creating a scale group, cannot find cloud vendor。


956081712:
Excuse me, when can I merge

PR #3453, opened 2020-02-19
fix(managed-delivery): Better error messages for ImportDeliveryConfig

luispollo:
This changes the following:
- Treats all 4xx HTTP errors are non-retryable since they're not likely to be resolved for the duration of the stage execution
- Adds special handling for 401 errors from igor, which are generally originating from the SCM system, and which indicate Spinnaker does not have access to the SCM repo to read the delivery config

Hopefully this will make for much better error messages in the UI.


orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -156,13 +169,18 @@ constructor(
...
 
   fun ImportDeliveryConfigContext.incrementAttempt() = this.also { attempt += 1 }
   fun ImportDeliveryConfigContext.toMap() = objectMapper.convertValue<Map<String, Any?>>(this)
 
   companion object {
     const val MAX_RETRIES = 5
+    const val UNAUTHORIZED_SCM_ACCESS_MESSAGE =
+      "HTTP 401 response received while trying to read your delivery config file. " +
+      "Spinnaker may be missing permissions in your source code repository to read the file. " +
+      "Please consult your Spinnaker operator for details on how to add those permissions."
```
emjburns: This last line is probably not necessary - I think it just adds confusion.

emjburns:
Error message change requested, but other than that looks great

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
luispollo: OK, will remove.

PR #3454, opened 2020-02-19
fix(Authorization): use GET method for collectJob

AbdulRahmanAlHamali:
This goes with the change in:
https://github.com/spinnaker/clouddriver/pull/4328


ttomsu:
@maggieneterval - this change (coupled with the Clouddriver change linked in the description) may have an effect on V1 users, but since that's going away, should we care? IIUC, this means that V1 users who used this method previously required WRITE permissions will now be allowed with just the READ permission.

ttomsu:
(@maggieneterval you can replace the do not merge label with the ready to merge if you're good with the changes)

maggieneterval:
> IIUC, this means that V1 users who used this method previously required WRITE permissions will now be allowed with just the READ permission. 

@ttomsu thanks for the heads-up. To confirm my understanding:

- The V1 Run Job deletes the pod, so should require WRITE permission
- The V2 Run Job does not delete the pod, so should not require WRITE permission
- The Clouddriver change makes it so that neither V1 nor V2 Run Jobs require WRITE permission

Shouldn't we still require WRITE permission for V1 Run Jobs? Presumably this change will be released with Spinnaker 1.19, and V1 will not be removed until 1.21.

ttomsu:
Your understanding matches my understanding. 

> Shouldn't we still require WRITE permission for V1 Run Jobs?

Ideally yes, but I think Clouddriver checks the permission of the call before it determines whether this applies to a V1 provider or a V2 provider. So it's either READ or WRITE for both providers.

So I guess technically this would be introducing a bug in the V1 provider, but since it's going away and we're actively discouraging and pushing people away from it, I figure it's a calculated risk. You have more insight who and how many people are planning on continuing to use the V1 provider given it's near-term EOL.

maggieneterval:
> Ideally yes, but I think Clouddriver checks the permission of the call before it determines whether this applies to a V1 provider or a V2 provider. So it's either READ or WRITE for both providers.

Ugh, this kind of entanglement is exactly the type of technical debt I can't wait to eliminate by removing V1.

@AbdulRahmanAlHamali what would it look like to bifurcate the V1 and V2 logic so they can check different permissions? Although we are actively discouraging V1 adoption, it seems unacceptable to knowingly introduce a bug while we still support it.

AbdulRahmanAlHamali:
Thanks @maggieneterval for expanding my vocabulary with the word "bifurcate" :smile: 

I'll give a shot to moving the authorization logic inside the function instead of the decorator, and using the provider type to make the authorization decision

AbdulRahmanAlHamali:
This and the other PR should be merged together
https://github.com/spinnaker/clouddriver/pull/4328
Since the other PR changes the method from `POST` to `GET`

ezimanyi:
Despite the above comment, the clouddriver PR was merged without this PR, breaking the run job stage.  As this PR doesn't allow edits from maintainers, I can't merge in master and thus merge this PR; closing it in favor of #3585 which is exactly the same code, just with master merged in.

AbdulRahmanAlHamali:
Thank you Eric

PR #3455, opened 2020-02-19
fix(health): When waiting for down instances, consider 'Starting' to be down.

christopherthielen:
If an app is stuck in Starting (due to a bug, etc), and an operator tries to fix this by deploying a new version, spinnaker would disable the cluster, but then time out waiting for the instances to go Down because it didn't consider Starting as a down state


dreynaud:
In principle, this makes sense.

In practice, we override the status to OutOfService on disable, is that not actually overriding the Starting state? Or is this to cover the case where we fail to override the status for some reason?

dreynaud:
Capturing meatspace discussion:
- you would think that the per-instance override works but it's actually not clear, and also it doesn't help if a new instance boots (hence with no override) but also stuck in Starting
- you would think the per-cluster disable override in discovery would take precedence over Starting but it's also actually not entirely clear
- there might be some weird side effects if we were to count Starting as down for  `someAreUpAndNoneAreDown` purposes, but with this PR it is only down for `someAreDownAndNoneAreUp` purposes. It makes it a little bit weird: is Starting an "up" or a "down"? Well, it depends...

In conclusion, ship it 👍

PR #3459, opened 2020-02-22
feat(interlink): handle executions in foreign partitions

dreynaud:
We can optionally inject an Interlink object in SqlExecutionRepository. When present, it can be used to notify a pubsub topic of a request that the execution repository could not handle (instead of throwing a `ForeignExecutionException`). Other orca clusters can then subscribe to that topic, unpack the message about the failed action and potentially replay it on their local repository if they handle that partition.

To avoid pathologic cases where 2 orca clusters end up endlessly bouncing the same message back and forth across the interlink, we add `MessageFlagger` to act as a circuit breaker. Based on a fixed size evicting queue, it inspects the fingerprint of the last ${maxSize} events published by this instance, and if it detects that we are over some threshold, it will not let the message get published on the topic. One example where this may happen is the following scenario with an orphaned execution that everybody considers as belonging to someone else:

1. execution repo 1 has an execution with id I and partition `apple`
2. execution repo 2 has an execution with the same id I and partition `banana`
3. when a user makes a request to orca 1 with partition `banana` to delete I, it is considered a foreign execution and a message is sent on the interlink
4. orca 2 receives the interlink message and tries to delete it from execution repo 2, but it is also considered foreign (because it has partition `apple`)
5. therefore, orca 2 sends a message on the interlink, and so on

Caveats:
- this change only handles pause, resume, cancel and delete execution events
- to cover every use case, we will also need to handle restart stage, skip wait, manual judgement...


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
```diff
@@ -156,6 +156,15 @@ Execution retrievePipelineForCorrelationId(@Nonnull String correlationId)
 
   List<String> retrieveAllExecutionIds(@Nonnull ExecutionType type);
 
+  @Nullable
+  default String getPartition() {
+    throw new UnsupportedOperationException();
```
marchello2000: should the default impl return `null`? that way default `handlesPartition` doesn't blow up?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
```diff
@@ -156,6 +156,15 @@ Execution retrievePipelineForCorrelationId(@Nonnull String correlationId)
...
+  }
+
+  default boolean handlesPartition(@Nullable String partition) {
```
marchello2000: nit:
```suggestion
  default boolean handlesPartition(@Nullable String partitionOfExecution) {
```

otherwise it's confusing whose partition you're talking about, esp in the `if`

orca-interlink/src/main/java/com/netflix/spinnaker/config/InterlinkConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+  @Bean
+  @ConditionalOnProperty({"pubsub.enabled", "pubsub.amazon.enabled"})
+  public AmazonPubsubMessageHandlerFactory amazonPubsubMessageHandlerFactory(
+      ObjectMapper objectMapper, ExecutionRepository repository) {
+    return new AmazonPubsubMessageHandlerFactory() {
+      @Override
+      public AmazonPubsubMessageHandler create(
+          AmazonPubsubProperties.AmazonPubsubSubscription subscription) {
+        if (!"interlink".equals(subscription.getName())) {
+          return null;
```
marchello2000: should this log?

orca-interlink/src/main/java/com/netflix/spinnaker/config/InterlinkConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+  PreprocessorConfiguration.class,
+  PluginsAutoConfiguration.class,
+  PubsubConfig.class,
+  AmazonPubsubConfig.class
+})
+@ConditionalOnProperty("interlink.enabled")
+@EnableConfigurationProperties(InterlinkConfigurationProperties.class)
+public class InterlinkConfiguration {
+  @Bean
+  @ConditionalOnProperty({"pubsub.enabled", "pubsub.amazon.enabled"})
```
marchello2000: who not put these `conditionalOnProp` on the whole config class since both beens share the same props?

orca-interlink/src/main/java/com/netflix/spinnaker/config/InterlinkConfigurationProperties.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package com.netflix.spinnaker.config;
+
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@Data
+@ConfigurationProperties(prefix = "interlink")
+public class InterlinkConfigurationProperties {
```
marchello2000: maybe worth a quick class comment/explanation (here or the flagger)?

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/Interlink.java
```diff
@@ -0,0 +1,80 @@
...
+      PubsubPublishers publishers,
+      ObjectMapper objectMapper,
+      MessageFlagger flagger,
+      Registry registry) {
+    this.objectMapper = objectMapper;
+    this.flagger = flagger;
+
+    publisher =
+        publishers.getAll().stream()
+            .filter(pubsubPublisher -> "interlink".equals(pubsubPublisher.getTopicName()))
```
marchello2000: maybe worth making `"interlink"` a const somewhere?

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/Interlink.java
```diff
@@ -0,0 +1,80 @@
...
+            .findFirst()
+            .orElse(null);
+
+    if (publisher == null) {
+      throw new SystemException(
```
marchello2000: I added a `ConfigurationException` in `kork` which I think would be appropriate here
https://github.com/spinnaker/kork/pull/527/files#diff-97ad90a3d82025731986034b6e2fc9cbR19

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/Interlink.java
```diff
@@ -0,0 +1,80 @@
...
+    this.flaggedCounter =
+        registry.counter(
+            "pubsub." + publisher.getPubsubSystem() + ".flagged",
+            "subscription",
+            publisher.getName());
+  }
+
+  public void publish(InterlinkEvent event) {
+    if (flagger.isFlagged(event)) {
+      log.warn("Event {} has been flagged, not publishing to interlink", event);
```
marchello2000: nit:
```suggestion
      log.warn("Event {} has been flagged for being published too many times, not publishing to interlink", event);
```

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/MessageFlagger.kt
```diff
@@ -0,0 +1,41 @@
...
+  val queue: EvictingQueue<HashCode> = EvictingQueue.create(maxSize)
+  val hasher = Hashing.goodFastHash(64)
+
+  init {
+    Preconditions.checkArgument(maxSize >= threshold, "maxSize (%s) has to be larger than threshold (%s)", maxSize, threshold)
+    Preconditions.checkArgument(threshold > 0, "threshold (%s) has to be > 0", threshold)
+  }
+
+  fun isFlagged(event: InterlinkEvent): Boolean {
+    val hashCode = hasher.hashString("${event.eventType}:${event.executionId}:${event.executionType}", Charsets.UTF_8)
```
marchello2000: i still think we need the "original submission" timestamp here.

Also, do you even need the hashcode here? you can just push the fingerprint into the queue - I don't think the hash buys you much, these are pretty short strings

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/aws/InterlinkAmazonMessageHandler.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  @VisibleForTesting
+  void handleInternal(InterlinkEvent event) {
+    if (executionRepository.handlesPartition(event.getPartition())) {
+      event.applyTo(executionRepository);
+    } else {
+      log.debug(
+          "Execution repository with local partition {} does not handle this event's partition {}, "
```
marchello2000: nit: 
```suggestion
          "Execution repository with local partition {} does not handle this events' partition {}, "
```

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/events/InterlinkEvent.java
```diff
@@ -0,0 +1,58 @@
...
+    DELETE,
+    RESUME,
+  }
+
+  @JsonIgnore
+  EventType getEventType();
+
+  Execution.ExecutionType getExecutionType();
+
+  String getExecutionId();
```
marchello2000: i think it's a bit weird to have the getters here but not the backing field... maybe it would make for lombok/jackson nightmare, but i almost feel like this `InterlinkEvent` is more of a base class than an interface.

orca-interlink/orca-interlink.gradle
```diff
@@ -0,0 +1,31 @@
+/*
```
ajordens: This module seems more like a `orca-interlink-aws` ... ?

Someone could come along and re-implement the pub/sub bits against something other than `aws` right?

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/MessageFlagger.kt
```diff
@@ -0,0 +1,41 @@
...
+class MessageFlagger(maxSize: Int, val threshold: Int) {
+  val queue: EvictingQueue<HashCode> = EvictingQueue.create(maxSize)
+  val hasher = Hashing.goodFastHash(64)
+
+  init {
+    Preconditions.checkArgument(maxSize >= threshold, "maxSize (%s) has to be larger than threshold (%s)", maxSize, threshold)
+    Preconditions.checkArgument(threshold > 0, "threshold (%s) has to be > 0", threshold)
+  }
+
+  fun isFlagged(event: InterlinkEvent): Boolean {
```
marchello2000: it's a bit awkward that `isFlagged` has a side-effect (of `queue.push`). But I am failing to come up with a better name

orca-interlink/src/test/groovy/com/netflix/spinnaker/orca/interlink/aws/InterlinkAmazonMessageHandlerSpec.groovy
```diff
@@ -0,0 +1,51 @@
...
+    1 * executionRepo.delete(PIPELINE, "id")
+
+    when:
+    handler.handleInternal(new DeleteInterlinkEvent(PIPELINE, "id").withPartition("foreign"))
+
+    then:
+    1 * executionRepo.handlesPartition("foreign") >> false
+    0 * executionRepo.delete(PIPELINE, "id")
+  }
+}
```
marchello2000: i assume this is still coming.... but i think it's worth adding similar tests for other events

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/cleanup/OldPipelineCleanupPollingNotificationAgent.kt
```diff
@@ -88,7 +88,7 @@ class OldPipelineCleanupPollingNotificationAgent(
       log.info("Agent {} started", notificationType)
       performCleanup()
     } catch (e: Exception) {
-      log.error("Agent {} failed to perform cleanup", e)
+      log.error("Agent {} failed to perform cleanup", javaClass, e)
```
marchello2000: 👍

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -124,48 +131,57 @@ class SqlExecutionRepository(
...
+      storeExecutionInternal(dslContext, execution)
+    }
+  }
+
+  override fun pause(type: ExecutionType, id: String, user: String?) {
+    doForeignAware(PauseInterlinkEvent(type, id, user)) {
+      execution, dslContext ->
+      if (execution.status != RUNNING) {
+        throw UnpausablePipelineException("Unable to pause pipeline that is not RUNNING " +
+          "(executionId: $id, currentStatus: ${execution.status})")
```
marchello2000: nit:
```suggestion
          "(executionId: ${execution.id}, currentStatus: ${execution.status})")
```

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -174,22 +190,16 @@ class SqlExecutionRepository(
...
-            execution.status = RUNNING
-            execution.paused?.resumedBy = user
-            execution.paused?.resumeTime = currentTimeMillis()
-            storeExecutionInternal(it, execution)
-          }
+    doForeignAware(ResumeInterlinkEvent(type, id, user, ignoreCurrentStatus)) {
+      execution, dslContext ->
+      if (!ignoreCurrentStatus && execution.status != PAUSED) {
+        throw UnresumablePipelineException("Unable to resume pipeline that is not PAUSED " +
+          "(executionId: $id, currentStatus: ${execution.status}")
```
marchello2000:
```suggestion
          "(executionId: ${execution.id}, currentStatus: ${execution.status}")
```

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
dreynaud: you're probably right, I was wondering about null vs unsupported

orca-interlink/src/main/java/com/netflix/spinnaker/config/InterlinkConfiguration.java
dreynaud: it wouldn't hurt, I'll add it

orca-interlink/src/main/java/com/netflix/spinnaker/config/InterlinkConfiguration.java
dreynaud: I think it makes sense to keep them on the beans, we could imagine somebody adding beans in the same config class that depend on `pubsub.google.enabled` or `pubsub.kafka.enabled`. I.e. this is the global interlink config, not the interlink config for Amazon

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/Interlink.java
dreynaud: ah yes indeed 💯

orca-interlink/src/main/java/com/netflix/spinnaker/config/InterlinkConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+@Configuration
+@Import({
+  PreprocessorConfiguration.class,
+  PluginsAutoConfiguration.class,
```
dreynaud: just realized that `PreprocessorConfiguration` and `PluginsAutoConfiguration` have copy-pasta'ed their way onto this, will remove

orca-interlink/src/test/groovy/com/netflix/spinnaker/orca/interlink/aws/InterlinkAmazonMessageHandlerSpec.groovy
dreynaud: I wasn't going to add these, it is testing for the foreign/vs local partition matching and there is nothing specific about the event type to test

orca-interlink/orca-interlink.gradle
dreynaud: yes, I was thinking the other pubsub systems could be handled in the same module. Do you think it's worth it to carve it out in `orca-interlink-aws`?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
dreynaud: not my code but fair :)

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/aws/InterlinkAmazonMessageHandler.java
dreynaud: I think my version is correct though, there is only one event. So we don't handle the partition of this event, aka this event's partition

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/MessageFlagger.kt
dreynaud: yes I see what you mean. Some alternatives:

`fun push(event: InterlinkEvent): Boolean` (push that message, returns if it's flagged or not)
`fun process(event: InterlinkEvent): Unit`, but throws a MessageFlaggedException instead of returning a boolean. This _is_ supposed to be exceptional, so maybe an exception makes sense here

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/aws/InterlinkAmazonMessageHandler.java
marchello2000: i always get this wrong, so you're prob right but it still feels wrong to me :)

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/MessageFlagger.kt
marchello2000: something about throwing an exception feels wrong. I can't articulate it. but it does make it explicit and it is pretty clean what's going with an exception...

orca-interlink/src/main/java/com/netflix/spinnaker/config/InterlinkConfiguration.java
dreynaud: but per @ajordens' comment, maybe this should leave in a separate `orca-interlink-aws` module? I'm started to get module fatigue a little bit 😅

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/MessageFlagger.kt
dreynaud: The strings are pretty short for now, but with the hashcode I feel it's a little more generic. Each fingerprint is also going to be hashed only once but compared dozens of times, so I don't find it shocking

Will add some kind of timestamp based on our offline discussion 👍

orca-interlink/src/test/groovy/com/netflix/spinnaker/orca/interlink/aws/InterlinkAmazonMessageHandlerSpec.groovy
marchello2000: it also tests that the event does the right think in the `applyTo` call - maybe there is a better way to test that though

orca-interlink/src/main/java/com/netflix/spinnaker/config/InterlinkConfiguration.java
marchello2000: 👍 oh, yeah, i was wondering about those but i gusss i forgot to say something, ooops!

orca-interlink/src/test/groovy/com/netflix/spinnaker/orca/interlink/aws/InterlinkAmazonMessageHandlerSpec.groovy
dreynaud: in 4468ce2230c3d22ed2137d9b26c35aca221b684c

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/MessageFlagger.kt
dreynaud: made it more clear that I intend the fingerprint to be more opaque/generic/subject to change here: 6e4393260bba1435e7c19ea2c3b911cfc4e2a10c

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
```diff
@@ -156,6 +156,23 @@ Execution retrievePipelineForCorrelationId(@Nonnull String correlationId)
...
+   * means that it handles any partition (in other words, this execution repository is not partition
+   * aware)
+   */
+  @Nullable
+  default String getPartition() {
+    return null;
+  }
+
+  default boolean handlesPartition(@Nullable String partitionOfExecution) {
+    return partitionOfExecution == null // executions with no partition can be handled by anyone
```
marchello2000: nit:
```suggestion
    return partitionOfExecution == null // executions with no partition are implied to be owned by current orca
```

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/Interlink.java
```diff
@@ -0,0 +1,83 @@
...
+            "pubsub." + publisher.getPubsubSystem() + ".flagged",
+            "subscription",
+            publisher.getName());
+  }
+
+  public void publish(InterlinkEvent event) {
+    try {
+      flagger.process(event);
+    } catch (MessageFlaggedException e) {
+      log.warn("Will not publish event to interlink", e);
```
marchello2000:
```suggestion
      log.warn("Will not publish event {} to interlink", event, e);
```

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/MessageFlagger.kt
```diff
@@ -0,0 +1,76 @@
...
+ * 3. when a user makes a request to orca 1 with partition `banana` to delete I, it is considered a foreign execution
+ *    and a message is sent on the interlink
+ * 4. orca 2 receives the interlink message and tries to delete it from execution repo 2, but it is also considered
+ *    foreign (because it has partition `apple`)
+ * 5. therefore, orca 2 sends a message on the interlink, and so on
+ *
+ * @param maxSize the maximum number of fingerprints to consider
+ * @param threshold a message with fingerprint F will be flagged if the number of times F is present in the queue is
+ *    over the threshold
+ */
```
marchello2000: lookback?

orca-interlink/orca-interlink.gradle
marchello2000: it seems like there is very little aws-y things in here, just the config or am I missing something if that's the case i think prob easier not to make another module?

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/MessageFlagger.kt
dreynaud: in 6e4393260bba1435e7c19ea2c3b911cfc4e2a10c

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/MessageFlagger.kt
dreynaud: in 486d43dd630d6b9973f6d67562d4f76e8d2eb3a0

PR #3460, opened 2020-02-24
feat(rosco): Add HELM3 Template Renderer

evansb:
Parts of https://github.com/spinnaker/spinnaker/issues/5474

This add `HELM3` as valid template renderer type.


ethanfrogers:
No need to address this right now, mostly just a comment for the future - I think we should do a little bit of refactoring in this area. There's probably a more generic way to handle creating those `BakeManifestRequests` than with a `switch`.

PR #3461, opened 2020-02-24
fix(peering): disallow changing partition name and fix cleanup agents

marchello2000:
This change adds three fixes for peering:
* Once a partition is set on the cluster/db it should never be changed (this is likely a misconfiguration issue)
* Added `where` clause to the pipeline/execution cleanup agents to only clean up executions in it's own partition (if one is set)
* Added prevention to not accidentally delete ALL executions during peering

### DIFF
--- a/orca-peering/README.md
+++ b/orca-peering/README.md
@@ -20,8 +20,8 @@ This is an semi-experimental approach to solve the problem of having multiple `o
 
 
 The peering mechanism accomplishes a few things:
-1. (complete) Peer (copy) executions (both pipelines and orchestrations) from a database of a peer to the local cluster database
-2. (in-progress) Allow for an `orca` cluster to perform actions on a foreign execution (e.g. executions running on a peer)
+1. Peer (copy) executions (both pipelines and orchestrations) from a database of a peer to the local cluster database
+2. Allow for an `orca` cluster to perform actions on a foreign execution (e.g. executions running on a peer, see `orca-interlink` module)
 3. (still to come) Take ownership and resume an execution previously operated on by a peer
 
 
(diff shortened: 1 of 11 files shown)


orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -84,6 +85,31 @@ class SqlExecutionRepository(
 
   private val log = LoggerFactory.getLogger(javaClass)
 
+  init {
+    log.info("Creating SqlExecutionRepository with partition=$partitionName with pool=$poolName")
+
+    withPool(poolName) {
+      jooq.transactional {
+        val record: Record? = jooq.fetchOne(table("partition_name"))
```
dreynaud: what happens if there are multiple rows in that table?

orca-peering/README.md
```diff
@@ -135,3 +136,4 @@ The following dynamic properties are exposed and can be controlled at runtime vi
 |----------|---------|-------|
 |`pollers.peering.enabled`          | `true` | if set to `false` turns off all peering |
 |`pollers.peering.<PEERID>.enabled` | `true` | if set to `false` turns off all peering for peer with given ID | 
+|`pollers.peering.max-allowed-delete-count`| 1000 | maximum number of executions that are allowed to be deleted in a single run of the agent |
```
dreynaud: It might be interesting to discuss that default value a little. I guess for the most part that's the volume of deletions we might expect from one of the cleanup agents at Netflix scale. But one could imagine that dropping 1000 executions on a smaller Spinnaker installation is dropping the entire thing, so how big 1000 is relative

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -160,6 +160,12 @@ class PeeringAgent(
 
     if (pipelineIdsToMigrate.isNotEmpty() || pipelineIdsToDelete.isNotEmpty()) {
       log.debug("Found ${completedPipelineKeys.size} completed $executionType candidates with ${migratedPipelineKeys.size} already copied for peering, ${pipelineIdsToMigrate.size} still need copying and ${pipelineIdsToDelete.size} need to be deleted")
+      val maxDeleteCount = dynamicConfigService.getConfig(java.lang.Integer::class.java, "pollers.peering.max-allowed-delete-count", Integer(1000))
+      if (pipelineIdsToDelete.size > maxDeleteCount.toInt()) {
+        log.error("Number of pipelines to delete (${pipelineIdsToDelete.size}) > threshold ($maxDeleteCount) - not performing deletes - if this is expected you can set the pollers.peering.max-allowed-delete-count property to a larger number")
```
dreynaud: for visibility: the third change is a safety mechanism to avoid deleting the world here

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
marchello2000: it should never happen... if it does it will throw, i will wrap the exception into something useful, good catch

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
marchello2000: thank you - amended the commit but github doesn't pick it up, fixing

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -84,6 +85,31 @@ class SqlExecutionRepository(
 
   private val log = LoggerFactory.getLogger(javaClass)
 
+  init {
+    log.info("Creating SqlExecutionRepository with partition=$partitionName with pool=$poolName")
```
robzienert:
```suggestion
    log.info("Creating SqlExecutionRepository with partition=$partitionName and pool=$poolName")
```

orca-sql/src/main/resources/db/changelog/20200221-partition-name-table.yml
```diff
@@ -0,0 +1,22 @@
...
+                  name: id
+                  type: int
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: name
+                  type: varchar(255)
+                  constraints:
+                    nullable: true
```
robzienert: Why would `name` be nullable?

marchello2000: yeah, i guess it shouldn't be since we won't insert unless we have a partition name, i will change it

marchello2000: actually, i think it makes it easier to fix up in case you actually want to go back to a `null` partition name

PR #3463, opened 2020-02-25
feat(managed-delivery): Include detailed error info in ImportDeliveryConfig

luispollo:
Include detailed error information added to keel in https://github.com/spinnaker/keel/pull/821 so it can be displayed in the UI.

Another piece to https://github.com/spinnaker/keel/issues/818

### DIFF
--- a/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
+++ b/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.orca.keel.task
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.convertValue
+import com.fasterxml.jackson.module.kotlin.readValue
 import com.netflix.spinnaker.orca.ExecutionStatus
 import com.netflix.spinnaker.orca.KeelService
 import com.netflix.spinnaker.orca.RetryableTask
@@ -115,6 +116,8 @@ constructor(
           // ...but give users a hint about 401 errors from igor/scm
           if (e.response?.status == 401 && URL(e.url).host.contains("igor")) {
             UNAUTHORIZED_SCM_ACCESS_MESSAGE
+          } else if (e.response?.status == 400 && URL(e.url).host.contains("keel") && e.response!!.body.length() > 0) {
+            objectMapper.readValue<Map<String, Any?>>(e.response!!.body.`in`())
           } else {
             "Non-retryable HTTP response ${e.response?.status} received from downstream service: ${e.friendlyMessage}"
           }
@@ -145,9 +148,13 @@ constructor(
     }
   }
 
-  private fun buildError(errorMessage: String): TaskResult {
-    log.error(errorMessage)
-    return TaskResult.builder(ExecutionStatus.TERMINAL).context(mapOf("error" to errorMessage)).build()
+  private fun buildError(error: Any): TaskResult {
+    if (error is Map<*, *>) {
+      log.error(error["message"]?.toString())
+    } else {
+      log.error(error.toString())
+    }
+    return TaskResult.builder(ExecutionStatus.TERMINAL).context(mapOf("error" to error)).build()
   }
 
   override fun getBackoffPeriod() = TimeUnit.SECONDS.toMillis(30)
--- a/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
+++ b/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
@@ -115,9 +115,9 @@ constructor(
         // just give up on 4xx errors, which are unlikely to resolve with retries
         buildError(
           // ...but give users a hint about 401 errors from igor/scm
-          if (response.status == 401 && URL(e.url).host.contains("igor")) {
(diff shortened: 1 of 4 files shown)


luispollo:
I messed up a rebase. Fixing it now.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -115,6 +116,8 @@ constructor(
           // ...but give users a hint about 401 errors from igor/scm
           if (e.response?.status == 401 && URL(e.url).host.contains("igor")) {
             UNAUTHORIZED_SCM_ACCESS_MESSAGE
+          } else if (e.response?.status == 400 && URL(e.url).host.contains("keel") && e.response!!.body.length() > 0) {
```
emjburns: it is safe to do `response!!` here? What do you loose by instead using `?` so we don't throw an NPE if the response does not exist? (lns 119 & 120)

luispollo: Yes because nullability is checked in the first condition (`e.response?.status == 400`). What I gain is that I don't have to add a nested `if` or similar to access `.length` on `e.response.body` (which might be null if I used the `?`).

emjburns: I'd rather another if or chain of `?` instead of `!!`. I don't think we can say for certain that when `e.response` is not null then `body` will not be null.

luispollo: It's the other way around, actually. The expression `e.response?.status == 400` will ensure `e.response` is not `null`. Given that `&&` is a short-circuit operator, the remaining expressions will never get evaluated if `e.response` is `null`. So, when we get to `e.response!!.body`, `e.response` is guaranteed not to be `null`.

I can change it, but wanted to make sure we're on the same page first.

luispollo: P.S. Actually, if you expand the block of code immediately above, you'll see that the only way we even get here is when the status code in the response is `400..499`, which means the response is never `null` here.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -111,13 +111,14 @@ constructor(
...
         buildError(
           // ...but give users a hint about 401 errors from igor/scm
-          if (e.response?.status == 401 && URL(e.url).host.contains("igor")) {
+          if (response.status == 401 && URL(e.url).host.contains("igor")) {
             UNAUTHORIZED_SCM_ACCESS_MESSAGE
-          } else if (e.response?.status == 400 && URL(e.url).host.contains("keel") && e.response!!.body.length() > 0) {
-            objectMapper.readValue<Map<String, Any?>>(e.response!!.body.`in`())
+          } else if (response.status == 400 && URL(e.url).host.contains("keel") && response.body.length() > 0) {
+            objectMapper.readValue<Map<String, Any?>>(response.body.`in`())
           } else {
```
luispollo: Is this better @emjburns?

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -145,9 +149,14 @@ constructor(
...
-    log.error(errorMessage)
-    return TaskResult.builder(ExecutionStatus.TERMINAL).context(mapOf("error" to errorMessage)).build()
+  private fun buildError(error: Any): TaskResult {
+    val normalizedError = if (error is Map<*, *>) {
+      error["error"] ?: error
+    } else {
+      error.toString()
+    }
+    log.error(normalizedError.toString())
+    return TaskResult.builder(ExecutionStatus.TERMINAL).context(mapOf("error" to normalizedError)).build()
```
marchello2000: does `deck` actually parse the `error` key in a stage context?
My general recommendation for making sure errors show up in the UI is to throw and let `RunTaskHandler` put the error info into the json in a consistent way that deck understands
(https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt#L174)

luispollo: Thanks, @marchello2000. The challenge here is precisely that these errors are not in a format that deck would render in a meaningful/helpful way without additional parsing. In fact, I've started a PR to just that (WIP): https://github.com/spinnaker/deck/pull/7959/

Would you stick by your recommendation in this case? For the record, I did discuss this with @erikmunson and, while I understand I could do the parsing in orca and include markdown in the text, it felt like that responsibility should lie in deck.

marchello2000: in an ideal world.... i would like us (on orca side) to have a nice API that sets error in a nice way (other than throwing) but alas... that is prob far away (thought maybe we will get there with plugins orca refactor).

So long as you're adding stuff to deck to parse it, i think it's ok. But it does mean that anyone who is consuming this via API will have to handle both cases...

PR #3465, opened 2020-02-26
fix(interlink): apply events to a queue+repo operator

dreynaud:
For some actions (like cancel, resume, pause), it is not sufficient to invoke the correspond execution repository call, we also need to create some events on the message queue (e.g. to actually cancel running tasks).

The task controller already has the logic to update both the queue and the execution repo, so I'm lifting some of it in a common but somewhat awkwardly named CompoundExecutionOperator. As a result, the same logic will be executed by the task controller and interlink events.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
@@ -91,16 +91,8 @@ public class CompoundExecutionOperator {
       ExecutionType executionType,
       String executionId) {
     try {
-      retrySupport.retry(
-          () -> {
-            runnerAction.run();
-            return true;
-          },
-          5,
-          Duration.ofMillis(100),
-          false);
-
-      repositoryAction.run();
+      runWithRetries(runnerAction);
+      runWithRetries(repositoryAction);
     } catch (Exception e) {
       log.error(
           "Failed to {} execution with executionType={} and executionId={}",
@@ -109,5 +101,16 @@ public class CompoundExecutionOperator {
           executionId,
           e);
     }
+  }
+
+  private void runWithRetries(Runnable action) {
+    retrySupport.retry(
+        () -> {
+          action.run();
+          return true;
+        },
+        5,
+        Duration.ofMillis(100),
+        false);
   }
 }
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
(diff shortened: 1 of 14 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/ExecutionRunner.java
```diff
@@ -21,22 +21,21 @@
...
-  void start(@Nonnull Execution execution) throws Exception;
+  void start(@Nonnull Execution execution);
 
-  default void restart(@Nonnull Execution execution, @Nonnull String stageId) throws Exception {
+  default void restart(@Nonnull Execution execution, @Nonnull String stageId) {
     throw new UnsupportedOperationException();
   }
 
-  default void reschedule(@Nonnull Execution execution) throws Exception {
+  default void reschedule(@Nonnull Execution execution) {
```
marchello2000: shouldn't these be `throws UnsupportedOperationException`? I guess it doesn't matter because it's an interface?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private void doInternal(
+      Runnable runnerAction,
+      Runnable repositoryAction,
+      String action,
+      ExecutionType executionType,
+      String executionId) {
+    try {
+      retrySupport.retry(
```
marchello2000: why do you need the retry here? I didn't see it on the controller side..?
If you get rid of it, you can pass in a single lambda/runnable instead of two

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/ExecutionRunner.java
dreynaud: yeah and also UnsupportedOperationException is an unchecked exception

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
dreynaud: I thought it made sense to add, especially in a somewhat central place like this. Do you think it's unnecessary?

marchello2000: all the runner does is push a thing onto the queue... Also, from this methods perspective it's just a runnable so it seems weird to distinguish one runnable and retry it but not the other one. I think it would be cleaner more obvious what's going on if a) there is one runnable and (optionally) b) it's either all retried or not

dreynaud: It is weird that the repo action is not retried, I think it is more consistent in 9da1e69f4579ad12166e9a8db59f67d78af941ab.

I kept the 2 runnables, because I don't want to retry updating the queue if the repo action fails. Seems reasonable?

PR #3466, opened 2020-02-26
fix(peering): Don't use `null` partition on local db during peering

marchello2000:
For historical reasons, `partition=null` is set on all current executions (since sqlrepo runs without a partition).
So when checking if an execution is local we need to check if `parition == MY_PARTITION` OR `partition is null`.
However, this check is not correct when copying from the peer. From the peer we need to get `partition == peerId or partition is null`,
but locally it must always be `partition == peerId` because we force set the partition during copy.
Without that we would think that we need to delete all our local executions since they would not show up in the peer query
This change fixes this issue

### DIFF
--- a/orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
+++ b/orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
@@ -112,6 +112,7 @@ class PeeringAgent(
     log.debug("Starting active $executionType copy for peering")
 
     val activePipelineIds = srcDB.getActiveExecutionIds(executionType, peeredId)
+      .plus(srcDB.getActiveExecutionIds(executionType, null))
 
     if (activePipelineIds.isNotEmpty()) {
       log.debug("Found ${activePipelineIds.size} active $executionType, copying all")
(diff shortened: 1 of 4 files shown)


orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -138,8 +138,8 @@ class PeeringAgent(
     log.debug("Starting completed $executionType copy for peering with $executionType updatedAfter=$updatedAfter")
 
     // Compute diff
-    val completedPipelineKeys = srcDB.getCompletedExecutionIds(executionType, peeredId, updatedAfter)
-    val migratedPipelineKeys = destDB.getCompletedExecutionIds(executionType, peeredId, updatedAfter)
+    val completedPipelineKeys = srcDB.getCompletedExecutionIds(executionType, peeredId, true, updatedAfter)
+    val migratedPipelineKeys = destDB.getCompletedExecutionIds(executionType, peeredId, false, updatedAfter)
```
dreynaud: Good catch, but I'm a little uncomfortable with an unassuming boolean getting passed to the same function here, the two function calls read almost the same but behave crucially differently. Since the boolean changes the semantics of `getCompletedExecutionIds`, I think it would make sense to create two higher-level functions with more descriptive names. For instance, something like:

```
getCompletedExecutionIdsToMigrate(executionType, peeredId, updatedAfter) -> calls with allowNullPartition = true
getMigratedExecutionIds(executionType, peeredId, updatedAfter) -> calls with allowNullPartition = false
```

marchello2000: yeah, i agree. What do you think about making two separate calls: e.g.
```java
srcDB.getCompletedExecutionIds(executionType, peeredId, updatedAfter)
srcDB.getCompletedExecutionIds(executionType, null, updatedAfter)
```
and concatting the result - that i think makes it obvious what's up and doesn't leak this `null` parition history into the sql access layer?

PR #3467, opened 2020-02-27
fix(peering): Capture the source execution BEFORE the source stage during peering

marchello2000:
It's important that we capture the executions BEFORE we capture their stages because we use the execution's `updated_at` for our diff calculation.
Before this change, it was possible to see a peered execution that was `SUCCEEDED` but a stage within that execution still showing as `RUNNING`.
This happens because we would grab the stages from the source DB, persist them to the dest DB (this can take some time) and only then we would
grab the executions and persist those. But in this flow our peer could have updated the stage in the time between we capture stages and capture
the execution and we would forever miss that update (if the execution was completed now)

It's totally fine for the stages to be "newer" than their execution since that will be fixed up in the next agent run


marchello2000:
going to mark it as "ready" since i want it to run in our env overnight. let me know if you have feedback

PR #3470, opened 2020-02-27
feat(peering): Add ability to peer a set of specific executions

marchello2000:
This change allows an external caller (coming soon!) to ask the peering agent to peer specific executions
This is useful for shortcircuiting the peering delay for user actions such as cancel/delete/etc

### DIFF
--- a/orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/ExecutionCopier.kt
+++ b/orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/ExecutionCopier.kt
@@ -133,6 +133,8 @@ open class ExecutionCopier(
       val executionRows = srcDB.getExecutions(executionType, idsToMigrate)
       val stagesToMigrate = srcDB.getStageIdsForExecutions(executionType, idsToMigrate)
 
+      val executionsNoLongerPresent = arrayListOf<String>()
+
       // Step 1: Copy over stages before the executions themselves -
       // if we saved executions first the user could request an execution but it wouldn't have any stages yet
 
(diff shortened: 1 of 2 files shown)


orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/ExecutionCopier.kt
```diff
@@ -152,13 +154,35 @@ open class ExecutionCopier(
...
+          latestUpdatedAt = max(latestUpdatedAt, r.get("updated_at", Long::class.java))
+          idsToMigrateHash.remove(r.get("id"))
+        }
+
+        executionsNoLongerPresent.addAll(idsToMigrateHash)
+      } else {
+        executionRows.forEach { r ->
+          r.set(partitionField, peeredId)
+          latestUpdatedAt = max(latestUpdatedAt, r.get("updated_at", Long::class.java))
+        }
```
dreynaud: The forEach introduces quite a bit of duplication in the 2 branches of the if/else, wouldn't it be the same as this?

```suggestion
      executionRows.forEach { r ->
        r.set(partitionField, peeredId)
        latestUpdatedAt = max(latestUpdatedAt, r.get("updated_at", Long::class.java))
      }
      
      if (executionRows.size < idsToMigrate.size) {
        // There must've been some executions that have been deleted during our diff calculation, find out which ones
        val idsToMigrateHash = idsToMigrate.toHashSet()

        executionRows.forEach { r ->
          idsToMigrateHash.remove(r.get("id"))
        }

        executionsNoLongerPresent.addAll(idsToMigrateHash)
      }```

Comment (dreynaud) [reply orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/ExecutionCopier.kt]: I believe this is maybe a slightly more efficient/readable way to figure out executions no longer present:

```
  private fun executionsNoLongerPresent(executionRows: Result<Record>, idsToMigrate: List<String>): Iterable<String> {
    return if (executionRows.size < idsToMigrate.size) {
      // There must've been some executions that have been deleted during our diff calculation, find out which ones
      val executionSet = executionRows.intoMap(DSL.field("id"))
      idsToMigrate.filter { id -> !executionSet.containsKey(id) }
    } else {
      emptyList()
    }
  }
```

and then it's a simple call

```
if (executionsNoLongerPresent(executionRows, idsToMigrate).any()) {
   ...
```

File: orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/ExecutionCopier.kt [hunk 2]
```diff
@@ -152,13 +154,35 @@ open class ExecutionCopier(
...
       }
       destDB.loadRecords(getExecutionTable(executionType).name, executionRows)
       peeringMetrics.incrementNumPeered(executionType, state, idsToMigrate.size)
 
+      // If we found any executions that are no longer there, delete them from the destination
+      if (executionsNoLongerPresent.any()) {
+        destDB.deleteExecutions(executionType, idsToMigrate)
```
dreynaud: doesn't this delete all `idsToMigrate`, not just the ones in `executionsNoLongerPresent`?

marchello2000:
gonna close for now, reopen laterz

PR #3475, opened 2020-02-28
fix(peering): Peer if there are no (newly) completed executions to peer

marchello2000:
If there are no "newly" completed executions to peer the peering agent would never peer active executions.
The reason was that that the "lastUpdated" timestamp wouldn't be updated if there was nothing completed to peer so it would sit at `0` indicating it's always first run

This change fixes that (and also fixes the tests)

### DIFF
--- a/orca-peering/src/test/groovy/com/netflix/spinnaker/orca/peering/PeeringAgentSpec.groovy
+++ b/orca-peering/src/test/groovy/com/netflix/spinnaker/orca/peering/PeeringAgentSpec.groovy
@@ -29,7 +29,7 @@ class PeeringAgentSpec extends Specification {
   MySqlRawAccess dest = Mock(MySqlRawAccess)
   PeeringMetrics metrics = Mock(PeeringMetrics)
   ExecutionCopier copier = Mock(ExecutionCopier)
-  def clockDrift = 100
+  def clockDrift = 10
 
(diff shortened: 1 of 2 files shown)


orca-peering/src/test/groovy/com/netflix/spinnaker/orca/peering/PeeringAgentSpec.groovy
```diff
@@ -92,18 +95,21 @@ class PeeringAgentSpec extends Specification {
...
+    copyCallCount * copier.copyInParallel(executionType, toCopy, ExecutionState.COMPLETED) >>
+        new ExecutionCopier.MigrationChunkResult(
+            (srcKeys + srcKeysNull).findAll {toCopy.contains(it.id)}.collect {it.updated_at}.max() ?: 0,
+            toCopy.size(),
+            false)
 
     if (executionType == PIPELINE) {
-      peeringAgent.completedPipelinesMostRecentUpdatedTime == (srcKeys + srcKeysNull).max { it.updated_at }?.updated_at ?: 1
-      peeringAgent.completedOrchestrationsMostRecentUpdatedTime == 2
+      assert peeringAgent.completedPipelinesMostRecentUpdatedTime == correctMax
```
marchello2000: this lack of `assert` was a problem...

marchello2000:
use whitespace ignore diff

marchello2000:
superseeded by https://github.com/spinnaker/orca/pull/3478

PR #3477, opened 2020-02-28
chore(java11): Switch orca to use the Java 11 JRE

plumpy:
This was waiting on the removal of a `kork` `PluginLoader` class that
wasn't compatible. It was removed in spinnaker/kork#528 and #3476

### DIFF
--- a/Dockerfile.java8
+++ b/Dockerfile.java8
@@ -1,7 +1,7 @@
 FROM alpine:3.11
 MAINTAINER sig-platform@spinnaker.io
 COPY orca-web/build/install/orca /opt/orca
-RUN apk --no-cache add --update bash openjdk11-jre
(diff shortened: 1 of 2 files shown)


Dockerfile.java8
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.11
+MAINTAINER sig-platform@spinnaker.io
+COPY orca-web/build/install/orca /opt/orca
+RUN apk --no-cache add --update bash openjdk8-jre
+RUN adduser -D -S spinnaker
+RUN mkdir plugins && chown -R 100:100 plugins
```
ezimanyi: Not excited to see these UIDs further propagated, but hopefully this will be cleaned up.

plumpy: Yeah, Brandon is working on it...

PR #3478, opened 2020-02-28
fix(peering): Peer if there are no (newly) completed executions to peer

dreynaud:
If there are no "newly" completed executions to peer the peering agent would never peer active executions.
The reason was that that the "lastUpdated" timestamp wouldn't be updated if there was nothing completed to peer so it would sit at 0 indicating it's always first run

This change fixes that (and also fixes the tests)


dreynaud:
at least 12% better than https://github.com/spinnaker/orca/pull/3475

PR #3482, opened 2020-02-29
config(core): Push halyard config defaults down to base

ezimanyi:
See spinnaker/clouddriver#4368 for more details; this is making the same changes to orca.  I'll hold off on doing more services until everyone has weighed in on these two PRs, but if all looks good I'll make the same changes to all other services.

* config(core): Push halyard config defaults down to base 

  This change pushes a number of config settings that were in the Halyard-specific config file down to the base config so they apply to all uesrs.

  The one potentially breaking change is that echo used to default to disabled but now defaults to enabled. Given that most users use Halyard and netflix uses echo, I suspect that nobody was relying on this to default to off. We can document this as a change in the release notes (as it's easy to override if desired).

* config(core): Update bakery defaults 

  Push the defaults for extractBuildDetails, roscoApisEnabled, and allowMissingPackageInstallation down to the properties class.
  (As defaults were there anyway, there is no need to specify them in clouddriver.yml.)

  This makes the following changes:
  * Removes references the services.rosco.* settings for two flags; only the Halyard config referenced them, but Halyard never wrote out those values.
  * Sets extractBuildDetails to default to true; this has always been the case for Halyard users.

* config(core): Update redis config setting 

  This change updates the core redis config setting to read from services.redis.baseUrl then default to localhost:6379.

  This does remove looking at the REDIS_URL environment variable. I strongly suspect this isn't being used by anyone (as definitely it is not set by Halyard users), though would like to confirm with Netflix before merging.

### DIFF
--- a/halconfig/orca.yml
+++ b/halconfig/orca.yml
@@ -1,4 +1,1 @@
 # halconfig
-
-redis:
-  connection: ${services.redis.baseUrl:redis://localhost:6379}
--- a/halconfig/orca.yml
+++ b/halconfig/orca.yml
@@ -1,8 +1,4 @@
 # halconfig
 
-bakery:
-  extractBuildDetails: ${services.rosco.extractBuildDetails:true}
-  allowMissingPackageInstallation: ${services.rosco.allowMissingPackageInstallation:false}
-
 redis:
   connection: ${services.redis.baseUrl:redis://localhost:6379}
--- a/halconfig/orca.yml
+++ b/halconfig/orca.yml
@@ -1,43 +1,8 @@
 # halconfig
 
-server:
-  port: ${services.orca.port:8083}
-  address: ${services.orca.host:localhost}
-
-oort:
-  baseUrl: ${services.clouddriver.baseUrl:http://localhost:7002}
-
-front50:
-  enabled: true
-  baseUrl: ${services.front50.baseUrl:http://localhost:8080}
-
-mort:
-  baseUrl: ${services.clouddriver.baseUrl:http://localhost:7002}
-
-kato:
-  baseUrl: ${services.clouddriver.baseUrl:http://localhost:7002}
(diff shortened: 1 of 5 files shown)


ezimanyi:
@robzienert : Things I want to explicitly check are fine with Netflix on this PR:
* That removing the env variable REDIS_URL as a source of the redis endpoint won't break anything
* That defaulting `extractBuildDetails` to `true` won't break anything.

halconfig/orca.yml
```diff
@@ -1,43 +1 @@
 # halconfig
```
plumpy: why does this file still need to exist? Does `halyard` get mad otherwise?

ezimanyi: I'm 99% sure that it will get mad...though I'll be honest that I didn't actually check.  If these changes look good and become my plan for all the services, I'll double-check if we can actually completely delete the files.

jervi:
Just a little notice from the sideline. This PR partially broke our bakery (it started to try to install DEB packages in Amazon Linux, and `extendedAttributes` were not propagated to Rosco any more). It was an easy fix (flip `roscoApisEnabled` to `true`; note to self: don't rely on defaults), but I was wondering if maybe `roscoApisEnabled` should default to `true`, as most organizations aren't Netflix, and it seems this is already the default in Halyard (we don't use it yet).

ezimanyi:
@jervi : Sorry about that!  That was entirely an accident...my intention was only to remove things from `orca.yml` that were redundant to defaults in the code.  But in this case, the code was defaulting to false and `orca.yml` was setting the value to `true`, so it was an unintentional breaking change.  The above-linked PR will default it to `true` in code so that there is no visible change to users (and I'll cherry-pick that to 1.19).  Thanks for letting me know, and sorry again about that!

PR #3488, opened 2020-03-04
feat(preconfiguredjob): Add 'ui' field to PreconfiguredJobStageProperties

christopherthielen:
`basic` will cause Deck to register the default preconfigured job stage config UI (which renders inputs for each parameter).

### DIFF
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/preconfigured/jobs/PreconfiguredJobStageProperties.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/preconfigured/jobs/PreconfiguredJobStageProperties.java
@@ -58,13 +58,13 @@ public abstract class PreconfiguredJobStageProperties {
 
   private String propertyFile;
 
-  public enum PreconfiguredJobUI {
+  public enum PreconfiguredJobUIType {
     BASIC,
     CUSTOM
   };
 
   /** "BASIC" to auto-generate a UI. "CUSTOM" if a custom UI is provided for Deck */
-  private PreconfiguredJobUI ui = PreconfiguredJobUI.BASIC;
+  private PreconfiguredJobUIType uiType = PreconfiguredJobUIType.BASIC;
 
   /** Indicates whether this job produces any artifacts. */
   private boolean producesArtifacts = false;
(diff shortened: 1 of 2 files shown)


orca-api/src/main/java/com/netflix/spinnaker/orca/api/preconfigured/jobs/PreconfiguredJobStageProperties.java
```diff
@@ -58,6 +58,9 @@
 
   private String propertyFile;
 
+  /** "basic" to auto-generate a UI. "custom" if a custom UI is provided for Deck */
+  private String ui = "basic";
```
christopherthielen: should this be an enum?

orca-api/src/main/java/com/netflix/spinnaker/orca/api/preconfigured/jobs/PreconfiguredJobStageProperties.java
```diff
@@ -58,6 +58,14 @@
 
   private String propertyFile;
 
+  public enum PreconfiguredJobUI {
+    BASIC,
+    CUSTOM
+  };
+
+  /** "BASIC" to auto-generate a UI. "CUSTOM" if a custom UI is provided for Deck */
```
srekapalli: instead of `ui` .. I think uiType maybe ?

orca-api/src/main/java/com/netflix/spinnaker/orca/api/preconfigured/jobs/PreconfiguredJobStageProperties.java
```diff
@@ -58,6 +58,14 @@
 
   private String propertyFile;
```
srekapalli: PreconfiguredJobUIType ?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -420,6 +420,7 @@ class OperationsController {
         noUserConfigurableFields: true,
         parameters: it.parameters,
         producesArtifacts: it.producesArtifacts,
+        ui: it.ui
```
robzienert: This needs to be updated for `uiType` change.

christopherthielen: groovy, thanks

PR #3492, opened 2020-03-05
fix(queue): avoid pushing message for foreign execution

dreynaud:
When manually cancelling an execution handled by a different orca cluster, we want to let the execution repo handle it. Also pushing messages on the queue for that foreign execution was an oversight, and it led to duplicate interlink messages getting published.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
@@ -99,7 +99,7 @@ public class CompoundExecutionOperator {
             if (repository.handlesPartition(execution.getPartition())) {
               runnerAction.accept(execution);
             } else {
-              log.warn(
+              log.info(
                   "Not pushing queue message action='{}' for execution with foreign partition='{}'",
                   action,
                   execution.getPartition());
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
@@ -17,10 +17,12 @@
 package com.netflix.spinnaker.orca.pipeline;
 
 import com.netflix.spinnaker.kork.core.RetrySupport;
+import com.netflix.spinnaker.orca.pipeline.model.Execution;
 import com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType;
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import java.time.Duration;
+import java.util.function.Consumer;
 import javax.annotation.Nullable;
 import lombok.NonNull;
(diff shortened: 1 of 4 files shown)


orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperatorSpec.groovy
```diff
@@ -29,6 +30,10 @@ class CompoundExecutionOperatorSpec extends Specification {
   ExecutionRunner runner = Mock(ExecutionRunner)
   def execution = Mock(Execution)
```
srekapalli: may be here also we can do setup and cleanup instead of the entire spec ?

dreynaud: I don't think this is necessary, the tests in this class don't put anything in the MDC

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/aws/InterlinkAmazonMessageHandler.java
```diff
@@ -59,14 +61,19 @@ public void handleMessage(Message message) {
...
-    if (executionRepository.handlesPartition(event.getPartition())) {
-      event.applyTo(executionOperator);
-    } else {
-      log.debug(
-          "Execution repository with local partition {} does not handle this event's partition {}, "
-              + "so it will not be applied",
-          executionRepository.getPartition(),
-          event.getPartition());
+    try {
+      MDC.put(Header.EXECUTION_ID.getHeader(), event.getExecutionId());
```
marchello2000: ❤️ 
maybe also put x-spinnaker-user in here

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
```diff
@@ -77,21 +79,32 @@ public void resume(
...
       String executionId) {
     try {
-      runWithRetries(runnerAction);
+      runWithRetries(
+          () -> {
+            Execution execution = repository.retrieve(executionType, executionId);
+            if (repository.handlesPartition(execution.getPartition())) {
+              runnerAction.accept(execution);
+            } else {
+              log.warn(
```
marchello2000: this will be happening a lot so prob not worth a `warn`?

marchello2000: esp since we will get an interlink publish log shortly after

dreynaud: handshake deal, making it an `info`

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/aws/InterlinkAmazonMessageHandler.java
dreynaud: we don't have the user on every interlink event yet, but we probably will have to have it to handle manual stages later on

PR #3493, opened 2020-03-05
fix(pipelinetemplate): Add template variables to module tag

jervi:
According to documentation, the module tag should be able to use variables both from the pipeline template and variables scoped strictly to the module, however only the latter was the case. Fixed by adding the template variables to the render context.


jervi:
@spinnaker/oss-approvers PTAL 🙂

jervi:
I'm missing the manual judgement at the end, plus blocket-remote-email and blocket-remote-merge. I'll add those ASAP.

robzienert:
Needs tests.

PR #3494, opened 2020-03-05
fix(logging): add logging for when we can't find the stage

marchello2000:
To aid debugging further an issue where the stage that should've been there was not


kevinheins:
I am possible hitting this issue (`InvalidStageId`) in Orca but can't find any other references to it within Slack or Github. Can you elaborate on what your issue was and if there was a fix?

PR #3502, opened 2020-03-06
feat(k8s): can change manifest stabilization timeout

bpowell:
fixes https://github.com/spinnaker/spinnaker/issues/5429


bpowell:
@maggieneterval

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTask.java
```diff
@@ -45,14 +46,17 @@
 
   private final OortService oortService;
 
+  @Value("${kubenetes.manifest.timeout-minutes:30}")
```
ezimanyi: Please use constructor injection instead of value injection, and make this field `final`.  You should be able to just annotate the appropriate parameter of the constructor with this `@Value` annotation.

Also, it looks like a number of other places that have added overridable timeouts have scoped them as `tasks.task-name-timeout`...though arguably it might even be better as `tasks.task-name.timeout-minutes`.  At least I think it would be best to have this scoped under `tasks` to avoid too many top-level things.

The other configurable timeouts also specify milliseconds...for consistency, maybe it's better for these to all be the same?  (Though I do see a point that people probably don't need that precision...)

agoyal-sfdc:
As someone who would like to have this feature, I think that having this at the stage level should be supported. Using the Salesforce example, one team runs and maintains Spinnaker for all of Salesforce, so any users of Spinnaker will not have the ability to change this value. Making it a stage-level configurable gives the actual end user the ability to control however long they want Spinnaker to attempt a deployment.

agoyal-sfdc:
Also, just as an aside, and this could be a different PR, but a similar configurable for `Run Job (Manifest)` would be very useful.

bpowell:
Hey @ezimanyi with this being a task and all of the information being contained in a Stage that is passed to the `execute` method, how would we get the timeout from the UI (which would be contained in the `stage` variable) be used to update the `getTimeOut` method that is on class and is used outside of the class?

ezimanyi:
@bpowell : The `RetryableTask` interface that `WaitForManifestStableTask` implements has both `long getTimeout()` and `long getDynamicTimeout(Stage stage)`.  I believe that overriding the latter would allow setting the timeout based on values in the stage context.

I don't have a strong opinion one way or another whether this should be static or configurable on the stage; configuring on the stage would definitely be a bit more work (and add more fields in the UI) but it does seem like this is intended to cover a small number of edge cases and allowing those particular pipelines to override it might be better than just increasing the value for all of them (similar to what @agoyal-sfdc noted).

bpowell:
@ezimanyi thanks very much for the help on this one! I think I got it all working having the stage control the timeout

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTask.java
```diff
@@ -55,6 +57,20 @@ public long getTimeout() {
     return TimeUnit.MINUTES.toMillis(30);
   }
 
+  @Override
+  public long getDynamicTimeout(StageExecution stage) {
+    ManifestTimeout timeout = stage.mapTo(ManifestTimeout.class);
+    if (timeout.timeoutInMinutes != null) {
```
ezimanyi: This seems to be mixing having `timeoutInMinutes` be `@Nullable` and having it be an `Optional`; you're checking that the `Optional<Long>` is not `null` but never checking `.isPresent()` before calling `.get()` which will throw an exception if it's an empty `Optional`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTask.java
```diff
@@ -55,6 +57,20 @@ public long getTimeout() {
...
+      return TimeUnit.MINUTES.toMillis(timeout.timeoutInMinutes.get());
+    }
+
+    return getTimeout();
+  }
+
+  static class ManifestTimeout {
+    @JsonProperty Optional<Long> timeoutInMinutes;
```
ezimanyi: Does Jackson properly handle the `Optional` here?  As in, will it always set this to an empty `Optional` if the field is missing, and to an optional containing the field if it's present?  I've never seen this before, though would definitely believe it worked if there was a test on it 😁 

Also, as a really minor point, I think generally we omit the `in` from fields like this, so it would be `timeoutMinutes`.  (Either would in general be fine, but I think consistency with other fields would be good here.)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTask.java
```diff
@@ -55,6 +57,20 @@ public long getTimeout() {
...
+  public long getDynamicTimeout(StageExecution stage) {
+    ManifestTimeout timeout = stage.mapTo(ManifestTimeout.class);
+    if (timeout.timeoutInMinutes != null) {
+      return TimeUnit.MINUTES.toMillis(timeout.timeoutInMinutes.get());
+    }
+
+    return getTimeout();
+  }
+
+  static class ManifestTimeout {
```
ezimanyi: This class should be `private` if it's not intended to be used outside of the outer class.

maggieneterval:
Thanks for making and reviewing this change, @bpowell and @ezimanyi! To clarify, are we going to add a UI field for `timeoutMinutes` on the Deploy and Patch Manifest stages (the original issues mentions wanting a configurable timeout on both) or should we document that users will need to edit the JSON manually?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTask.java
bpowell: I'm not sure on the Jackson side of things, but looking at other examples that's what they have: https://github.com/spinnaker/orca/blob/062e60fdea15e1ce75316535edfef09e61caf370/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableContext.java#L43

And I'll go ahead and remove the `in` from the field

bpowell:
@maggieneterval yes, we plan on adding a deck piece to make this easier! I'm not sure when I'll get to that PR, hoping to soon

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTask.java
```diff
@@ -55,6 +57,20 @@ public long getTimeout() {
     return TimeUnit.MINUTES.toMillis(30);
   }
 
+  @Override
+  public long getDynamicTimeout(StageExecution stage) {
+    ManifestTimeout timeout = stage.mapTo(ManifestTimeout.class);
+    if (timeout.timeoutMinutes != null && timeout.timeoutMinutes.isPresent()) {
```
ezimanyi: I don't think there's much benefit to using an `Optional` here if we still need to check whether it's null.  It looks like Jackson will set `timeoutMinutes` to `Optional.empty()` if the field `timeoutMinutes` is present but is set to something it considers empty (ex: null or empty string), whereas it will just leave it as `null` if the field is entirely absent from the input.  I think that `ManifestTimeout` should handle all this complexity and let callers assume that `timeoutMinutes` is non-null.

I think the best way of doing that is to change the `ManifestTimeout` class to something like this:
```
    @Nonnull private final OptionalLong timeoutMinutes;

    ManifestTimeout(@JsonProperty("timeoutMinutes") OptionalLong timeoutMinutes) {
      this.timeoutMinutes = timeoutMinutes;
    }
```

This will guarantee that Jackson will pass an `OptionalLong.empty()` to the constructor if the field is not present, as opposed to using setter-based construction which will only set fields that are present.  It also has the advantage of making the class immutable.  Then you can remove the `== null` check and only do the `.isPresent` check.

I think there should be another test of what happens if `timeoutMinutes` is present but set to an empty string (as would probably happen if it was left unset in the UI).  I would also suggest adding a test where `timeoutMinutes` is set to the string `"1"` as I believe that will be what comes back from deck in this case.

With those additional tests we should be in pretty good shape around making sure this works for all cases (and even if something like a Jackson upgrade breaks it, we'll find out right away).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/WaitForManifestStableTask.java
```diff
@@ -55,6 +57,24 @@ public long getTimeout() {
...
+    ManifestTimeout timeout = stage.mapTo(ManifestTimeout.class);
+    if (timeout.timeoutMinutes.isPresent()) {
+      return TimeUnit.MINUTES.toMillis(timeout.timeoutMinutes.get());
+    }
+
+    return getTimeout();
+  }
+
+  private static class ManifestTimeout {
+    @Nonnull private final Optional<Long> timeoutMinutes;
```
ezimanyi: The benefit here is really small, so no need to change, but in general it's often better to use `OptionalLong` instead of `Optional<Long>` as putting a `long` in an `Optional` requires first boxing it to a `Long` then putting it in the `Optional`; getting the value out requires the same two steps.  An `OptionalLong` just has the one step of directly putting the `long` in the `Optional`.  In this case, given that we're not doing this in a tight loop or anything, I think the difference is negligible so it's fine as is.

maggieneterval:
@bpowell just to save you some time continuously merging master into your branch, the `ready to merge` label will actually do that for you!

imakewebthings:
Thank you for tackling this @bpowell

marchello2000:
Hey @bpowell (cc: @maggieneterval / @ezimanyi ) Same comment as https://github.com/spinnaker/orca/pull/3587#issuecomment-616934451
What am I missing. Why `stageTimeoutMs` (already exposed in deck see linked comment) not enough here?

PR #3505, opened 2020-03-11
fix(config): Fix accidentally changed bakery config

ezimanyi:
The commit to clean up the config file by pushing some config down to code accidentally changed the default value of roscoApisEnabled. This was set as true in orca.yml, but I forgot to flip the default to true in code when removing the value in orca.yml, which ended up being a breaking change.

This change flips the default value of the boolean to true in code, restoring the deafult to what it was prior to the earlier change.


ezimanyi:
This flips a flag that should have been flipped when the default was removed from `orca.yml` in #3482.

ezimanyi:

PR #3507, opened 2020-03-12
feat(traffic-guards): bypass check for pinned server groups

dreynaud:
A common use case that trips up traffic guards is things like:
1. staging clusters with a fixed size (e.g. 3/3/3)
2. a deploy fails for some reason, leaving multiple server groups
3. from this point on, all deploys will fail the traffic guard capacity check
   (as you would delete N*3 instances and only leave 3 up)

It seems reasonable to punch an exception through traffic guards when this happens:
- there are multiple server groups going away, all with the same fixed size
- the action leaves _some_ instances up


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/Capacity.groovy
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support
+
+import groovy.transform.Canonical
+
+import javax.annotation.Nonnegative
+
+@Canonical
+class Capacity {
```
ezimanyi: As noted by `spinnakerbot` this should be Java...also probably can be an immutable class as well as I don't think we ever change the fields after creation.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/Capacity.groovy
```diff
@@ -0,0 +1,44 @@
...
+  @Nonnegative
+  int max
+
+  @Nonnegative
+  int desired
+
+  /**
+   * @return true if the capacity of this server group is fixed, i.e min, max and desired are all the same
+   */
+  def Boolean isPinned() {
```
ezimanyi: Can be a `boolean` here as I don't think this would ever be `null`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
```diff
@@ -64,6 +64,17 @@ class TargetServerGroup {
...
   }
 
+  Capacity getCapacity() {
+    return new Capacity(
+        toInt(serverGroup.capacity.min),
+        toInt(serverGroup.capacity.max),
+        toInt(serverGroup.capacity.desired))
+  }
+
+  private int toInt(Object field) {
```
ezimanyi: (minor) can be static

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/Capacity.groovy
marchello2000: it's amazing how many `capacity` classes we have... If you can use an existing one (e.g. ResizeStrategy.Capacity)...
Or... we just go through and change all the other ones to use this one, since it's nicer anyway

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -280,6 +280,22 @@ public void verifyTrafficRemoval(
...
 
     int futureCapacity = currentCapacity - capacityGoingAway;
+
+    int someDesiredSize = someServerGroup.getCapacity().getDesired();
+    if (futureCapacity > 0
+        && serverGroupsGoingAway.size() > 1
+        && serverGroupsGoingAway.stream().allMatch(sg -> sg.getCapacity().isPinned())
+        && serverGroupsGoingAway.stream()
+            .allMatch(sg -> sg.getCapacity().getDesired() == someDesiredSize)) {
+      log.debug(
```
marchello2000: def `info` level to me (at least)

marchello2000: also, maybe `...with all server groups pinned with matching capacity`? would be quicker to tell whats up in logs if this ever bites us

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/Capacity.groovy
dreynaud: Yes the idea here was to create a more generic `Capacity` class instead of "stealing" the one from `ResizeStrategy`. Ideally it would also be a straight-up refactor but for some reason IntelliJ won't let me refactor an inner class in Groovy (?)

dreynaud: How do we handle converting class to Java when the entire `orca-clouddriver` module is under `src/main/groovy`? Create a `src/main/java` with that lonely class in there?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
dreynaud: made this `debug` for consistency with all the other logs from the `TrafficGuard` class. Would you motion for all of them to become `info` or just this one?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/Capacity.groovy
ezimanyi: `.java` classes work fine even under `groovy` so I'd just put the `.java` class under `groovy` (looks like there are already a number of other java classes in there).

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
marchello2000: hm... i see. leave as `debug` then

dreynaud:
Thanks for the feedback @marchello2000 and @ezimanyi, I addressed things in 87e9e5b2

PR #3508, opened 2020-03-12
fix(FindImageFromCluster): update context with inferred regions

asher:


asher:
This is relevant to `AmazonServerGroupCreator.createServerGroupOperation()` which calls `withImageFromPrecedingStage()` followed by `withImageFromDeploymentDetails()` if the prior call didn't find an image. Prior to this change, the first call only found ami details from a parent `FindImage` stage for regions explicitly provided in the stage context. Inferred regions could only be discovered via `withImageFromDeploymentDetails()` which may behave differently for some pipeline configurations. This PR should ensure that the same image search logic is used for all regions when the "inferred region" feature is used.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -294,10 +294,17 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
...
-      artifacts: artifacts
-    ]).outputs([
+    // When additional regions are inferred for downstream deploy stages, adding them back
+    // to Stage.context.regions enables direct use via DeploymentDetailsAware.withImageFromPrecedingStage()
+    // instead of via a subsequent code path that may exhibit different behavior.
+    Map<String, Object> context = [amiDetails: deploymentDetails, artifacts: artifacts]
+    if (!inferredRegions.isEmpty()) {
+      inferredRegions.addAll(config.regions)
+      context.put("regions", new ArrayList<String>(inferredRegions))
+    }
```
marchello2000: nit: I think it would be clearer/simpler if this was just
```suggestion
    context.put("regions", config.regions + inferredRegions)
```

PR #3509, opened 2020-03-13
Revert "fix(FindImageFromCluster): update context with inferred regio…

maggieneterval:
…ns (#3508)"

This reverts commit 96384a32c80f77169488d0228ecab7cd3414e2b8.

The end to end tests are failing with an NPE for the [`kube_smoke_tests`](https://builds.spinnaker.io/job/Validate_BomMultiPlatform/4370/CLOUD_PLATFORM=gce/artifact/logs/citest_logs/kube_smoke_test.html): `java.lang.NullPointerException: Cannot execute null+[]` from line 301 of `FindImageFromClusterTask` . I'm guessing this is because `config.regions` is null for non-AWS providers. I don't have enough context around this change to determine whether the correct solution is to only execute the new logic if the provider is AWS, or if this change was intended to apply to other non-AWS providers and we should default `config.regions` to an empty set. Going to revert this change for now so that we can get the builds back in a healthy state.


asher:
Thanks for the quick revert and explanation, @maggieneterval

PR #3517, opened 2020-03-14
fix(rollback): Should be passing target groups and security groups in request

jeyrschabu:
- include request target groups and security groups

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/PreviousImageRollback.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/PreviousImageRollback.groovy
@@ -97,6 +97,8 @@ class PreviousImageRollback implements Rollback {
       credentials                  : parentStageContext.credentials,
       cloudProvider                : parentStageContext.cloudProvider,
       delayBeforeDisableSec        : delayBeforeDisableSeconds ?: 0,
+      targetGroups                 : parentStageContext.targetGroups,
+      securityGroups               : parentStageContext.securityGroups,
       source                       : [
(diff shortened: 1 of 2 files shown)


jeyrschabu:
Not passing target groups through causes the rollback to fail in titus. This fixes that.
`Deck` should ensure the parameters are passed in every request.

ajordens:
Didn't actually run this but looks reasonable to me.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/rollback/PreviousImageRollback.groovy
```diff
@@ -97,6 +97,8 @@ class PreviousImageRollback implements Rollback {
       credentials                  : parentStageContext.credentials,
       cloudProvider                : parentStageContext.cloudProvider,
       delayBeforeDisableSec        : delayBeforeDisableSeconds ?: 0,
+      targetGroups                 : parentStageContext.targetGroups,
```
marchello2000: what if this deploy changes the `targetGroup` or `securityGroup` to something new, shouldn't we restore it to the original state?

marchello2000: Actually, looks like only `deck` uses `PREVIOUS_IMAGE`. Do you have a corresponding `deck` PR for this as well?

jeyrschabu: Will fire up a deck PR as a follow up to re-enable Titus previous image rollbacks and ensure target groups are passed through the request.

PR #3519, opened 2020-03-16
fix(peering): Trigger pending executions across orca peers

marchello2000:
Imagine this scenario with peering:
* orca `peer A` is running execution `E1`
* a request (from `deck`) comes into orca `peer B` to start execution `E2`
  since an execution (`E1`) is already running, `peer B` queues `E2`
* `peer A` completes `E1` and checks if there are any pending executions
* since the pending table is not peered (that would be basically impossible to get right) `peer A` doesn't see any pending executions
* and so `E2` stays forever in a "pending" state

There were a few thoughts on how to fix this:
1. Pending execution agent
2. Use dual pending execution service
3. Peering pending executions table
4. Handle this in the peering agent when it peers completed executions
5. Use interlink

Basically, all of these approaches are flawed and interlink seems least problematic:
*Pending execution* has issues because:
* it feels disconnected from peering&interlink - it's another thing to configure and make sure it's on, etc
* the timing is non-deterministic
* due to current design, it doesn't know whether the queue should be purged or not (since this info is embedded in the prior execution)

*Dual pending execution service* breaks down when
* there are more than 2 peers (at least in current incarnation)
* it also adds the need for every peer to know every other peer which is not the case today and adds a ton of complexity to config/operation of orca

*Peering pending execution* table scared me!
* due to the inherent lag in peering the table could be horribly out of date
* to ensure we don't trigger something twice, we would also need to peer correlation ids (and again, the lag problem)

*Handing completions* in the peering agent, e.g. when we peer completed execution we see if we need to kick off something else:
* introduces requirements that peering agents run with queue (not the end of the world)
* adds this "weird" knowledge to peering agent

*Using Interlink* seemed like the least evil option:
* it fits in with the rest of implementation, nothing extra to configure
* fairly simple implementation

cons:
* it does add coupling between queue -> interlink (which is a bit gross)
* it has an issue in that interlink is not a guaranteed delivery, but I think it's ok/good enough

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueExecutionRunner.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueExecutionRunner.kt
@@ -45,4 +45,8 @@ class QueueExecutionRunner(
   override fun cancel(execution: PipelineExecution, user: String, reason: String?) {
     queue.push(CancelExecution(execution, user, reason))
   }
+
+  override fun startPending(pipelineConfigId: String, purge: Boolean) {
+    queue.push(StartWaitingExecutions(pipelineConfigId, purge))
+  }
 }
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartWaitingExecutionsHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartWaitingExecutionsHandler.kt
@@ -16,6 +16,9 @@
 
 package com.netflix.spinnaker.orca.q.handler
 
+import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType
+import com.netflix.spinnaker.orca.interlink.Interlink
+import com.netflix.spinnaker.orca.interlink.events.StartPendingInterlinkEvent
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
 import com.netflix.spinnaker.orca.q.CancelExecution
 import com.netflix.spinnaker.orca.q.ExecutionLevel
@@ -32,7 +35,8 @@ import org.springframework.stereotype.Component
 class StartWaitingExecutionsHandler(
   override val queue: Queue,
   override val repository: ExecutionRepository,
-  private val pendingExecutionService: PendingExecutionService
+  private val pendingExecutionService: PendingExecutionService,
+  private val interlink: Interlink?
 ) : OrcaMessageHandler<StartWaitingExecutions> {
 
   private val log: Logger get() = LoggerFactory.getLogger(javaClass)
@@ -40,7 +44,7 @@ class StartWaitingExecutionsHandler(
   override val messageType = StartWaitingExecutions::class.java
 
   override fun handle(message: StartWaitingExecutions) {
-    if (message.purgeQueue) {
+    val pendingMessage = if (message.purgeQueue) {
       // when purging the queue, run the latest message and discard the rest
       pendingExecutionService.popNewest(message.pipelineConfigId)
         .also { _ ->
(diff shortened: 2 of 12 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueExecutionRunner.kt
```diff
@@ -45,4 +45,8 @@ class QueueExecutionRunner(
   override fun cancel(execution: PipelineExecution, user: String, reason: String?) {
     queue.push(CancelExecution(execution, user, reason))
   }
+
+  override fun startPending(pipelienConfigId: String, purge: Boolean) {
+    queue.push(StartWaitingExecutions(pipelienConfigId, purge))
```
dreynaud:
```suggestion
   override fun startPending(pipelineConfigId: String, purge: Boolean) {
    queue.push(StartWaitingExecutions(pipelineConfigId, purge))
```

dreynaud:
```suggestion
   override fun startPending(pipelineConfigId: String, purge: Boolean) {
    queue.push(StartWaitingExecutions(pipelineConfigId, purge))
```

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartWaitingExecutionsHandler.kt
```diff
@@ -61,12 +65,21 @@ class StartWaitingExecutionsHandler(
...
-        if (it is ExecutionLevel) {
-          log.info("Starting queued pipeline {} {} {}", it.application, it.executionId)
-        }
-        queue.push(it)
+
+    if (pendingMessage != null) {
+
+      // spoiler, it always is!
+      if (pendingMessage is ExecutionLevel) {
+        log.info("Starting queued pipeline {} {} {}", pendingMessage.application, pendingMessage.executionId)
```
dreynaud: I know it's not yours but isn't this weird? It always pushes `pendingMessage`, but _only_ logs if it is `ExecutionLevel` 🤔

`queue.push(pendingMessage)` probably belongs in the body of the if statement

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartWaitingExecutionsHandler.kt
```diff
@@ -61,12 +65,21 @@ class StartWaitingExecutionsHandler(
...
       }
+      queue.push(pendingMessage)
+    } else {
+      interlink?.publish(
+        StartPendingInterlinkEvent(
+          ExecutionType.PIPELINE,
+          "NO_EXECUTION_ID",
```
dreynaud: nit: we might want to make it a symbolic constant in an interface somewhere, maybe in `InterlinkEvent`?

dreynaud:
Pretty straightforward, I like it. One observation: this will result in one interlink event per completed execution, right? So in terms of volume, this is likely to completely drown out the other interlink messages by some orders of magnitudes. Depending on how fast they are transmitted and processed, we probably want to double check that the queues are not getting full and that it does not delay the processing of "manual" messages. 

Not necessarily a problem, just thinking out loud.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartWaitingExecutionsHandler.kt
marchello2000: i think i can actually eliminate, let me see

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartWaitingExecutionsHandler.kt
marchello2000: it is weird... despite themessage always being executionlevel... i am scared of changing this :)

PR #3520, opened 2020-03-17
feat(interlink): patch stage support for foreign executions

dreynaud:
Some operations like skip wait or manual approvals result in calls to

    PATCH /pipelines/pipelineId/stages/stageId

with a context map values used to augment the designated stage context. The TaskController also updates the stage's `lastModified` field with user and account information, so that manual approval stages can be used to propagate auth info. This is all great as long as we operate on local executions, but for foreign executions (owned by a different orca cluster), we can't just patch read-only stages.

To support this, we:
- make `ExecutionRepository::storeStage` foreign-aware
- for foreign executions, we create a `PatchStageInterlinkEvent` (if supported) that contains the body of the stage
- the stage body includes the updated context map values and `lastModified` information
- we add an `updateStage` method in `CompoundExecutionOperator` and update the `TaskController` to call it
- we also call it from the event's `applyTo` method

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
@@ -17,28 +17,28 @@
 package com.netflix.spinnaker.orca.pipeline;
 
 import com.netflix.spinnaker.kork.core.RetrySupport;
+import com.netflix.spinnaker.kork.web.exceptions.NotFoundException;
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType;
 import com.netflix.spinnaker.orca.api.pipeline.models.PipelineExecution;
+import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import java.time.Duration;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 import javax.annotation.Nullable;
 import lombok.NonNull;
+import lombok.Value;
+import lombok.experimental.NonFinal;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
+@Value
+@NonFinal
 public class CompoundExecutionOperator {
-  private ExecutionRepository repository;
-  private ExecutionRunner runner;
-  private RetrySupport retrySupport;
-
-  public CompoundExecutionOperator(
-      ExecutionRepository repository, ExecutionRunner runner, RetrySupport retrySupport) {
-    this.repository = repository;
-    this.runner = runner;
-    this.retrySupport = retrySupport;
-  }
+  ExecutionRepository repository;
+  ExecutionRunner runner;
+  RetrySupport retrySupport;
 
   public void cancel(ExecutionType executionType, String executionId) {
     cancel(
@@ -66,7 +66,7 @@ public class CompoundExecutionOperator {
       @NonNull String executionId,
       @Nullable String pausedBy) {
     doInternal(
-        (PipelineExecution execution) -> runner.reschedule(execution),
+        runner::reschedule,
         () -> repository.pause(executionType, executionId, pausedBy),
         "pause",
         executionType,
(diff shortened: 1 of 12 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
```diff
@@ -79,32 +79,67 @@ public void resume(
...
+              execution.getStages().stream()
+                  .filter(s -> s.getId().equals(stageId))
+                  .findFirst()
+                  .orElseThrow(
+                      () ->
+                          new NotFoundException(
+                              "Could not find stage with id "
+                                  + stageId
+                                  + " in execution "
+                                  + executionId));
```
marchello2000:
```suggestion
          StageExecution stage = execution.stageById(stageId);
```

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -465,26 +460,19 @@ class TaskController {
...
+          stage.context.putAll(context)
+          validateStageUpdate(stage)
+          stage.lastModified = new StageExecution.LastModifiedDetails(
+              user: AuthenticatedRequest.getSpinnakerUser().orElse("anonymous"),
+              allowedAccounts: AuthenticatedRequest.getSpinnakerAccounts().orElse(null)?.split(",") ?: [],
+              lastModifiedTime: System.currentTimeMillis()
+          )
+
+          // `lastModifiedBy` is deprecated (pending a update to deck)
+          stage.context["lastModifiedBy"] = AuthenticatedRequest.getSpinnakerUser().orElse("anonymous")
```
marchello2000: i know it was here before... (and supposedly needs to go away...) but
```suggestion
          stage.context["lastModifiedBy"] = stage.lastModified.user
```

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/events/PatchStageInterlinkEvent.java
```diff
@@ -0,0 +1,102 @@
...
+    Preconditions.checkNotNull(mapper, "applyTo requires an ObjectMapper");
+
+    try {
+      StageExecution stageFromMessage = mapper.readValue(stageBody, StageExecutionImpl.class);
+
+      executionOperator.updateStage(
+          executionType,
+          executionId,
+          stageId,
+          stage -> {
```
marchello2000: tbh, i find it very hard to grok with all the labda's being passed around, what's the value add for not just adding this to the `CompoundExecutionOperator::updateStage`? I guess you're trying to generalize with `TaskController`... but i feel it would be easier if this just took a map to use in `putAll`.

I am also not sure I follow why you need to check for the recency of the lastmodified? because the rest of the patches will still be applied.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
dreynaud: OMG thank you lol

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
```diff
@@ -335,7 +322,9 @@ public void setScheduledTime(@Nullable Long scheduledTime) {
 
   private LastModifiedDetails lastModified;
 
-  public @Nullable LastModifiedDetails getLastModified() {
+  @Nullable
```
robzienert: Ah, thanks.

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/events/PatchStageInterlinkEvent.java
dreynaud: The recency check is maybe a bit overkill, the idea is that going through interlink may add quite a bit of latency and it would be weird to throw away something that's more recent.

dreynaud: tried to make the lambda clearer and the time check more permissive (warn-only) in cbc504e94e4fc1f804078afc292d0a454079c60c, wdyt?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -465,26 +460,19 @@ class TaskController {
...
-
-      // `lastModifiedBy` is deprecated (pending a update to deck)
-      stage.context["lastModifiedBy"] = AuthenticatedRequest.getSpinnakerUser().orElse("anonymous")
-
-      executionRepository.storeStage(stage)
-
-      executionRunner.reschedule(pipeline)
-    }
-    pipeline
+    executionOperator.updateStage(PIPELINE, id, stageId,
```
marchello2000: can you add a return statement here please, i just spent 5 minutes utterly confused where is the return from `doInternal` used... thought i knew it should be here it took me a while to "remember"... ahhh... groovy

```suggestion
    return executionOperator.updateStage(PIPELINE, id, stageId,
```

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -465,26 +460,19 @@ class TaskController {
...
+        { stage ->
+          stage.context.putAll(context)
+          validateStageUpdate(stage)
```
marchello2000: i still think you need to pull `validateStageUpdate` out of this lambda. Imagine a case where `validateStageUpdate` fails. In the new implementation, since you perform the runner action first you might reschedule the message but fail to modify the stage context. but in the old implementation if validate fails nothing is done

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
```diff
@@ -79,32 +78,56 @@ public void resume(
...
+          stageUpdater.accept(stage);
+
+          repository.storeStage(stage);
+        },
+        "reschedule",
+        executionType,
+        executionId);
+  }
+
+  private PipelineExecution doInternal(
```
marchello2000: wait.. this just occurred to me... doesn't this method pose a race?
we push a message onto the queue before we mutate the repository. so if the message gets picked up before we manage to save the repo we are going to be in a weird state where the stage/execution have the right data but the action didn't take place because the queue processed the message before the state was updated?

i think we can just reverse the order and we will be fine

orca-interlink/src/main/java/com/netflix/spinnaker/orca/interlink/events/PatchStageInterlinkEvent.java
marchello2000: i still don't love it, but i am failing to find a better way, so 🚢 it :)

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
dreynaud: Fixed and tested in 0cfef4d0d (`validateStageUpdate` is still here, but if it fails we won't touch the queue)

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
dreynaud: You're right, fixed and tested in 0cfef4d0d!

PR #3527, opened 2020-03-18
fix(cleanupagents): fix for issue #5583

marchello2000:
resolves https://github.com/spinnaker/spinnaker/issues/5583 which is a regression of https://github.com/spinnaker/orca/pull/3461/files
don't require partition names in cleanup agents


marchello2000:

PR #3529, opened 2020-03-18
fix(execution): Persist `Execution.spelEvaluator` in Redis

nicolasff:
When using Redis as the execution repository, the spelEvaluator field of
the Execution object was not serialized or deserialized.
This change adds the field to the object stored in Redis, and reads it
back out when deserializing an Execution object.


nicolasff:
@marchello2000 This is a fix for the issue I mentioned today in Slack `#dev`: https://spinnakerteam.slack.com/archives/C0DPVDMQE/p1584562261086000?thread_ts=1584481356.080100&cid=C0DPVDMQE

PR #3534, opened 2020-03-20
feature(Pipeline Expressions): add pipelineIdOrNull SpEl helper function

AbdulRahmanAlHamali:
I don't think it's a good idea to throw an exception when a pipeline does not exist, because exceptions cannot be handled in Spinnaker Expressions.

We should just return `null`, and let the user decide what to do with that

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
@@ -58,6 +58,15 @@ public class PipelineExpressionFunctionProvider implements ExpressionFunctionPro
                 "execution",
                 "The execution containing the currently executing stage"),
             new FunctionParameter(
+                String.class, "pipelineName", "A valid stage reference identifier")),
+        new FunctionDefinition(
+            "pipelineIdOrNull",
+            "Lookup pipeline ID given the name of the pipeline in the current application",
+            new FunctionParameter(
+                PipelineExecution.class,
+                "execution",
+                "The execution containing the currently executing stage"),
+            new FunctionParameter(
                 String.class, "pipelineName", "A valid stage reference identifier")));
   }
 
@@ -66,7 +75,7 @@ public class PipelineExpressionFunctionProvider implements ExpressionFunctionPro
    *
    * @param execution the current execution
    * @param pipelineName name of the pipeline to lookup
-   * @return the id of the pipeline or null if pipeline not found
+   * @return the id of the pipeline or exception if pipeline not found
    */
   public static String pipelineId(PipelineExecution execution, String pipelineName) {
     if (Strings.isNullOrEmpty(pipelineName)) {
@@ -106,6 +115,25 @@ public class PipelineExpressionFunctionProvider implements ExpressionFunctionPro
       throw e;
     } catch (Exception e) {
       throw new SpelHelperFunctionException("Failed to evaluate #pipelineId function", e);
+    }
+  }
+
+  /**
+   * Function to convert pipeline name to pipeline ID (within current application), will return Null
+   * if pipeline ID not found
+   *
+   * @param execution the current execution
+   * @param pipelineName name of the pipeline to lookup
+   * @return the id of the pipeline or null if pipeline not found
+   */
+  public static String pipelineIdOrNull(PipelineExecution execution, String pipelineName) {
+    try {
+      return pipelineId(execution, pipelineName);
+    } catch (SpelHelperFunctionException e) {
+      if (e.getMessage().startsWith("Pipeline with name ")) {
+        return null;
+      }
+      throw e;
     }
   }
 }
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
@@ -16,8 +16,6 @@
 
 package com.netflix.spinnaker.orca.front50.pipeline;
 
-import static java.lang.String.format;
-
 import com.google.common.base.Strings;
 import com.netflix.spinnaker.kork.core.RetrySupport;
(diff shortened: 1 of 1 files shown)


marchello2000:
an exception from this function is most certainly a pipeline bug (user error) so I think exception here is more appropriate.

if the user does in fact not care then they can always check to not fail the stage on eval errors and they will get the result to be `null`

AbdulRahmanAlHamali:
My use case is that I have a parent pipeline, which calls a child pipeline, but right before calling it, it sets its configuration by using the `SavePipelineStage`:
https://github.com/spinnaker/orca/blob/master/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/SavePipelineStage.java

The child pipeline could exist prior to the call, in which case we need to provide its ID, or could be a totally new pipeline, in which case the ID should be null.

I have succeeded in doing that by writing this code
```python
pipeline_config_dict = asdict(pipeline_config)
pipeline_config_dict["id"] = "ID_TO_BE_CALCULATED_AT_RUNTIME"
pipeline_config_json = json.dumps(pipeline_config_dict)

self.pipeline = (
            "${" + f"#toBase64('{pipeline_config_json}'.replace('ID_TO_BE_CALCULATED_AT_RUNTIME', #pipelineId('{pipeline_config.name}')))" + "}"
        )
```

But the only thing stopping me now is that `#pipelineId` throws an exception instead of returning `null`, for cases when the child pipeline is not yet created. 

If you want, we can keep the old behaviour, and add a parameter that says `throwOnNotFound`. What do you think?

AbdulRahmanAlHamali:
Any thoughts?

marchello2000:
@AbdulRahmanAlHamali : where are you calling this code from? Is it all orchestrated by a Spinnaker pipeline? It seems that your better bet would be to
a) use EvaluateVariables stage right before your SavePipeline stage to compute what the ID should be, this eval stage can have `failOnFailedExpressions = false` which would mean that ID will be `null` on failure
b) use the result of evalVars in your save pipeline.

This will also make your pipeline more understandable, I think. What do you think?

AbdulRahmanAlHamali:
@marchello2000 yeah I think that does solve the problem indeed! Albeit less convenient. Why don't we allow the function to take another parameter called `throwOnNotFound`, that is set to `true` by default, but could be set to false? This feels better than the hack you're suggesting

On a side note: 
> where are you calling this code from?

I'm running this code in an AWS Lambda. In our installation we have a legacy "deployment configuration" that users provide in their code repositories. We take that configuration, and translate it in the Lambda to a Spinnaker pipeline, then run it on Spinnaker.

marchello2000:
@AbdulRahmanAlHamali we've made spel helpers a bit of a kitchen sink/trashcan thing that has very little consistency (unfortunately). Some functions fail one way some another, etc. Then there is also a master check at the stage level "should I fail on failed expressions?" which is false by default, unless it's an eval vars stage... lots of edge cases.
Adding another param to a function (prob boolean in this case) will make spel expression with it harder to grok. Also, I am not sure if SpEL actually supports function parameter overloading..?
(I am also a bit hesitant because there is a quite a bit usage of this function internally for us and I am not sure what all the usages are so I'd rather not making any potentially breaking changes here)
If it will make your life much easier, how about adding a `pipelineIdOrNull()` function? or we can start a pattern for `pipelineIdNoThrow()`, I don't have a preference here

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -66,7 +75,7 @@ public Functions getFunctions() {
    *
    * @param execution the current execution
    * @param pipelineName name of the pipeline to lookup
-   * @return the id of the pipeline or null if pipeline not found
+   * @return the id of the pipeline or exception if pipeline not found
```
AbdulRahmanAlHamali: @marchello2000 Ironically the documentation said it returned `null` here

AbdulRahmanAlHamali:
Alright, I added a `pipelineIdOrNull` function.

That said, I think it would be good to start following a pattern for SpEl functions not to throw exceptions when not absolutely necessary, because you can't easily handle an exception without doing the trick of "do not fail on failed expressions", etc.

marchello2000:
> Alright, I added a `pipelineIdOrNull` function.
> 
> That said, I think it would be good to start following a pattern for SpEl functions not to throw exceptions when not absolutely necessary, because you can't easily handle an exception without doing the trick of "do not fail on failed expressions", etc.

Very valid, thanks for following through and if you have any ideas on patterns - let me know

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -53,6 +53,15 @@ public Functions getFunctions() {
...
             "Lookup pipeline ID given the name of the pipeline in the current application",
+            new FunctionParameter(
+                PipelineExecution.class,
+                "execution",
+                "The execution containing the currently executing stage"),
+            new FunctionParameter(
+                String.class, "pipelineName", "A valid stage reference identifier")),
+        new FunctionDefinition(
+            "pipelineIdOrNull",
+            "Lookup pipeline ID given the name of the pipeline in the current application",
```
marchello2000:
```suggestion
            "Lookup pipeline ID (or null if not found) given the name of the pipeline in the current application",
```

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
marchello2000: 🤦‍♂ thank you!

AbdulRahmanAlHamali:
@marchello2000 my pleasure :) Thank you for your help. Can we also cherrypick this into 1.19?

marchello2000:
@AbdulRahmanAlHamali can you allow others to commit to your branch? otherwise it can't be rebased and merged

also once merged... you can `@spinnakerbot cherry-pick 1.19` and it will be cherrypicked

AbdulRahmanAlHamali:

PR #3539, opened 2020-03-23
fix(interlink): Add ability to skip flagging of certain interlink events

marchello2000:
Message flagging was designed to detect issues with events bouncing back and forth.
The StartPendingExecutions event is very voluminous and because it doesn't really have a unique and identifiable id it will
a) trip flagging all the time
b) will make flagging of all other events essentially impossible (since they are much much less frequent)

One solution is to fix the StartPendingExecutions event to be actually unique, but that requires a bit of surgery on the orca queue
Another solution is to skip flagging these events, I opted for skipping


marchello2000:
Posting for discussion

marchello2000:
will be reworking based on a conversation with @dreynaud

PR #3540, opened 2020-03-23
fix(kubernetes): Fix artifact binding for run job

ezimanyi:
Fixes spinnaker/spinnaker#5562

The Kubernetes V2 Run Job stage doesn't bind artifacts if the manifest source is text. This is because we only populate artifacts if the source is an artifact; luckily the logic to populate artifacts is independent of the aritfact source, so we can run it in either case by just removing the if condition.


ezimanyi:
@maggieneterval : Had to change enough of this that I wouldn't mind if you took another look.  Thanks!

PR #3541, opened 2020-03-23
feat(api-test): Initial scaffolding of orca-api-test harnesses

robzienert:
Initial groundwork for an Orca testkit that plugin developers can use to assert that their Orca-based plugins will work before deploying into an environment. This doesn't really provide anything meaningful, other than providing some additional in-memory variants of persistence classes that didn't exist before. 

My next step is to work on creating a test harness that will allow developers to create a `PipelineExecution` with a particular state (using `PipelineBuilder`) and step-wise assertions of Pipeline and Stage execution state.


orca-redis/src/main/java/com/netflix/spinnaker/orca/config/RedisConfiguration.java
```diff
@@ -35,6 +35,7 @@
 import rx.Scheduler;
 
 @Configuration
+@ConditionalOnProperty(value = "redis.enabled", matchIfMissing = true)
```
robzienert: This is the biggest change in this PR. We didn't have a mechanism for avoiding this configuration entirely, for whatever reason.

PR #3542, opened 2020-03-24
fix(interlink): Don't rebroadcast StartPending interlink event

marchello2000:
Previous implementation of triggering executions across peers (https://github.com/spinnaker/orca/pull/3519) had a flaw in that
the `StartPendingInterlinkEvent` would kept being rebroadcast if there are no pending executions causing an infinite stream of events until
`MessageFlagger` broke the chain.

This change makes it so that queue messages that result from processing the `StartPendingInterlinkEvent` are not rebroadcast if they aren't acted upon


marchello2000:
going to close this... this needs to be redone in some other way and will revert https://github.com/spinnaker/orca/pull/3519

PR #3548, opened 2020-03-25
feat(peering): Pending Executions Agent

marchello2000:
Introduces a pending execution agent which kicks off executions in the pending state.
Normally, such executions will be started by the act of completing a pipeline.
However, when there is a waiting execution in a different orca DB from where the current pipeline completed
it doesn't work.

With this change there is an agent that runs every ~15s (configurable) to see if there are any pending executions
which need to be kicked off and if so, kicks them off.

This is a continuation of failed attempt at implementing such functionality here:
https://github.com/spinnaker/orca/pull/3544

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
@@ -25,8 +25,8 @@ import com.netflix.spinnaker.security.AuthenticatedRequest;
 import java.time.Duration;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import lombok.NonNull;
 import lombok.Value;
 import lombok.experimental.NonFinal;
 import lombok.extern.slf4j.Slf4j;
@@ -61,8 +61,8 @@ public class CompoundExecutionOperator {
   }
 
   public void pause(
-      @NonNull ExecutionType executionType,
-      @NonNull String executionId,
+      @Nonnull ExecutionType executionType,
+      @Nonnull String executionId,
       @Nullable String pausedBy) {
     doInternal(
         runner::reschedule,
@@ -73,10 +73,10 @@ public class CompoundExecutionOperator {
   }
 
   public void resume(
-      @NonNull ExecutionType executionType,
-      @NonNull String executionId,
+      @Nonnull ExecutionType executionType,
+      @Nonnull String executionId,
       @Nullable String user,
-      @NonNull Boolean ignoreCurrentStatus) {
+      @Nonnull Boolean ignoreCurrentStatus) {
     doInternal(
         runner::unpause,
         () -> repository.resume(executionType, executionId, user, ignoreCurrentStatus),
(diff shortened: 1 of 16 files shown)


orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/pending/PendingExecutionAgentTest.kt
```diff
@@ -0,0 +1,134 @@
+/*
```
marchello2000: this file shouldn't be here, will remove later

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
```diff
@@ -25,6 +25,7 @@
 import java.time.Duration;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import lombok.NonNull;
```
dreynaud: is `lombok.NonNull` unused now?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/pending/PendingExecutionAgent.kt
```diff
@@ -0,0 +1,87 @@
...
+      for (configId in pendingConfigIds) {
+        val runningPipelines = executionRepository.retrievePipelinesForPipelineConfigId(configId, singleRunningCriteria)
+          .toList().toBlocking().single()
+
+        if (runningPipelines.isEmpty()) {
+          val lastCompletedPipeline = executionRepository.retrievePipelinesForPipelineConfigId(configId, lastCompletedCriteria)
+            .toList().toBlocking().single()
+
+          val purgeQueue = if (lastCompletedPipeline.any()) {
+            !(lastCompletedPipeline.first().isKeepWaitingPipelines)
```
dreynaud: why check the last completed and not the current pipeline config?

bonus question: is there always a last completed?

dreynaud:
LGTM, seems a lot more simple to manage and understand.

Just thinking through a scenario:

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/CompoundExecutionOperator.java
marchello2000: missed a spot, but fixed it now, thanks!

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/pending/PendingExecutionAgent.kt
marchello2000: yeah, this boggles my mind too... the way it works today is that the pipeline that completes decides if we purge or not. seems counter-intuitive to me, should we just change it - it would make everything simpler

marchello2000:
> LGTM, seems a lot more simple to manage and understand.
> 
> Just thinking through a scenario:
> 
> * some pipeline is running in orca A
> * it is peered in orca B
> * there is a new execution request in orca A, it becomes pending
> * there is a new execution request in orca B, it also becomes pending because the running one was already peered
> * as soon as the execution completes, it kicks off `StartWaitingExecutions` in orca A
> * the completed execution is peered in orca B, but the new execution is not peered yet
> * if the pending executions agent runs often enough, it also kicks off `StartWaitingExecutions` in orca B, so we now have 2 concurrent executions in orca A and B
> 
> What do you think, seems like a realistic race condition?

very true. but that is true today as well, if someone is submitting many workloads across both orcas... i thought we decided this "concurrent" business is best effort anyway?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/pending/PendingExecutionAgent.kt
dreynaud: I think we can change it, it makes sense

PR #3552, opened 2020-03-26
perf(peering): improve peering perf

marchello2000:
A big chunk of time was spent storing all stages of running executions.
This is not necessary because most of those stages don't change (only about 8% change with ~30s agent interval)
Only store stages that have a different `updated_at` timestamp (or those that don't exist).
This improves peering performance by about 25%

### DIFF
--- a/orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/ExecutionCopier.kt
+++ b/orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/ExecutionCopier.kt
@@ -131,7 +131,8 @@ open class ExecutionCopier(
       // We can't have the execution update in the time we capture its stages and the time we capture the execution it self
       // It's totally fine for the stages to be "newer" than the execution since that will be fixed up in the next agent run
       val executionRows = srcDB.getExecutions(executionType, idsToMigrate)
-      val stagesToMigrate = srcDB.getStageIdsForExecutions(executionType, idsToMigrate)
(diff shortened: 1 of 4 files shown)


orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/ExecutionCopier.kt
```diff
@@ -131,29 +131,36 @@ open class ExecutionCopier(
...
       // It is possible that the source stage list has mutated. Normally, this is only possible when an execution
       // is restarted (e.g. restarting a deploy stage will delete all its synthetic stages and start over).
       // We delete all stages that are no longer in our peer first, then we update/copy all other stages
-      val stagesPresent = destDB.getStageIdsForExecutions(executionType, idsToMigrate)
-      val stagesToMigrateHash = stagesToMigrate.toHashSet()
-      val stagesToDelete = stagesPresent.filter { !stagesToMigrateHash.contains(it) }
+      val stagesInDest = destDB.getStageIdsForExecutions(executionType, idsToMigrate)
+      val stagesInDestMap = stagesInDest.map { it.id to it }.toMap()
+
+      val stagesToDelete = stagesInDest.filter { !stagesInSourceHash.contains(it.id) }.map { it.id }
```
dreynaud: nit: `stageIdsToDelete`

marchello2000: thank you!

PR #3562, opened 2020-03-31
fix(bakery): only conditionally override input artifact account in CreateBakeManifestTask

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4737, https://github.com/spinnaker/spinnaker/issues/5533

Helm `inputArtifacts` have unfortunately never worked with `artifactsRewrite` enabled: we are unconditionally overwriting the resolved artifact's account with the unresolved input artifact's account (which will always be empty if configured with `artifactsRewrite`). Let's only overwrite the resolved artifact's account when there is a non-empty account defined on the input artifact (in other words, when it was configured with `artifactsRewrite` disabled).

I think we should cherry-pick this back into each supported release given the number of users who have commented on the first issue linked above.


maggieneterval:
@Mergifyio backport release-1.18.x
@Mergifyio backport release-1.17.x

maggieneterval:
@Mergifyio backport release-1.18.x

maggieneterval:
@Mergifyio backport release-1.17.x

PR #3569, opened 2020-04-01
fix(peering): Improve peering of deletes

marchello2000:
Currently, in order to detect deletes, we need to perform a period FULL diff of the DB. This is EXTREMELY expensive and causes peering delay to be inconsistent.
This change adds what is essentially a log of all "recent" deletes into the new `deleted_executions` table.
The peering agent can then scan that table (remembering the last known "deletion" id) and delete all executions listed in the table.
Deletes older than some time (e.g. 2days) are cleaned up from the table during a delete.

Because execution deletes are now more complicated (need to add a row into the `deleted_executions` table) I had to change how cleanup agents work (in order not to duplicate the deletion code).
While there, I refactored them a little to extract [some] of the common functionality.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
@@ -164,6 +164,14 @@ class DualExecutionRepository(
     return select(type, id).delete(type, id)
   }
 
+  override fun delete(type: ExecutionType, idsToDelete: MutableList<String>) {
+    // NOTE: Not a great implementation, but this method right now is only used on SqlExecutionRepository which has
+    // a performant implementation
+    idsToDelete.forEach { id ->
+      delete(type, id)
+    }
+  }
+
   override fun retrieve(type: ExecutionType): Observable<PipelineExecution> {
     return Observable.merge(
       primary.retrieve(type),
--- a/orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
+++ b/orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
@@ -20,10 +20,12 @@ import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType
 import com.netflix.spinnaker.orca.notifications.AbstractPollingNotificationAgent
 import com.netflix.spinnaker.orca.notifications.NotificationClusterLock
 import org.slf4j.LoggerFactory
+import java.time.Duration
+import java.time.Instant
 import kotlin.math.max
 
 /**
- *
+ * The orca agent that performs peering of executions across different orca DBs
  */
 class PeeringAgent(
   /**
@@ -69,25 +71,22 @@ class PeeringAgent(
   private val log = LoggerFactory.getLogger(javaClass)
   private var completedPipelinesMostRecentUpdatedTime = 0L
   private var completedOrchestrationsMostRecentUpdatedTime = 0L
-  private var runCount = 0
+  private var deletedExecutionCursor = 0
 
   override fun tick() {
-    // Temporary "hack" to replicate deletes, for now we run a full diff (very expensive) every 10 ticks of the agent
-    // Better solution forth coming
-    if (runCount++ > 10) {
-      completedOrchestrationsMostRecentUpdatedTime = 0L
-      completedPipelinesMostRecentUpdatedTime = 0L
-      runCount = 0
-    }
-
     if (dynamicConfigService.isEnabled("pollers.peering", true) &&
       dynamicConfigService.isEnabled("pollers.peering.$peeredId", true)) {
-      peerExecutions(ExecutionType.PIPELINE)
-      peerExecutions(ExecutionType.ORCHESTRATION)
+      peeringMetrics.recordOverallLag() {
+        peerExecutions(ExecutionType.PIPELINE)
+        peerExecutions(ExecutionType.ORCHESTRATION)
+        peerDeletedExecutions()
+      }
     }
   }
 
   private fun peerExecutions(executionType: ExecutionType) {
+    val start = Instant.now()
+
     val mostRecentUpdatedTime = when (executionType) {
       ExecutionType.ORCHESTRATION -> completedOrchestrationsMostRecentUpdatedTime
       ExecutionType.PIPELINE -> completedPipelinesMostRecentUpdatedTime
@@ -99,11 +98,11 @@ class PeeringAgent(
     if (isFirstRun) {
       peerCompletedExecutions(executionType)
     } else {
-      peeringMetrics.recordLag(executionType) {
-        peerCompletedExecutions(executionType)
-        peerActiveExecutions(executionType)
-      }
+      peerCompletedExecutions(executionType)
+      peerActiveExecutions(executionType)
     }
+
+    peeringMetrics.recordLag(executionType, Duration.between(start, Instant.now()))
   }
 
   /**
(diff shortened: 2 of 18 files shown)


orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -69,25 +71,22 @@ class PeeringAgent(
...
-      completedOrchestrationsMostRecentUpdatedTime = 0L
-      completedPipelinesMostRecentUpdatedTime = 0L
-      runCount = 0
-    }
-
     if (dynamicConfigService.isEnabled("pollers.peering", true) &&
       dynamicConfigService.isEnabled("pollers.peering.$peeredId", true)) {
-      peerExecutions(ExecutionType.PIPELINE)
-      peerExecutions(ExecutionType.ORCHESTRATION)
+      peeringMetrics.recordOverallLag() {
```
marchello2000, line 79: @dreynaud added this to capture accurate lag as a simple metric

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -99,11 +98,11 @@ class PeeringAgent(
...
     } else {
-      peeringMetrics.recordLag(executionType) {
-        peerCompletedExecutions(executionType)
-        peerActiveExecutions(executionType)
-      }
+      peerCompletedExecutions(executionType)
+      peerActiveExecutions(executionType)
     }
+
+    peeringMetrics.recordLag(executionType, Duration.between(start, Instant.now()))
```
marchello2000, line 105: i changed this because it was a pain to mock/test otherwise

orca-sql/src/main/resources/db/changelog/20200327-deleted-executions-table.yml
```diff
@@ -0,0 +1,11 @@
+databaseChangeLog:
+  - changeSet:
+      id: create-deleted-executions-table
+      author: mvulfson
+      changes:
+        - sql:
+            dbms: mysql
+            sql: CREATE TABLE `deleted_executions` (id INT AUTO_INCREMENT, execution_id char(26) NOT NULL, execution_type ENUM("pipeline", "orchestration") NOT NULL, deleted_at DATETIME NOT NULL, CONSTRAINT deleted_executions_pk PRIMARY KEY (id))
```
marchello2000, line 8: i couldn't figure out how to make `enum` columns... hence this became plain SQL, if you know how to do that, let m eknow

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/cleanup/OldPipelineCleanupPollingNotificationAgent.kt
```diff
@@ -39,64 +36,27 @@ import java.time.temporal.ChronoUnit
 import java.util.concurrent.atomic.AtomicInteger
 
 @Component
-@ConditionalOnExpression("\${pollers.old-pipeline-cleanup.enabled:false} && !\${execution-repository.redis.enabled:false}")
+@ConditionalOnExpression("\${pollers.old-pipeline-cleanup.enabled:false} && \${execution-repository.sql.enabled:false}")
```
marchello2000, line 39: @robzienert I flipped the logic here, it made more sense to me to base this on the presence of SQL rather than lack of presence of redis, or was there a specific reason for the logic you had before?

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -149,6 +148,39 @@ class PeeringAgent(
...
 
+  /**
+   * Propagate deletes
+   * NOTE: ids of executions (both orchestrations and pipelines) that have been deleted are stored in the deleted_executions table
+   * the "id/primarykey" on that table is an auto-incrementing int, so we use that as a "cursor" to know what we've deleted and
+   * what still needs to be deleted.
+   * There is no harm (just some wasted RDS CPU) to "deleting" an execution that doesn't exist
+   */
+  private fun peerDeletedExecutions() {
+    val deletedExecutionIds = srcDB.getDeletedExecutions(deletedExecutionCursor)
```
dreynaud, line 159: any concern about the growth of the `deleted_executions` table and how much data this call might end up returning with `deletedExecutionCursor = 0`?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -267,30 +269,59 @@ class SqlExecutionRepository(
...
+   * NOTE: this method explicitly does not check if the execution is foreign or not.
+   */
+  override fun delete(type: ExecutionType, idsToDelete: List<String>) {
+    jooq.transactional { tx ->
+      deleteInternal(tx, type, idsToDelete)
     }
   }
 
+  private fun deleteInternal(dslContext: DSLContext, type: ExecutionType, idsToDelete: List<String>) {
+    val correlationField = if (type == PIPELINE) "pipeline_id" else "orchestration_id"
```
dreynaud, line 290: Nit: I'd rather be defensive and not assume that there are only 2 possible values here:

```kotlin
val correlationField = when (type) {
    PIPELINE -> "pipeline_id"
    ORCHESTRATION -> "orchestration_id"
    else -> throw IllegalStateException("Unexpected field $type")
}
```

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -267,30 +269,59 @@ class SqlExecutionRepository(
...
+      insertQueryBuilder = insertQueryBuilder
+        .values(id, type.toString(), now())
+    }
+
+    insertQueryBuilder
+      .execute()
+
+    dslContext
+      .deleteFrom(table("deleted_executions"))
+      .where(field("deleted_at").lt(timestampSub(now(), 2, DatePart.DAY)))
```
dreynaud, line 321: oh we only keep things deleted in the last 2 days in `deleted_executions`, am I reading this right? Sounds good but maybe we want to make that duration configurable

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
marchello2000: the `SqlExectionRepository` deletes rows older than 2days. But as you said, I will make it configurable

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
marchello2000: will do

marchello2000: actually... it adds a bunch of extra complexity that I am not sure it's worth it. I really can't imagine a situation where this would matter. If we delete too quickly (e.g. before the peer was able to peer them)... oh, well, when the peer is red/blacked it will do a full diff anyway and will pick up deletes. if the concern is that the table will fill up... i don't think that's a problem. even if there is an execution that is deleted every second that's 172k rows which is of course a bit crazy, but not really a big deal (since the peer will only read since its known cursor). I will reduce the retentio to 1 day

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
```diff
@@ -164,6 +164,14 @@ class DualExecutionRepository(
     return select(type, id).delete(type, id)
   }
 
+  override fun delete(type: ExecutionType, idsToDelete: MutableList<String>) {
+    // NOTE: Not a great implementation, but this method right now is only used on SqlExecutionRepository which has
+    // a performant implementation
```
robzienert, line 169: I don't follow - why can't we use `delete(type, idsToDelete)` here? Why do we have to iterate over each ID?

marchello2000: we don't know which repos to perform the deletes on. I guess i could call:
```java
primary.delete(type, idsToDelete)
previous.delete(type, idsToDelete)
```

but depending on what primary and previous are (e.g. not SQL) then this is even less performant, probably?

marchello2000: eh, i take it back, it's the same since there is a select somewhere no matter what

PR #3570, opened 2020-04-01
feat(core): Add DynamicStageResolver to help stage plugin migrations

robzienert:
See class docs for description.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/DefaultStageResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/DefaultStageResolver.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca;
+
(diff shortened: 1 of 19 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/DefaultStageResolver.java
```diff
@@ -0,0 +1,81 @@
...
+  private final Map<String, StageDefinitionBuilder> stageDefinitionBuilderByAlias = new HashMap<>();
+
+  public DefaultStageResolver(
+      Collection<StageDefinitionBuilder> stageDefinitionBuilders,
+      Collection<SimpleStage> simpleStages) {
+    for (StageDefinitionBuilder stageDefinitionBuilder : stageDefinitionBuilders) {
+      stageDefinitionBuilderByAlias.put(stageDefinitionBuilder.getType(), stageDefinitionBuilder);
+      for (String alias : stageDefinitionBuilder.aliases()) {
+        if (stageDefinitionBuilderByAlias.containsKey(alias)) {
+          throw new DuplicateStageAliasException(
```
marchello2000, line 44: I guess this is how it was before... but it seems a bit weird to me that we don't check that the `type` is unique.
And yet someone could override the `getType` (and they do) and mess it up..perhaps we make it not overrideable (though it's used in tests...)?
Anyway, i think checking `getType` is something we should do

marchello2000:
> See class docs for description.

not to be a pain in the butt, but i prefer this message to be duplicated in the commit message. Commit messages are searchable and it's pretty useful, i think (https://dhwthompson.com/2019/my-favourite-git-commit)

orca-core/src/main/java/com/netflix/spinnaker/orca/DefaultStageResolver.java
srekapalli: stageDefinitionBuilderByAlias map is keyed with type of the builder and not by alias value..?

robzienert: I'd rather do this as part of a separate change, I think. It would be valuable to check this, for sure, but I don't want to squish a primary path change in with this experimental stuff.

PR #3572, opened 2020-04-02
fix(plugins): allow more than one SimpleStage impl

claymccoy:
Multiple SimpleStages keeps Orca from starting because SimpleTask had a Component annotation and took a SimpleStage in its constructor. Orca would try to construct all the Tasks for taskResolver and when it got to SimpleTask it wouldn't know which SimpleStage to inject. So SimpleTask obviously shouldn't be a Component. But once the Component annotation was removed SimpleStages would not execute.
The TaskResolver now needs to load all the SimpleStages and wrap each one in a SimpleTask. They need to be keyed by the SimpleStage impl class name, they can't all be keyed as a SimpleTask. But you could only use classes of type Task to refer to them in the task graph, even though it ultimately just gets turned into a String. I made some modification to stick the SimpleStage impl name in the task graph instead.
On the other side the RunTaskHandler looked up tasks based on a list of Task components and SimpleTask is no longer in that list now that it is not a Component. There would only be the one if it was anyway. Now RunTaskHandler looks up tasks using the TaskResolver. Doesn't it make more sense to use the TaskResolver to resolve tasks anyway? I also use the implementingClass on the TaskModel to look up the task rather than the RunTask.taskType because it would always be SimpleTask when someone is using SimpleStages. This seems cleaner but we still use RunTask.taskType and resolve the task based on type as a backup mainly for existing tests.

### DIFF
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/graph/TaskNode.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/graph/TaskNode.java
@@ -110,6 +110,17 @@ public interface TaskNode {
     }
 
     /**
+     * Adds a task to the current graph.
+     *
+     * @param task the task node to add
+     * @return this builder with the new task appended.
+     */
+    public Builder withTask(TaskNode task) {
+      graph.add(task);
+      return this;
+    }
+
+    /**
      * Adds a sub-graph of tasks that may loop if any of them return {@link
      * ExecutionStatus#REDIRECT}. The sub-graph will run after any previously added tasks and before
      * any subsequently added ones. If the final task in the sub-graph returns {@link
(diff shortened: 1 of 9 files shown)


claymccoy:
I get a lot of "Pipeline did not complete" when running tests in `orca-queue-redis` and `orca-queue-sql` now, and I'm not sure why yet.

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/graph/TaskNode.java
```diff
@@ -162,8 +173,16 @@ public GraphType getType() {
     }
   }
 
+  interface DefinedTask {
```
robzienert, line 176: Need docs on this interface & its methods.

PR #3575, opened 2020-04-04
Fix stage resolver

srekapalli:
Stage resolver requires alias to be passed in if present on the stage context to resolve the builders.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/StageNavigator.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/StageNavigator.java
@@ -44,7 +44,12 @@ public class StageNavigator {
    */
   public List<Result> ancestors(StageExecution startingStage) {
     return startingStage.ancestors().stream()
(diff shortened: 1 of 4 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/StageNavigator.java
```diff
@@ -44,7 +44,12 @@ public StageNavigator(StageResolver stageResolver) {
    */
   public List<Result> ancestors(StageExecution startingStage) {
     return startingStage.ancestors().stream()
-        .map(it -> new Result(it, stageResolver.getStageDefinitionBuilder(it.getType())))
+        .map(
+            it ->
+                new Result(
+                    it,
```
srekapalli, line 50: Passing in alias to the resolver. This is the main change.

srekapalli: Weird format with spotless :(

PR #3579, opened 2020-04-08
feat(plugins): Add orca-plugins-test using kork-plugins-tck and test implementations for existing orca-api extensions

jonsie:
I think there's the potential to unify the efforts here and in `orca-api-tests` to produce basically one `OrcaFixture` that could be exported and used in downstream plugin repositories to verify plugins load and verify plugin functionality.  However, this does not attempt to do that.  I'm willing to refactor this to support that pattern if we can identify a good way to do it, either before or after this merges.


gradle.properties
```diff
@@ -1,7 +1,7 @@
 #Tue Mar 31 20:41:32 UTC 2020
 fiatVersion=1.18.0
 enablePublishing=false
-korkVersion=7.31.0
+korkVersion=7.34.0-SNAPSHOT
```
jonsie, line 4: Will not build until this PR merges and a release is cut:

https://github.com/spinnaker/kork/pull/595

orca-plugins-test/src/test/kotlin/com/netflix/spinnaker/orca/plugins/test/OrcaPluginsFixture.kt
```diff
@@ -0,0 +1,116 @@
...
+import org.springframework.context.annotation.Primary
+import org.springframework.test.context.ContextConfiguration
+import org.springframework.test.context.TestPropertySource
+import java.io.File
+import java.time.Clock
+import java.time.Duration
+
+class OrcaPluginsFixture : PluginsTckFixture, OrcaTestService() {
+  val objectMapper: ObjectMapper = ObjectMapper(YAMLFactory())
+    .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
```
robzienert, line 53: Should this use `OrcaObjectMapper`?

jonsie: `OrcaObjectMapper` uses the standard `JsonFactory`, this needs to use the `YAMLFactory`.  We don't really have a use for an ObjectMapper that uses the `YAMLFactory` in Orca, except for in this test to verify configs are loaded correctly.  So I think it's fair to just leave this here.

orca-plugins-test/src/main/kotlin/com/netflix/spinnaker/orca/plugins/OrcaPlugin.kt
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.plugins
+
+import org.pf4j.Plugin
+import org.pf4j.PluginWrapper
+
+class OrcaPlugin(pluginWrapper: PluginWrapper?) : Plugin(pluginWrapper)
```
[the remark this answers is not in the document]
jonsie: Ah, true.

orca-plugins-test/orca-plugins-test.gradle
```diff
@@ -0,0 +1,41 @@
...
+
+  testImplementation("com.netflix.spinnaker.keiko:keiko-mem:$keikoVersion")
+  testImplementation("com.netflix.spinnaker.kork:kork-plugins-tck")
+
+  testImplementation("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml")
+  testImplementation("com.fasterxml.jackson.dataformat:jackson-dataformat-properties")
+  testImplementation "com.fasterxml.jackson.module:jackson-module-kotlin"
+
+  testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+  testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine")
```
robzienert, line 34: I think you’ll want a lot of these deps as implementation, not just for test.

orca-plugins-test/orca-plugins-test.gradle
```diff
@@ -0,0 +1,45 @@
...
+}
+
+test {
+  filter {
+    includeTestsMatching "com.netflix.spinnaker.orca.plugins.test.OrcaPluginsTest"
```
jonsie, line 39: Without filtering down to a specific test, the task `orca-plugins-test:test` fails.  The reason it fails is because the Orca `Main` class package does not have an Implementation-Version set (which causes tests to fail as the plugin that should not load due to an unmet system requirement constrain actually does load).  I'm not sure why filtering down to a specific test means that the `Main` class package has Implementation-Version set, the only thing I can think of is we read the orca-web dependency from a compiled JAR when filtering down to one test, whereas running all tests via `orca-plugins-test:test` means we run `orca-web` at runtime.  

Anyways, need to figure this out, just capturing my thoughts for later.

PR #3587, opened 2020-04-11
feat(runJob): Add configurable timeout to Run Job stage

agoyal-sfdc:
As discussed in the k8s SIG, it can be desirable to configure the timeout for `Run Job (Manifest)`, similar to #3502


agoyal-sfdc:
@ezimanyi @maggieneterval looks like there's no unit testing that currently exists for this task (or any other job-related task for that matter), let me know if you'd like me to add those. Thanks!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
```diff
@@ -52,6 +53,23 @@ public class WaitOnJobCompletion extends AbstractCloudProviderAwareTask implemen
 
   static final String REFRESH_TYPE = "Job"
 
+  @Override
+  long getDynamicTimeout(@Nonnull StageExecution stage) {
+    JobTimeout timeout = stage.mapTo(JobTimeout.class)
+    if (timeout.timeoutMinutes.isPresent()) {
+      return TimeUnit.MINUTES.toMillis(timeout.timeoutMinutes.getAsLong())
```
agoyal-sfdc, line 60: The [existing default timeout](https://github.com/spinnaker/orca/pull/3587/files#diff-75848b16a2174760f7f312a0b888d6c1R40) is in hours, but the dynamic one is in minutes. Everything gets converted to milliseconds, but if it makes sense, I can change the default to minutes as well, for better readability maybe?

ezimanyi: Up to you...I guess it might slightly increase readability, so might not be a bad idea.  One thing would be great regardless is to update that default timeout to be `final` as we should definitely never be changing that global value.

agoyal-sfdc: Makes sense - will update that

agoyal-sfdc:
> I wonder if we should take the JobTimeout class (which looks almost identical to ManifestTimeout) and make a top-level class that can be shared (somewhere in com.netflix.spinnaker.orca.api.pipeline). I think it will probably be reasonable common for other stages to start implementing this, and it would be better to not need all of them to define the same model class.

@ezimanyi Makes sense, however, I have a question about how to implement that. the constructor right now has the `@JsonProperty` annotation to deserialize the stage definition and get the `timeoutMinutes` field
```
JobTimeout(@JsonProperty("timeoutMinutes") OptionalLong timeoutMinutes)
```

However, `orca-api` does not have this package defined. Should I add it there? Or maybe this class belongs elsewhere? Wasn't sure, so figured I'd ask.

Sounds good about the unit tests, will add those as well.

ezimanyi:
One option would be to add that dependency to `orca-api` but I know that's a fairly new module that is intended to provide an API surface for plugins, and given that this is new functionality we should probably but in in a non-extensible place until there's a clear need for it to be in the API.  What about in `com.netflix.spinnaker.orca.pipeline.model`in `orca-core`?  That looks like it might be a good place for it.

agoyal-sfdc:
@ezimanyi added `OverridableStageTimeout` to `com.netflix.spinnaker.orca.pipeline.model`

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletionTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Test
+  void jobTimeoutOneMinute() {
+    WaitOnJobCompletion task = new WaitOnJobCompletion();
+    StageExecutionImpl myStage = createStageWithContext(ImmutableMap.of("timeoutMinutes", 1));
+    assertThat(task.getDynamicTimeout(myStage)).isEqualTo(60000);
+  }
+
+  @Test
+  void jobTimeoutMissingTimeoutShouldBe30Minutes() {
```
ezimanyi, line 26: (minor) I think the timeout in the test name doesn't match what's in the test

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletionTest.java
```diff
@@ -0,0 +1,50 @@
+package com.netflix.spinnaker.orca.clouddriver.tasks.job;
```
ezimanyi, line 1: Missing copyright header.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/OverridableStageTimeout.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2020 Netflix, Inc.
```
ezimanyi, line 2: This should be you or your company instead of Netflix.

agoyal-sfdc:
Thanks for the help and thorough review @ezimanyi!

ezimanyi:
Github actions seems to be broken; going to close and re-open to see if that helps.

ezimanyi:
Oh, looks like there is an ongoing incident for Github actions; will re-open and I guess this should auto-merge once the tests run.

marchello2000:
@agoyal-sfdc : I know this is merged already/i am late to the party.. why is this needed?
I must be missing something, because there is already an override functionality on every stage (so long as the task implements `OverridableTimeoutRetryableTask` see (https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt#L295) was this functionality not enough?

agoyal-sfdc:
@marchello2000 this PR (and #3502) uses the functionality enabled by `OverridableTimeoutRetryableTask` (and `RetryableTask`), but there was currently no way to actually provide this timeout. Since the `getDynamicTimeout` was not defined, it defaulted to `getTimeout` which returns the hardcoded timeout value.

So the pieces were in place for the timeout to be provided by the user, but there was no way to actually provide it without this.

marchello2000:
I think the functionality is just confusing. Both this and the PR you mention #3502 aren't needed. If you look at the link I provided (https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt#L295) if the stage has `stageTimeoutMs` specified on its context, that value will be used instead of the value in the task. Does that make sense. Am I missing something? This is even exposed in `deck` 
image

Unless i am missing something we should back to these changes to reduce confusion

agoyal-sfdc:
Ah interesting - I guess none of us realized that this stage-level timeout effected the specific tasks as well.

agoyal-sfdc:
@marchello2000 I think there potentially could be an issue with using the stage-level timeout though. If you have `n` `OverridableTimeoutRetryableTask`s that comprise a certain stage, then providing the timeout in such a way would make the effective timeout for the stage `n*stageTimeoutMs` which is not desirable either. Could be wrong about this, but wanted to put it out there.

marchello2000:
that's not quite how the timeouts work in orca. The `stageTimeoutMs` is the totalexecution time for the stage, not each task.
Before checking task timeout orca will check stage timeout: https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt#L283-L284

I agree that the timeout story is a bit confusing, but I am not sure that's a reason to add more knobs to the confusing system. 
Because at this point here is what can happen: you will plumb this value through in `deck` for the two stages it's on. Your enduser will now have two places to specify a timeout for these two stages (WaitForJobCompletion and ManifestStable). One of those values (`stageTimeoutMs`) will win and it's going to be pain to debug/very error-prone for users.

At a micro level, the change is pretty simple. But at a macro level, you had to make the change because you didn't know about the generic timeout functionality, someone else comes along a month from now and adds another timeout because they are confused by the two that already exist and then we have more confusion for the end-user, complexity for the operators, and complexity for developers.

I think the options here are: if generic functionality is sufficient for the problem you're trying to solve we should undo these two changes. If it's not, it's worth thinking about a better overall timeout strategy.

ezimanyi:
@agoyal-sfdc @marchello2000 : I did not realize that this functionality existed on all stages!  (I guess I should pay more attention to all those checkboxes at the bottom of each stage.)  Based on that, I think it makes sense to revert this and instead rely on the existing functionality.  What do you think @agoyal-sfdc?

I think it would be best to revert this quickly and get the reversion into 1.20 so users don't rely on this in an actual released version of Spinnaker.  (I guess the fact it's not in the UI mitigates this somewhat, but still would reduce the chances that people start using this.)

agoyal-sfdc:
@marchello2000 @ezimanyi yes totally agree - I don't think it makes sense to add more knobs when something that exists can achieve the same functionality. We should, however, maybe make it more clear in documentation etc how this works.

So yeah, let's revert this and #3502

agoyal-sfdc:
Thanks for pointing this out @marchello2000 btw!!

PR #3594, opened 2020-04-14
fix(aws/findimage): Show warning if search results were truncated

german-muzquiz:
Helps to fix https://github.com/spinnaker/spinnaker/issues/3355

Added a new parameter to `ImageFinder.byTags` for saving any warnings that should be presented to the user in `Find Image` stage. When the returned list from clouddriver matches the max result size, it probably means that the result list was truncated and the stage is not producing  the desired result.
This translates in the following warning in the stage results execution:

image

This is currently only implemented for AWS provider.

Clouddriver PR: https://github.com/spinnaker/clouddriver/pull/4508


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/AmazonImageFinder.java
```diff
@@ -85,6 +91,11 @@ public boolean test(AmazonImage amazonImage) {
             .sorted()
             .collect(Collectors.toList());
 
+    if (allMatchedImages.size() == MAX_SEARCH_RESULTS) {
+      warningsCollector.add(
+          "Truncated too many results matching search criteria. Consider refining the search.");
```
robzienert, line 96:
```suggestion
          "Too many results matching search criteria: Consider refining the search.");
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/AmazonImageFinder.java
```diff
@@ -85,6 +91,11 @@ public boolean test(AmazonImage amazonImage) {
             .sorted()
             .collect(Collectors.toList());
 
+    if (allMatchedImages.size() == MAX_SEARCH_RESULTS) {
```
robzienert, line 94:
```suggestion
    if (allMatchedImages.size() >= MAX_SEARCH_RESULTS) {
```

Typically speaking, I'm pretty pessimistic of equality operators in these cases not introducing a regression in months, years to come.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/AmazonImageFinder.java
```diff
@@ -33,6 +33,9 @@
 
 @Component
 public class AmazonImageFinder implements ImageFinder {
+
+  private static final int MAX_SEARCH_RESULTS = 1000;
```
robzienert, line 37: Any thoughts on making this configurable? 1000 seems like a lot?

german-muzquiz: I think in the beginning the 1000 limit was set in clouddriver for reading "enough" entries from cache before applying some filters in memory. Now that this proposal reads _all_ matching entries from cache, the limit is just useful for avoiding network congestion, so it can be decreased both here and in clouddriver.

Ideally, clouddriver's response should include a way for orca to know if the results were truncated or not, to avoid having them to agree in a common number. Here are a couple of alternatives without changing too much the existing api:

* Include in the response a [Link Header](https://tools.ietf.org/html/rfc5988#page-6). Link headers are a way for supplying pagination information instead of embedding that data in the response payload.
* Include in the response a custom header, like `X-SPINNAKER-TOTAL-COUNT` or `X-SPINNAKER-PARTIAL-RESULT`, useful for identifying if the given response is complete or not.

robzienert: With how Retrofit works, having response envelopes are far more preferential to using headers.

```
{
  "results": [ ... ]
  "totalCount": 12312312312312312312
}
```

robzienert:
Changing the payload would be more work... incremental improvement is good.

robzienert:
Does this PR depend on https://github.com/spinnaker/clouddriver/pull/4508 being merged?

german-muzquiz:
> Does this PR depend on [spinnaker/clouddriver#4508](https://github.com/spinnaker/clouddriver/pull/4508) being merged?

Any of the two PRs (orca, clouddriver) can be merged first, since today clouddriver already returns 1000 results to orca. Clouddriver PR is needed to return the correct images even if there are less than 1000 results in the payload.

PR #3599, opened 2020-04-14
refactor(core): Clean up dependencies

ezimanyi:
An upcoming kork bump will remove some things that were api dependencies and make them implementation; update places that were relying on these to explicitly include them.

The places where groovy is added as an API dependency is because there are classes that due to groovy magic subclass a groovy type, so consumers need core groovy on the classpath.

### DIFF
--- a/orca-front50/orca-front50.gradle
+++ b/orca-front50/orca-front50.gradle
@@ -19,6 +19,9 @@ apply from: "$rootDir/gradle/groovy.gradle"
 dependencies {
   implementation(project(":orca-core"))
   implementation(project(":orca-retrofit"))
+
+  api("org.codehaus.groovy:groovy")
+
   implementation("com.netflix.spinnaker.fiat:fiat-api:$fiatVersion")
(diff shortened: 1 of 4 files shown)


orca-front50/orca-front50.gradle
```diff
@@ -19,6 +19,9 @@ apply from: "$rootDir/gradle/groovy.gradle"
 dependencies {
   implementation(project(":orca-core"))
   implementation(project(":orca-retrofit"))
+
+  api("org.codehaus.groovy:groovy")
```
plumpy, line 23: Does this really need to be API? I don't really see anything that needs to be exported, but maybe there are things in the bytecode?

ezimanyi: Yeah, I added a comment above in the PR message...but when you try to consume the `Application` class it says that the superclass (which is a groovy object) is not on the classpath.

plumpy: > Yeah, I added a comment above in the PR message...

Whoops, so you did.

PR #3603, opened 2020-04-15
fix(kubernetes): correctly write Patch Manifest patchBody (#3602)

ezimanyi:
Prior to [this refactor](https://github.com/spinnaker/orca/pull/3402), PatchManifestTask read patchBody directly from the ManifestEvaluator's result. Now, we are attempting to read it off the stage's context, but we deserialized the context with the same model, PatchManifestContext, that we used to deserialize the initial stage inputs, which tries to read manifests from patchBody (at this point, we should be reading manifests from the manifests key instead). This is intended to be a cherry-pickable fix, but a better more invasive fix would define different models for the inputs and outputs of each task and assert that the context had required values before beginning the task.


ezimanyi:
Manual cherry pick of #3602; the only conflict was in imports.

PR #3605, opened 2020-04-15
fix(kubernetes): Fix artifact binding in patch manifest stage

ezimanyi:
The patch manifest stage was incorrectly setting allArtifacts to just the required artifacts, which was breaking artifact resolution in clouddriver. Set this to all of the artifacts (both required and optional).


ezimanyi:

PR #3612, opened 2020-04-16
feat(cleanup agents): Add ability for special treatment of apps

marchello2000:
Add the ability to treat certain applications differently from pipeline clean up perspective.
Sometimes (e.g. for compliance reasons), it is useful to have different pipeline clean up threshold for certain applications.
This change adds `exceptionalApplications` setting to the `OldPipelineCleanupAgent` to handle this case.

The config looks like this:

```yaml
pollers:
  oldPipelineCleanup:
    enabled: true
    exceptionalApplicationsThresholdDays: 365
    exceptionalApplications:
      - application1
      - application2
```


marchello2000:
recommend reviewing with whitespace diff off:
https://github.com/spinnaker/orca/pull/3612/files?diff=split&w=1

PR #3616, opened 2020-04-19
chore(build): update to newer spinnaker project plugin

jonsie:
Also enables the orca-plugins-test module.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,31 +14,16 @@
  * limitations under the License.
  */
 
-buildscript {
-  repositories {
-    jcenter()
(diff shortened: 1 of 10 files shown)


orca-api-tck/src/main/resources/orca-test-app.yml
```diff
@@ -0,0 +1,31 @@
+front50:
```
jonsie, line 1: Oops though I stashed this.

build.gradle
```diff
@@ -16,29 +16,21 @@
 
 buildscript {
```
cfieber, line 17: you shoudl be able to delete the whole buildscript block here

PR #3624, opened 2020-04-22
fix(titus): custom job timeout should match task timeout

marchello2000:
There is a lot of confusion for users when they specify a timeout on the batch job on Titus if that timeout is larger than 2hrs.
In those cases, if the user forgets to update the `stageTimeoutMs`.
This change allows the job runner to "hint" how long the task will run, if the hint is present the `WaitOnJobCompletionTask` will wait that amount of time + small buffer instead of the default 2hrs


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobRunner.java
```diff
@@ -0,0 +1,37 @@
+package com.netflix.spinnaker.orca.clouddriver.tasks.job;
```
marchello2000, line 1: converted to java to get a `default` interface implementation for `getJobTimeoutMs`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
```diff
@@ -52,13 +53,26 @@ public class WaitOnJobCompletion extends AbstractCloudProviderAwareTask implemen
...
+   * Extra minute to pad the timing supplied by the job provider.
+   * E.g. if TitusJobRunner says this task is limitted to 50minutes we will wait 50m + 5m(padding),
+   * we should wait a bit longer to allow for any inaccuracies of the clock across the systems
+   */
+  static final Long PROVIDER_PADDING_MS = Duration.ofMinutes(5).toMillis()
 
   @Override
   long getDynamicTimeout(@Nonnull StageExecution stage) {
+    Long jobTimeoutFromProvider = (stage.context.get("jobRuntimeLimitMs") as Long)
+
```
jeyrschabu, line 66: Oh the hinted runout takes precedence.

jeyrschabu:
Ignore my earlier question. lgtm

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/RunJobTask.groovy
```diff
@@ -75,9 +75,14 @@ class RunJobTask extends AbstractCloudProviderAwareTask implements RetryableTask
         "notification.type"   : "runjob",
         "kato.result.expected": creator.katoResultExpected,
         "kato.last.task.id"   : taskId,
-        "deploy.account.name" : credentials,
+        "deploy.account.name" : credentials
     ]
 
+    Optional<Long> actualJobTimeout = creator.getJobTimeoutMs(stage)
+    if (actualJobTimeout != null) {
```
cfieber, line 82: unless there is some groovy magic at play here I think you want actualJobTimeout.isPresent() instead of a null check (and creator.getJobTimeoutMs should ideally never return null)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
marchello2000: yeah, i was debating taking the max, but at the end, since this is to reduce confusion, i think job timeout makes sense to be the "main" param since the users know about it but not the stageoverride

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/RunJobTask.groovy
marchello2000: thanks!! im confused how did it work... fixing!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobRunner.java
```diff
@@ -0,0 +1,37 @@
...
+  boolean isKatoResultExpected();
+
+  /** @return The cloud provider type that this object supports. */
+  String getCloudProvider();
+
+  /**
+   * @return the timeout in millis, if any, of the job so we can set the timeout of the task to the
+   *     same
+   */
+  default Optional<Long> getJobTimeoutMs(StageExecution stage) {
```
cfieber, line 32: can you stick a `@Nonnull` on this return value

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/RunJobTask.groovy
jeyrschabu: +1

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobRunner.java
cfieber: also maybe worth making this `Optional<Duration>` and ditching all the suffixes of various time formats. The value if populated in the stage could use the duration syntax (PT1S etc) from Duration.parse

marchello2000: yeah, that's fair. i should've just done that upfront... fixing up!

PR #3627, opened 2020-04-23
fix(bake): Bake respect regions

edgarulg:
Helps to fix: spinnaker/spinnaker#4641

When we specify a bake stage to happen in a specific region, it is looking for the regions to downstairs stages (in this case, deploy stages). At the end, it is taking the region of the bake and the deploy and try to bake on all regions found.

We added a new field that will come from Deck named `autoDetectRegions` for the `BakeStage`.
By default `autoDetectRegions` will be true in order to preserve existing functionality, otherwise it will prevent `BakeStage` to look for other regions in downstair stages (e.g. deploy stages).

Will make a new Deck PR for add the `autoDetectRegions` checkbox.


orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStageSpec.groovy
```diff
@@ -63,28 +63,33 @@ class BakeStageSpec extends Specification {
...
+    [cloudProviderType: "aws", region: "us-west-2", autoDetectRegions: false]                             | deployAz("aws", "cluster", "us-west-1") + deployAz("aws", "cluster", "us-west-2")  || expectedContexts("aws", "19700101011512", "us-west-2")
+    [cloudProviderType: "aws", regions: ["us-east-1", "us-west-1"], autoDetectRegions: false]             | deployAz("aws", "clusters", "eu-central-1")                                        || expectedContexts("aws", "19700101011512", "us-east-1", "us-west-1")
+    [cloudProviderType: "aws", regions: ["us-east-1", "us-west-1"], autoDetectRegions: true]              | deployAz("aws", "cluster", "us-west-1") + deployAz("aws", "cluster", "us-west-2")  || expectedContexts("aws", "19700101011512", "us-east-1", "us-west-1", "us-west-2")
+    [cloudProviderType: "aws", region: "us-west-2", autoDetectRegions: true]                              | deployAz("aws", "cluster", "us-west-2")                                            || expectedContexts("aws", "19700101011512", "us-west-2")
```
german-muzquiz, line 92: Can we add another region to this last case on the second column? right now the output is the same as the input regardless of `autoDetectRegions` variable

edgarulg: The extra region on the deploy context were added, the test completed without issues, returning the expected behavior.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -94,7 +94,8 @@ class BakeStage implements StageDefinitionBuilder {
     Set<String> deployRegions = (stage.context.region ? [stage.context.region] : []) as Set<String>
     deployRegions.addAll(stage.context.regions as Set<String> ?: [])
 
-    if (!deployRegions.contains("global")) {
+    Boolean autoDetectRegions = Boolean.FALSE != stage.context.autoDetectRegions
```
marchello2000, line 97: nit: I think most people's expectation is that values not present will be treated as `false`. So I would prefer this to be 'skipRegionDetection` (or something like that) instead (and then you'd test for `true` instead)

edgarulg:
@marchello2000 Thanks for you input! I already changed the name to `skipRegionDetection` of the input and set it `false` as the default value

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -104,7 +106,7 @@ class BakeStage implements StageDefinitionBuilder {
       stage.context.amiSuffix = clock.instant().atZone(UTC).format("yyyyMMddHHmmss")
     }
     return deployRegions.collect {
-      stage.context - ["regions": stage.context.regions] + ([
+      stage.context - ["regions": stage.context.regions, "skipRegionDetection": skipRegionDetection] + ([
```
marchello2000, line 109: nice, thanks for changing this.

another issue, i just realized - i don't think it's a huge deal, but best to address before it causes weird issues.

```suggestion
      stage.context - ["regions": stage.context.regions, "skipRegionDetection": stage.context.skipRegionDetection] + ([
```

otherwise. if `stage.context.skipRegionDetection = 0` this subtraction won't work

edgarulg:
@marchello2000 I just commit your suggested change, I also added a new test to validate if we got a zero value on skipRegionDetection field. Thanks for the input

link108:

PR #3628, opened 2020-04-24
feat(plugins): Add Orca API tck and modify existing fixture, use as dependency in orca-plugins-test

jonsie:
This is a draft as I'm still working through issues of using it in downstream plugin repositories.  I have plugins and extensions loading, but I'm hitting a class not found exception when pulling in some 3rd party dependencies.  I also plan on using this to actually submit stages to test plugin stages, and I think that'll still require a few changes here.


orca-api-tck/src/main/resources/orca-test-app.yml
```diff
@@ -29,3 +29,7 @@ redis:
 spring:
   application:
     name: orca
+
+spinnaker:
+  extensibility:
+    plugins-root-path: build/plugins
```
jonsie, line 35: I currently have test configuration in the plugin repo as such:

```
test {
  libsDirName = 'plugins'
  dependsOn("assemblePlugin")
}
```
I'll probably move something like that to our Grardle plugin test configuration so developers don't need to do this.

orca-api-tck/src/main/kotlin/com/netflix/spinnaker/orca/api/test/OrcaFixture.kt
```diff
@@ -29,7 +35,22 @@ import org.springframework.test.context.TestPropertySource
 @SpringBootTest(classes = [Main::class])
 @ContextConfiguration(classes = [TestKitConfiguration::class])
 @TestPropertySource(properties = ["spring.config.location=classpath:orca-test-app.yml"])
-abstract class OrcaFixture
+abstract class OrcaFixture {
```
jonsie, line 38: In the downstream plugin repo I am doing:

```
@TestPropertySource(properties = ["spinnaker.extensibility.plugins.netflix.winston.enabled=true"])
class Fixture : OrcaFixture()
```

This works, but I'd like to provide a better experience so plugin developers don't need to think about the config structure (`spinnaker.extensibility.plugins`) and can just provide the ID and we wire things up for them.

orca-plugins-test/src/test/kotlin/com/netflix/spinnaker/orca/plugins/test/OrcaPluginsFixture.kt
```diff
@@ -102,19 +88,3 @@ class OrcaPluginsFixture : PluginsTckFixture, OrcaTestService() {
...
-@TestConfiguration
-internal class PluginTestConfiguration {
-
-  @Bean
-  @Primary
-  fun queue(clock: Clock?, publisher: EventPublisher?): Queue {
-    return InMemoryQueue(
-      clock!!, Duration.ofMinutes(1), emptyList(), false, publisher!!)
-  }
-}
```
jonsie, line 120: It makes sense to me to lift this out of here and use the Orca instance provided by `orca-api-tck`.

PR #3629, opened 2020-04-24
fix(delete): Add index and chunk deletes of old executions

marchello2000:
In my prior change (https://github.com/spinnaker/orca/pull/3569) I introduced a concept of a `deleted_executions` table to keep track of ~1day worth of deletes so that peering has much easier time to peer deletes.
This change causes the deletion of executions to fail sometimes because of a SQL deadlock.

To fix it:
1. Add an index on `deleted_at` - this column is used to delete executions older than `x` time
2. Take the truncation of the `deleted_executions` table out of the big delete transaction


marchello2000:
thank you both!

PR #3630, opened 2020-04-24
feat(queue): support for dynamic activation of the queue shovel

dreynaud:
This makes it possible to have a QueueShovel bean enabled by default, but not active.

Once the `queue.shovel.active` flag becomes true, it will start actively shoveling messages for some time then stop and check again.

Example config to support this:

```
queue:
  shovel:
    enabled: true
    kind: sql-to-sql
    source-pool-name: previous
```


dreynaud:
This is a follow up to https://github.com/spinnaker/keiko/pull/113

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
```diff
@@ -34,15 +38,34 @@ class QueueShovel(
...
 ) {
   private val log = LoggerFactory.getLogger(javaClass)
 
   private val pollOpsRateId = registry.createId("orca.nu.shovel.pollOpsRate")
   private val shoveledMessageId = registry.createId("orca.nu.shovel.pushedMessageRate")
   private val shovelErrorId = registry.createId("orca.nu.shovel.pushedMessageErrorRate")
 
-  @Scheduled(fixedDelayString = "\${queue.shovel.poll-frequency.ms:50}")
+  @Scheduled(fixedDelay = 5_000)
+  fun pollStatus() {
```
marchello2000, line 51: `pollStatus` isn't really what this does... no? it's more like `migrate`

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
```diff
@@ -34,15 +38,34 @@ class QueueShovel(
...
+    if (isActive()) {
+      thread(name = "queue-shovel") {
```
marchello2000, line 53: why does this go on a separate thread? what if that thread takes longer than 5s, won't spring just kick off the next one? is that ok?
seems that `migrateSome` can just go inside this method and it would be simpler

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
```diff
@@ -34,15 +38,34 @@ class QueueShovel(
...
+        migrateSome(Duration.ofMillis(4_900))
```
marchello2000, line 54: also, because of this, this number is (4900 should always be stricktly less than the fixedDelay)... maybe a const?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
dreynaud: yeah I'll rename, that's because at the time my mind was really set on how to check if we were active or not

dreynaud: looks better in 5c36aa71d9c9b5b1328bf69f06487c587436420e?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
```diff
@@ -34,32 +40,48 @@ class QueueShovel(
...
-          registry.counter(shovelErrorId).increment()
-        }
+    registry.counter(pollOpsRateId).increment()
+    previousQueue.poll { message, ack ->
+      try {
+        queue.push(message)
+        ack.invoke()
+        registry.counter(shoveledMessageId).increment()
+      } catch (e: Throwable) {
+        log.error("Failed shoveling message from previous queue to active (message: {})", message, e)
```
marchello2000, line 78: nit:
```suggestion
        log.error("Failed shoveling message from previous queue to active (message: $message)", message, e)
```

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
```diff
@@ -34,32 +40,48 @@ class QueueShovel(
...
+      return
+    }
+
+    log.info("Actively shoveling from $previousQueue to $queue")
+    val workDuration = Duration.ofMillis(workDurationMs)
+    val start = Instant.now()
+
+    while (Duration.between(start, Instant.now()) < workDuration) {
+      migrateOne()
+      Thread.sleep(50)
```
marchello2000, line 64: do we still want to "respect" `queue.shovel.poll-frequency.ms`

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
```diff
@@ -34,32 +40,48 @@ class QueueShovel(
...
   private val shovelErrorId = registry.createId("orca.nu.shovel.pushedMessageErrorRate")
 
-  @Scheduled(fixedDelayString = "\${queue.shovel.poll-frequency.ms:50}")
+  @Scheduled(fixedRate = rateMs)
+  fun migrateIfActive() {
+    if (!isActive()) {
+      return
+    }
+
+    log.info("Actively shoveling from $previousQueue to $queue")
```
marchello2000, line 58: what does the `toString` here print for the `queue` objects, is it useful?

dreynaud: at the moment it's not terribly useful (`shoveling from SqlQueue@X to SqlQueue@Y`), but it has the potential to be a bit more useful. I was thinking we could have SqlQueue print the pool name.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
dreynaud: it didn't seem terribly useful to me and was just "snuck" in there via `@Scheduled`. It would make the bean wiring more complicated for little return IMO

marchello2000: yeah, that makes sense

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
dreynaud: now useful with keiko 3.6.1 :)

PR #3632, opened 2020-04-27
chore(mptv1): Unpin jinjava

marchello2000:
Validated that the version of jinjava specified in `kork` (2.5.2) doesn't break MPTv1
Hopefully, the MPTv1 jinjava renderer tests will catch any future jinjava issues


marchello2000:
Going to let it simmer in our prod env for a bit longer but so far all looks good. Will merge by EOD if all looks good on our side

PR #3636, opened 2020-04-27
chore(java11): Compile with Java 11 (but targeting Java 8)

plumpy:
* chore(java11): Compile with Java 11 (but targeting Java 8)

release.yml looks scary in the GitHub diff view, but it's mostly indentation changes, and one other unrelated changes (removing two redundant If conditions) that already exist in every other repository. It's just copied from clouddriver.

* fix(java11): don't compare times with nanoseconds to those with millis

* fix(java11): Define getType for all anonymous StageDefinitionBuilders

Under Java 11, Class#getSimpleName() returns the empty string for an anonymous object literal. I verified that these are the only places in the entire codebase where we have an anonymous definition of StageDefinitionBuilder that doesn't override getType().

### DIFF
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/graph/StageDefinitionBuilder.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/graph/StageDefinitionBuilder.java
@@ -16,11 +16,9 @@
 
 package com.netflix.spinnaker.orca.api.pipeline.graph;
 
-import static com.google.common.base.Preconditions.checkState;
 import static com.netflix.spinnaker.orca.api.pipeline.graph.TaskNode.Builder;
 import static com.netflix.spinnaker.orca.api.pipeline.graph.TaskNode.GraphType.FULL;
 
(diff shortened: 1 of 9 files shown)


orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/graph/StageDefinitionBuilder.java
```diff
@@ -108,6 +110,9 @@ default void prepareStageForRestart(@Nonnull StageExecution stage) {}
    */
   static String getType(Class<? extends StageDefinitionBuilder> clazz) {
     String className = clazz.getSimpleName();
+    checkState(
```
cfieber, line 113: I'd prefer to see this without the guava dependency. I'm sure its already somehow transitively in orca-api but this is a dependency we strongly want to avoid in API packages so lets not introduce it in this class to save 2 lines of code

plumpy: Good call, I removed it.

PR #3638, opened 2020-04-28
fix(health): OutOfService is a Down health state

dreynaud:
When we're waiting for up instances, OutOfService was considered neither Up nor Down, which means that as long as some other health was Up, having one OutOfService would not keep the task waiting.
    
With this change, we consider OutOfService a Down status, which means that someAreUpAndNoneAreDown will return false, which will keep the task waiting.


robzienert:
Edited the description so it would line-wrap. Hope you don't mind. :)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/model/HealthState.java
```diff
@@ -0,0 +1,46 @@
...
+public enum HealthState {
+  Failed,
+  Down,
+  OutOfService,
+  Unknown,
+  Starting,
+  Succeeded,
+  Up;
+
+  // HealthState.valueOf does:
```
marchello2000, line 29: make this into a javadoc `/**` ?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/HealthHelper.groovy
```diff
@@ -96,22 +98,19 @@ class HealthHelper {
     }
 
     // no health indicators is indicative of being down
-    boolean someAreDown = !healths || healths.any { it.state == 'Down' || it.state == 'OutOfService' || it.state == 'Starting' }
-    boolean noneAreUp = !healths.any { it.state == 'Up' }
+    boolean someAreDown = !healths || healths.any { HealthState.fromString(it.state) in [Down, OutOfService, Starting] }
```
marchello2000, line 101: nit: these `.any` operations are now much more expensive and create more GC gen 0 objects.
Not sure how often this gets called. But might be worth making a new `healths` list that is just 
```java
healths.map { HealthState.fromString(it.state) }
```

and then check against that

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/HealthHelper.groovy
```diff
@@ -96,22 +98,19 @@ class HealthHelper {
...
   static boolean someAreUpAndNoneAreDownOrStarting(Map instance, Collection<String> interestingHealthProviderNames) {
     List<Map> healths = filterHealths(instance, interestingHealthProviderNames)
-    boolean someAreUp = healths.any { Map health -> health.state == 'Up' }
-    someAreUp = areSomeUpConsideringPlatformHealth(healths, interestingHealthProviderNames, someAreUp)
 
-    boolean noneAreDownOrStarting = !healths.any {
-      Map health -> health.state == 'Down' || health.state == 'Starting'
-    }
+    boolean someAreUp = healths.any { HealthState.fromString(it.state) == Up }
+    boolean noneAreDown = !healths.any { HealthState.fromString(it.state) in [Down, OutOfService, Starting] }
```
cfieber, line 111: worth making upness/downness an attribute of the enum?  (and I think Failed is maybe a down but I dunno..)

dreynaud: I was thinking about that, it probably makes sense. `Failed` is referenced in the k8s, dcos and azure providers, it seems like it should be a down status indeed.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/HealthHelper.groovy
dreynaud: I considered it, keep in mind that most instances will only have 2 or 3 healths so I don't think it would change much

dreynaud: ideally we would be working with a model from the get go as opposed to parsing strings from maps in here 😢 

I'll make a pass at a nicer cleanup later

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/model/HealthState.java
dreynaud: does it matter? In what way is a javadoc better?

marchello2000: I feel like in this case you're describing what this method does/returns and Javadoc is exactly for that (and will show up in IDE)

PR #3641, opened 2020-04-30
fix(gremlin): Fix gremlin config

ezimanyi:
The feature flag for gremlin was removed from deck (along with a number of other feature flags) in 1.20 but we missed that this flag was also checked in orca. This change removes checking the flag in orca (and removes setting the flag in orca.yml).

As orca.yml already sets the base url, we don't need to default the Value using it (though in the future maybe it's better to push the defaulting to there rather than have it in the default config).


ezimanyi:
Maybe since this is more than just a flag gating a stage, it should stay behind a flag?  Not sure, could go either way.  If we think it should stay behind the flag we can not merge this and leave the flag in Halyard.

ezimanyi:

PR #3643, opened 2020-04-30
fix(build): 2 minute Gradle HTTP client socket and connection timeout

jonsie:
Release build has been failing with connection timeouts to bintray, let's see if this resolves that.


jonsie:
@cfieber Yea that's kinda what I was thinking for next steps.  👍

PR #3647, opened 2020-05-01
fix(peering): Log error metrics more better

marchello2000:
Not all exceptions were caught and thus would leak out of the `PeeringAgent` and weren't logged.
Additionally, I threw a retry on the most egregious SQL query in case it fails due to timeout.


orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/MySqlRawAccess.kt
```diff
@@ -3,6 +3,10 @@ package com.netflix.spinnaker.orca.peering
 import com.netflix.spinnaker.kork.exceptions.SystemException
 import com.netflix.spinnaker.kork.sql.routing.withPool
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType
+import io.github.resilience4j.retry.Retry
+import io.github.resilience4j.retry.RetryConfig
+import io.vavr.control.Try
```
dreynaud, line 8: ha, TIL

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -156,43 +160,49 @@ class PeeringAgent(
...
+        log.error("Failed to delete some orchestrations", e)
+        peeringMetrics.incrementNumErrors(ExecutionType.ORCHESTRATION)
+        hadFailures = true
+      }
 
-    log.debug("Found ${deletedExecutionIds.size} (orchestrations: ${orchestrationIdsToDelete.size} pipelines: ${pipelineIdsToDelete.size} deleted candidates after cursor: $deletedExecutionCursor")
-    var hadFailures = false
-    var orchestrationsDeleted = 0
-    var pipelinesDeleted = 0
+      try {
```
dreynaud, line 182: I just noticed the code duplication here. I would suggest the following:

  data class DeletionResult(val numDeleted: Int, val hadFailures: Boolean)

  private fun delete(executionType: ExecutionType, idsToDelete: List<String>): DeletionResult {
    var numDeleted = 0
    var hadFailures = false
    try {
      numDeleted = destDB.deleteExecutions(executionType, idsToDelete)
      peeringMetrics.incrementNumDeleted(executionType, numDeleted)
    } catch (e: Exception) {
      log.error("Failed to delete some $executionType", e)
      peeringMetrics.incrementNumErrors(executionType)
      hadFailures = true
    }

    return DeletionResult(numDeleted, hadFailures)
  }

And then in `peerDeletedExecutions`, it becomes simply:

    val (orchestrationsDeleted, orchestrationsHadFailures) = delete(ExecutionType.ORCHESTRATION, orchestrationIdsToDelete)
    val (pipelinesDeleted, pipelinesHadFailures) = delete(ExecutionType.PIPELINE, pipelineIdsToDelete)

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -76,7 +75,7 @@ class PeeringAgent(
   override fun tick() {
     if (dynamicConfigService.isEnabled("pollers.peering", true) &&
       dynamicConfigService.isEnabled("pollers.peering.$peeredId", true)) {
-      peeringMetrics.recordOverallLag() {
+      peeringMetrics.recordOverallLag {
```
dreynaud, line 78: is it the same with and without the parens? I'm confused

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -156,43 +160,49 @@ class PeeringAgent(
...
+        // deleted here so it's OK for the actual delete counts to not match the "requested" count
+        log.debug("Deleted orchestrations: $orchestrationsDeleted (of ${orchestrationIdsToDelete.size} requested), pipelines: $pipelinesDeleted (of ${pipelineIdsToDelete.size} requested), new cursor: $deletedExecutionCursor")
+      } else {
+        log.error("Failed to delete some executions, not updating the cursor location to retry next time")
+      }
     } catch (e: Exception) {
-      log.error("Failed to delete some pipelines", e)
+      log.error("Failed to delete some executions", e)
+      peeringMetrics.incrementNumErrors(ExecutionType.ORCHESTRATION)
       peeringMetrics.incrementNumErrors(ExecutionType.PIPELINE)
```
dreynaud, line 205: ah hmmm... it wouldn't be terribad to pass an execution type to `peerDeletedExecutions` and call it twice, right? Then the metrics wouldn't have to lie 😅

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
marchello2000: thanks, i like it!

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
marchello2000: yes, i guess spotless did it.. but yeah, i guess that's kotlin convention: https://kotlinlang.org/docs/reference/lambdas.html#passing-a-lambda-to-the-last-parameter

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
marchello2000: it's a bit sucky since there are both pipelines and orchestrations in that `deleted_executions` table...

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/MySqlRawAccess.kt
```diff
@@ -204,4 +210,16 @@ open class MySqlRawAccess(
 
     return persisted
   }
+
+  private fun <T> withRetry(action: () -> T): T {
```
cfieber, line 214: you can also do this with a retry annotation:
https://github.com/spinnaker/fiat/blob/64021e98f8d55c11a83149dc8aacdb854342c777/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/Front50DataLoader.java#L39

And then make the options configurable/overridable via spring:
https://github.com/spinnaker/fiat/blob/64021e98f8d55c11a83149dc8aacdb854342c777/fiat-roles/src/main/resources/resilience4j-defaults.properties

cfieber: now that I say that, the annotation approach would change this from returning a function to executing the retry so never mind

marchello2000: there was also some weird thing that @jonsie and I tried to figure out (the annotation wasn't working somewhere... river maybe?) anyway, it was driving me mad so I didn't want to chance it here :)

PR #3648, opened 2020-05-02
feat(build): adds project properties for composite builds.

cfieber:
This adds flags for the related projects that we bring in
as dependencies, to enable referencing the related project
as a composite build via gradle.

This method of enabling composite builds seems to work
significantly better with IntelliJ. (As opposed to
importing multiple projects into the IDE and configuring
the composite relationship after the fact).

This holds the opinion that all the projects are checked
out in the same parent directory, and can be found at
`'../<projectName>'` relative to this project.

To enable a composite reference to a project, set the
project property `'<projectName>Composite=true'`.

This can be done either as
* a command line flag, e.g. `-PkorkComposite=true`
* a project property via gradle.properties
* a global project property via ~/.gradle/gradle.properties

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -14,6 +14,14 @@
  * limitations under the License.
  */
 
+['fiat', 'keiko', 'kork'].each { prj ->
+  String propName = "${prj}Composite"
+  String projectPath = "../$prj"
+  if (settings.ext.has(propName) && Boolean.parseBoolean(settings.ext.get(propName) as String)) {
+    includeBuild projectPath
+  }
+}
+
 include "orca-api",
   "orca-api-tck",
   "orca-applications",
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,3 +5,19 @@ korkVersion=7.41.1
 kotlinVersion=1.3.71
 org.gradle.parallel=true
 spinnakerGradleVersion=8.0.6
+
+# To enable a composite reference to a project, set the
+#  project property `'<projectName>Composite=true'`.
+#
+# This can be done either as
+#  * a command line flag, e.g. `-PkorkComposite=true`
+#  * a project property via gradle.properties
+#  * a global project property via ~/.gradle/gradle.properties
+#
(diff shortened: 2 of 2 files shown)


cfieber:
my intention is to push this change everywhere, since it nicely improves the local dev experience when trying to iterate on something in a library and validate or test out those changes in a service (as opposed to doing the `pTML` sad dance)

settings.gradle
```diff
@@ -14,6 +14,14 @@
  * limitations under the License.
  */
 
+['fiat', 'keiko', 'kork'].each { prj ->
```
marchello2000, line 17: ❤️  maybe worth adding a note here (similar to the one you had in the description of the PR) or `gradle.properties` (or readme) prob gradle.props is the most likely to get noticed 

> a command line flag, e.g. -PkorkComposite=true
> a project property via gradle.properties
> a global project property via ~/.gradle/gradle.properties

cfieber: good idea, will do!

plumpy: I have a feeling bumpdeps won't preserve comments in`gradle.properties`, but I don't know for sure...

cfieber: ahh yeah you are correct..

I will move this into the README for now

we could always solve bumpdeps with a regex instead of using java.util.Properties, what do you think about that?

plumpy: Yeah, seems reasonable.

plumpy:
Is there a way to get this into a file that is in `.gitingore`? So we could just always have it enabled locally? I dunno maybe that's a bad idea...

cfieber:
if you want it always enabled locally you can set it in ~/.gradle/gradle.properties (though that will affect all your local projects)

plumpy:
Oh yeah, that makes sense, thanks.

settings.gradle
cfieber: https://github.com/spinnaker/bumpdeps/pull/6/files

PR #3649, opened 2020-05-04
feat(plugins): Use SpinnakerExtensionPoint for orca-api extension points, skipping Task for now

jonsie:
Depends on: https://github.com/spinnaker/kork/pull/632

This implementation uses `SpinnakerExtensionPoint` which is added in the above PR.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 enablePublishing=false
 fiatVersion=1.18.3
-keikoVersion=3.5.0
-korkVersion=7.41.2
+keikoVersion=3.6.1
+korkVersion=7.41.3
 kotlinVersion=1.3.71
 org.gradle.parallel=true
 spinnakerGradleVersion=8.0.6
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 enablePublishing=false
 fiatVersion=1.18.3
 keikoVersion=3.5.0
-korkVersion=999.0.0-local-SNAPSHOT
+korkVersion=7.41.2
 kotlinVersion=1.3.71
 org.gradle.parallel=true
 spinnakerGradleVersion=8.0.6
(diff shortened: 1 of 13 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/TaskResolver.java
```diff
@@ -113,7 +113,8 @@ public Task getTask(@Nonnull Class<? extends Task> taskType) {
   @Nonnull
   public Class<? extends Task> getTaskClass(@Nonnull String taskTypeIdentifier) {
     try {
-      return getTask(taskTypeIdentifier).getClass();
+      Task task = getTask(taskTypeIdentifier);
+      return (Class<? extends Task>) task.getExtensionClass();
```
jonsie, line 117: When the extension is proxied, `TaskExecution.type` class was being serialized as a `Proxy` object.  This fixes that, and allows `TaskTypeDeserializer` to deserialize the proxied object to the correct Task type.

jonsie:
When spinnaker/kork#632 merges and we cut a kork release, this test here will pass:

https://github.com/spinnaker/orca/blob/master/orca-plugins-test/src/test/kotlin/com/netflix/spinnaker/orca/plugins/test/OrcaPluginsTest.kt#L74

gradle.properties
```diff
@@ -1,7 +1,7 @@
 enablePublishing=false
 fiatVersion=1.18.3
 keikoVersion=3.5.0
-korkVersion=7.41.1
+korkVersion=999.0.0-local-SNAPSHOT
```
srekapalli, line 4: korkVersion seems odd .....?

cfieber: this was built on spinnaker/kork#632

jonsie:
Backed out the `TaskResolver` and `Task` changes here due to extensive unit test changes that are required.  I'll put up a separate PR for that work, this will at least un-block the Kork bump.

PR #3650, opened 2020-05-04
feat(bake): Expose amiName in bake stage output (if present)

jervi:


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -119,6 +119,7 @@ class BakeStage implements StageDefinitionBuilder {
   static class CompleteParallelBakeTask implements Task {
     public static final List<String> DEPLOYMENT_DETAILS_CONTEXT_FIELDS = [
       "ami",
+      "amiName",
```
marchello2000, line 122: isn't `amiName` same as `imageName`?

jervi: Sadly, no. amiName is a field that is settable from Deck. I haven’t looked at the specifics, but the gist is that ami’s are named after the first package you specify in the `packages` field. Sometimes that package name isn’t very descriptive, so you can override the name of the resulting ami using amiName. And we have a use case where it would be very convenient to be able to access this field using SPeL from a parent pipeline, but it disappeared from the context when using the pipeline stage.

PR #3651, opened 2020-05-04
feat(core): Merge collections when merging stage outputs

jervi:
Currently, when merging stage outputs, if equal keys are found, one of them will overwrite the other one. This can be a problem if you, for instance, run a baking pipeline that produces several different bakes, and you start this baking pipeline as a pipeline stage from a parent pipeline (i.e. not as a trigger). Each bake stage will produce a `deploymentDetails: [ bakeContext... ]` and put this into the stage outputs. When the pipeline stage completes, only the last of these `deploymentDetails` will be present in the execution context of the parent pipeline.

This PR fixes this by checking if the duplicated key contains collections, and if so, it will concatenate them instead of overwriting. For everything else the behaviour will be like before.


orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/model/PipelineExecutionImplSpec.groovy
```diff
@@ -76,6 +80,10 @@ class PipelineExecutionImplSpec extends Specification {
         refId = "2"
         outputs["baz"] = "foo-2"
         outputs["qux"] = "qux-2"
+        outputs["myList"] = [[
```
marchello2000, line 83: worth adding a duplicate to see `distinct` working

marchello2000:
@jervi , thank you! You are brave for modifying these fundamental parts of orca. I think it should be ok, but I want to think about it some more, will also throw a few folks on the review to get their opinion

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/PipelineExecutionImpl.java
```diff
@@ -150,12 +162,24 @@ public void setKeepWaitingPipelines(boolean keepWaitingPipelines) {
...
         .map(StageExecution::getOutputs)
         .map(Map::entrySet)
         .flatMap(Collection::stream)
-        .collect(toMap(Map.Entry::getKey, Map.Entry::getValue, (o, o2) -> o2));
+        .collect(
+            toMap(
+                Map.Entry::getKey,
+                Map.Entry::getValue,
+                (o, o2) -> {
+                  if (o instanceof List && o2 instanceof List) {
```
robzienert, line 176: General question, should we instead be looking for `Collection`? Are we confident that `Set` is never used within a context?

jervi: Fair point, I'm gonna change it to Collection

orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/model/PipelineExecutionImplSpec.groovy
jervi: Done 👍

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/PipelineExecutionImpl.java
```diff
@@ -150,12 +162,24 @@ public void setKeepWaitingPipelines(boolean keepWaitingPipelines) {
...
-        .collect(toMap(Map.Entry::getKey, Map.Entry::getValue, (o, o2) -> o2));
+        .collect(
+            toMap(
+                Map.Entry::getKey,
+                Map.Entry::getValue,
+                (o, o2) -> {
+                  if (o instanceof Collection && o2 instanceof Collection) {
+                    return Stream.concat(((Collection) o).stream(), ((Collection) o2).stream())
+                        .distinct()
+                        .collect(Collectors.toList());
```
ezimanyi, line 179: Are we worried about this changing the type of collections in the context?  In particular, if we had a `Set` in the context this would now change it to a `List`.  This wouldn't matter (except maybe for performance) if everywhere we fetch it from the context we get it as a `Collection` but it might if there were places that we got it out expecting a `Set`.  Obviously I don't know that we want to over-complicate the solution and special case all collections, but just worth considering whether this could cause subtle (or maybe not that subtle if it's a cast error) bugs.

jervi: I did consider the same thing, and actually started to implement it, but then threw it away because i didn’t want to complicate it too much, and I suspect this only ends up as Jain anyway, erasing any types anyway (but I’m not 100% sure).
I’m happy to reimplement it if you think it is a good idea.

ezimanyi: I think this should be fine the way you've written it, assuming nobody else can think of an edge case this would break.

robzienert:
@marchello2000 You expressed you wanted to think about this more. Are you satisfied with the change?

jervi:
Btw: We have been running this PR in production for about a week. Only related issue we had was a team  that used a SPeL expression to select an AMI from a list using `[0]`, and after merging this the order of the list had changed, so they had to edit their pipelines to use a proper selector instead of assuming a specific order.

marchello2000:
@robzienert I couldn't think of anything that would be broken... will add a merge label
@jervi thanks for the heads up!

PR #3652, opened 2020-05-04
fix(bake): Bake stage output may contain output from unrelated stages

jervi:
As @luispollo notes in the code I removed, the current implementation may include output from completely unrelated bakes that are running at the same time as the bake stage in question. I fail to see any reason this can be correct.

After applying this fix, each bake stage will contain the bakes created in that bake stage, and nothing else.

Coupled with [this PR](https://github.com/spinnaker/orca/pull/3651), this fix will make the output of pipelines with multiple bake stages correct, as all the `deploymentDetails` will be merged instead of overwritten, and thus contain all images created, and not just an arbitrary list which is timing dependent.


luispollo:
Hi @jervi. I'm also willing to give this a shot, and suspect this _will_ break a handful of pipelines at Netflix, but hopefully we can try to redesign those pipelines instead of keeping this skeleton in the closet. 😃

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -141,15 +141,8 @@ class BakeStage implements StageDefinitionBuilder {
...
-      def bakeInitializationStages = stage.execution.stages.findAll {
-        it.parentStageId == stage.parentStageId && it.status == ExecutionStatus.RUNNING
-      }
-
-      // FIXME? (lpollo): I don't know why we do this, but we include in relatedStages all parallel bake stages in the
-      //  same pipeline, not just the ones related to the stage passed to this task, which means this list actually
-      //  contains *unrelated* bake stages...
       def relatedBakeStages = stage.execution.stages.findAll {
-        it.type == PIPELINE_CONFIG_TYPE && bakeInitializationStages*.id.contains(it.parentStageId)
+        it.type == PIPELINE_CONFIG_TYPE && stage.id == it.parentStageId
```
luispollo, line 145: Do we want to add `it.status == ExecutionStatus.RUNNING`  to the filter, since it was present before?

cfieber: I suspect the status==RUNNING check *might* have been a dirty workaround to sort of limit the related bake stages that were found

With the check for parent stage Id I think the current code is correct

jervi: Yes, as far as I was able to figure out, the RUNNING check was part of the problem. For instance, if you had multiple parallel bake stages, and one or more of those bakes was already baked (causing Rosco to reuse the existing bake), only those reused bakes would be present in the `outputs`. The proper bakes would be stripped away.

PR #3655, opened 2020-05-04

spinnakerbot:


plumpy:
@jonsie this is broken because of spinnaker/kork#632

> there should be no reason to make this public I think

orca disagrees with you :grin:

jonsie:
@plumpy It's actually slightly different - this specific class that it's complaining about IS deleted.  The other one moved to a different package and is only internally necessary.  I can fix this.

jonsie:
This is fixed here: https://github.com/spinnaker/orca/pull/3649

Just need that to build and merge.

PR #3667, opened 2020-05-07
feat(shovel): override partition of foreign executions

dreynaud:
This should allow the shovel to be used in failover scenarios, between orcas that have a different partition.


dreynaud:
(I will fix the tests and add some more in a bit, wanted to get this out for feedback early)

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
```diff
@@ -81,7 +90,33 @@ class QueueShovel(
...
+      return
+    }
+
+    if (message !is ExecutionLevel) {
+      log.warn("Message $message does not implement ExecutionLevel, can not inspect partition")
+      return
+    }
+
+    // don't catch exceptions on retrieve/store (e.g. ExecutionNotFoundException), so that we can short-circuit shoveling
+    // of this message
```
robzienert, line 104: If the execution isn't found, we'll log an error and then do nothing with the message: Won't we just enter into a loop of logging out an error for N messages, but never actually doing anything with them?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
```diff
@@ -81,7 +90,33 @@ class QueueShovel(
...
+    if (executionRepository == null) {
+      return
+    }
+
+    if (message !is ExecutionLevel) {
+      log.warn("Message $message does not implement ExecutionLevel, can not inspect partition")
+      return
+    }
+
+    // don't catch exceptions on retrieve/store (e.g. ExecutionNotFoundException), so that we can short-circuit shoveling
```
cfieber, line 103: is it worth catching for expected cases so we don't end up logging scary looking error logs?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
dreynaud: Yes, another way to think about it is that we'll keep retrying shoveling that message. It may eventually succeed, for instance if the execution was missing because of peering lag

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
dreynaud: I see, so for instance we could avoid logging the stack trace for something like an `ExecutionNotFound` 👍

orca-web/orca-web.gradle
```diff
@@ -90,7 +90,6 @@ dependencies {
   testImplementation("cglib:cglib-nodep")
   testImplementation("org.objenesis:objenesis")
   testImplementation("org.junit.jupiter:junit-jupiter-api")
-  testImplementation("org.hamcrest:hamcrest-core:1.3")
```
dreynaud, line 93: I snuck this change in here, it is not related to the rest of the PR... but I also don't think we use hamcrest anywhere in orca

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
```diff
@@ -71,17 +75,54 @@ class QueueShovel(
...
+
+        // transfer the ownership _before_ pushing the message on the queue
+        // we don't want a task handler running that message if the execution is not local
+        transferOwnership(message)
+
         queue.push(message)
         ack.invoke()
         registry.counter(shoveledMessageId).increment()
+      } catch (e: ExecutionNotFoundException) {
+        // no need to log the stack trace on ExecutionNotFoundException, which can be somewhat expected
```
marchello2000, line 88: why because it might not have been peered yet?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
marchello2000: but won't the message have been taken and ack'd already so, in essence, we would just dump it on the floor in the peering lag case?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
```diff
@@ -71,17 +75,54 @@ class QueueShovel(
...
+      executionRepository.store(execution)
+    }
+  }
+
   @PostConstruct
-  fun confirmShovelUsage() =
+  fun confirmShovelUsage() {
     log.info("${javaClass.simpleName} migrator from $previousQueue to $queue is enabled")
+    if (executionRepository == null) {
+      log.warn("${javaClass.simpleName} configured without an ExecutionRepository, won't be able to check execution partitions")
```
marchello2000, line 125:
```suggestion
      log.warn("${javaClass.simpleName} configured without an ExecutionRepository, won't be able to transfer ownership")
```

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
dreynaud: yes, that's what I was thinking

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/QueueShovel.kt
dreynaud: @marchello2000 not sure what you mean, I don't see how we dump anything on the floor? There is a test that makes sure we don't ack messages if there is a problem reading/updating their execution

PR #3675, opened 2020-05-08
fix(Authorization): allow users without `WRITE` permission to execute SavePipelineStage

AbdulRahmanAlHamali:
The [SavePipelineStage](https://github.com/spinnaker/orca/blob/master/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/SavePipelineStage.java) is a pipeline stage that gets executed as part of a pipeline. This should mean that users with `EXECUTE` privilege should be able to execute it. However, they cannot, because the front50 save pipeline operation requires `WRITE` permission to the application.

This PR allows users to provide `asAnonymous` parameter, which will end up, executing the process anonymously, solving the problem.

If this is accepted, I would like to cherry-pick it into 1.20 as well please

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
@@ -93,14 +93,20 @@ public class SavePipelineTask implements RetryableTask {
         .forEach(m -> m.mutate(pipeline));
 
     Response response;
-    if ((Boolean) Optional.ofNullable(stage.getContext().get("asAnonymous")).orElse(false)) {
-      User unrestrictedUser = new User();
-      unrestrictedUser.setUsername("anonymous");
-      response =
-          AuthenticatedRequest.propagate(
-              () -> front50Service.savePipeline(pipeline), unrestrictedUser);
-    } else {
-      response = front50Service.savePipeline(pipeline);
+    try {
+      if ((Boolean) Optional.ofNullable(stage.getContext().get("asAnonymous")).orElse(false)) {
+        User unrestrictedUser = new User();
+        unrestrictedUser.setUsername("anonymous");
+        response =
+            AuthenticatedRequest.propagate(
+                    () -> front50Service.savePipeline(pipeline), unrestrictedUser)
+                .call();
+      } else {
+        response = front50Service.savePipeline(pipeline);
+      }
+    } catch (Exception e) {
+      log.error("Unable to save pipeline: " + e.getMessage());
+      return TaskResult.builder(ExecutionStatus.TERMINAL).context(new HashMap<>()).build();
     }
 
     Map<String, Object> outputs = new HashMap<>();
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
@@ -23,6 +23,7 @@ import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
 import com.netflix.spinnaker.orca.front50.Front50Service;
 import com.netflix.spinnaker.orca.front50.PipelineModelMutator;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
+import com.netflix.spinnaker.security.User;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
```diff
@@ -90,7 +91,12 @@ public TaskResult execute(StageExecution stage) {
         .filter(m -> m.supports(pipeline))
         .forEach(m -> m.mutate(pipeline));
 
-    Response response = front50Service.savePipeline(pipeline);
+    Response response;
+    if ((Boolean) Optional.ofNullable(stage.getContext().get("asAnonymous")).orElse(false)) {
+      response = AuthenticatedRequest.allowAnonymous(() -> front50Service.savePipeline(pipeline));
```
AbdulRahmanAlHamali, line 96: Does `allowAnonymous` mean that anonymous will over-rule even if a user exists?

marchello2000: hm... I don't think this will work/do anything. (allowAnonymous is confusing)
`allowAnonymous` is purely for logging purposes (e.g. if there is no user in the `RequestContext` and the call isn't made from within an `allowAnonymous` context a warning will be produced with a stack trace as it's likely going to cause issues down the road.

It's a bit weird if a case when you have a savepipeline stage in the pipeline that you can execute but can't edit the application... If you truly want, in that situation, for the user to edit the pipeline, i believe the correct fix is to update the permission check in front50:
https://github.com/spinnaker/front50/blob/c721dfa7046ee22d50816b5084fa1ab9a885b66b/front50-web/src/main/groovy/com/netflix/spinnaker/front50/controllers/PipelineController.groovy#L114

AbdulRahmanAlHamali: Hi Mark, I hope you're doing well. Thanks for the review!

For the usecase, it's a bit complex, but to simplify, the admin who has `WRITE` permission has created the pipeline for the user. The pipeline has `SavePipelineStage` in it, which will update another pipeline right before triggering it, to confirm that it always has the right configuration. The user running the pipeline only has `EXECUTE` permission, but the one who put everything in place was actually the admin

Regarding `allowAnonymous`, I am adding this here because from what I've seen in `fiat`, they have a concept of `unrestricted_user`, and it seems to be linked to the `anonymous` user:
https://github.com/spinnaker/fiat/blob/d91e70e8f66caf50484c5e5e931c0b2a9aa5bdf4/fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/ControllerSupport.java#L25

Do you know how I can use that `unrestricted_user`?

AbdulRahmanAlHamali: Perhaps I can add a wrapper function similar to `allowAnoynmous`, called `asAnonymous`, which will rip out the user headers, and add the anonymous header (to avoid the warning). What do you think?

marchello2000: Got it, I am not too familiar with fiat, does it mean that if the user is `anonymous` fiat won't check permissions (I seem to recall that being the case, but just triple confirming)
`allowAnonymous` was a bit of a misnomer that I meant to rename/think of a better name but didn't get to. I think adding another function as you describe makes sense, but call it something like `asUnrestrictedUser` or `asSystemUser` to make it more clear.

Also cc: @cfieber as he might have some additional comments about allowing the user who runs the pipeline but can't save it to actually save it. It's an edge case but seems like it could be exploited somehow...

ajordens:
This strikes me as an edge case, at this point I'm not entirely comfortable with a change like this landing.

Would a more out-of-the-box solution to have another pipeline trigger on completion that runs with a service account granting sufficient permissions to call `savePipeline` or whatever else needs to be done?

AbdulRahmanAlHamali:
@marchello2000 @cfieber After looking around, I found out that I should use `propagate`, which allows me to use a different `principal` for the specific request. I can set the principal to the "anonymous" user, which has unrestricted privileges.

I'm still not sure if this overrides whatever user already exists in the `SecurityContext`, but if I understand correctly, the only place where there is a user in the `SecurityContext` is gate. Is that right?

One more question, does the PR build some image that I can test in my `dev` setup? simulating a full scenario on my local machine is gruesome, and I would like to give this a proper test to know if it works. It would be great if an image is built with the PR number as a tag for example

AbdulRahmanAlHamali:
@ajordens I wanted to land this PR because it would enable me to solve a problem I'm having very cleanly, but since this solution is not preferred, I'll go with something else on my end.

I'll close this PR then, thank you all for the discussion!

PR #3679, opened 2020-05-12
chore(titus): Break titus provider code out into new module

robzienert:
Providing great value by moving code from one spot in the repo to a new spot. 🥇 

I'm looking into the level of effort that it would take to remove all cloud provider specific logic out of Orca and move it into Clouddriver, or in some cases, deprecating cloud providers. As a means of organizing my thoughts and coming up with a plan, I'm creating modules that take the custom code cloud providers have jammed into `orca-clouddriver` and putting them in their own code corner.

This is incomplete. My next task is to look through all those nasty conditionals of `if (cloudProvider == "titus")` and finding new interfaces that may be introduced.


robzienert:
Yeah, I agree. I'll probably close this and re-open without the refactors. I just feel dirty adding Groovy to new modules. Two-phase approach is bester, though.

robzienert:
Re-opening with minimum required refactors.

PR #3681, opened 2020-05-12
feat(plugins): Load plugin release stage

robzienert:
Helpful stage for administering preferred plugin versions in pipelines.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/LoadPluginReleaseTask.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/LoadPluginReleaseTask.java
@@ -15,10 +15,14 @@
  */
 package com.netflix.spinnaker.orca.front50.tasks;
 
+import static java.lang.String.format;
+
 import com.netflix.spinnaker.orca.api.pipeline.RetryableTask;
 import com.netflix.spinnaker.orca.api.pipeline.TaskResult;
+import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus;
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
 import com.netflix.spinnaker.orca.front50.Front50Service;
+import com.netflix.spinnaker.orca.front50.model.PluginInfo;
 import java.time.Duration;
 import java.util.Objects;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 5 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/LoadPluginReleaseTask.java
```diff
@@ -0,0 +1,77 @@
...
+      pluginInfo = front50Service.getPluginInfo(pluginId);
+    } catch (Exception e) {
+      log.error("Failed to retrieve plugin info for '{}'", pluginId, e);
+      return TaskResult.RUNNING;
+    }
+
+    return pluginInfo.getReleases().stream()
+        .filter(it -> it.getVersion().equals(version))
+        .findFirst()
+        .map(r -> TaskResult.builder(ExecutionStatus.SUCCEEDED).context("release", r).build())
```
marchello2000, line 61: why not put it to `outputs` (instead of `context`)?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/LoadPluginReleaseTask.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public LoadPluginReleaseTask(Front50Service front50Service) {
+    this.front50Service = front50Service;
+  }
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull StageExecution stage) {
+    final String pluginId = Objects.requireNonNull((String) stage.getContext().get("pluginId"));
+    final String version = Objects.requireNonNull((String) stage.getContext().get("version"));
```
jonsie, line 48: I think this should be `preferredVersion`.

jonsie: Wait I guess this is just to load any version not necessarily the preferred version, nevermind.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/LoadPluginReleaseTask.java
robzienert: Yeah, I think it's just because I typo'd that.

robzienert: Ha. No, it's because I load the `release` into the context so that `ExtractRequiredPluginDependenciesTask` can extract the release requirements field. If I do `outputs`, `stage.mapTo()` doesn't work because it only looks at the context.

Perhaps it'd be worthwhile having a similar `mapTo` for outputs?

PR #3695, opened 2020-05-21
fix(aws/cfn): irrecoverable stack status

ariadnarouco:
In a Deploy Cloudformation stage, when a stack creation fails, it goes to an irrecoverable status ROLLBACK_COMPLETE.  AWS rollbacks the changes made. Even if the changes are made to the template to solve the issue why the stack creation failed, the stack needs to be deleted in order to re-run the stage successfully.
Before this changes, when a CFN stage tried to create a stack (without a changeset) and creation failed, error message explained the reason why the creation failed; but it didn't explain that the stack needs to be deleted to re-run the stage.
When the same scenario happened and a change set creation was requested, the error message was  "NOT_YET_READY", which it does not reflect what actually happened.
With this changes, error message will explain steps to take and reason why the creation failed.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/WaitForCloudFormationCompletionTask.java
```diff
@@ -122,6 +115,22 @@ public long getTimeout() {
...
+    List unrecoverableStatuses =
+        Arrays.asList(
+            CloudFormationStates.ROLLBACK_COMPLETE.name(),
+            CloudFormationStates.ROLLBACK_FAILED.name(),
+            CloudFormationStates.DELETE_FAILED.name(),
+            CloudFormationStates.UPDATE_ROLLBACK_FAILED.name());
+    if (unrecoverableStatuses.contains(stack.get("stackStatus"))) {
+      return "Irrecoverable stack status - Review the error, make changes in template and delete the stack to re-run the pipeline successfully; Reason: "
+          + getStackInfo(stack, "stackStatusReason");
+    }
```
xavileon, line 128: This is probably only related to stack operations (not change sets). Should this block be inside the `getStackInfo` method? In any case, this clearly calls for a refactor as you suggested to split the logic between stacks and changesets 👍

ariadnarouco: I think this logic belongs to both cases, stack and changeset. 

*When the same scenario happened and a change set creation was requested, the error message was "NOT_YET_READY", which it does not reflect what actually happened.*

So when this happens, the flow goes through the change set logic and it returns NOT_YET_READY because there's not changeset information. 
So instead, we should return the stack status.

xavileon: Ah cool, makes sense. Thanks!

xavileon:
@spinnaker/oss-approvers PTAL

ariadnarouco:
@spinnakerbot cherry-pick 1.20

PR #3701, opened 2020-05-26
feat(peering): add ability to author custom peering agents

marchello2000:
This change adds an extension point (in the form of beans injected into the `PeeringAgent`) to allow peering of custom data in the `orca` [SQL] DB.
Currently, the custom peering agents are called without much context as to what the main peering agent is doing (they are simply called at the end of every agent tick, after all main peering agent logic completes).
This can be changed later if needed but for now I wanted to keep the two as independant and simple as possible.

### DIFF
--- a/orca-peering/orca-peering.gradle
+++ b/orca-peering/orca-peering.gradle
@@ -18,6 +18,7 @@ apply from: "$rootDir/gradle/kotlin.gradle"
 apply from: "$rootDir/gradle/spock.gradle"
 
 dependencies {
+  api("com.netflix.spinnaker.keiko:keiko-core:$keikoVersion")
   implementation(project(":orca-core"))
 
   implementation("com.netflix.spinnaker.kork:kork-sql")
--- a/orca-peering/orca-peering.gradle
+++ b/orca-peering/orca-peering.gradle
@@ -18,7 +18,6 @@ apply from: "$rootDir/gradle/kotlin.gradle"
 apply from: "$rootDir/gradle/spock.gradle"
 
 dependencies {
-  api("com.netflix.spinnaker.keiko:keiko-core:$keikoVersion")
   implementation(project(":orca-core"))
 
   implementation("com.netflix.spinnaker.kork:kork-sql")
--- a/orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
+++ b/orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
@@ -64,6 +64,8 @@ class PeeringAgent(
 
   private val executionCopier: ExecutionCopier,
 
+  private val customPeerers: List<CustomPeerer>,
+
   clusterLock: NotificationClusterLock
 ) : AbstractPollingNotificationAgent(clusterLock) {
 
@@ -79,6 +81,7 @@ class PeeringAgent(
         peerExecutions(ExecutionType.PIPELINE)
         peerExecutions(ExecutionType.ORCHESTRATION)
         peerDeletedExecutions()
+        invokeCustomPeerers()
       }
     }
   }
@@ -186,6 +189,24 @@ class PeeringAgent(
       log.error("Failed to delete some executions", e)
       peeringMetrics.incrementNumErrors(ExecutionType.ORCHESTRATION)
       peeringMetrics.incrementNumErrors(ExecutionType.PIPELINE)
+    }
+  }
+
+  /**
+   * If we have any custom peerers, invoke them
+   */
+  private fun invokeCustomPeerers() {
+    customPeerers.forEach { customPeerer ->
+      val peererName = customPeerer.javaClass.simpleName
+
+      try {
+        log.info("Starting peering with custom peerer '$peererName'")
+        customPeerer.doPeer(srcDB, destDB, peeredId)
+        log.info("Completed peering with custom peerer '$peererName'")
+      } catch (e: Exception) {
+        peeringMetrics.incrementCustomPeererError(peererName)
+        log.error("Custom peerer '$peererName' failed", e)
+      }
     }
   }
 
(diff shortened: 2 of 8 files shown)


orca-peering/orca-peering.gradle
```diff
@@ -18,6 +18,7 @@ apply from: "$rootDir/gradle/kotlin.gradle"
 apply from: "$rootDir/gradle/spock.gradle"
 
 dependencies {
+  api("com.netflix.spinnaker.keiko:keiko-core:$keikoVersion")
```
dreynaud, line 21: keiko! What for?

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -189,6 +192,24 @@ class PeeringAgent(
     }
   }
 
+  /**
+   * If we have any custom peerers, invoke them
+   */
+  private fun invokeCustomPeerers() {
+    customPeerers.forEach { customPeerer ->
```
dreynaud, line 199: it might make sense to define a `getPriority` method in `CustomPeerer` so that we can control the order in which they run, independently of the order Spring orders them beans

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -189,6 +192,24 @@ class PeeringAgent(
...
+      val peererName = customPeerer.javaClass.simpleName
+
+      try {
+        log.info("Starting peering with custom peerer '$peererName'")
+        customPeerer.doPeer(srcDB, destDB, peeredId)
+        log.info("Completed peering with custom peerer '$peererName'")
+      } catch (e: Exception) {
+        peeringMetrics.incrementCustomPeererError(peererName)
```
dreynaud, line 207: could pass the exception in there too and add the simple class name as a tag

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
```diff
@@ -189,6 +192,24 @@ class PeeringAgent(
...
+  /**
+   * If we have any custom peerers, invoke them
+   */
+  private fun invokeCustomPeerers() {
+    customPeerers.forEach { customPeerer ->
+      val peererName = customPeerer.javaClass.simpleName
+
+      try {
+        log.info("Starting peering with custom peerer '$peererName'")
+        customPeerer.doPeer(srcDB, destDB, peeredId)
```
dreynaud, line 204: would be neat to wrap this call in a `Timer::record` call

orca-peering/src/test/groovy/com/netflix/spinnaker/orca/peering/PeeringAgentSpec.groovy
```diff
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.orca.peering
 
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
 import com.netflix.spinnaker.orca.notifications.NotificationClusterLock
+import org.jetbrains.annotations.NotNull
```
dreynaud, line 21: not that it matters too much, but we probably don't want the jetbrains `NotNull`

marchello2000: auto-generated code... that's what i get for using that, thanks - removed!

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
marchello2000: I thought about it, and the ability to specify if you want to run before or after the rest of the peering runs. Ultimately, I decided it's added complexity for no reason.
If you have custom data that needs peering, you should do that in a single peerer. Reasoning about the order in which they run across multiple classes seems fraught.

I guess as I say this, I think the right approach here is to change List<CustomPeerer> to CustomPeerer to only allow a single one. Thoughts?

orca-peering/orca-peering.gradle
marchello2000: I needed the queue :( explanation forthcoming in our repo

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
ajordens: If we did actually need that wouldn't using `@Order` on the bean itself do the trick?

marchello2000: i'm changing this to only take a single custom peerer, it doesn't make sense to have multiple - if it does, we can change it then

orca-peering/orca-peering.gradle
marchello2000: actually, removed...

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
dreynaud: Either an explicit order or a single one are fine with me

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
marchello2000: yeah, i was going back and forth... at the end settled that the implementer should do that so they know what metric to track... but i can be convinced otherwise... marinading

marchello2000:
@dreynaud any more feedback you have?

orca-peering/src/main/kotlin/com/netflix/spinnaker/orca/peering/PeeringAgent.kt
dreynaud: I really feel strongly that it should be done here to enforce that it is actually done so that there is a consistent metric naming/tagging scheme. Plus if you're implementing the custom peerer the timing comes for free out of the box, which is a nice touch.

dreynaud:
@marchello2000 looks good, gave you the thumb of approval

PR #3706, opened 2020-05-29
fix(mpt-v1): prevent leak of information through template resolution …

danielpeach:
…endpoint

You can use Orca to search the internet! Or read its ec2 metadata or whatever. This change fixes that. The provided test fails without the rest of the changes and is a good end-to-end demonstration of the vulnerability.

The other proposed solutions were a blacklist or a whitelist for the template resolver. A blacklist is going to be incomplete. A whitelist is going to be either default open (in which case it doesn't solve the vulnerability for those that don't read this PR) or default closed, which would break people. A whitelist might still be a good idea eventually, though I'm not sure how many people are still using this feature.


danielpeach:

danielpeach:

ezimanyi:
💯 Great job on this fix, and tests to demonstrate it!

rkreddypandu:
Is there a CVE assigned to this vulnerability?

bcornils:
We don't have a CVE assigned yet.

PR #3711, opened 2020-05-30
refactor(artifacts): Update places that mutate artifacts

ezimanyi:
Artifacts will soon become immutable to eliminate the confusion about whether mutating an artifact is actually updating the artifact on the pipeline. This is confusing because we very often use objectMapper.convertValue() on artifacts to convert back and forth between maps. It is thus far from clear whether a mutation at any given point is acting on a copy of the pipeline artifact or the artifact itself.

This commit updates places that mutated artifacts to instead return a copy. In all cases this is setting the account for legacy workflows that have this specified separately from the artifact itself. We should not be adding more of these, and ideally these would at some point be migrated to use the new workflow of having the account specified on the artifact.

In all cases updated here were were either always or sometimes operating on a copy of the artifact anyway because of the frequent Map <-> Artifact conversion we do (mostly to evaluate SpEL). This will make things a lot clearer by having these mutations always return a mutated copy to avoid depending on the fragile behavior of mutating a sometimes-copy.

(At one point a Spring Boot upgrade which upgraded Jackson broke artifact workflows because it changed whether objectMapper.convertValue short-circuited in the event of no conversion being necessary, for an example of the fragility of mutating artifacts.)

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
@@ -17,7 +17,6 @@
 
 package com.netflix.spinnaker.orca.bakery.tasks.manifests;
 
-import com.google.common.base.Strings;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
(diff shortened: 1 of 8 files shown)


ezimanyi:
With this change there are no longer any places in orca that mutate artifacts.  I'll still leave the deprecated mutating functions in `kork` until at least the next release so it's easier to roll back any regressions.

PR #3717, opened 2020-06-02
refactor(artifacts): Update call to getMetadata

ezimanyi:
I'm working on better encapuslating artifacts, and one change is removing direct access to the map of metadata via .getMetadata() in favor of having clients specifically ask for keys via getMetadata(String key). In addition to hiding the implementation from clients, this usually also simplifies client code as generally client code first checks that the metadata map is not null then checks a specific key.

This updates the one place in orca where we were fetching the full metadata map to instead look up the desired key.


ezimanyi:
@clareliguori @Kaixiang-AWS : Mind taking a look at this tiny change to support a refactor I'm doing of artifacts?  (Tagging you as this was added in #3424.)  Thanks!

Kaixiang-AWS:
LGTM.

PR #3722, opened 2020-06-04
feat(aws/findimage): pass account to clouddriver when retrieving images

dbyron0:
if present in the stage json

See https://spinnakerteam.slack.com/archives/CB78MFQPP/p1591123948104700 for background.

If there are some docs to update, I could use a pointer to them.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/AmazonImageFinder.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/AmazonImageFinder.java
@@ -83,8 +83,15 @@ public class AmazonImageFinder implements ImageFinder {
       Map<String, String> tags,
       List<String> warningsCollector) {
     StageData stageData = (StageData) stage.mapTo(StageData.class);
+
     List<AmazonImage> allMatchedImages =
-        oortService.findImage(getCloudProvider(), packageName, null, null, prefixTags(tags))
+        oortService
+            .findImage(
+                getCloudProvider(),
+                packageName,
+                stageData.imageOwnerAccount,
+                null,
+                prefixTags(tags))
             .stream()
             .map(image -> objectMapper.convertValue(image, AmazonImage.class))
             .filter(image -> image.tagsByImageId != null && image.tagsByImageId.size() != 0)
@@ -131,6 +138,7 @@ public class AmazonImageFinder implements ImageFinder {
 
   static class StageData {
     @JsonProperty Collection<String> regions;
+    public String imageOwnerAccount;
   }
 
   static class AmazonImage implements Comparable<AmazonImage> {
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/AmazonImageFinder.java
```diff
@@ -131,6 +133,7 @@ public String getCloudProvider() {
 
   static class StageData {
     @JsonProperty Collection<String> regions;
+    public String account;
```
dbyron0, line 136: Is it correct/required to add `@JsonProperty` here?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/AmazonImageFinder.java
```diff
@@ -83,8 +83,10 @@ public boolean test(AmazonImage amazonImage) {
       Map<String, String> tags,
       List<String> warningsCollector) {
     StageData stageData = (StageData) stage.mapTo(StageData.class);
+
     List<AmazonImage> allMatchedImages =
-        oortService.findImage(getCloudProvider(), packageName, null, null, prefixTags(tags))
+        oortService
+            .findImage(getCloudProvider(), packageName, stageData.account, null, prefixTags(tags))
```
dreynaud, line 89: LGTM, @cfieber do you see any reason to avoid passing the account to clouddriver if we have it?

cfieber:
can you elaborate a bit on the specific behavior you are trying to get out of this?

In a Spinnaker install there is typically configured a single account that is responsible for image creation (I think there are some config options floating around like `default.bake.account` etc), and then images are shared from that account to other accounts via an allow launch operation.

I think the reason we currently don't pass an account is we expect that we are going to find images from that default account, and if we try to look for them specifically in other accounts they won't show up.

That said - if you are doing something else for image creation or trying to locate public images or something else, could you describe a bit more?

Also if we do go ahead with this change I'd propose we clarify the `account` attribute to make its purpose a bit more explicit (`imageOwnerAccount` or something - naming things is hard...)

dbyron0:
Thanks for getting involved.

Sure, let me try to explain where this is coming from.

We typically bake our images outside of spinnaker -- jenkins jobs that run packer.
So our pipelines typically have a Find Image From Tags stage and then a Deploy stage.  And probably due to some less than ideal AMI naming practices, spinnaker found "the wrong image" in "the wrong account".  After cleaning up some bunk images spinnaker ended up finding the right image in the right account, but it seems cleaner to be able make the account explicit.

I totally support `imageOwnerAccount`.

cfieber:
cool, I'm okay if we go ahead with this with the change to imageOwnerAccount

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/AmazonImageFinder.java
cfieber: it isn't necessary (your test case proves that)

dbyron0: Yeah, was curious whether there's some gap so that even with a passing test, it could break in the 'real world.'

dbyron0:
I was out of date with master, so I force pushed all as one commit with `imageOwnerAccount`: https://github.com/spinnaker/orca/pull/3722/commits/8b88c5df927f66b457b107ec02f460eeaec5627e

dbyron0:
Anyone know how to kick the PR builder...not sure why it timed out / was cancelled.

cfieber:
kicked it

dbyron0:
PR builder struggling yet again :(

dreynaud:
> PR builder struggling yet again :(

kicked it again, it's building now

PR #3723, opened 2020-06-04
fix(misc): Make default retry statuses configurable.

srekapalli:
Make default retry statuses configurable


ajordens:
Any value in just making the retryable status codes configureable (and defaulting to [429])?

Whatever is causing the occasional 403 blip is Netflix-specific right.

robzienert:
+1 to making this configurable.

srekapalli:
> lgtm, i am curious though why we want to retry on 403's?

Doing this as we see 403's intermittently while invoking monitor webhooks tasks for a longer period of time and halts large pipelines from completion.

Instead of specifying explicitly, making it configurable.

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/CreateWebhookTask.groovy
```diff
@@ -48,8 +49,14 @@ class CreateWebhookTask implements RetryableTask {
   long backoffPeriod = 10000
   long timeout = 300000
 
-  @Autowired
   WebhookService webhookService
+  int[] defaultRetryStatusCodes
+
+  @Autowired
+  CreateWebhookTask(WebhookService webhookService, @Value('${webhook.default-retry-status-codes:429}') int[] defaultRetryStatusCodes) {
```
robzienert, line 56: Do we like `@Value` annotations these days? We've historically steered clear of them quite aggressively. I think they're hard to discover and document, but I'm open to countering opinions.

PR #3728, opened 2020-06-08
fix(clouddriver): Use application owner as user in EphemeralServerGroupsPoller cleanup executions

robzienert:
Will attempt to use the application owner of a server group as the user, or else fallback to the original behavior of a static `spinnaker` user if an application owner cannot be found.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pollers/EphemeralServerGroupsPoller.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pollers/EphemeralServerGroupsPoller.java
@@ -25,13 +25,16 @@ import com.google.common.collect.ImmutableMap;
 import com.netflix.spectator.api.Counter;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.kork.core.RetrySupport;
+import com.netflix.spinnaker.kork.exceptions.SystemException;
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType;
 import com.netflix.spinnaker.orca.clouddriver.OortService;
+import com.netflix.spinnaker.orca.clouddriver.config.PollerConfigurationProperties;
+import com.netflix.spinnaker.orca.front50.Front50Service;
+import com.netflix.spinnaker.orca.front50.model.Application;
 import com.netflix.spinnaker.orca.notifications.AbstractPollingNotificationAgent;
 import com.netflix.spinnaker.orca.notifications.NotificationClusterLock;
 import com.netflix.spinnaker.orca.pipeline.ExecutionLauncher;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
-import com.netflix.spinnaker.security.User;
 import groovy.util.logging.Slf4j;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pollers/EphemeralServerGroupsPoller.java
```diff
@@ -128,17 +135,25 @@ protected void tick() {
...
+                      return "spinnaker";
+                    });
+
+        AuthenticatedRequest.runAs(
+            taskUser,
+            Collections.singletonList(ephemeralServerGroupTag.account),
+            () ->
+                executionLauncher.start(
+                    ExecutionType.ORCHESTRATION,
+                    objectMapper.writeValueAsString(cleanupOperation)));
```
marchello2000, line 156: i think you still need the `.call()` at the end, no?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pollers/EphemeralServerGroupsPoller.java
```diff
@@ -128,17 +135,25 @@ protected void tick() {
...
+                .orElseGet(
+                    () -> {
+                      log.warn(
+                          "Failed to find application owner for server group '{}', "
+                              + "will use static 'spinnaker' user as fallback.",
+                          ephemeralServerGroupTag.serverGroup);
+                      return "spinnaker";
+                    });
+
+        AuthenticatedRequest.runAs(
```
cfieber, line 150: the opinion that the app owner is always a member of the app WRITE permissions is something we have implemented internally but isn't generally the case

also the allowedAccounts bit here is used as a fallback, but if we successfully look up the app owner as a service account then this would run with that service account's permissions. hopefully it succeeds, and we can via configuration make it so, but we may see failures for service accounts that are not set up to access restricted accounts (hopefully those become less of a thing as people start using those service accounts for managed delivery)

trying to think if I can offer any suggestions aside from just raising potential problems.. heh

ajordens: I doubt anyone else is using entity tags right, should we just consider this Netflix-only?

I suppose an alternative would be to look up the `executionId` and grab the authorization out of it ... at this point I don't know if we'd be splitting hairs tho.

robzienert: Look up the `executionId` of what? This is a poller, so there is no parent execution that we can lookup.

I will make the default for this task configurable, rather than `"spinnaker"`. That should check the box of many of the issues Cam brought up... in case anyone actually uses entity tags out there.

Good thoughts, thanks.

robzienert: Look up the `executionId` of what? This is a poller, so there is no parent execution that we can lookup.

I will make the default for this task configurable, rather than `"spinnaker"`. That should check the box of many of the issues Cam brought up... in case anyone actually uses entity tags out there.

Good thoughts, thanks.

robzienert: Updated. A configuration property will be preferred, but if unset, the original logic of using the application user will be used, still falling back to `spinnaker` if still nothing is found.

ajordens: The entity tag has the original execution id ... configuration property is reasonable to me tho.

PR #3729, opened 2020-06-09
fix(traffic-guards): make front50 look up first

dreynaud:
fix(traffic-guards): make front50 look up first
    
We need to check if traffic guards are enabled for this cluster in front50 first.
    
If they are not enabled, no need to make any clouddriver calls.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
@@ -163,6 +163,16 @@ public class TrafficGuard {
       Location location,
       String cloudProvider,
       String operationDescriptor) {
+
+    if (!hasDisableLock(serverGroupMoniker, account, location)) {
+      log.debug(
+          "No traffic guard configured for '{}' in {}/{}",
+          serverGroupName,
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -163,6 +163,16 @@ public void verifyTrafficRemoval(
       Location location,
       String cloudProvider,
       String operationDescriptor) {
+
+    if (!hasDisableLock(serverGroupMoniker, account, location)) {
```
gal-yardeni, line 167: will this condition prevent throwing `TrafficGuardException` if traffic guard is not configured? (not sure what `hasDisableLock` is doing)

gal-yardeni: oh I see, this call is calling `front50` to verify it.

dreynaud: Yes, `hasDisableLock` checks the trafficGuard section of the app config in front50. No traffic guard configured, no exception

PR #3733, opened 2020-06-11
fix(provider/cf): use moniker objects for task operations to support future non-frigga names

zachsmith1:
In the future we will be using Moniker to support non-frigga based naming conventions. This change ensures that the cloudfoundry tasks which use application name are sending the moniker object, as well as the old. The old will be removed once all clouddriver operations have been changed as well.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryServerGroupCreator.java
```diff
@@ -60,6 +62,8 @@
             .build();
     ManifestEvaluator.Result evaluatedManifest = manifestEvaluator.evaluate(stage, manifestContext);
 
+    Moniker moniker = buildMoniker(context);
```
german-muzquiz, line 65: I don't see this new variable being used (so the new `buildMoniker` method is also not used)

zachsmith1: Thanks -- must have missed it when I switched branches and re-did the changes.

PR #3734, opened 2020-06-11
fix(misc): determine host using authority if host returned is null du…

srekapalli:
`URI.getHost()` returns null if the host part of the  url contains **_** character, so using `URI.getAuthority` api to determine the host if we receive null from `getHost` call.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
@@ -19,7 +19,14 @@ package com.netflix.spinnaker.orca.config;
 import java.net.InetAddress;
 import java.net.NetworkInterface;
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -103,34 +110,45 @@ public URI validateURI(String url) throws IllegalArgumentException {
       if (!allowedSchemes.contains(u.getScheme().toLowerCase())) {
         throw new IllegalArgumentException("unsupported URI scheme " + url);
       }
-      if (!allowedHostnames.matcher(u.getHost()).matches()) {
+
+      // If host contains '_', getHost is returning null, so falling back to authority to determine
```
ajordens, line 114:
```
// fallback to `getAuthority()` in the event that the hostname contains an underscore
```

ajordens:
LGTM

PR #3737, opened 2020-06-12
fix(orca-clouddriver): Do not mark ResizeAsgStage as SUCCESSFUL in taskGraph()

christopherthielen:


.idea/compiler.xml
```diff
@@ -3,6 +3,27 @@
   <component name="CompilerConfiguration">
     <annotationProcessing>
       <profile default="true" name="Default" enabled="true" />
+      <profile name="Gradle Imported" enabled="true">
```
christopherthielen, line 6: What the?  How did this get here

jeyrschabu: Yeah these files get modified by intelliJ. You can tell git to care about them. `git update-index --no-assume-unchanged compiler.xml`

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/ResizeAsgStage.groovy
```diff
@@ -56,14 +56,6 @@ class ResizeAsgStage implements StageDefinitionBuilder {
 
   @Override
   void taskGraph(@Nonnull StageExecution stage, @Nonnull TaskNode.Builder builder) {
-    if (!stage.parentStageId || stage.execution.stages.find {
```
marchello2000, line 59: i don't even... 🤯

PR #3738, opened 2020-06-12
fix(perf): retrievePipelinesForPipelineConfigIdsBetweenBuildTimeBound…

marchello2000:
…ary shouldn't get all executions

Because of "elaborate" ability to filter in `searchForPipelinesByTrigger` it must combine the filtering capabilities of execution repository and plain code.
To do that, it would retrieve ALL executions first and then filter to the subset desired (default 10).

This change adds filter on a page-by-page execution retrieval, so only the minimum number of results are retrieved from the DB

Ideally, this endpoint just goes away. But this is the minimum change to make it palatable perf wise


marchello2000:
fixing tests

PR #3739, opened 2020-06-15
fix(provider/appengine): groovy.lang.MissingMethodException when referencing github config artifact for GAE deploy - spinnaker/spinnaker#5836

guido9j:
**Bug description:**
groovy.lang.MissingMethodException thrown when referencing github config artifact for GAE deploy. (see spinnaker/spinnaker#5836 for more details)

**Fix:**
The config artifacts  were of instance of HashMap not an Artifact object. Added logic to translate HashMap to Artifact.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
@@ -25,9 +25,7 @@ import com.netflix.spinnaker.orca.pipeline.util.ArtifactUtils
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.PIPELINE
-import groovy.util.logging.Slf4j
 
-@Slf4j
 @Component
 class AppEngineServerGroupCreator implements ServerGroupCreator {
   boolean katoResultExpected = false
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
@@ -25,7 +25,9 @@ import com.netflix.spinnaker.orca.pipeline.util.ArtifactUtils
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.PIPELINE
+import groovy.util.logging.Slf4j
 
+@Slf4j
 @Component
 class AppEngineServerGroupCreator implements ServerGroupCreator {
   boolean katoResultExpected = false
@@ -75,8 +77,19 @@ class AppEngineServerGroupCreator implements ServerGroupCreator {
       List<ArtifactAccountPair> configArtifacts = operation.configArtifacts
       if (configArtifacts != null && configArtifacts.size() > 0) {
         operation.configArtifacts = configArtifacts.collect { artifactAccountPair ->
-          def artifact = artifactUtils.getBoundArtifactForStage(stage, artifactAccountPair.id, artifactAccountPair.artifact)
-          artifact.artifactAccount = artifactAccountPair.account
+          log.debug ("appendArtifactData: artifactAccountPair.id="+artifactAccountPair.id+" artifactAccountPair.artifact="+artifactAccountPair.artifact+" className="+artifactAccountPair.artifact.getClass().getName()+" account="+artifactAccountPair.account)
+          String c_id = artifactAccountPair.id
+          def Artifact c_artifact
+          if (artifactAccountPair.artifact instanceof Map){
+            def map = artifactAccountPair.artifact
+            c_artifact = Artifact.builder().reference(map.get("reference")).metadata(map.get("metadata")).artifactAccount(map.get("artifactAccount")).uuid(map.get("id")).type(map.get("type")).build()
+          } else if (artifactAccountPair.artifact instanceof Artifact){
+            c_artifact = artifactAccountPair.artifact
+          }
+          def artifact = artifactUtils.getBoundArtifactForStage(stage, c_id, c_artifact)
+          if (artifactAccountPair.account != null){
+            artifact.artifactAccount = artifactAccountPair.account
+          }
           return artifact
         }
       }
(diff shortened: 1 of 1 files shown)


guido9j:
@ezimanyi : here is the pull request

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
```diff
@@ -75,8 +77,19 @@ class AppEngineServerGroupCreator implements ServerGroupCreator {
       List<ArtifactAccountPair> configArtifacts = operation.configArtifacts
       if (configArtifacts != null && configArtifacts.size() > 0) {
         operation.configArtifacts = configArtifacts.collect { artifactAccountPair ->
-          def artifact = artifactUtils.getBoundArtifactForStage(stage, artifactAccountPair.id, artifactAccountPair.artifact)
-          artifact.artifactAccount = artifactAccountPair.account
+          log.debug ("appendArtifactData: artifactAccountPair.id="+artifactAccountPair.id+" artifactAccountPair.artifact="+artifactAccountPair.artifact+" className="+artifactAccountPair.artifact.getClass().getName()+" account="+artifactAccountPair.account)
```
ezimanyi, line 80: Now that this issue is resolved, do we still need to log all of this information (even at debug level)?  If it is necessary, I would recommend just logging the artifact itself rather than building a long string like this.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
```diff
@@ -75,8 +77,19 @@ class AppEngineServerGroupCreator implements ServerGroupCreator {
...
+          String c_id = artifactAccountPair.id
+          def Artifact c_artifact
+          if (artifactAccountPair.artifact instanceof Map){
+            def map = artifactAccountPair.artifact
+            c_artifact = Artifact.builder().reference(map.get("reference")).metadata(map.get("metadata")).artifactAccount(map.get("artifactAccount")).uuid(map.get("id")).type(map.get("type")).build()
```
ezimanyi, line 85: Instead of manually creating the artifact, you can just do `objectMapper.convertValue(artifactAccountPair.artifact, Artifact.class)`.  This also will short-circuit if this is already an `Artifact` so you don't even need the `if/else` block here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
```diff
@@ -75,8 +77,19 @@ class AppEngineServerGroupCreator implements ServerGroupCreator {
       List<ArtifactAccountPair> configArtifacts = operation.configArtifacts
       if (configArtifacts != null && configArtifacts.size() > 0) {
         operation.configArtifacts = configArtifacts.collect { artifactAccountPair ->
-          def artifact = artifactUtils.getBoundArtifactForStage(stage, artifactAccountPair.id, artifactAccountPair.artifact)
-          artifact.artifactAccount = artifactAccountPair.account
+          log.debug ("appendArtifactData: artifactAccountPair.id="+artifactAccountPair.id+" artifactAccountPair.artifact="+artifactAccountPair.artifact+" className="+artifactAccountPair.artifact.getClass().getName()+" account="+artifactAccountPair.account)
+          String c_id = artifactAccountPair.id
```
ezimanyi, line 81: Do we need this as a separate variable?  It looks like this is only used in one place, so it would be clearer to just directly access `artifactAccountPair.id` in that one place of use.

guido9j:
@ezimanyi I applied all your suggestions.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
```diff
@@ -25,7 +25,9 @@ import com.netflix.spinnaker.orca.pipeline.util.ArtifactUtils
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 import static com.netflix.spinnaker.orca.pipeline.model.Execution.ExecutionType.PIPELINE
+import groovy.util.logging.Slf4j
 
+@Slf4j
```
ezimanyi, line 30: I think this can be removed now.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
```diff
@@ -75,8 +77,10 @@ class AppEngineServerGroupCreator implements ServerGroupCreator {
       List<ArtifactAccountPair> configArtifacts = operation.configArtifacts
       if (configArtifacts != null && configArtifacts.size() > 0) {
         operation.configArtifacts = configArtifacts.collect { artifactAccountPair ->
-          def artifact = artifactUtils.getBoundArtifactForStage(stage, artifactAccountPair.id, artifactAccountPair.artifact)
-          artifact.artifactAccount = artifactAccountPair.account
+          def artifact = artifactUtils.getBoundArtifactForStage(stage, artifactAccountPair.id, objectMapper.convertValue(artifactAccountPair.artifact, Artifact.class))
+          if (artifactAccountPair.account != null){
```
ezimanyi, line 81: I'm curious about the need for this change here (which seems unrelated)...did you encounter a bug where this was set even when null?

This seems reasonable, but for a change on a release branch I'd want the minimal change that fixes the bug.

ezimanyi:
I think the plan here should be:
* Port this to `master` which will require a manual port because an adjacent line was changed.
* Once that is merged to `master`, (and the comments here addressed) we can merge this here and port to 1.20.

That way it will be fixed on 1.19, 1.20, and master.  (This was a regression introduced in #3013 which is why I think we should accept the cherry pick.)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
guido9j: @ezimanyi:
 I added this logic:
`if (artifactAccountPair.account != null) { ... }`
Because in this branch, it breaks the code. The `account` is already set in the Artifact object, and the assignment of account in Artifact was actually setting the account to `null` instead of keeping the value already set in Artifact.

The master branch actually already has this fix, when it calls `ArtifactUtil.withAccount()`, which does the check for null, instead of blindly setting it.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/appengine/AppEngineServerGroupCreator.groovy
guido9j: This is removed along with `import`

ezimanyi:

PR #3746, opened 2020-06-16
feature(Pipeline Expressions): Add function pipelineIdInApplication hich allows to fetch the ID of a pipeline given its name and its application name

AbdulRahmanAlHamali:
This function is useful when we are running a dependent pipeline that falls in a different application


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineExpressionFunctionProvider.java
```diff
@@ -83,39 +92,18 @@ public static String pipelineId(PipelineExecution execution, String pipelineName
           "pipelineName must be specified for function #pipelineId");
     }
```
AbdulRahmanAlHamali, line 94: I extracted the logic here into a common function

AbdulRahmanAlHamali:
It's sad that we cannot overload pipeline expression functions

zachsmith1:
@AbdulRahmanAlHamali Please merge master into your branch in order to complete this PR.

AbdulRahmanAlHamali:
Done :)

PR #3749, opened 2020-06-17
feat(clouddriver): Add generic cloud operation stage

robzienert:
This is the first draft of a generic cloud operation stage. The goal of this
stage, and the surrounding work, is to migrate all cloud provider logic from
orca into Clouddriver. The "vision" of this would be that to add a cloud
provider, or new operations for a cloud provider, would no longer require
modification to Orca; only requiring changes to Clouddriver.

This PR is submitting an incomplete stage, but I don't want to drop a mega-huge PR once everything is buttoned up. I'm not really concerned about it being incomplete since nothing will be using it.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoRestService.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoRestService.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.orca.clouddriver
 
+import com.netflix.spinnaker.orca.clouddriver.model.OperationContext
 import com.netflix.spinnaker.orca.clouddriver.model.Task
 import com.netflix.spinnaker.orca.clouddriver.model.TaskId
 import io.github.resilience4j.retry.annotation.Retry
@@ -44,6 +45,12 @@ interface KatoRestService {
   TaskId requestOperations(@Query("clientRequestId") String clientRequestId,
                            @Path("cloudProvider") String cloudProvider,
                            @Body Collection<? extends Map<String, Map>> operations)
+
+  @POST("/{cloudProvider}/ops/{operationName}")
+  Response submitOperation(@Query("clientRequestId") String clientRequestId,
+                           @Path("cloudProvider") String cloudProvider,
+                           @Path("operationName") String operationName,
+                           @Body OperationContext operation);
 
   @GET("/applications/{app}/jobs/{account}/{region}/{id}")
   Response collectJob(@Path("app") String app,
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/CloudOperationStage.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/CloudOperationStage.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.clouddriver.pipeline;
+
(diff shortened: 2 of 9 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoRestService.groovy
```diff
@@ -45,6 +46,11 @@ interface KatoRestService {
                            @Path("cloudProvider") String cloudProvider,
                            @Body Collection<? extends Map<String, Map>> operations)
 
+  @POST("/{cloudProvider}/ops")
+  Response submitOperations(@Query("clientRequestId") String clientRequestId,
+                            @Path("cloudProvider") String cloudProvider,
+                            @Body Collection<Map<String, OperationContext>> operations);
```
robzienert, line 52: Not really different from the other `submitOperations` methods, except this one returns the `Response`, as the response code is significant for this stage to work correctly.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/CloudOperationStage.java
```diff
@@ -0,0 +1,159 @@
...
+import org.springframework.util.StringUtils;
+
+@Component
+public class CloudOperationStage implements StageDefinitionBuilder, Nameable {
+
+  private static final String LIFECYCLE_KEY = "operationLifecycle";
+
+  protected TaskGraphConfigurer configureTaskGraph(@Nonnull StageExecution stage) {
+    return new TaskGraphConfigurer();
+  }
```
robzienert, line 41: `ResizeServerGroupStage` will work as-is without further configuration of the task graph, but for more complex stages, such as Deploy and so-on, they'll need to extend this stage to decorate the task graph more - this method is forward-looking to those use cases.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorCloudOperationTask.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.netflix.spinnaker.orca.api.pipeline.RetryableTask;
+import com.netflix.spinnaker.orca.api.pipeline.TaskResult;
+import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
+import java.time.Duration;
+import javax.annotation.Nonnull;
+import org.springframework.stereotype.Component;
+
+@Component
+public class MonitorCloudOperationTask implements RetryableTask {
```
robzienert, line 26: This will be roughly equivalent to `MonitorKatoTask`, but I have an eye towards simplification. I haven't written this bit yet (all functional testing has been fire-and-forget). Maybe `MonitorKatoTask` is the way to go, but I'm eager to evaluate if we can have something simpler that gets the same job done.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/SubmitCloudOperationTask.java
```diff
@@ -0,0 +1,64 @@
...
+  public TaskResult execute(@Nonnull StageExecution stage) {
+    OperationContext context = stage.mapTo(OperationContext.class);
+
+    SubmitOperationResult result =
+        katoService.submitOperation(context.getCloudProvider(), Collections.singletonList(context));
+
+    TaskResult.TaskResultBuilder builder = TaskResult.builder(ExecutionStatus.SUCCEEDED);
+    if (result.getId() != null) {
+      builder.context("kato.last.task.id", result.getId());
+    }
```
robzienert, line 50: There's more work to be done in here to correctly extract result objects from clouddriver. I'm hoping that it will just be dropping values into the outputs direct from clouddriver's response.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoRestService.groovy
cfieber: Thoughts on whether this should support the list of operations model here rather than just hitting a single operation `/{cloudProvider}/ops/{operationName}`?

I think the most we've used the ability to submit a list of ops for was to do things like injecting an allow launch before an AWS deploy and things like that which will move to clouddriver with these changes

It makes it harder to reason about what the Response code means when there are multiple operations in the payload.

Not sure if you have some specific cases in mind where we might need to submit multiple ops to clouddriver?

robzienert: Oh, no. I was just trying to keep consistent with existing use cases. I'd love to see multiple ops disappear. Nice! I'll switch over to that.

PR #3752, opened 2020-06-17
fix(provider/cf): fix form based manifest resolution in deploy stage

zachsmith1:
Originally, when doing a CloudFoundry form-based manifest deployment, the form-based manifest would be set in the context under "manifest". This manifest object contains the actual details that are presented in the form. When this was passed to the resolveManifest method, it would try to map it to a StageContextArtifactView which contains an artifactId and an Artifact. This is then passed to artifactUtils.getBoundArtifactForStage(*) so that we can correctly resolve the manifest if its from a prior stage, statically typed in the stage, etc, etc. The problem was that when we tried to map a direct manifest (form-based) to the StageContextArtifactView, it would result in all its field being null, hence breaking. 

The fix is to use the pre-existing DeploymentManifest object and map the direct manifest to it. This will give us either the artifactId or the resolved artifact. The DeploymentManifest has logic to handle it out of the box. We then pass this object to the new resolverUtil that will allow the stage to resolve the manifest if its from a previous stage, etc. Also, rather than sharing the logic for the resolve method and mapping each artifact before-hand, I kept them separate so that we could throw an error thats specific to the artifact being resolved.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryServerGroupCreator.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryServerGroupCreator.java
@@ -100,7 +100,8 @@ class CloudFoundryServerGroupCreator implements ServerGroupCreator {
             stageContextArtifactView.getArtifactId(),
             stageContextArtifactView.getArtifact());
     if (artifact == null) {
-      throw new IllegalArgumentException("Unable to bind the application artifact");
+      throw new IllegalArgumentException(
+          "Unable to bind the application artifact. Either the application artifact doesn't exist or this stage doesn't have access to fetch it");
     }
 
     return artifact;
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryServerGroupCreator.java
```diff
@@ -104,6 +106,18 @@ private Artifact resolveArtifact(StageExecution stage, Object input) {
     return artifact;
   }
 
+  private Artifact resolveManifestArtifact(StageExecution stage, Object input) {
+    DeploymentManifest manifest = mapper.convertValue(input, DeploymentManifest.class);
+    Artifact artifact =
+        artifactUtils.getBoundArtifactForStage(
+            stage, manifest.getArtifactId(), manifest.getArtifact());
+    if (artifact == null) {
+      throw new IllegalArgumentException("Unable to bind the manifest artifact");
```
dogonthehorizon, line 115: Do we have any more context for why the manifest failed to bind?

zachsmith1: @dogonthehorizon at this point in the code, no. But the artifactUtils.getBoundArtifactForStage has plenty of different catches for problems that could arise. This is a last resort throw

zachsmith1: I could, however, do something a bit more specific like : "Unable to bind the manifest artifact because artifactId or Artifact is null"

zachsmith1: Added more detail to the error message

zachsmith1:

kevinawoo:

PR #3758, opened 2020-06-22
Fixing mine canaries not waiting after disabling the canary clusters

dwest-netflix:
This is to address RSLNC-365 and, in particular, mine canaries failing to wait after disabling the baseline and canary cluster. Ideally this would use the disableCluster stage. However it looked like that might require a larger refactor since Mine doesn't deploy things in a normal fashion and the disableCluster stage was silently filtering out the cluster I gave it.

### DIFF
--- a/orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/pipeline/CanaryDisableClusterStage.groovy
+++ b/orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/pipeline/CanaryDisableClusterStage.groovy
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.kayenta.pipeline
+
+import com.netflix.spinnaker.orca.api.pipeline.CancellableStage
+import com.netflix.spinnaker.orca.api.pipeline.graph.StageDefinitionBuilder
+import com.netflix.spinnaker.orca.api.pipeline.graph.StageGraphBuilder
+import com.netflix.spinnaker.orca.api.pipeline.graph.TaskNode
+import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
+import com.netflix.spinnaker.orca.clouddriver.tasks.MonitorKatoTask
+import com.netflix.spinnaker.orca.mine.pipeline.CanaryDestroyClusterStage
+import com.netflix.spinnaker.orca.mine.pipeline.CanaryStage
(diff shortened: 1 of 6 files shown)


orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/pipeline/MonitorCanaryStage.groovy
```diff
@@ -37,22 +39,22 @@ class MonitorCanaryStage implements StageDefinitionBuilder, CancellableStage {
...
   @Autowired MineService mineService
   @Autowired CanaryStage canaryStage
+  @Autowired CanaryDisableClusterStage canaryDisableClusterStage
 
   @Override
   void taskGraph(@Nonnull StageExecution stage, @Nonnull TaskNode.Builder builder) {
     builder
       .withTask("registerCanary", RegisterCanaryTask)
       .withTask("monitorCanary", MonitorCanaryTask)
-      .withTask("disableCanaryCluster", DisableCanaryTask)
```
dwest-netflix, line 46: These tasks were broken up into two extra stages. The disable canary stage and the destroy canary stage.

orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/DisableCanaryTask.groovy
```diff
@@ -70,8 +65,7 @@ class DisableCanaryTask extends AbstractCloudProviderAwareTask implements Task {
     return TaskResult.builder(ExecutionStatus.SUCCEEDED).context([
       'kato.last.task.id'    : taskId,
       'deploy.server.groups' : dSG,
-      disabledCluster        : selector,
-      waitTime               : waitTime
```
dwest-netflix, line 74: No longer needed since this is now specified in the disableCanaryStage itself

orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/pipeline/CanaryDisableClusterStage.groovy
```diff
@@ -0,0 +1,79 @@
...
+        .withTask("disableCanaryCluster", DisableCanaryTask)
+        .withTask("monitorDisable", MonitorKatoTask)
+        .withTask("disableBaselineCluster", DisableCanaryTask)
+        .withTask("monitorDisable", MonitorKatoTask)
+  }
+
+  @Override
+  void afterStages(StageExecution parent, StageGraphBuilder graph) {
+    Integer waitTime = parent.context.clusterDisableWaitTime != null ? stage.context.clusterDisableWaitTime : DEFAULT_CLUSTER_DISABLE_WAIT_TIME
+    graph.append {
```
dwest-netflix, line 55: This is the real change here. We now use a wait stage instead of the wait task. The wait task was taking the stage's "startTime" as the basis for how long it should wait. Since the stage was the monitorCanary stage, if the canary ran longer than 180 seconds the wait task did nothing.

kevinawoo:
Hey hey, I think this is causing an error on master which is blocking cutting the release branches

Here's the error:
```
Description:

Field canaryStage in com.netflix.spinnaker.orca.kayenta.pipeline.CanaryDisableClusterStage required a bean of type 'com.netflix.spinnaker.orca.mine.pipeline.CanaryStage' that could not be found.


Action:

Consider defining a bean of type 'com.netflix.spinnaker.orca.mine.pipeline.CanaryStage' in your configuration.
```

I'm not exactly sure what mime is >_> lol

marchello2000:
@dwest-netflix can you take a look please?

dwest-netflix:
@kevinawoo @marchello2000 Sorry about that. Oversight when I moved a file around. Looks like I had the namespace incorrect for that particular class. PR incoming

PR #3759, opened 2020-06-22
fix(runjob): Only cancel runjob if one was started successfully

marchello2000:
If there is no `jobStatus` object there was no job and we will NPE


marchello2000:
review with whitespace off: https://github.com/spinnaker/orca/pull/3759/files?diff=split&w=1

PR #3760, opened 2020-06-22
fix(leak): fix metaspace leak form preconfigured jobs

marchello2000:
Every execution of preconfigured job would run groovy.eval which in turn generates a script whose Klass stuffs is never released from metaspace
(similar issues: https://github.com/elastic/elasticsearch/issues/18572 and https://hoangx281283.wordpress.com/2014/07/24/unload-groovy-classes/)

Unclear why there is a need to run `eval` here anyway, so removing it all-together


marchello2000:
thanks @cfieber !

PR #3763, opened 2020-06-23

sanopsmx:


sanopsmx:
Closing this PR as i have raised a new PR

https://github.com/spinnaker/orca/pull/3829

PR #3764, opened 2020-06-23
fix(rollback): fix capacity casting

emjburns:
We were failing to rollback because Capacity also has an "isPinned" field ([here](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/Capacity.java#L20)). This fixes the casting so that it won't fail.


marchello2000:
🤦  groovy...
i am wondering if there are other places like this. Thanks for hunting it down!

PR #3766, opened 2020-06-23
chore(build): gradle 6.5

cfieber:


cfieber:
the initial build failure isn't something I've yet repro'd locally, trying to re-run the jobs and see if cosmic rays are to blame..

PR #3770, opened 2020-06-24
fix(log noise - mptv2): denoise some MPTv2 errors

marchello2000:
Currently, if there are any templated pipelines that are not fully specified (which is very common as users are iterating on pipeline design) planning those pipelines will generate lots of noise logs.
This is especially annoying when this is done as part of dependentpipeline starter and so it looks like an exception occurred during an execution when in fact it had nothing to do with this particular execution.

This change silences those logs for dependent pipeline starter, there will be a part 2 of this change where better logic is applied to `/v2/pipelineTemplates/plan` to not log exceptions for similar cases (this EP is called by echo alot)

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/exceptions/PipelineMissingTemplateVariabledException.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/exceptions/PipelineMissingTemplateVariabledException.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/exceptions/PipelineMissingTemplateVariabledException.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.pipelinetemplate.exceptions;
+
+public class PipelineMissingTemplateVariabledException
+    extends IllegalTemplateConfigurationException {
```
marchello2000, line 19: this needs to eventually be `UserException` - that i will tackle in part 2

PR #3771, opened 2020-06-24
refactor(preconfigureJobs): Move KubernetesPreconfiguredJobProperties to orca-api

edgarulg:
The preconfiguredJob implementations were moved to `orca-api` except `KubernetesPreconfiguredJobProperties`, This PR is about moving it to `orca-api` to avoid have dependencies pointing to `orca-clouddriver` only for use `KubernetesPreconfiguredJobProperties` on plugins.


edgarulg:
Closing this PR because move `KubernetesPreconfiguredJobProperties` class to `orca-api` force us to implement Kubernetes client that is not a good approach

PR #3776, opened 2020-06-24
fix(monitored): split up resize and unpin operations for monitored deploy

marchello2000:
In the situations when the user is deploying via monitored to, say, [min:1, desired:100, max:100] capacity and there is downward pressure at the end of deploy (due to reduced traffic)
we might fail the deploy because we will never reach the desired state at the last (100%) step since we unpin at the same time as we scale up to 100%.
This change removes the "unpin" part of the 100% step of monitored and introduces a separate unpin operation after the deploy reaches 100% healthy


gal-yardeni:
@marchello2000 I'm having a hard time follow the logic, so I'll defer to Luis/Daniel for reviewing. Question - does it fix a filed bug?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/strategies/MonitoredDeployStrategy.groovy
```diff
@@ -261,6 +264,26 @@ class MonitoredDeployStrategy implements Strategy {
       }
     })
 
+    // Unpin target
```
marchello2000, line 267: this is the real change, just split up unpin from last resize

PR #3777, opened 2020-06-24
feat(echo): add a dynamic toggle for full executions

dreynaud:
Big executions can generate a lot of tasks, which means a lot of echo traffic that all include the _full_ execution payload.

I think we can definitely skip these at least for task events as a first step. We can also probably be more selective with stage events, but leaving that for later.


orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingStageListener.groovy
```diff
@@ -127,6 +131,16 @@ class EchoNotifyingStageListener implements StageListener {
         event.content.taskName = "${stage.type}.${task.name}".toString()
       }
 
+      if (dynamicConfigService.isEnabled("orca.echoEvents.includeFullExecution.enabled", true)) {
```
marchello2000, line 134:
```suggestion
      if (dynamicConfigService.isEnabled("orca.echoEvents.includeFullExecution", true)) {
```

isEnabled will add `.enabled` for you

orca-echo/src/test/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingStageListenerSpec.groovy
```diff
@@ -121,6 +123,7 @@ class EchoNotifyingStageListenerSpec extends Specification {
     echoListener.afterTask(persister, stage, task, executionStatus, wasSuccessful)
 
     then:
+    1 * dynamicConfigService.isEnabled("orca.echoEvents.includeFullExecution.enabled", true) >> fullExecutionToggle
```
marchello2000, line 126: same as above

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingStageListener.groovy
ajordens: Curious if we'd rather this be `echo.events.includeFullExecution`

```
echo:
  baseUrl: blahblahblah
  events:
    includeFullExecution: true
```

I don't think we have a config namespace right now for `orca` so it might be a bit odd-ball?

Otherwise this looks alright to me.

dreynaud: @marchello2000 I was intentional about adding the `.enabled` explicitly, I think it's important that we can easily grep for the full name of properties. A while back I updated `DynamicConfigService` so that it doesn't append `.enabled` if it's already there.

dreynaud: @ajordens I went back and forth between multiple kludgy names, `echo.events.includeFullExecution` is nicer 👍

dreynaud: in e4390b448

PR #3753, opened 2020-06-19
fix(provider/cf): fix form based manifest resolution in deploy stage (#3752)

zachsmith1:
Please see https://github.com/spinnaker/orca/pull/3752. 

Here is the breaking [commit](https://github.com/spinnaker/orca/commit/98152c833631a84203f700099149c955e4e4cbb1#diff-9c8b0013a03ef69b40307d50445a4acd)

The object mapping on line 50 was expecting an Artifact or/and an ArtifactId. The original was an actual manifest so it didn't map correctly. This would work fine with any deployments that weren't form based but all form based broke after this change. The new change uses the original working mapper wrapped in a method to add some error handling and more specific throwables. 

Co-authored-by: Zach Smith
Co-authored-by: Zach Smith
Co-authored-by: mergify[bot] <37929162+mergify[bot]@users.noreply.github.com>

PR #3782, opened 2020-06-25
fix(provider/cf): adding test to a fix

zachsmith1:
Added a quick test to confirm that the manifest is correctly getting converted to an artifact that will get resolved. Functionality is working as expected so this test should fail in the future if a breaking change is introduced.


kevinawoo:
fyi for others, this was related to adding a test on https://github.com/spinnaker/orca/pull/3753

PR #3785, opened 2020-06-27
fix(resizeServerGroup): Use targetHealthyDeployPercentage when available

vigneshm:
`WaitForCapacityMatchTask` doesn't make use of `targetHealthyDeployPercentage` when available and there has been some reports of users desiring this behavior specifically for cases where they're trying to deploy a large number of instances.

~~Note that this task would [succeed](https://github.com/spinnaker/orca/blob/ae2b8043e538b3ed0db39b61c8da0d2e36ee8a5a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy#L67-L70) if the server group is disabled. I'm not exactly sure if that is acceptable behavior with `targetHealthyDeployPercentage` so wanted to call that out.~~ It wouldn't since [`targetDesiredSize` would not be set](https://github.com/spinnaker/orca/blob/b217bd51aa6896f63c94f6d3536353eef0368b38/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy#L36-L38) under that condition.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy
```diff
@@ -58,8 +58,15 @@ class WaitForCapacityMatchTask extends AbstractInstancesCheckTask {
         desired = ((Map<String, Integer>)serverGroup.capacity).desired
       }
 
-      splainer.add("checking if capacity matches (desired=${desired}, instances.size()=${instances.size()}) ")
-      if (desired != instances.size()) {
+      splainer.add("checking if capacity matches (desired=${desired}, target=${stage.context.targetDesiredSize ?: "none"} instances.size()=${instances.size()}) ")
```
marchello2000, line 61: I am missing something, where is the `targetDesiredSize` computed to take `targetHealthyDeployPercentage` into account?

vigneshm: It is computed [here](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForCapacityMatchTask.groovy#L38)

marchello2000: got it, thanks! I think it looks reasonable. But i would get @dreynaud input as he is the ultimate expert here

dreynaud: I'm a bit rusty on this but the explanation makes sense and I don't see issues 👍

robzienert: nit: `instances.size()` loses meaning when viewed just as a log message: Anyone unfamiliar with the code would have to code dive to figure out what `instances.size()` is referring to - is it the desired or target?

dreynaud: `instances.size()` is meaningful to me, it is the current number of instances. As in, "your desired is X, your target is Y, but you currently have Z"

vigneshm: I can update that to `current` if that makes it easier to read.

PR #3787, opened 2020-06-30
feat(bake): Toggleable pausing for bake stage

luispollo:
Introduces the ability to "pause" bakes through a feature toggle.

Pausing is achieved by the insertion of a simple task at the top of the list which waits until the feature toggle has been cleared. Note that turning on the feature flag will not affect currently running bakes, only those that start after it's been set.

The immediate need for this is to support our migration of the Bakery to ElasticSearch 6.x. It will require temporary "downtime" as we finish migrating the data, deploy the new server groups, and verify everything is OK. During that period, the plan is to use this toggle to pause all new bakes, then allow running bakes to complete, before the deployment. Once the migration is completed, we will "unpause".

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
@@ -17,22 +17,20 @@
 package com.netflix.spinnaker.orca.bakery.pipeline
 
 import com.google.common.base.Joiner
-import com.netflix.spinnaker.kork.exceptions.ConstraintViolationException
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
-import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
-import com.netflix.spinnaker.orca.api.pipeline.graph.StageGraphBuilder
-import com.netflix.spinnaker.orca.pipeline.StageExecutionFactory
-
-import java.time.Clock
-import javax.annotation.Nonnull
-import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
+import com.netflix.spinnaker.kork.exceptions.ConstraintViolationException
 import com.netflix.spinnaker.orca.api.pipeline.Task
 import com.netflix.spinnaker.orca.api.pipeline.TaskResult
+import com.netflix.spinnaker.orca.api.pipeline.graph.StageDefinitionBuilder
+import com.netflix.spinnaker.orca.api.pipeline.graph.StageGraphBuilder
+import com.netflix.spinnaker.orca.api.pipeline.graph.TaskNode
+import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
+import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
 import com.netflix.spinnaker.orca.bakery.tasks.CompletedBakeTask
 import com.netflix.spinnaker.orca.bakery.tasks.CreateBakeTask
 import com.netflix.spinnaker.orca.bakery.tasks.MonitorBakeTask
-import com.netflix.spinnaker.orca.api.pipeline.graph.StageDefinitionBuilder
-import com.netflix.spinnaker.orca.api.pipeline.graph.TaskNode
+import com.netflix.spinnaker.orca.pipeline.StageExecutionFactory
+import com.netflix.spinnaker.orca.pipeline.tasks.ToggleablePauseTask
 import com.netflix.spinnaker.orca.pipeline.tasks.artifacts.BindProducedArtifactsTask
 import com.netflix.spinnaker.orca.pipeline.util.RegionCollector
 import groovy.transform.CompileDynamic
@@ -41,6 +39,8 @@ import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
+import javax.annotation.Nonnull
+import java.time.Clock
 import java.util.stream.Collectors
 
 import static com.netflix.spinnaker.orca.api.pipeline.SyntheticStageOwner.STAGE_BEFORE
@@ -53,12 +53,18 @@ import static java.time.ZoneOffset.UTC
 class BakeStage implements StageDefinitionBuilder {
 
   public static final String PIPELINE_CONFIG_TYPE = "bake"
+  public static final String BAKE_PAUSE_TOGGLE = "stages.bake-stage.pause"
+
+  private RegionCollector regionCollector
+  private Clock clock
+  private DynamicConfigService dynamicConfigService
 
   @Autowired
-  RegionCollector regionCollector
-
-  @Autowired
-  Clock clock = systemUTC()
+  BakeStage(RegionCollector regionCollector, DynamicConfigService dynamicConfigService, Clock clock = systemUTC()) {
+    this.regionCollector = regionCollector
+    this.clock = clock
+    this.dynamicConfigService = dynamicConfigService
+  }
 
   @Override
   void taskGraph(@Nonnull StageExecution stage, @Nonnull TaskNode.Builder builder) {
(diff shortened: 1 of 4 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/ToggleablePauseTask.java
```diff
@@ -0,0 +1,62 @@
...
+        stage.getContext().containsKey("pauseToggle")
+            ? stage.getContext().get("pauseToggle").toString()
+            : null;
+
+    if (pauseToggle != null && dynamicConfigService.isEnabled(pauseToggle, false)) {
+      log.debug(
+          stage.getName()
+              + " stage currently paused based on "
+              + pauseToggle
+              + " toggle. Waiting...");
```
luispollo, line 41: This formatting was spotless, not me... 🤷‍♂️

robzienert: Using the logger string replacement functionality is preferential to string concatenation as you’re doing here. It makes reading and writing the log statement easier in code.

robzienert: Changing the stage name is an interesting way to address this. Executions have a “system notification” concept that could be used instead that (i think) gets displayed as a notice on the executions view if you want to use that instead. Its like a pipeline execution-level blesk.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/ToggleablePauseTask.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Override
+  public long getBackoffPeriod() {
+    return Duration.ofMinutes(1).toMillis();
+  }
+
+  @Override
+  public long getTimeout() {
+    return Long.MAX_VALUE;
```
marchello2000, line 56: i would probably bound this somewhat. Maybe a few hours? 4 (can also make it configurable)?

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -53,19 +53,37 @@ import static java.time.ZoneOffset.UTC
...
     if (isTopLevelStage(stage)) {
       builder
         .withTask("completeParallel", CompleteParallelBakeTask)
     } else {
+      if (dynamicConfigService.isEnabled(BAKE_PAUSE_TOGGLE, false)) {
+        log.info("Baking is currently paused. Adding pause task to ${stage.name} stage.")
+        stage.context.put("pauseToggle", BAKE_PAUSE_TOGGLE)
+        builder.withTask("pauseBake", ToggleablePauseTask)
+
+        if (stage.context.containsKey("stageTimeoutMs")) {
```
marchello2000, line 80: instead of this i would put this warning in `ToggleablePauseTask`, just override the `onTimeout` method.

That way, if we will only see these messages if the bake actually timeout out - this will make things easier to find in logs if we need to do a surgical rescue mission.
There is, of course, the case of the stage being unpaused with, say, 1 minute left on the clock so the bake task itself will fail, which is unfortunate, but again, we can search for that in the logs pretty easily.

I think, ideally, `RunTaskHandler` doesn't account for this pause time in timeout calculations, but that's probably not worth the effort
Ideall

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -53,19 +53,37 @@ import static java.time.ZoneOffset.UTC
...
   @Override
   void taskGraph(@Nonnull StageExecution stage, @Nonnull TaskNode.Builder builder) {
     if (isTopLevelStage(stage)) {
       builder
         .withTask("completeParallel", CompleteParallelBakeTask)
     } else {
+      if (dynamicConfigService.isEnabled(BAKE_PAUSE_TOGGLE, false)) {
+        log.info("Baking is currently paused. Adding pause task to ${stage.name} stage.")
+        stage.context.put("pauseToggle", BAKE_PAUSE_TOGGLE)
+        builder.withTask("pauseBake", ToggleablePauseTask)
```
marchello2000, line 78: nit: `deck` will the task name verbatim, so *maybe* `delayBake` is more helpful text for the user.? up to you

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/ToggleablePauseTask.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  @Nonnull
+  public TaskResult execute(@Nonnull final StageExecution stage) {
+    final String pauseToggle =
+        stage.getContext().containsKey("pauseToggle")
+            ? stage.getContext().get("pauseToggle").toString()
+            : null;
+
+    if (pauseToggle != null && dynamicConfigService.isEnabled(pauseToggle, false)) {
```
marchello2000, line 36: prob worth spitting out a warning if `pauseToggle` is `null`
also *maybe* `pauseToggleKey`?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/ToggleablePauseTask.java
luispollo: Is this in milliseconds?

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
luispollo: Wait, I'm confused... I thought `Task.onTimeout()` would be called if the task's own `getTimeout()` had elapsed, not if the parent stage timed out. Which one is it?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/ToggleablePauseTask.java
luispollo: Also, I think I disagree with this in principle. The whole point of this task is to run indefinitely until the toggle has been turned off. Adding a timeout would defeat the purpose. The task can still be cancelled if someone got tired of waiting.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
marchello2000: The currently executing task will have `onTimeout` [called](https://github.com/spinnaker/orca/blob/master/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt#L115) if the stage times out (due to `stageTimeoutMs`)

luispollo: I see. It checks for _both_ stage and task timeout. Got it.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/ToggleablePauseTask.java
marchello2000: I hear what you're saying. It's just weird to have waits that are unbounded. We could end up with a bunch of executions that are waiting indefinitely and we have to cancel them all manually. While I understand it's not really possible (or easily recoverable) with your current implementation. Someone else who might find this task useful and want to use may not realize it/have a bug.
I am ok not bounding this for the given use case, but the idea that there are no guard rails in the "default" case is uncomfortable to me

luispollo: OK. I'll add a 24 hour timeout as a compromise.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/ToggleablePauseTask.java
```diff
@@ -0,0 +1,77 @@
...
+import java.time.Duration;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * A task that just waits indefinitely based on a feature toggle specified in the `pauseToggleKey`
```
marchello2000, line 16:
```suggestion
 * A task that waits based on a feature toggle specified in the `pauseToggleKey`
```

marchello2000:
looks good!

PR #3788, opened 2020-07-01
fix(traffic guards): avoid server group lookup

cfieber:
On verifyInstanceTermination, short circuit checks for
server group instance state if there are no enabled
traffic guards for the server group

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
@@ -70,6 +70,7 @@ public class TrafficGuard {
       Location location,
       String cloudProvider,
       String operationDescriptor) {
+    Front50Cache front50Cache = new Front50Cache(this);
     // TODO rz - Expose a single `instance server groups` endpoint in clouddriver; returns a
     // map<instanceid, servergroup>.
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -231,7 +260,8 @@ public void verifyTrafficRemoval(
     TargetServerGroup someServerGroup = serverGroupsGoingAway.stream().findAny().get();
     Location location = someServerGroup.getLocation();
 
-    if (!hasDisableLock(someServerGroup.getMoniker(), account, location)) {
+    Front50Cache front50Cache = new Front50Cache(this);
```
marchello2000, line 263: doesn't creating a new cache here defeat some of the cache benefit vs passing it in from line 248?

cfieber: the intent of the cache is it should just live through one public call path through TrafficGuard, so this one is at the start of one of the public verifyTrafficRemoval signatures that eventually calls the same verifyTrafficRemovalInternal as line 248 is calling

There was one case where (for instance termination) the entrypoint from one public method was calling into another public method so that is where I introduced a private method that allows propagating the cache along

PR #3789, opened 2020-07-06
refactor(core): Make task timeout methods return durations instead of longs

marchello2000:
This is a breaking change (of questionable value... so maybe not worth doing at all) especially for plugins

### DIFF
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/RetryableTask.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/RetryableTask.java
@@ -25,21 +25,31 @@ import java.time.Duration;
  */
 @Beta
 public interface RetryableTask extends Task {
-  /** TODO(rz): Use Duration. */
-  long getBackoffPeriod();
+  Duration getBackoffPeriod();
 
-  /** TODO(rz): Use Duration. */
-  long getTimeout();
+  Duration getTimeout();
 
-  default long getDynamicTimeout(StageExecution stage) {
+  default Duration getDynamicTimeout(StageExecution stage) {
     return getTimeout();
   }
 
-  default long getDynamicBackoffPeriod(Duration taskDuration) {
-    return getBackoffPeriod();
+  default Duration getDynamicBackoffPeriod(Duration taskDuration) {
+    Duration backoff = getBackoffPeriod();
+
+    if (backoff == null) {
+      backoff = Duration.ZERO;
+    }
+
+    return backoff;
   }
 
-  default long getDynamicBackoffPeriod(StageExecution stage, Duration taskDuration) {
-    return getDynamicBackoffPeriod(taskDuration);
+  default Duration getDynamicBackoffPeriod(StageExecution stage, Duration taskDuration) {
+    Duration result = getDynamicBackoffPeriod(taskDuration);
+
+    if (result == null) {
+      return Duration.ZERO;
+    }
+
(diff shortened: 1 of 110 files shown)


marchello2000:
@robzienert After I got a few q's around what should i return from "getTimeout" and I saw [this todo](https://github.com/spinnaker/orca/blob/master/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/RetryableTask.java#L28) I started on this refactor; and then halfway through I realized.. maybe it's not a great idea to introduce this breaking change. Anyway. Here it is, feel free to tell me to just close it and forget it

robzienert:
Thanks for starting this rabbit hole. I think it would be a positive change, but it is indeed a rabbit hole:

Serde: Doing the type migration is one thing, but `Duration` is serialized by Jackson differently than a `long` is: `PT30S` vs `30000`. Going forward with this, we'd need to create a new deserializer for these fields so we can continue to load existing executions. I'm unsure whether special attention would be needed for stage configuration. I suspect we'd need to serialize the durations back into longs for the sake of Deck, unless we wanted to refactor how stages are configured as well. I think the configuration aspect is a more difficult problem to solve; we already have a good pattern for JIT serialized type migrations.

Plugins: We've not encountered an API change yet. Since the entire interface is marked `@Beta`, we do have the contractual flexibility of breaking API changes, but the issue of coordinating breaking changes across plugins and the service is is a hard point. In my opinion, the only real path forward would be to introduce new methods that would be mapped to the same backing property:

```java
public interface RetryableTask extends Task {
  /**
   * @deprecated use {@link #getBackoffDuration()}
   */
  @Deprecated
  long getBackoffPeriod();

  /**
   * @deprecated use {@link #getTimeoutDuration()}
   */
  @Deprecated
  long getTimeout();

  Duration getBackoffDuration();

  Duration getBackoffDuration();
```

I suspect we'd need some additional annotations to help with tooling / notifying plugin developers of breaking API changes, like `@DeprecationNotice(scheduledRemoval = "1.30.0")` or something that we could spew compiler warnings if uses are discovered. Only after that scheduled removal version would we actually be able to remove the method(s).

Just riffing here.

robzienert:
All that to say, I don't think we should close it (yet? depends on our appetite for this change), since it pulls on a thread that we need to pull at some point anyway.

marchello2000:
thanks for the thoughts!
specifically re: serialization. I don't think these values are serialized anywhere, are they? they are not in the json and not in the message payload (as far as I can tell) what am I missing? 

I think it might be a useful exercise to see how breaking changes like this should be rolled out for plugins, but you're right in that we should prob go the `@Deprecated` route

robzienert:
Oh. Right you are. I was thinking of stage timeouts. Whew, well, that makes this a lot simpler!

cfieber:
There is still the matter of this being a breaking change in orca-api by changing those return types on RetryableTask, and I think we do have some plugins adding tasks somewhere?

It is probably worth doing this non-breakingly though that would mean the naming of the new methods probably has to suck since the 'good' names are already taken..

Theoretically you could add new methods defaulted to wrap the existing methods in Durations, and update all the consumer code for timeouts to use the duration methods and not have to touch any existing task (and then include all the existing task refactors away from the deprecated methods)

jonsie:
I'm in favor of this.  This is a good opportunity to learn how to roll out a breaking orca-api change eventually.  I think the example code @robzienert added makes sense, and would allow us to move forward for now.

PR #3799, opened 2020-07-08
feat(SpEL): Add applicationMetadata SpEL helper function

marchello2000:
Function to retrieve application metadata (from `front50`) via SpEL


luispollo:
Good to know it's this easy to add helpers! Thanks for Cc'ing me.

PR #3800, opened 2020-07-08
feat(manifests): add property to make failing on unknown manifest variables optional

zachsmith1:
This new flag will allow users to decide if they want their pipelines to fail when evaluating spinnaker manifest expressions/variables. This use case came from OS community members needing a solution that would allow the manifest to be passed on to the cloudprovider so that it could also do variable substitution, on top of spinnakers evaluation. This solution does exactly that. It will allow the pipeline to continue with a warning if it was unable to resolve any of the variables in the manifest. The default value for this new property (manifest.evaluator.failOnUnknownKeys) is true, so that existing behavior isn't affected for users who don't know about it or don't want to change anything. If a user wants to turn off hard failing for this behavior, they would need to set this property to false by doing something similar to this [doc](https://spinnaker.io/reference/halyard/custom/) for the orca service. Two new tests were created to ensure the default and new behavior work as expected.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
@@ -49,6 +49,7 @@ import java.util.stream.StreamSupport;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.SafeConstructor;
@@ -66,17 +67,20 @@ public class ManifestEvaluator implements CloudProviderAware {
   private final ContextParameterProcessor contextParameterProcessor;
   private final OortService oortService;
   private final RetrySupport retrySupport;
+  private final Boolean failOnUnknownKeys;
 
   @Autowired
   public ManifestEvaluator(
       ArtifactUtils artifactUtils,
       ContextParameterProcessor contextParameterProcessor,
       OortService oortService,
-      RetrySupport retrySupport) {
+      RetrySupport retrySupport,
+      @Value("${manifest.evaluator.failOnUnknownKeys:true}") Boolean failOnUnknownKeys) {
     this.artifactUtils = artifactUtils;
     this.contextParameterProcessor = contextParameterProcessor;
     this.oortService = oortService;
     this.retrySupport = retrySupport;
+    this.failOnUnknownKeys = failOnUnknownKeys;
   }
 
   @RequiredArgsConstructor
@@ -195,7 +199,7 @@ public class ManifestEvaluator implements CloudProviderAware {
             contextParameterProcessor.buildExecutionContext(stage),
             /* allowUnknownKeys= */ true);
 
-    if (processorResult.containsKey(PipelineExpressionEvaluator.SUMMARY)) {
+    if (failOnUnknownKeys && processorResult.containsKey(PipelineExpressionEvaluator.SUMMARY)) {
       throw new IllegalStateException(
           String.format(
               "Failure evaluating manifest expressions: %s",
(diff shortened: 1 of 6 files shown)


zachsmith1:
@marchello2000 @kevinawoo 👆

aimeeu:
This will have a related docs PR

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluatorTest.java
```diff
@@ -58,6 +65,9 @@
   private final TypedString manifestString =
       new TypedString("{'metadata': {'name': 'my-manifest'}}");
 
+  private final TypedString spelManifestString =
+      new TypedString("{'metadata': {'name': '${manifest}'}}");
```
kevinawoo, line 69: this is suppose to be JSON right? JSON doesn't accept single quotes 😞

zachsmith1: yep, done!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java
```diff
@@ -66,17 +67,20 @@
...
+  private final Boolean failOnUnknownKeys;
 
   @Autowired
   public ManifestEvaluator(
       ArtifactUtils artifactUtils,
       ContextParameterProcessor contextParameterProcessor,
       OortService oortService,
-      RetrySupport retrySupport) {
+      RetrySupport retrySupport,
+      @Value("${manifest.evaluator.failOnUnknownKeys:true}") Boolean failOnUnknownKeys) {
```
marchello2000, line 78: Two things:
1. As of Spring 2 you need to use the snake case otherwise the props won't load
```suggestion
      @Value("${manifest.evaluator.fail-on-unknown-keys:true}") Boolean failOnUnknownKeys) {
```

2. (you don't need to change this, just for future reference) we prefer the use of `ConfigurationProperties` pattern over `@Value` annotations because a) they don't suffer from these casing issues and b) they are much more easily discoverable and docs for the can be generated automatically

ezimanyi:
@zachsmith1 : I wish I had seen this before it merged. There are already a number of ways to configure whether SpEL evaluation happens for a manifest, and I fear that adding this other way will make things even more confusing.

In particular, we have the field "Skip SpEL evaluation" already in the "Deploy Manifest" context:
image

The use case for that is to handle exactly the case noted above---where users have expressions containing `$` in their manifest and don't want them evaluated by Spinnaker.  If that box is checked, SpEL evaluation is skipped entirely for the fetched manifest.

Is there a use case where users *do* want SpEL evaluation to happen, but want to ignore errors?  In all the use cases we'd discussed when adding the above feature, what users really wanted was to skip SpEL evaluation (and ignoring errors was just a workaround).  So I'd definitely want to understand better what the use case is if we're going to allow this.

Overall, given that this has a long and confusing history that has been discussed in detail at the Kubernetes SIG, I'd really like to have another discussion there before we add yet another parameter/setting to configure this.  If there is consensus that this is a use case we'd like to support, we can then decide with input from the SIG how to support it.  (In particular the other SpEL configuration happens on a per-stage level via checkboxes, so I'd want to understand why this is a global setting instead of matching the others.)

Are you able to make the Kubernetes SIG next Tuesday so we can discuss this in more detail?  (CC @ethanfrogers @maggieneterval ).

zachsmith1:
@ezimanyi (as discussed) and others reading this:

The goal with this PR was to allow manifest evaluation the ability NOT to fail when it can't resolve variables. The use case is because there are variables in the manifest that spinnaker can resolve, and should for that matter. But, when it can't resolve them, it may be because those variables are to be resolved by the cloud provider. In CloudFoundry's case, there are variables in the manifest that cloudfoundry will resolve during deployment. So, there is a use case that would want spinnaker to resolve the variables it knows about and then allow the other variables to remain without failing the stage/pipeline. 

Happy to discuss this further in the Kube SIG, but note we don't intend this solution to solve any k8 issues. Eric mentioned there may be an alternative way of doing this without needing the flag so this can be a potential solution as well.

PR #3804, opened 2020-07-09
Revert "chore(deprecate): Remove `resizeAsg` stage/task (#3786)"

marchello2000:
This reverts commit 8587846bb4226c58f5398fc12786e23492dbc04e.


marchello2000:
This breaks an internal user, need to temporarily revert this change

PR #3811, opened 2020-07-09
feat(orca): swap ManiestEvaluator to use configurationProperties

zachsmith1:
Please see: https://github.com/spinnaker/orca/pull/3800

"This new flag will allow users to decide if they want their pipelines to fail when evaluating spinnaker manifest expressions/variables. This use case came from OS community members needing a solution that would allow the manifest to be passed on to the cloudprovider so that it could also do variable substitution, on top of spinnakers evaluation. This solution does exactly that. It will allow the pipeline to continue with a warning if it was unable to resolve any of the variables in the manifest. The default value for this new property (manifest.evaluator.failOnUnknownKeys) is true, so that existing behavior isn't affected for users who don't know about it or don't want to change anything. If a user wants to turn off hard failing for this behavior, they would need to set this property to false by doing something similar to this doc for the orca service. Two new tests were created to ensure the default and new behavior work as expected."


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/config/ManifestEvaluatorConfigurationProperties.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.tasks.manifest.config;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Component
```
marchello2000, line 24: i think the preferred route is to not make this class a component, but instead, mark the consumer (`ManifestEvaluator`) with `@EnableConfigurationProperties`

zachsmith1: @marchello2000 Sure - we can do it that way. I looked at most of the way CP's are done in Orca today and saw both patterns and went with the former. From what I have seen there isn't any real benefit from doing it one way or the other. Do you know why @EnableConfigurationProperties with a POJO is the preferred?

zachsmith1:
For those reading this, please see the original PR as called out by Eric for continuing discussions.

PR #3812, opened 2020-07-10
fix(sql): PostgreSQL compatibility for executions and partitions

jcavanagh:
Includes the following changes:

- Removing bigint display qualifiers - these are MySQL specific, and display-only
- Avoid using `forceIndex` on PostgreSQL - it does not support that feature
- Remove MySQL-specific column backticks for `partition`, and replace with `DSL.name` throughout
- A number of queries used MySQL-specific upsert features. These have been expanded with PostgreSQL-compatible varieties.
- Add new PostgreSQL module with runtimeOnly driver dependency
- Add PostgreSQL tests for orca-sql

Depends on:
https://github.com/spinnaker/kork/pull/716
https://github.com/spinnaker/keiko/pull/176


orca-sql/src/main/resources/db/changelog/20190911-notification-cluster-lock.yml
```diff
@@ -14,6 +14,6 @@ databaseChangeLog:
                     nullable: false
               - column:
                   name: expiry
-                  type: bigint(13)
+                  type: bigint
```
robzienert, line 17: Won't `validCheckSum` be required here as well?

orca-sql/src/main/resources/db/changelog/20200327-deleted-executions-table.yml
```diff
@@ -2,10 +2,38 @@ databaseChangeLog:
...
+                  nullable: false
+            - column:
+                name: execution_type
+                type: varchar(26)
+                constraints:
+                  nullable: false
         - sql:
             dbms: mysql
-            sql: CREATE TABLE `deleted_executions` (id INT AUTO_INCREMENT, execution_id char(26) NOT NULL, execution_type ENUM("pipeline", "orchestration") NOT NULL, deleted_at DATETIME NOT NULL, CONSTRAINT deleted_executions_pk PRIMARY KEY (id))
+            sql: ALTER TABLE `deleted_executions` MODIFY COLUMN `execution_type` ENUM("pipeline", "orchestration") NOT NULL
```
robzienert, line 36: 👏

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/notifications/SqlNotificationClusterLock.kt
```diff
@@ -52,8 +53,19 @@ class SqlNotificationClusterLock(
       jooq.insertInto(lockTable)
         .set(lockField, notificationType)
         .set(expiryField, now.plusSeconds(lockTimeoutSeconds).toEpochMilli())
-        .onDuplicateKeyIgnore()
-        .execute()
+        .run {
+          when (jooq.dialect()) {
+            SQLDialect.POSTGRES -> {
+              onConflict(DSL.field("id"))
```
nicolasff, line 59: This should be `lock_name`, not `id` which is not a field in this table. `lock_name` is the primary key.

orca-sql/src/main/resources/db/changelog/20190911-notification-cluster-lock.yml
jcavanagh: Indeed it would be - I've added the original checksum to the migration.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/notifications/SqlNotificationClusterLock.kt
jcavanagh: This is fixed, thank you.

jcavanagh:
@robzienert Thanks for the review!  This is still dependent on https://github.com/spinnaker/keiko/pull/176 as well - would it be possible to get eyes on that sometime as well?

marchello2000:
merged keiko, it should produce a release https://github.com/spinnaker/keiko/releases/tag/v3.8.0

jcavanagh:
Thanks @marchello2000 !  I'll wait for the next Kork release, and then this PR should (hopefully) test correctly.

PR #3817, opened 2020-07-13
fix(orca): Push service account permission message to deck

lyzhang1999:
complete TODO - services account permission message.
before add message to deck when save pipeline error it will show:
```
Pipeline save operation did not succeed:xxxxxxx(status: TERMINAL)
```
after add message it will show in deck:
```
User xxxx is not authorized with all roles for pipeline
```


lyzhang1999:
@cfieber please help me review,thanks

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java
```diff
@@ -128,9 +131,19 @@ public TaskResult execute(@Nonnull StageExecution stage) {
     }
 
     if (!isUserAuthorized(user, roles)) {
-      // TODO: Push this to the output result so Deck can show it.
       log.warn("User {} is not authorized with all roles for pipeline", user);
-      return TaskResult.ofStatus(ExecutionStatus.TERMINAL);
+      Map<String, Object> operation = new HashMap<>();
```
ajordens, line 135: I believe we could just throw an exception here ... `throw new UserException(format("User '%s' is not authorized with all roles for pipeline", user))` and let `orca` map it correctly in the context and set the right `TERMINAL` status.

lyzhang1999: I have change the code, thanks for you advice

lyzhang1999:
@ajordens

PR #3822, opened 2020-07-14
feat(manifest): use failOnFailedExpressions for ManifestEvaluator

zachsmith1:
This change is to ensure the stage flag failOnFailedExpressions will work as expected when set. The manifestEvaluator class will now check that this flag is true/false before throwing an error. This is similar to a change that was merged and rolled back in favor of using the stage level flag over the global flag. [Here](https://github.com/spinnaker/orca/pull/3821) is the PR to revert the global flag.


zachsmith1:
@maggieneterval @ezimanyi @kevinawoo @ethanfrogers ^^ based on the discussion from k8 sig

PR #3825, opened 2020-07-17
fix(logging): Wrap all of RunTaskHandler in a `withAuth` for context

marchello2000:
This hoists up the `.withAuth` and `.withLoggingContext` calls such that all logging (specifically those for thrown exceptions)
contains the execution/stage/task ids


marchello2000:
review with ignore whitespace:
https://github.com/spinnaker/orca/pull/3825/files?diff=split&w=1

PR #3830, opened 2020-07-23
fix(preconfigured): Allow PreconfiguredJobStage parameters to map to one-dimensional array property values

jaydee864:
Addresses https://github.com/spinnaker/spinnaker/issues/5889 by checking for property nodes which end in array notation and extracting them from the parent node as a List<Object>, then returning the Object at the desired index.

Notably, this PR does not support 2+ dimensional arrays (eg. "foo[0][1]"). I struggled to come up with a clean and viable solution for handling that scenario, but I'm also not entirely sure it's needed for the object models that Orca is currently working with.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.groovy
```diff
@@ -95,13 +97,38 @@ class PreconfiguredJobStage extends RunJobStage {
...
+      index = Integer.parseInt(parts[1])
+    } catch (NumberFormatException ex) {
+      throw new IllegalArgumentException("Unable to parse index from expresion ${expression}", ex)
+    }
+    List<Object> nextList = root[propName]
+    if (nextList == null) {
+      throw new IllegalArgumentException("no property ${propName} on $root")
+    }
+    if (nextList.size() <= index || index < 0) {
+      throw new IllegalArgumentException("Invalid index $index for list $nextList")
```
marchello2000, line 129: maybe
```suggestion
      throw new IllegalArgumentException("Invalid index $index for list $propName")
```

since that will actually have the list name

jaydee864:
@marchello2000  @cfieber @srekapalli Are there any further changes needed from my side on this PR?

kevinawoo:
@mergifyio backport release-1.19.x release-1.20.x release-1.21.x

PR #3831, opened 2020-07-23
chore(deprecate): Remove `resizeAsg` stage/task

marchello2000:
Reapply changes in #3786

It has been deprecated in September 2015, pretty sure nothing uses it.
If anyone is using it, they should migrate to resizeServerGroup instead


marchello2000:
Already reviewed in #3786 so labeling for merge

PR #3833, opened 2020-07-23
refactor(task): Refactor DetermineRollbackCandidatesTask

lorin:
Refactor the logic in DetermineRollbackCandidatesTask. There is no (intentional) behavior change with this PR.

This is work I did in preparation for fixing a bug in the monitored deployment behavior where automated rollback doesn't behave as expected if the new ASG reaches 100% and the old ASG is disabled before the rollback is attempted.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -105,172 +105,120 @@ public long getTimeout() {
...
+    return buildResult(imagesToRestore, rollbackTypes, rollbackContexts);
+  }
 
-      if (rollbackDetails != null) {
-        logger.info(
-            "Found rollback candidate in cluster {}, region {}: {}",
-            moniker.get().getCluster(),
-            region,
-            rollbackDetails.rollbackContext.get("restoreServerGroupName"));
+  private RollbackDetails findBestCandidate(
```
lorin, line 177: Even though this function doesn't contain much code, I pulled it out because my next PR should hopefully just touch this function without changing anything else.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -105,172 +105,120 @@ public long getTimeout() {
...
-      } catch (Exception e) {
-        logger.warn(
-            "Failed to fetch server group, retrying! (account: {}, region: {}, serverGroup: {})",
-            stageData.credentials,
-            stageData.regions.get(0),
-            stageData.serverGroup,
-            e);
-        return TaskResult.RUNNING;
-      }
+    boolean success =
```
luispollo, line 111: To use a consistent approach to communicate failed returns, maybe return the moniker or `null` here instead of a boolean?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -281,6 +229,171 @@ public TaskResult execute(@Nonnull StageExecution stage) {
...
+  @Nonnull
+  private List<ServerGroup> toServerGroups(Object obj) {
+    List<ServerGroup> serverGroups =
+        objectMapper.convertValue(obj, new TypeReference<List<ServerGroup>>() {});
+    serverGroups.sort(Comparator.comparing((ServerGroup o) -> o.createdTime).reversed());
+    return serverGroups;
+  }
+
+  /** Verify that a rollback is actually possible */
+  private boolean IsRollbackPossible(
```
luispollo, line 251:
```suggestion
  private boolean isRollbackPossible(
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: Yeah, that was my initial implementation, but the downside there is that it's less clear that the monitor object being passed in is being returned.

I also considered throwing an exception (which would be the more Java-ish way), but I didn't think that would improve readability.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -105,172 +105,120 @@ public long getTimeout() {
...
     }
 
-    List<ServerGroup> serverGroups =
-        objectMapper.convertValue(
-            cluster.get("serverGroups"), new TypeReference<List<ServerGroup>>() {});
-    serverGroups.sort(Comparator.comparing((ServerGroup o) -> o.createdTime).reversed());
+    return getResult(stageData, moniker.get().getCluster(), serverGroups);
+  }
+
+  private TaskResult getResult(
```
luispollo, line 127: Consider naming this after what it actually does, something like `determineRollbackCandidates`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -105,172 +105,120 @@ public long getTimeout() {
   @Nonnull
   @Override
   public TaskResult execute(@Nonnull StageExecution stage) {
-    Map<String, String> rollbackTypes = new HashMap<>();
-    Map<String, Map> rollbackContexts = new HashMap<>();
-
     StageData stageData = stage.mapTo(StageData.class);
-    Map<String, Object> cluster;
-
     AtomicReference<Moniker> moniker = new AtomicReference<>(stageData.moniker);
```
lorin, line 109: I'm not sure why this is an `AtomicReference`.  Could multiple threads be accessing this code simultaneously?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: Fixed in d074dd424

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
luispollo: To clarify, I meant you would instantiate the moniker object inside the function instead of populating a pre-created one, since there's nothing done to that object after creation right above, in other words, there's no advantage to creating it outside the function.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -105,172 +105,120 @@ public long getTimeout() {
...
-                .put("region", region)
-                .put("image", rollbackDetails.imageName)
-                .put("rollbackMethod", rollbackDetails.rollbackType.toString());
+    return getBestCandidate(cluster, region, serverGroupToRollBack, candidates, imageDetails);
+  }
 
-        if (rollbackDetails.buildNumber != null) {
-          imageToRestore.put("buildNumber", rollbackDetails.buildNumber);
-        }
+  private ServerGroup getServerGroupToRollBack(List<ServerGroup> enabledServerGroupsInRegion) {
```
luispollo, line 197: Consider annotating the parameter with `@Nonnull` since you're assuming that below.
```suggestion
  private ServerGroup getServerGroupToRollBack(@Nonnull List<ServerGroup> enabledServerGroupsInRegion) {
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -281,6 +229,171 @@ public TaskResult execute(@Nonnull StageExecution stage) {
         .build();
   }
 
+  @Nullable
+  private List<ServerGroup> getServerGroups(
+      AtomicReference<Moniker> moniker, String credentials, String cloudProvider) {
+    return Optional.ofNullable(fetchClusterInfoWithRetry(moniker, credentials, cloudProvider))
+        .map(c -> c.get("serverGroups"))
```
luispollo, line 236:
```suggestion
        .map(clusterInfo -> clusterInfo.get("serverGroups"))
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -105,172 +105,120 @@ public long getTimeout() {
...
+  /** Retrieve the details for the best rollback candidate */
+  @Nonnull
+  private DetermineRollbackCandidatesTask.RollbackDetails getBestCandidate(
+      String cluster,
+      String region,
+      ServerGroup serverGroupToRollBack,
+      List<ServerGroup> candidates,
+      ImageDetails imageDetails) {
+    return Optional.ofNullable(imageDetails)
+        .map(d -> getDetailsUsingEntityTags(candidates, serverGroupToRollBack, d, cluster, region))
```
luispollo, line 210:
```suggestion
        .map(imgDetails -> getDetailsUsingEntityTags(candidates, serverGroupToRollBack, imgDetails, cluster, region))
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: Ohhhh! Yes, that's definitely better. I'll do that.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: @ajordens Do you have any idea why this was originally an atomic reference?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: Fuxed un 0d664470d

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
marchello2000: yeah, this doesn't make sense to me. all this code is single threaded

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -281,6 +234,172 @@ public TaskResult execute(@Nonnull StageExecution stage) {
...
+  }
+
+  /**
+   * Retrieve info about the server group and use it to populate a Moniker object
+   *
+   * @return true on success, false on failure
+   */
+  private AtomicReference<Moniker> populateMonikerWithServerGroupInfo(
+      Moniker m, String credentials, String region, String serverGroupName) {
+    AtomicReference<Moniker> moniker = new AtomicReference<>(m);
```
marchello2000, line 358: like you said, i don't see the point of the `AtomicReference`, but 🤷

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -105,172 +105,125 @@ public long getTimeout() {
...
-            moniker.get().getCluster(),
-            region);
+      if (!isRollbackPossible(allServerGroupsInRegion, cluster, region)) {
         continue;
       }
 
       List<ServerGroup> enabledServerGroupsInRegion =
           allServerGroupsInRegion.stream()
-              .filter(s -> s.disabled == null || !s.disabled)
+              .filter(DetermineRollbackCandidatesTask::isEnabled)
```
marchello2000, line 147: nit: i would rename this to `isServerGroupEnabled` otherwise it feels like we are talking about the task being enabled

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -281,6 +234,172 @@ public TaskResult execute(@Nonnull StageExecution stage) {
...
+        .map(clusterInfo -> clusterInfo.get("serverGroups"))
+        .map(this::toServerGroups)
+        .orElse(null);
+  }
+
+  /** Deserialize a list of server groups. The list is sorted by creation time, newest first */
+  @Nonnull
+  private List<ServerGroup> toServerGroups(Object obj) {
+    List<ServerGroup> serverGroups =
+        objectMapper.convertValue(obj, new TypeReference<List<ServerGroup>>() {});
```
robzienert, line 250: nit: If this is a hot path, it may be worthwhile moving the `TypeReference` object into a static member of the class.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
robzienert: +1 remove.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -281,6 +234,172 @@ public TaskResult execute(@Nonnull StageExecution stage) {
...
+   * @return true on success, false on failure
+   */
+  private AtomicReference<Moniker> populateMonikerWithServerGroupInfo(
+      Moniker m, String credentials, String region, String serverGroupName) {
+    AtomicReference<Moniker> moniker = new AtomicReference<>(m);
+    if (moniker.get() == null && serverGroupName != null) {
+      try {
+        Map<String, Object> serverGroup =
+            retrySupport.retry(
+                () -> fetchServerGroup(credentials, region, serverGroupName), 5, 1000, false);
```
robzienert, line 363: later: `RetrySupport` usage should be removed in favor of `Retry` from resilience4j.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -350,13 +469,13 @@ private static RollbackDetails fetchRollbackDetails(
       throw new IllegalStateException(
           format(
               "Found more than one server group with the same name! (serverGroupName: %s)",
```
marchello2000, line 471: @lorin I know this is not yours, but this error message is misleading. It's not that there are multiple with same name it's that we didn't find a server group to roll back to.
Can you update it please, since you already have this PR out?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: Fixed in 3eeb3c326591ef58dd0f16b7676e11260d6cd600

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: Fixed in b6e3f91b1

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: Fixed in b6e3f91

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: Fixed in 512516fb42c4800ff6f0f9330f72510c097b6247

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: Fixed in 722ebd4d3627444ac91c40c25f80b4de48f01daa

PR #3835, opened 2020-07-24
chore(exceptions): Fix up what exceptions are thrown

marchello2000:
In an effor to clean up logs to improve signal to noise ratio of bug vs non-essential/expected failures:
* clean up what exceptions are thrown from a bunch of places
* anything deriving from UserException won't get logged as a failure
* for now, change capacity short circuit to warn instead of error because it is mostly noise. I will work on making it provide accurate signal separately

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/exceptions/PipelineTemplateValidationException.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/exceptions/PipelineTemplateValidationException.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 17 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/exceptions/PipelineTemplateValidationException.java
```diff
@@ -0,0 +1,64 @@
...
+    super(message, cause);
+    this.errors = errors;
+  }
+
+  public PipelineTemplateValidationException(Throwable cause, Collection errors) {
+    super(cause);
+    this.errors = errors;
+  }
+
+  public PipelineTemplateValidationException(
```
maggieneterval, line 48: Do we need all these constructors?

orca-core/src/main/java/com/netflix/spinnaker/orca/exceptions/PipelineTemplateValidationException.java
```diff
@@ -0,0 +1,64 @@
...
+
+package com.netflix.spinnaker.orca.exceptions;
+
+import com.netflix.spinnaker.kork.exceptions.HasAdditionalAttributes;
+import com.netflix.spinnaker.kork.exceptions.UserException;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+
+public class PipelineTemplateValidationException extends UserException
```
maggieneterval, line 25:
```suggestion
public final class PipelineTemplateValidationException extends UserException
```

orca-core/src/main/java/com/netflix/spinnaker/orca/exceptions/PipelineTemplateValidationException.java
marchello2000: silly copy-paste, thank you! removing a bunch of them

PR #3836, opened 2020-07-24
fix(disable): enforce serverGroup.disabled is true

dreynaud:
Let's simplify the logic in WaitForClusterDisableTask:
1. first make sure that the server group disabled flag is set
2. then delegate to waitForRequiredInstancesDownTask to check the various healths and the more subtle points like whether interestingHealthProviderNames are set or not

Motivation: it was possible to get out of WaitForClusterDisableTask with a server group that had Down instances but `disabled: false`, therefore it would still count as an active server group and would be filtered out by downstream stages like scaleDownServerGroup.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/WaitForClusterDisableTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/WaitForClusterDisableTask.groovy
@@ -23,13 +23,15 @@ import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.support.Targe
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.ServerGroupCreator
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.WaitForRequiredInstancesDownTask
 import com.netflix.spinnaker.orca.clouddriver.utils.CloudProviderAware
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/WaitForClusterDisableTask.groovy
```diff
@@ -68,33 +70,22 @@ class WaitForClusterDisableTask extends AbstractWaitForClusterWideClouddriverTas
...
     if (!serverGroup.isPresent()) {
-      return false
+      return COMPLETED
     }
 
-    // Even if the server group is disabled, we want to make sure instances are down
-    // to prevent downstream stages (e.g. scaleDownCluster) from having to deal with disabled-but-instances-up server groups
+    // make sure that we wait for the disabled flag to be set
     def targetServerGroup = serverGroup.get()
-    if (targetServerGroup.isDisabled() || stage.context.desiredPercentage) {
```
ajordens, line 81: Is there something special here about the partial disables for RRB?

dreynaud: not sure why there used to be, `waitForRequiredInstancesDownTask.hasSucceeded` knows how to deal with partial disables

PR #3838, opened 2020-07-28
chore(spotless): upgrade spinnaker gradle plugin

plumpy:
This includes updates to google-java-format and ktlint, so requires some
formatting changes. These were made with a simple `./gradlew spotlessApply`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,7 +4,7 @@ keikoVersion=3.8.0
 korkVersion=7.56.0
 kotlinVersion=1.3.72
 org.gradle.parallel=true
-spinnakerGradleVersion=8.3.0
+spinnakerGradleVersion=8.4.0
 
 # To enable a composite reference to a project, set the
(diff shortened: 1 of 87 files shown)


plumpy:
@marchello2000 Well there were 17 of these reviews, so I just tried to spread them around somewhat randomly :)

dreynaud:
@plumpy after this change I run into this running `./gradlew clean build`:

```> Task :orca-clouddriver:spotlessJava FAILED
Step 'google-java-format' found problem in 'orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/CloudDriverTaskStatusService.java':
com/google/googlejavaformat/java/JavaFormatterOptions has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0```

dreynaud:
yeah that worked. I bet most of us were still on java 8, so the switchening is happening

PR #3840, opened 2020-07-29
feat(stages): Add orca-remote-stage module with initial implementation in RunJobStage, specifically targeting Titus jobs

jonsie:
To kick off a remote stage, we submit the `RemoteStageExecution` object to a remote target (in this case, specifically targeting Titus run jobs).  This object provides the following information about the stage execution:

- Type:  The type identifier of the stage.
- ID: The stage execution ID.
- Pipeline execution ID: The overall pipeline execution ID.
- Context:  The full context object of the stage execution.
- Upstream stage outputs: The graph of upstream outputs.
- Callback:  A callback object which will include both an HTTP callback and a mechanism for calling back via a queue (AWS SQS implemented here).  The HTTP callback will be a Gate endpoint that will essentially just put the message on the configured queue.  The Gate endpoint is part 2 to this change.

Todo:
- [x] docs on models and interfaces
- [ ] tests


orca-api/src/main/java/com/netflix/spinnaker/orca/api/preconfigured/jobs/PreconfiguredJobStageProperties.java
```diff
@@ -29,6 +31,13 @@
   /** If enabled, a pipeline can be configured to use this configuration for running a job. */
   private boolean enabled = true;
 
+  /**
+   * If enabled, the stage executes as a remote stage and is expected to be compatible with the
+   * Orca remote stage API.  See the orca-remote-stage module for more details.
+   */
+  @Alpha
+  private boolean remoteStage = false;
```
jonsie, line 39: My POC plan is to use the preconfigured job stage.  Eventually, it would be nice to provide a mechanism to provide a preconfigured job configuration in a dynamic way (perhaps as metadata in front50).

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/stage/messages/messages.kt
```diff
@@ -0,0 +1,34 @@
...
+  val pipelineExecutionId: String
+}
+
+data class RemoteStageStatusMessage(
+  override val stageExecutionId: String,
+  override val pipelineExecutionId: String,
+  val status: ExecutionStatus,
+  val outputs: MutableMap<String, Any>,
+  val startTime: Long,
+  val endTime: Long,
```
robzienert, line 16: Let's represent time and durations as rich objects, rather than primitives... i.e. `Instant`.

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/stage/messages/messages.kt
```diff
@@ -0,0 +1,34 @@
...
+data class RemoteStageTaskStatusMessage(
+  override val stageExecutionId: String,
+  override val pipelineExecutionId: String,
+  val name: String = "remoteNoOpTask",
+  val description: String = "The default remote stage no-op task",
+  val status: ExecutionStatus = ExecutionStatus.SKIPPED,
+  val outputs: MutableMap<String, Any> = mutableMapOf(),
+  val startTime: Long = 0,
+  val endTime: Long = 0,
+  val error: String = "none"
```
robzienert, line 33:
```suggestion
  val error: String? = null
```

I quite dislike the idea of an error value that isn't an error.

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/remote/execution/RemoteStageExecution.java
```diff
@@ -0,0 +1,41 @@
...
+  @Nonnull
+  private String pipelineExecutionId;
+
+  /** The stage context provided at stage initialization. */
+  @Nonnull
+  private Map<String, Object> context;
+
+  /** All upstream stage outputs ordered by stage refId. */
+  @Nonnull
+  private Map<String, Map<String, Object>> upstreamStageOutputs;
```
christopherthielen, line 36: I'm wondering if it would be better to hide the details of upstream outputs (from the stage author) and just emulate what we do in StageContext?

https://github.com/spinnaker/orca/blob/ae2b8043e538b3ed0db39b61c8da0d2e36ee8a5a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageContext.java#L57-L67

jonsie: Hm, that's an interesting idea.  So basically just provide the `context` which is the merger of the current stage (in `StageContext` represented by `delegate`) and the `ancestors` stage outputs?  I think the key in the current stage context would have to override a matching key in the ancestors outputs.

christopherthielen: Traverse the ancestors backwards and add all stage context keys to a map: 

```java
      ArrayList<StageExecution> ancestors = new ArrayList<>(stage.ancestors());
      Collections.reverse(ancestors);
      Map<String, Object> result = new HashMap<>();
      ancestors.forEach(stage -> result.putAll(stage.getContext()));
```

... I think

jonsie: Yeah that looks legit-ish to me. I'll try it out - thanks!

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/remote/messages/RemoteStageStatusMessage.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @Override
+  public String getStageExecutionId() {
+    return null;
+  }
+
+  @Override
+  public String getPipelineExecutionId() {
+    return null;
+  }
```
robzienert, line 51: ???

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnRemoteJobStageCompletion.java
```diff
@@ -0,0 +1,23 @@
...
+@Alpha
+public class WaitOnRemoteJobStageCompletion extends WaitOnJobCompletion {
+
+  @Nonnull
+  @Override
+  public TaskResult execute(@Nonnull StageExecution stage) {
+    return TaskResult
+        .builder(stage.getStatus())
+        .context(stage.getContext())
+        .outputs(stage.getOutputs())
```
robzienert, line 20: I'm confused why you're doing this. Code comment would be nice for sure, but I don't think we want to be copying context and output of a stage down to a task - that's backwards.

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/remote/execution/RemoteStageExecution.java
ajordens: Curious -- this could be quite large, any restrictions on actually getting it into the target runtime?

In some future time, it'd be neat to allow for a stage to define _what_ data it'd like passed to it as a more strongly defined contract. If that data's not available, than we fail.

orca-clouddriver-provider-titus/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/TitusRunJobStageDecorator.java
```diff
@@ -40,4 +60,25 @@ public void modifyDestroyJobContext(
...
+        .getOrDefault("remoteStage", "false")
+        .toString()
+        .equalsIgnoreCase("true");
+
+    if (isRemoteStage) {
+      RemoteStageExecution remoteStageExecution = remoteStageExecutionProvider.get(stageExecution);
+
+      try {
+        TitusPreconfiguredJobProperties.Cluster titusCluster = objectMapper.convertValue(stageExecution.getContext().get("cluster"), TitusPreconfiguredJobProperties.Cluster.class);
+        titusCluster.getEnv().put("REMOTE_STAGE_EXECUTION", Base64.getEncoder().encodeToString(objectMapper.writeValueAsBytes(remoteStageExecution)));
```
ajordens, line 77: To the above point ... any know limitations on this? Will this also result in the current stage context getting blown up?

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/remote/execution/RemoteStageExecution.java
jonsie: We could do that instead, and then just provide a mechanism for querying for upstream stage outputs.  Maybe that's better?

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/stage/messages/RemoteStageStatusMessageHandler.kt
```diff
@@ -0,0 +1,44 @@
...
+
+  override fun handle(message: RemoteStageStatusMessage) {
+    log.info("Remote stage execution '{}' status received with '{}'", message.stageExecutionId, message.status.toString())
+    val pipelineExecution = executionRepository.retrieve(ExecutionType.PIPELINE, message.pipelineExecutionId)
+    val stageExecution = pipelineExecution.stages.find { it.id == message.stageExecutionId }
+
+    if (stageExecution != null) {
+      stageExecution.status = message.status
+      stageExecution.outputs = message.outputs
+      stageExecution.context["remoteTasks"] = message.tasks
```
robzienert, line 30: I would've expected that tasks received from a message would be applied directly to the stage. If left like this, then users would have to open the execution source from the UI to see what was actually going on.

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/stage/messages/RemoteStageStatusMessageHandler.kt
```diff
@@ -0,0 +1,44 @@
...
+      stageExecution.context["remoteStartTime"] = message.startTime
+      stageExecution.context["remoteEndTime"] = message.endTime
+      stageExecution.context["remoteError"] = message.error
```
robzienert, line 33: Similarly here, I'm confused why an error in the remote stage would not impact the stage inside of Orca?

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/remote/execution/RemoteStageExecution.java
robzienert: > provide a mechanism for querying for upstream stage outputs

Meaning, a remote stage would say, "I need context on X, Y, Z" and then the query mechanism would satisfy this and pass the results into the remote stage inputs, or were you thinking something else?

ajordens:
> One reason why you'd want to not merge contexts, is if you have a pipeline like this

That is a current problem right now I think. We may have some specialized support for looking certain types of data upstream that navigates the graph as well as jumping into previously forked off executions.

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/remote/execution/RemoteStageExecution.java
christopherthielen: > Which bake stage wins?

What would currently happen for in-process stages? 

---

In general, I think we should minimize what we pass to the stage.  If they need to do fancy things like query the execution graph, then maybe an escape hatch would be to fetch that execution from gate (within the remote stage).

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/remote/messages/RemoteStageStatusMessage.java
jonsie: Ha, yea cleaning this up.

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/remote/execution/RemoteStageCallback.java
```diff
@@ -0,0 +1,59 @@
...
+@Data
+@Builder
+public class RemoteStageCallback {
+
+  /** {@link Http} callback configuration. */
+  @Nonnull private Http http;
+
+  /** {@link Pubsub} callback configuration. */
+  @Nonnull
+  private Pubsub pubsub;
```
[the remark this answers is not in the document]
jonsie: I was actually thinking of removing the Pubsub callback here and just supporting HTTP for the time being.  I'd rather do all the authz and message validation in Gate before these messages hit Orca.

orca-clouddriver-provider-titus/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/TitusRunJobStageDecorator.java
jonsie: Yeah, 16k is the total space available for declared env variables to be injected (32kb is the total overall size for env variables).

I think there's probably some value in limiting what we pass to the remote stage, and then providing some mechanism for querying for upstream stage output.

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/stage/messages/RemoteStageStatusMessageHandler.kt
ajordens: Don't we have fairly standardized handling for errors, is there a need to go off the path here?

If we go this messages on queue route, `outputs` could be quite large. Are messages intended to contain context or just pointers to context?

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/remote/execution/RemoteStageExecution.java
ajordens: @robzienert That is basically what I was thinking ... perhaps with some formalization around certain types of data we already know about (deployment details, etc.).

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/config/CallbackConfigurationProperties.kt
```diff
@@ -0,0 +1,19 @@
...
+
+  inner class HttpConfigurationProperties {
+    var url: String = "configure-me"
+  }
+
+  inner class PubsubConfigurationProperties {
+    var id: String = "configure-me"
+    var provider: String = "configure-me"
+    var providerConfig: Map<String, String> = mapOf()
+  }
```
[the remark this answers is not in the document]
jonsie: Oh, yea I am actually planning on removing the pubsub callback since I think we will need some message validation and authz prior to hitting Orca and I think doing that in Gate is the best option for now.

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/stage/messages/RemoteStageStatusMessageHandler.kt
jonsie: Applied to the `stageExecution.tasks`?  I think that runs the risk of queuing the task, and in this case we don't want to do that.  This is more similar to a kato task than an internal Orca task.  I was thinking if we standardize around a key in the context (`remote.tasks`/`remoteTasks` for example) then we can do some additional UI work in Deck to display these tasks to the end user.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnRemoteJobStageCompletion.java
jonsie: Yea this approach was very naive the more I learn about Orca message handling - the stage status could be overridden by the last task status pretty easily.  I'm also going to leave context and outputs out of this monitor task regardless of the solution I settle on.

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/stage/messages/RemoteStageStatusMessageHandler.kt
ajordens: I see this as just arbitrary context that could be shown back to a user ... `tasks` seems very specific and overloaded. 

I think there's value in having some sort of a contract around what this is (ie. is it a `List<String>` or something else?)

orca-clouddriver-provider-titus/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/TitusRunJobStageDecorator.java
ajordens: Definitely some value in limiting (may also vary across runtimes).

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/stage/messages/RemoteStageStatusMessageHandler.kt
ajordens: Also question whether we should align the way these updates get applied with [TaskController. updatePipelineStage()](https://github.com/spinnaker/orca/blob/master/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy#L474).

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/stage/messages/RemoteStageStatusMessageHandler.kt
robzienert: I don't think `List<String>` will be good enough - these remote stages are likely going to want to have custom deck UI as well, and will need to pass information that the UI can use to render things as needed.

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/remote/execution/RemoteStageExecution.java
marchello2000: one more in favor of passing in concrete type (and not allowing the stage to look stuff up) if we "leak" the "lookup something from some random stage" concept into plugins it will be that much harder for us to refactor this painful aspect of orca later.

jonsie:
Thanks for the feedback here everyone!  I'm going to close this PR since I decided to approach the problem differently (by building out some fundamental remote extension configuration components first).  However, when I get back to a remote Orca stage this feedback is going to be super valuable - so it was not a waste.  Thanks again!

PR #3841, opened 2020-08-04
feat(pipelines): Pass 'staleCheck' through from savePipeline context to front50

christopherthielen:


christopherthielen:
need to update tests...

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
```diff
@@ -77,6 +77,8 @@ public TaskResult execute(StageExecution stage) {
     final Boolean isSavingMultiplePipelines =
         (Boolean)
             Optional.ofNullable(stage.getContext().get("isSavingMultiplePipelines")).orElse(false);
+    final Boolean staleCheck =
+        (Boolean) Optional.ofNullable(stage.getContext().get("staleCheck")).orElse(false);
```
marchello2000, line 81: this is sprinkled here across the whole task (and orca...) but really this should be 
```suggestion
        (Boolean) Optional.ofNullable(stage.getContext().getCurrentOnly("staleCheck")).orElse(false);
```

otherwise if there is an evalVars upstream that has a variable `stageCheck` we might end up using that if none is present on this stage context

christopherthielen: hmm, in StageExecutionImpl:

```
  /** The context driving this stage. Provides inputs necessary to component steps */
  private Map<String, Object> context = new StageContext(this);
```
... feels pretty nasty to cast this to a `StageContext`
Since StageContext implements Map, would it be reasonable to change `StageExecution` to return a StageContext?

```
  /** TODO(rz): Try to use StageContext instead? */
  @Nonnull
  Map<String, Object> getContext();
```

christopherthielen: welp, `StageContext` isn't available in `orca-api` so we'd have to extract an interface for it to do ☝️

christopherthielen: Looks like we *are* casting in one spot already, the pipeline stage https://github.com/spinnaker/orca/blob/ae2b8043e538b3ed0db39b61c8da0d2e36ee8a5a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java#L72

christopherthielen: after discussing with mark offline, we're leaving it as-is

PR #3843, opened 2020-08-06
fix(exceptions): Don't let errors in `trackResult` leak out

marchello2000:
It's possible for trackResult metrics call to fail and we don't want that failure to leak out and fail orca tasks

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/clouddriver/utils/CloudProviderAware.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/clouddriver/utils/CloudProviderAware.java
@@ -58,9 +58,12 @@ public interface CloudProviderAware {
 
   // may return a list with 0, 1 or more regions (no guarantees on the ordering)
   default List<String> getRegions(Map<String, Object> context) {
-    String region = (String) context.getOrDefault("region", null);
+    Object region = context.getOrDefault("region", null);
     if (region != null) {
-      return ImmutableList.of(region);
+      if (region instanceof List) {
+        return ImmutableList.copyOf((List<String>) region);
+      }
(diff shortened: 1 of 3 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/clouddriver/utils/CloudProviderAware.java
```diff
@@ -58,9 +58,12 @@ default String getCredentials(Map<String, Object> context) {
 
   // may return a list with 0, 1 or more regions (no guarantees on the ordering)
   default List<String> getRegions(Map<String, Object> context) {
-    String region = (String) context.getOrDefault("region", null);
+    Object region = context.getOrDefault("region", null);
```
marchello2000, line 61: this sometimes is a string and sometimes a list. I tried to make it handle both. but also wrapped in `try/catch` in the  other files

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -169,7 +170,7 @@ class RunTaskHandler(
             } else if (e is TimeoutException && stage.context["markSuccessfulOnTimeout"] == true) {
               queue.push(CompleteTask(message, SUCCEEDED))
             } else {
-              if (e !is TimeoutException) {
+              if (e !is TimeoutException && e !is UserException) {
```
marchello2000, line 173: oh, this is actually from a related but separate change. but might as well go together. Basically, we shouldn't log errors for exceptiosn that are not in "our" control. E.g. bad SpEL expression and any other `UserExceptions`

PR #3845, opened 2020-08-06
fix(exceptions): add exception handler for exceptional situations

marchello2000:
When a user submits a bad pipeline e.g.
* with jenkins trigger that has an invalid properties file or job that doesn't exist)
* with a broken pipeline template

we don't want to 500 because:
a) the calling service (echo in most cases here) will keep retrying which makes no sense
b) it's an erroneous HTTP 500 metric that is noise

### DIFF
--- a/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
+++ b/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
@@ -23,8 +23,7 @@ import com.netflix.spinnaker.fiat.shared.FiatService
 import com.netflix.spinnaker.fiat.shared.FiatStatus
 import com.netflix.spinnaker.kork.exceptions.ConfigurationException
 import com.netflix.spinnaker.kork.exceptions.SpinnakerException
-import com.netflix.spinnaker.kork.web.exceptions.InvalidRequestException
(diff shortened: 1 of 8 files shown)


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -478,4 +480,11 @@ class OperationsController {
       pipeline.origin = AuthenticatedRequest.spinnakerUserOrigin.orElse('unknown')
     }
   }
+
+  @ExceptionHandler(UserException.class)
+  static void handleUserExceptions(UserException e, HttpServletResponse response) {
```
cfieber, line 485: do our generic error handlers in kork not handle UserException as a 400 already?

I think if they do, they will serve a json payload with some details instead of a text response - might play slightly nicer with handling API responses if they come back that way

marchello2000: oh... i didn't see it let me look/try, thanks!

marchello2000: doesn't look like it, but i can add one to kork, that will obviate the need for them here

marchello2000:
https://github.com/spinnaker/kork/pull/737/files will update this PR

PR #3848, opened 2020-08-06
fix(destroy): add disable stage to destroy server group stage

dreynaud:
Let's add it as a beforeStage instead of duplicating the tasks in the disable stage.

In particular, the task graph of DestroyServerGroupStage was missing a determineHealthProviderTask
and using the now-deprecated WaitForAllInstancesNotUpTask.


ajordens:
Does this all wire up nice in the UI?

christopherthielen:
> Does this all wire up nice in the UI?

yeah it should be good

PR #3851, opened 2020-08-07
Default values for template vars not defined in pipeline

jmalvarezf-lmes:
Hi,

This PR basically addresses the issue that I left a month ago: https://github.com/spinnaker/spinnaker/issues/5842

The idea here is that when you create a template pipeline (in V2), you can define template variables, and you define a default value for them. That default value is used if you create the pipeline from the UI (only the first time), but if you use directly spin, or update your template with new variables, the default value is never used, and your pipeline will fail if a value is not defined for that variable. This change allows you to get the default value from the template if the variable is not defined in your template.

The PR is small, but it is very useful for us, as we are constantly evolving templates and it is very difficult to maintain those templates if we have to modify the pipelines that use them all every time we add a variable to address a new feature.

Hope it is useful!

Regards,


robzienert:
Test asserting this behavior would be good.

jmalvarezf-lmes:
@robzienert you are completely right. I've added a test to assert the behaviour.

Thanks!

robzienert (approved):

PR #3857, opened 2020-08-11
feat(cleanup agents): Add ability for special treatment of apps for orchestration cleanup

marchello2000:
Similar to https://github.com/spinnaker/orca/pull/3612 but for orchestrations

Add the ability to treat certain applications differently from orchestration clean up perspective.
Sometimes (e.g. when an application produces a lot of tasks via API), it is useful to have different
orchestration clean up threshold for certain applications.
This change adds setting to the `TopApplicationExecutionCleanupAgent` to, optionally, specify custom policy per application.

The config looks like this:

```yaml
pollers:
  topApplicationExecutionCleanup:
    enabled: true
    exceptionApplicationThresholds:
      application1: 1000
      application2: 30000
```


ajordens:
I'll take a better read come morning but at a glance this looks good. 

Minor comment as to whether we could do with some per-application flexibility?

```
exceptionApplicationThresholds:
  - application1: 10000
  - application2: 5000
```

I don't have strong feelings and nothing in this config precludes a change in the future (I suppose).

marchello2000:
@ajordens good idea, i wanted to originally model on the work I did for oldpipelinecleanupagent, but it think your approach make more sense for this case. Updated

orca-sql/src/test/groovy/com/netflix/spinnaker/orca/sql/cleanup/TopApplicationExecutionCleanupPollingNotificationAgentSpec.groovy
```diff
@@ -0,0 +1,161 @@
...
+import spock.lang.Shared
+import spock.lang.Specification
+
+import java.time.Instant
+
+import static com.netflix.spinnaker.kork.sql.test.SqlTestUtil.*
+import static com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType.ORCHESTRATION
+import static java.time.temporal.ChronoUnit.DAYS
+
+abstract class TopApplicationExecutionCleanupPollingNotificationAgentSpec extends Specification {
```
robzienert, line 43: It doesn't look like this is used anywhere?

PR #3858, opened 2020-08-11
feat(destroy): Add a flag to be able to skip the disable of destroy stage

marchello2000:
Add an experimental flag `forceSkipDisable` to the `DestroyServerGroupStage` so that applications that don't participate in discovery can skip disable step which should speed up their destruction operation.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DestroyServerGroupStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DestroyServerGroupStage.groovy
@@ -25,11 +25,15 @@ import com.netflix.spinnaker.orca.clouddriver.tasks.MonitorKatoTask
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.DestroyServerGroupTask
 import com.netflix.spinnaker.orca.clouddriver.tasks.servergroup.WaitForDestroyedServerGroupTask
 import com.netflix.spinnaker.orca.pipeline.graph.StageGraphBuilderImpl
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DestroyServerGroupStage.groovy
```diff
@@ -40,10 +40,14 @@ class DestroyServerGroupStage extends TargetServerGroupLinearStageSupport implem
   }
 
   private static void addDisableStage(Map<String, Object> context, StageGraphBuilderImpl graph) {
-    graph.add {
-      it.name = "disableServerGroup"
-      it.type = getType(DisableServerGroupStage)
-      it.context.putAll(context)
+    boolean skipDisable = (boolean)context.getOrDefault("forceSkipDisable", false)
```
ajordens, line 43: nit ... maybe `shouldDisableBeforeDestroy` (default to `true`) vs `forceSkipDisable` (default to `false`).

Otherwise I think this is a useful flag!

marchello2000: I prefer props that aren't there to be `false`. But in this very limited (from code perspective) use case it feels less consequential. I also thought having more strong language e.g. `force` would be good since most people shouldn't be setting the flag so when it's set, hopefully it stands out a bit

ajordens: Agree with the default-to-false idea, makes sense to me.

PR #3859, opened 2020-08-12
fix(artifact): include subpath in location field only for git artifact

edgarulg:
**Spinnaker Version**
1.19.x

**Issue Description**
When we have multiple expected artifacts in a parent pipeline configuration with type `git/repo` using the same `reference` and `version` but with a different `subPath`, `ArtifactResolver` throws an `InvalidRequestException` because they are more than 1 artifact matches

The problem is that an `ExpectedArtifact` only uses `type`, `name`, `version`, `location` and `reference` to validate if an `Artifact` match another `Artifact`, the subpath field in a `git/repo` artifact is added as metadata and is not validated.

Here is the docs for `git/repo` artifact: https://spinnaker.io/reference/artifacts-with-artifactsrewrite/types/git-repo/

**How to reproduce it**
1. Create a pipeline with name "child", add a wait stage and finally in the configuration include an expected artifact as the following:
```
"expectedArtifacts": [
    {
      "defaultArtifact": {
        "artifactAccount": "test-account",
        "id": "171ca875-ef04-4b8a-8ac5-c6308003d3b8",
        "metadata": {
          "subPath": "files"
        },
        "reference": "https://github.com/edgarulg/spinnaker-operator-files",
        "type": "git/repo",
        "version": "master"
      },
      "displayName": "wonderful-lizard-67",
      "id": "16fdc428-f193-4721-b94e-a874bbcb37c4",
      "matchArtifact": {
        "artifactAccount": "test-account",
        "id": "7635f7cf-80e1-4f97-8495-6be4c17f4586",
        "metadata": {
          "subPath": "files"
        },
        "reference": "https://github.com/edgarulg/spinnaker-operator-files",
        "type": "git/repo",
        "version": "master"
      },
      "useDefaultArtifact": true,
      "usePriorArtifact": false
    }
  ]
```
2. Create a pipeline with name "parent", add a pipeline stage and execute the "child" pipeline, finally in the configuration include two expected artifacts as the expected artifact of the child pipeline but only change the subpath to a different thing.
3. Execute the parent pipeline.

**About the fix**
The fix includes the `subpath` value in the `location` field of a `git/repo` artifact, by default, the location field is evaluated to match an artifact. The `location` field is not used by clouddriver when fetch `git/repo` artifacts.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
@@ -202,6 +202,7 @@ public class ArtifactUtils {
             .map(Collection::stream)
             .orElse(Stream.empty())
             .map(it -> objectMapper.convertValue(it, ExpectedArtifact.class))
+            .map(this::updateExpectedArtifacts)
             .collect(toImmutableList());
 
     ImmutableSet<Artifact> receivedArtifacts =
@@ -213,6 +214,7 @@ public class ArtifactUtils {
                     .map(Collection::stream)
                     .orElse(Stream.empty()))
             .map(it -> objectMapper.convertValue(it, Artifact.class))
+            .map(this::updateLocationOnGitArtifact)
             .collect(toImmutableSet());
 
     ArtifactResolver.ResolveResult resolveResult =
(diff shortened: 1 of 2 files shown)


ezimanyi:
This still feels to me like we're solving the problem at the wrong level of abstraction.  Just as the generic `Artifact` class shouldn't have logic that depends on the implementation of a particular artifact type, I don't think that the general artifact resolution logic that runs for every pipeline should have this extra logic that only applies to artifacts of type `git/repo`.  The idea of an `Artifact` is that it's an abstraction and matching logic should not need to switch on details of the particular artifact.

Ideally there would be some change we could make to only `git/repo` artifacts that allows them to fit into the model, while fixing this bug.  Where are these artifacts being created?  Could we change the source of the artifacts to put the subpath in the `location` field so that we don't need this type-specific logic in the general artifact resolver?

Is the artifact coming from people's configuration in the UI?  If so, perhaps we should change it to put the sub-path in the `location` field directly rather than in the `metadata`.  This might require people encountering the bug to update their expected artifacts, but that might not be a bad thing as we *will* be changing the behavior if we start using that field to match.

edgarulg:
@ezimanyi yes, these artifacts are created in the UI. I think this could be fixed by adding the subpath in the location field on deck but I still need to let the subpath in the metadata because it's used by clouddriver here https://github.com/spinnaker/clouddriver/blob/66d2a48af5fa0ef1baaa8ee3904d91119ff1f13b/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java#L100

ezimanyi:
Could we change that code in clouddriver to use the `location` field...or even to try `location` but fall back to `metadata.subPath` if `location` is not provided?  In that case we'd be changing code that only affects `GitRepoArtifactCredentials` which would avoid general artifact-matching logic need to know about this.

So I guess then the tentative solution would be:
* Have the UI start putting the `subPath` in the `location` field.  (We'd need to think about how to make this backwards-compatible; if people already have it in the `metadata` field should we switch it for them on load?  We should definitely think about how to handle this case.)
* Update the `GitRepoArtifactCredentials` code to use either `location` or `metadata.subpath` (the latter for backwards compatibility) when setting the subpath.

I guess I also want to quickly set up this pipeline myself to make sure I understand the use case.  I am still wondering if using default artifacts and avoiding the need to match at all would be the better solution, but I am probably not fully understanding as I have not actually set up the pipeline.

edgarulg:
I will start taking a look at Deck to see how I can put the subpath in the location with backwards- compatible. Also I attached the pipelines that i am using to reproduce the issue.
[pipelines.zip](https://github.com/spinnaker/orca/files/5071038/pipelines.zip)

ezimanyi:
Thanks for the pipelines!  I just loaded them into my Spinnaker, and thought about this for a bit.  Based on that, I do think that the proposed solution of updating `git/repo` artifacts to use the `location` field for a `subpath` is the right fix.

The reason is that ultimately `subpath` is necessary to fully identify the data that the artifact represents; two artifacts with the same git repo but different subpaths will load completely different data.  (And for users that have a number of different things in the same repo, it may be common to need two paths from the same repo in a single pipeline.)

As it's important to be able to distinguish these two artifacts (as they represent completely different things) we should be using a top-level field instead of metadata.  Thinking about it more that field seems appropriate because we are distinguishing which location within the repo (though admittedly we don't have very strong or consistent definitions about what artifact types put in these fields).

As a workaround until your fix is merged, moving to use inline artifacts might still work.  But after looking at your example and thinking about this some more, I do think that we'll want to make a fix to handle this generally.

Let's continue to discuss next week the clouddriver/deck changes that would move `subPath` to be in the `location` field.

edgarulg:
Hi @ezimanyi I opened three PRs to move `subpath` to be in the `location` field for `git/repo` artifacts:
https://github.com/spinnaker/clouddriver/pull/4815
https://github.com/spinnaker/deck/pull/8507
https://github.com/spinnaker/front50/pull/930

I am going to close this PR in Orca.

PR #3867, opened 2020-08-20
refactor(kubernetes): minimize getDynamicManifest contract

maggieneterval:
* refactor(kubernetes): minimize getDynamicManifest contract 

  The only client of Clouddriver's dynamic manifest endpoint is Orca's ResolveDynamicManifestTask, which immediately discards the entire Manifest except for its name.

  Adds a ManifestCoordinates model, whose analog ezimanyi recently added to Clouddriver. As we continue to stop passing entire Kubernetes manifests among Orca tasks and stages and between Orca tasks and Clouddriver, we will likely re-use this soon.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/model/ManifestCoordinates.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/model/ManifestCoordinates.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2020 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
+import lombok.Builder;
+import lombok.Value;
+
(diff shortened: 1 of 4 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/model/ManifestCoordinates.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.Builder;
+import lombok.Value;
+
+@JsonDeserialize(builder = ManifestCoordinates.ManifestCoordinatesBuilder.class)
+@NonnullByDefault
+@Value
+public class ManifestCoordinates {
+  String kind;
+  String namespace;
+  String name;
```
ezimanyi, line 33: This is really just a personal preference, but I like to write out the `private final` even though `@Value` already does that for me (even though by default IntelliJ complains if you do, an inspection I've turned off).  I think it's probably because I'm not super comfortable with an annotation processor changing attributes on things I've written.  (It's already a little strange that Lombok injects stuff into your class, though I guess I've sort of gotten used to that...but actually changing the attributes of things I wrote is another level.)

I think also partly it's because this might get used as an example somewhere else and I'd like as many explicit `private final` things in our code to serve as good examples.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/model/ManifestCoordinates.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Builder(toBuilder = true)
+  private ManifestCoordinates(String kind, String namespace, String name) {
+    this.kind = kind;
+    this.namespace = namespace;
+    this.name = name;
+  }
```
ezimanyi, line 40: In general you'll want to annotated constructors called by a `Builder` as `@ParametersAreNullableByDefault` because there's not really a way of controlling whether all parameters get passed.  (As you could just do `ManifestCoordinates.builder().build()` which would implicitly call this constructor with `(null, null, null)`.)

That would require handling things that are null here...either `with an Objects.requireNonNull` if we want to NPE in that case (maybe for `kind` we do?) or defaulting to an appropriate fallback.

ezimanyi: Oh, and I guess to be more clear in case that was confusing...things that are checked with `Objects.requireNonNull` in the constructor can (and should) stay `@Nonnull` as the parameter so if we decide to do that for all the fields, we can leave the annotation as is.  The issue is that in that case we should eagerly throw the NPE at object construction rather than let the null object end up in a field annotated as `@Nonnull` (and eventually given someone an NPE when they call the accessor and use the result).

PR #3874, opened 2020-08-25
fix(monitored-deploy): roll back to original server group

lorin:
## Problem

In some cases, when a stage in a monitored deploy pipeline execution fails and attempts to automatically
roll back, the rollback stage fails. The associated error reported by the rollback stage is that Spinnaker cannot find a valid server group to roll back to. However, the original server group still exists.

## Solution requirements

If the original server group still exists, a rollback triggered by a monitored deployment should roll back to the original server group.

## Implemented solution

The *MonitoredDeployStrategy* class has been modified to add the name of the original server group to the context of the *RollbackClusterStage* in the pipeline.

The *DetermineRollbackCandidatesTask* class, which executes inside of a *RollbackClusterStage*, now determines which server group to roll back to by retrieving the original server group from its stage's context.  If the task fails to retrieve this information, it falls back to the previously implemented behavior.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
@@ -31,6 +31,7 @@ import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus;
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
 import com.netflix.spinnaker.orca.clouddriver.FeaturesService;
 import com.netflix.spinnaker.orca.clouddriver.OortService;
+import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.CreateServerGroupStage;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.RollbackServerGroupStage;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.rollback.PreviousImageRollbackSupport;
 import com.netflix.spinnaker.orca.clouddriver.tasks.AbstractCloudProviderAwareTask;
@@ -220,7 +221,8 @@ public class DetermineRollbackCandidatesTask extends AbstractCloudProviderAwareT
 
   /** Return the name of the original server group we should roll back to */
   private Optional<String> getOriginalServerGroup(StageExecution stage) {
-    return Optional.ofNullable(stage.findAncestor(isType("createServerGroup")))
+    return Optional.ofNullable(
+            stage.findAncestor(isType(CreateServerGroupStage.PIPELINE_CONFIG_TYPE)))
         .map(StageExecution::getContext)
         .map(this::getSourceServerGroup);
   }
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
@@ -31,7 +31,6 @@ import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus;
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
 import com.netflix.spinnaker.orca.clouddriver.FeaturesService;
 import com.netflix.spinnaker.orca.clouddriver.OortService;
-import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.CreateServerGroupStage;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.RollbackServerGroupStage;
 import com.netflix.spinnaker.orca.clouddriver.pipeline.servergroup.rollback.PreviousImageRollbackSupport;
 import com.netflix.spinnaker.orca.clouddriver.tasks.AbstractCloudProviderAwareTask;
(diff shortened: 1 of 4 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -194,7 +196,50 @@ private RollbackDetails findBestCandidate(
...
+   * Given context info from a createServerGroup stage, extract the field:
+   *
+   * <p>{"source": {"serverGroupName": "`ASG-NAME-HERE`" }}
+   */
+  @Nullable
+  private String getSourceServerGroup(Map<String, Object> contextFromCreateServerGroupStage) {
+    Object source = contextFromCreateServerGroupStage.get("source");
+    if (source instanceof Map) {
+      Object serverGroupName = ((Map<?, ?>) source).get("serverGroupName");
+      if (serverGroupName instanceof String) {
```
luispollo, line 238: Check for `null`? I guess a simpler way to do this, since you're already calling `toString()` on the value below, would be to just replace this with:
```suggestion
      if (serverGroupName != null) {
```

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTaskSpec.groovy
```diff
@@ -55,6 +55,71 @@ class DetermineRollbackCandidatesTaskSpec extends Specification {
     ]
   }
 
+  def "should EXPLICIT-ly roll back to original ASG when parent createServerGroup has source info"() {
+
+    given: "a parent create server group stage in the pipeline"
```
luispollo, line 60:
```suggestion
    given: "an ancestor create server group stage in the pipeline"
```

luispollo:
> I'm also not certain whether:
> 
> * createServerGroup is always a parent stage of the rollbackCluster stage (which seems to be the stage associated with the DetermineRollbackCandidatesTask task).
> * there is always only one createServerGroup parent stage of the rollbackCluster stage (is it possible that this code will pick up the wrong ASG)?

I can't really comment on those, sorry... Other than that, PR looks good to me, especially the test.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: The problem with just checking for null is that if this points to something that isn't a string (e.g., an object,list,number), then this call will succeed, but the resulting string almost certainly won't be a valid server group name. I'd prefer to return `null` in that case.

Now that I'm writing this, I think that I should add a log statement upon returning null here, since we always expect the "source" field to be present.

lorin: Added a log statement in 878bee228

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -194,7 +197,52 @@ private RollbackDetails findBestCandidate(
...
+            .build();
+
+    return new RollbackDetails(
+        EXPLICIT, context, imageDetails.getImageName(), imageDetails.getBuildNumber());
+  }
+
+  /** Return the name of the original server group we should roll back to */
+  private Optional<String> getOriginalServerGroup(StageExecution stage) {
+    return Optional.ofNullable(
+            stage.findAncestor(isType(CreateServerGroupStage.PIPELINE_CONFIG_TYPE)))
```
lorin, line 225: Reading the code, I discovered that the CloneServerGroupStage and the (deprecated) CopyLastAsgStage could also potentially be ancestors.  I'll need to revise this code to handle those cases.

lorin: I'm going to use a different strategy for implementing this: I'm going to modify the MonitoredDeployStrategy to set the original server group as context on this stage, rather than calling findAncestor here. Updates coming, stay tuned...

lorin:
Updated and now ready for review.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
lorin: This is now done.

marchello2000:
> > I'm also not certain whether:
> > 
> > * createServerGroup is always a parent stage of the rollbackCluster stage (which seems to be the stage associated with the DetermineRollbackCandidatesTask task).
> > * there is always only one createServerGroup parent stage of the rollbackCluster stage (is it possible that this code will pick up the wrong ASG)?
> 

Not sure where these came from and if you still need an answer but..
* `createServerGroup` is not always the parent. `deck` might initiate a rollback without a create. Actually, curious what happens in that case. Does deck explicitly specify an ASG to rollback to? Might worth checking/double checking that there isn't another path that tries to accomplish the same as what your PR does (I don't believe so, but talking it out "loud" makes me wonder what i might be missing)

* there should only be one `createServerGroup` parent (if there is a parent). but either way, context.get() will return the most proximate parent

lorin:
> Not sure where these came from and if you still need an answer but..

The original questions from an earlier version of the PR, which had a very different implementation.

> 
> * `createServerGroup` is not always the parent. `deck` might initiate a rollback without a create. Actually, curious what happens in that case. Does deck explicitly specify an ASG to rollback to? Might worth checking/double checking that there isn't another path that tries to accomplish the same as what your PR does (I don't believe so, but talking it out "loud" makes me wonder what i might be missing)

I looked at the logic in deck in the file [app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.controller.js](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.controller.js).

From what I can tell, rolling back a server group via deck does not invoke the DetermineRollbackCandidatesTask code at all. Instead, deck contains its own logic in rollbackServerGroup.controller.js to determine what to roll back to. It generates the same kinds of outputs (`rollbackType`, `rollbackContext`) as DetermineRollbackCandidateTask class does, in order to provide the necessary rollback details.

PR #3875, opened 2020-08-25
feat(plugins): Initial commit of orca-remote-stage

jonsie:
Does not include declaring callback information or handling callback via pubsub or a specific endpoint - planning on that being a separate PR once I've verified this end-to-end flow.

Depends on: https://github.com/spinnaker/kork/pull/759

There's also various other things we can do here such as providing an endpoint to query for remote stage extensions (so we can do some fancy stuff in Deck, for example).


orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/model/RemoteStageExtensionConfig.kt
```diff
@@ -0,0 +1,10 @@
+package com.netflix.spinnaker.orca.remote.model
+
+import com.netflix.spinnaker.kork.plugins.remote.extension.RemoteExtensionConfigType
+
+data class RemoteStageExtensionConfig(
+  val type: String,
+  val description: String,
+  val label: String,
+  val parameters: MutableMap<String, Any> = mutableMapOf()
```
robzienert, line 9: Are `null` parameter values not allowed?

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/model/RemoteStageExtensionPayload.kt
```diff
@@ -0,0 +1,10 @@
+package com.netflix.spinnaker.orca.remote.model
+
+import com.netflix.spinnaker.kork.plugins.remote.extension.RemoteExtensionPayload
+
+data class RemoteStageExtensionPayload(
+  val type: String,
+  val id: String,
+  val pipelineExecutionId: String,
+  val context: MutableMap<String, Any>
```
robzienert, line 9: Same here, are `null` values not allowed?

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/model/RemoteStageExtensionPayload.kt
```diff
@@ -0,0 +1,10 @@
+package com.netflix.spinnaker.orca.remote.model
+
+import com.netflix.spinnaker.kork.plugins.remote.extension.RemoteExtensionPayload
+
+data class RemoteStageExtensionPayload(
```
robzienert, line 5: Need docs.

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/pipeline/RemoteStage.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
+import com.netflix.spinnaker.orca.remote.model.RemoteStageExtensionConfig
+import com.netflix.spinnaker.orca.remote.service.RemoteStageExtensionService
+import com.netflix.spinnaker.orca.remote.tasks.MonitorRemoteStageTask
+import com.netflix.spinnaker.orca.remote.tasks.StartRemoteStageTask
+import org.springframework.stereotype.Component
+
+@Beta
+@Component
+@Aliases("remoteStage")
```
robzienert, line 16: Prior convention is that we do not include `Stage` in stage types, as that's redundant. I feel like we can get rid of this annotation entirely, as it'll already get the `remote` type from [`StageDefinitionBuilder.getType`](https://github.com/spinnaker/orca/blob/master/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/graph/StageDefinitionBuilder.java#L90)

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/pipeline/RemoteStage.kt
```diff
@@ -0,0 +1,37 @@
...
+  override fun taskGraph(stage: StageExecution, builder: TaskNode.Builder) {
+    val remoteExtensionConfig = remoteStageExtensionService
+      .getByStageType(stage.type)
+      .getTypedConfig<RemoteStageExtensionConfig>()
+
+    // Set default parameters if necessary
+    remoteExtensionConfig.parameters.forEach { (k, v) ->
+      if (stage.context[k] == null) {
+        stage.context[k] = v
+      }
```
robzienert, line 30: Why do we only set parameters if the context doesn't have the value? Wouldn't we always want the config to be applied to the stage?

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/service/RemoteStageExtensionService.kt
```diff
@@ -0,0 +1,34 @@
+package com.netflix.spinnaker.orca.remote.service
+
+import com.netflix.spinnaker.kork.plugins.remote.RemotePluginsProvider
+import com.netflix.spinnaker.kork.plugins.remote.extension.RemoteExtension
+import com.netflix.spinnaker.orca.remote.RemoteStageExtensionDefinition
+import com.netflix.spinnaker.orca.remote.model.RemoteStageExtensionConfig
+
+class RemoteStageExtensionService(
```
robzienert, line 8: Docs.

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/service/exceptions.kt
```diff
@@ -0,0 +1,16 @@
...
+
+class RemoteStageNotFoundException : IntegrationException(
+  "No remote stage found.  Have you published a plugin with a remote extension stage configuration?"
+)
+
+class RemoteStageTypeNotFoundException(stageType: String) : IntegrationException(
+  "Remote stage type $stageType not found.  Check if stage type exists in plugin info configuration."
+)
+
+class DuplicateRemoteStageTypeException(stageType: String, pluginIdA: String, pluginIdB: String) : SystemException(
```
cfieber, line 14: feels like this could possibly be List<String> of plugin ids

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/tasks/StartRemoteStageTask.kt
```diff
@@ -0,0 +1,35 @@
...
+  override fun execute(stage: StageExecution): TaskResult {
+    val remoteExtension = remoteStageExtensionService.getByStageType(stage.type)
+    val remoteStageExtensionPayload = RemoteStageExtensionPayload(
+      type = stage.type,
+      id = stage.id,
+      pipelineExecutionId = stage.execution.id,
+      context = stage.context
+    )
+
+    remoteExtension.invoke(remoteStageExtensionPayload)
```
cfieber, line 26: it might make sense to allow the invoke to return a result that gets added to the context and that would get passed back in subsequent monitor calls

then again that might encourage doing non async things in the invoke, so maybe that is bad

but just thinking of something similar to how we issue a request to clouddriver to start an atomic operation, and that returns a taskId and we can use that taskId to monitor the state of the operation

without supporting something like that, a remote stage is going to have to keep some state on its side that maps pipelineExecutionId to whatever work tracking it is doing

orca-remote-stage/src/main/kotlin/com/netflix/spinnaker/orca/remote/pipeline/RemoteStage.kt
jonsie: No, I think we only want to use the defaults from the config if the stage is not run with the same parameters and values already defined in the context.  Does that make sense?

robzienert (approved):

PR #3879, opened 2020-08-27
feat(kubernetes): DeployManifestStage makes live calls to determine cluster members for rollout strategies 

maggieneterval:
Groundwork for: https://github.com/spinnaker/spinnaker/issues/5906
Requires: https://github.com/spinnaker/clouddriver/pull/4841

* feat(kubernetes): DeployManifestStage makes live calls to determine cluster members for rollout strategies 

  In order to make the `liveManifestCalls` behavior the default for all users, we must change all pipeline operations that rely on cache consistency to instead call Clouddriver endpoints that make live queries against the Kubernetes cluster.

  This commit replaces the existing `getClusters` call in the DeployManifestStage, used to hydrate the context of any Disable and Delete stages necessary in order to enact a Red/Black or Highlander rollout strategy, with a new `getClusterManifests` call, which returns a list of ManifestCoordinates for members of a given cluster. Also adds tests asserting on the basic correctness of the stage graph when using rollout strategies.

* refactor(kubernetes): remove raw context Maps in DeployManifestStage 

  Leaving the other non-context raw Maps as-is because those are all a result of deeply inspecting the list of full manifests currently present in the stage context as raw Maps, which we'll replace in an upcoming PR with a list of ManifestCoordinates.

### DIFF
--- a/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStageTest.java
+++ b/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStageTest.java
@@ -1,0 +1,212 @@
+/*
+ * Copyright 2020 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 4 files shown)


orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStageTest.java
```diff
@@ -0,0 +1,212 @@
...
+import com.netflix.spinnaker.orca.pipeline.model.PipelineExecutionImpl;
+import com.netflix.spinnaker.orca.pipeline.model.StageExecutionImpl;
+import java.util.Map;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
```
ezimanyi, line 43: Interesting, I don't think I've used this before!

maggieneterval: I started using it when I saw @plumpy using it...just trying to fit in. 🍡

PR #3880, opened 2020-08-27
fix(saga): Avoid retrying if we get a 404 back from clouddriver

ajordens:


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -178,6 +178,14 @@ class MonitorKatoTask implements RetryableTask, CloudProviderAware {
       try {
         kato.resumeTask(katoTask.id)
       } catch (Exception e) {
+        if (e instanceof RetrofitError) {
+          RetrofitError retrofitError = (RetrofitError) e
+          if (retrofitError?.response?.status == 404) {
+            // unexpected -- no sense attempting to resume a saga that `clouddriver` has no knowledge about
+            throw e
```
ajordens, line 185: I don't think this is going to leave the greatest exception in the execution ... but it doesn't make sense to retry indefinitely when we're getting a 404?

Any legitimate reason to get a 404 back?

jonsie: I don't think so - if Orca thinks it should resume a task then that task should exist.

PR #3882, opened 2020-08-27

spinnakerbot:


ezimanyi:
@bpowell @kevinawoo : The test failures here are due to spinnaker/kork#755.  It looks like some of these test controllers have switched to returning XML instead of JSON after that change, breaking the tests.  My guess is that something new that was added to the runtime classpath in that PR is autoconfiguring something differently, but I haven't looked in too much detail other than to reproduce and confirm which kork PR introduced this.

Can you please take a look?

jonsie:
@bpowell @kevinawoo Any status update here regarding fixing the tests that @ezimanyi noted above?

bpowell:
@jonsie will take a look today!

kevinawoo:
Oh very interesting thing mergify did.... Since it force pushed the autobump branch, it looks like the original bump of [kork 7.61.3](https://github.com/spinnaker/orca/commit/d3be6855d2fe797c5a0c11dab4dfcfd275868916) got merged, but really, it' bumped it to `7.62.1`

FYI: `7.62.1` includes the [Azure secrets revert PR](https://github.com/spinnaker/kork/pull/765)

jonsie:
@kevinawoo Yeah that's how it always works - if there is an existing bump PR out there it'll push force push to that branch as it updates the version.  That's part of the autobump release process, not mergify.

PR #3884, opened 2020-08-31
chore(plugins): Enable v2 framework in orca-plugins-test

jonsie:
The tests will fail until the framework bug is resolved, but this is a good thing to do.


jonsie:
Closing as this was enabled in another PR.

PR #3885, opened 2020-09-01
fix(qos): Make DefaultExecutionPromoter a real PollingAgent

marchello2000:
This way only one instance promotes a buffered execution at a time


orca-qos/src/main/kotlin/com/netflix/spinnaker/orca/qos/ExecutionPromoter.kt
```diff
@@ -58,61 +51,44 @@ class DefaultExecutionPromoter(
...
-    }
-
-  @Scheduled(fixedDelayString = "\${pollers.qos.promote-interval-ms:5000}")
-  fun promote() {
-    if (!discoveryActivated.get()) {
-      return
-    }
-
+  @VisibleForTesting
+  public override fun tick() {
```
robzienert, line 55: nit: `public` is the default visibility for Kotlin.

marchello2000: yea.,.. i wasn't sure why, but the test didn't work without it... maybe cuz it's overriding a protected one?

robzienert: Oh. Hmm... escalating the visibility shouldn't even be possible in that case. 😆

PR #3889, opened 2020-09-02
feat(plugins): Orca v2 framework compatibility

jonsie:
Use of `Task`, `StafeDefinitionBuilder`, and `PreconfiguredJobConfigurationProvider` (so all `SpinnakerExtensionPoint` implementations) must be called from an `ObjectProvider` since the v2 framework does not make guarantees about when beans are instantiated.  Additionally, we need to support loading these extensions at anytime so there are some tweaks here to re-compute caches if necessary.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/DefaultStageResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/DefaultStageResolver.java
@@ -19,10 +19,15 @@ package com.netflix.spinnaker.orca;
 import static java.lang.String.format;
 
 import com.netflix.spinnaker.orca.api.pipeline.graph.StageDefinitionBuilder;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.ObjectProvider;
 
 /**
  * {@code StageResolver} allows for {@code StageDefinitionBuilder} retrieval via bean name or alias.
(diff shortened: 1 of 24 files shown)


orca-plugins-test/src/test/resources/orca-plugins-test.yml
```diff
@@ -22,8 +22,13 @@ spring:
   application:
     name: orca
 
+dynamic-stage-resolver:
+  enabled: true
+
 spinnaker:
   extensibility:
+    framework:
+      version: v2
```
jonsie, line 31: I verified these tests pass on both v1 and v2, as well as verified they pass with both the `DynamicStageResolver` and `DefaultStageResolver`.

orca-core/src/main/java/com/netflix/spinnaker/orca/DefaultStageResolver.java
```diff
@@ -61,15 +91,32 @@ public DefaultStageResolver(Collection<StageDefinitionBuilder> stageDefinitionBu
...
-            type, stageDefinitionBuilderByAlias.get(typeAlias));
+    StageDefinitionBuilder stageDefinitionBuilder = getOrDefault(type, typeAlias);
 
     if (stageDefinitionBuilder == null) {
-      throw new NoSuchStageDefinitionBuilderException(
-          type, typeAlias, stageDefinitionBuilderByAlias.keySet());
+      log.debug(
+          "Stage definition builder for '{}' not found in initial stage definition builder cache, re-computing...",
+          type);
+      computeStageDefinitionBuilders();
```
robzienert, line 100: perf: Computing these every time we want a `StageDefinitionBuilder` is going to be fairly detrimental to performance given how often we're calling this. We should be able to fairly easily check for staleness so we're only creating the temp hashmap and iterating all `SDB`s when things actually change (which will more than likely only be during startup times).

orca-core/src/main/java/com/netflix/spinnaker/orca/DynamicStageResolver.kt
```diff
@@ -38,30 +40,33 @@ import org.slf4j.LoggerFactory
...
-      builder.aliases().forEach { alias ->
-        putOrAdd(alias, builder)
+  init { computeStageDefinitionBuilders() }
+
+  override fun getStageDefinitionBuilder(type: String, typeAlias: String?): StageDefinitionBuilder {
+    var builder = stageDefinitionBuilder(type, typeAlias)
+
+    if (builder == null) {
+      log.debug("Stage definition builder for $type not found in initial stage definition builder cache, re-computing...")
+      computeStageDefinitionBuilders()
```
robzienert, line 58: perf: Same goes here.

orca-core/src/main/java/com/netflix/spinnaker/orca/DefaultStageResolver.java
jonsie: I don't know how we'd check for staleness without seeing what `StageDefintionBuilder` objects exist in the `ObjectProvider` (which we can't know unless we get all the stages from it) so I think this is necessary.  However, this will probably only be called once and AT MOST twice through the lifecycle of an application.  Same goes for the `DynamicStageResolver`.  This wont be called on every call to get a `StageDefinitionBuilder` (unless someone is looking up a stage that doesn't exist over and over again - which, looking through the logs, doesn't happen).

PR #3893, opened 2020-09-03
feat(aws): Adding support for launch templates in UpdateSecurityGroup…

jeyrschabu:
…sForServerGroupStage

- added UpdateLaunchTemplateTask to handle security groups for launch template backed server groups

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/UpdateLaunchConfigTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/UpdateLaunchConfigTask.groovy
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.orca.api.pipeline.Task
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
 import com.netflix.spinnaker.orca.api.pipeline.TaskResult
 import com.netflix.spinnaker.orca.clouddriver.KatoService
+import com.netflix.spinnaker.orca.clouddriver.model.TaskId
 import com.netflix.spinnaker.orca.clouddriver.utils.CloudProviderAware
 import com.netflix.spinnaker.orca.kato.tasks.DeploymentDetailsAware
 import groovy.util.logging.Slf4j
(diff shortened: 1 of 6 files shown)


jeyrschabu:
Ref: https://github.com/spinnaker/clouddriver/pull/4710

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/UpdateLaunchConfigTask.groovy
```diff
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.orca.api.pipeline.Task
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
```
cfieber, line 21: I'd be tempted to rename this class to something like AbstractSomethingSomethingLaunchTask and then have two concrete subclasses.

Feels like some slight risk that someone could accidentally break launch template by doing something for launch config

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/UpdateLaunchTemplateTask.java
```diff
@@ -0,0 +1,55 @@
...
+    ctx.put("kato.last.task.id", taskId);
+    ctx.put(
+        "deploy.server.groups",
+        Collections.singletonMap(region, Collections.singleton(serverGroupName)));
+    return ctx;
+  }
+
+  @Override
+  public String getOperation() {
+    return "updateLaunchTemplate";
```
cfieber, line 53:
```suggestion
    return OPERATION;
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/pipeline/UpdateSecurityGroupsForServerGroupStage.java
```diff
@@ -45,9 +46,15 @@
...
   public void taskGraph(@Nonnull StageExecution stage, @Nonnull TaskNode.Builder builder) {
-    builder
-        .withTask("updateLaunchConfig", UpdateLaunchConfigTask.class)
-        .withTask("updateInstances", UpdateInstancesTask.class);
+    Boolean hasLaunchTemplate =
+        (Boolean) stage.getContext().getOrDefault("hasLaunchTemplate", false);
+    if (hasLaunchTemplate) {
+      builder.withTask("updateLaunchTemplate", UpdateLaunchTemplateTask.class);
+    } else {
+      builder.withTask("updateLaunchConfig", UpdateLaunchConfigTask.class);
```
cfieber, line 54: could probably use the OPERATION constant here too

cfieber (approved):

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/UpdateLaunchConfigTask.groovy
jeyrschabu: That's a good point.

PR #3895, opened 2020-09-04
chore(monitoring): Log and emit a metric when a retried kato task is ultimately unsuccessful

jonsie:


robzienert (approved):
🎉

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -224,6 +233,20 @@ class MonitorKatoTask implements RetryableTask, CloudProviderAware {
     }
   }
 
+  /**
+   * Log and emits a metric when a kato task retry is finally terminal - typically either from
+   * the task timing out or from an unexpected error, like a 404, when attempting to retry.
+   */
+  private void monitorFinalTerminalRetry(StageExecution stage, String reason) {
```
marchello2000, line 240: nit: i am not sure i "get" the name `monitorFinalTerminalRetry` maybe more like `logFinalTerminalRetry` but obv not a strong feeling here

marchello2000 (commented):
this is awesome, thanks!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
jonsie: Ah yea, I went with `monitor` since we also emit a metric from this function.  I was toying with `observeFinalTerminalRetry` too.

jonsie: But yea the term `monitor` and `observe` can be sort of overloaded.

PR #3897, opened 2020-09-04
chore(dependencies): Merging Keiko back into Orca

robzienert:
Keeping history. This would mean we'd be able to deprecate the keiko repo.


jonsie:
I wonder if it would make sense to have an `orca-keiko` project that is itself multi-project for the various keiko modules?

robzienert:
Why?

jonsie:
I think all it would really achieve is providing a root for keiko in orca (for things like docs, primarily).  Probably not necessary.

robzienert:
I had just put the docs into `keiko-core`. I'm not sure I want to get into the business of nested subprojects at this point - it's not something we do anywhere else and while I don't think it'd negatively affect things like composite build, I simply don't trust that functionality enough to introduce a new potential variable.

jonsie (approved):

srekapalli (approved):

AbdulRahmanAlHamali:
Hi Rob, I noticed that the document here still refers to the old keiko repo:
https://spinnaker.io/guides/developer/service-overviews/orca/#runtime

Maybe it requires an update?

PR #3898, opened 2020-09-08
chore(dependencies): Add keiko-spring to orca-web and fixup some dependency implementations

jonsie:


orca-web/orca-web.gradle
```diff
@@ -33,6 +33,8 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-actuator")
   implementation("net.logstash.logback:logstash-logback-encoder")
 
+  implementation(project(":keiko-spring"))
```
jonsie, line 36: Without this Orca wasn't able to load up `ObjectMapperSubtypeProperties` from `keiko-spring` on startup.

robzienert: whoop. thanks

robzienert (approved):

PR #3902, opened 2020-09-10
feat(stages): Add stage to monitor specific execution(s)

marchello2000:
This change adds a `MonitorPipelineStage` stage that allows one to monitor a given set of executions, similar to how `PipelineStage` starts and monitors that pipelins.
The difference is that this stage doesn't start any pipelines, it simply performs the monitoring.

Ideally, I would use the `MonitorPipelineStage` as a synthetic child of `PipelineStage` but that requires a bit more finesse and refactor that I don't think is worth it right now.

I will also be making corresponding `deck` changes to visualize the stage execution

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/MonitorPipelineStage.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/MonitorPipelineStage.java
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.orca.front50.pipeline;
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.netflix.spinnaker.kork.annotations.DeprecationInfo;
 import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact;
 import com.netflix.spinnaker.orca.api.pipeline.graph.StageDefinitionBuilder;
 import com.netflix.spinnaker.orca.api.pipeline.graph.TaskNode;
@@ -77,6 +78,12 @@ public class MonitorPipelineStage implements StageDefinitionBuilder {
      * DO NOT USE: Legacy from PipelineStage Use {@link executionIds} instead (even if you only have
      * one execution ID to monitor)
      */
+    @Deprecated
+    @DeprecationInfo(
+        reason = "Used for backwards compat with PipelineStage",
+        replaceWith = "Use executionIds instead even if you only have one execution ID",
+        since = "2020.4",
+        eol = "n/a")
     public String executionId;
 
     /** List of executions IDs to monitor */
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/MonitorPipelineStage.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/MonitorPipelineStage.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.orca.front50.pipeline;
+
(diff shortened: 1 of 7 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/MonitorPipelineStage.java
```diff
@@ -0,0 +1,135 @@
...
+    /** As soon as (at least) one monitored pipeline fails, terminate this stage */
+    FailFast
+  }
+
+  public static class StageParameters {
+    /**
+     * DO NOT USE: Legacy from PipelineStage Use {@link executionIds} instead (even if you only have
+     * one execution ID to monitor)
+     */
+    public String executionId;
```
robzienert, line 80: I'm confused. How would anyone know not to use this unless they're sleuthing code? At minimum, I'd say we need `@Deprecated` & `@DeprecationInfo` on this. To be tackled later, I'd like a means of being able to provide this feedback of deprecation to end users (perhaps via `PipelineExecution.systemNotifications`?)

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
```diff
@@ -43,8 +43,12 @@
   public static final String PIPELINE_CONFIG_TYPE =
       StageDefinitionBuilder.getType(PipelineStage.class);
 
-  @Autowired(required = false)
-  ExecutionRepository executionRepository;
+  final ExecutionRepository executionRepository;
+
+  @Autowired
+  public PipelineStage(ExecutionRepository executionRepository) {
```
robzienert, line 49: This _appears_ to be a breaking change. I don't know when we wouldn't have an `ExecutionRepository`, but in this change, we're making it required, whereas in the past it was optional.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy
```diff
@@ -33,74 +35,134 @@ import static com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType.PIPEL
...
-              .findResults { k ->
-                String message = k.exception?.message ?: k.history ? ((List<String>) k.history).last() : null
-                return message ? buildExceptionMessage(childPipeline.name, message, s) : null
+    HashMap<String, MonitorPipelineStage.ChildPipelineStatusDetails> pipelineStatuses = new HashMap<>(pipelineIds.size())
+    PipelineExecution firstPipeline
+
+    pipelineIds.forEach { pipelineId ->
+      PipelineExecution childPipeline = executionRepository.retrieve(PIPELINE, pipelineId)
+      if (firstPipeline == null) {
+        // Capture the first pipeline, since if there is only, we will return its context as part of taskresult
```
robzienert, line 69:
```suggestion
        // Capture the first pipeline, since if there is only one, we will return its context as part of taskresult
```

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy
```diff
@@ -33,74 +35,134 @@ import static com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType.PIPEL
...
+        return buildTaskResult(ExecutionStatus.TERMINAL, context, result)
       }
+    }
 
-      return TaskResult.builder(terminalStatus(childPipeline.status)).context([
-        status   : childPipeline.status,
-        exception: exceptionDetails
-      ]).outputs(childPipeline.getContext())
-        .build()
+    // Finally, if all completed and we didn't catch that case yet that means they all were cancelled
```
robzienert, line 141: nit: "that" and "they" in this comment are ambiguous.

robzienert (commented):
Neat!

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
marchello2000: ah, good call. Not that this stage will work without a repo though.... not sure what to do. I think making it optional like it was before will give a wrong impression that we can work without it. and making the whole stage/bean conditional on repo seems weird, what do you think?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/MonitorPipelineStage.java
marchello2000: thanks, updating

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
robzienert: Yeah - I'd say the change you have is correct. I didn't take a look myself, so it just jumped out at me as something that _could_ break. If the stage won't work without it, you're just fixing things. :)

PR #3903, opened 2020-09-10
feat(core): Add new "stageStatus" spel helper

robzienert:
Small quality of life improvement.

Today, if you want to interact with a stage status, you must `#stage("foo").status.toString()` before doing conditionals, since `status` is an `ExecutionStatus` enum type. This helper function just makes it so you don't need to `toString()`: `#stageStatus("foo") == "SUCCEEDED"`


marchello2000:
Has this gotten broken, cuz i fixored this before... https://github.com/spinnaker/orca/pull/3272

caseyhebebrand (approved):
Awesome thanks!

marchello2000:
(adding to my prior hasty comment) This seems fine, but given we already have a way (just verified it is still working) to do `#stage('quax').status == 'SUCCEEDED` and `#stage('quax').hasSucceeded` and `#stage('quax').hasFailed` as part of the #3272 I am not sure this change is necessary.

robzienert:
Oh, agreed. I wasn't aware of those changes (or had forgotten). Cool, I like your way of things.

PR #3910, opened 2020-09-15
fix(aws): do not filter scaling processes to resume/suspend

dreynaud:
- not filtering means we might issue a task that says "resume Launch/Terminate", when in fact they already are enabled. Which in the spirit of idempotency should be quick and error free anyway (a noop)
- filtering means we occasionally think we don't need to reenable, when in fact we should, leading to a pipeline not behaving correctly and leaving things in a bad state


ajordens:
LGTM

marchello2000 (approved):
less code = less chance for errors, I love it!

robzienert (approved):

PR #3912, opened 2020-09-17
refactor(kubernetes): remove ManifestForceCacheRefreshTask 

maggieneterval:
* refactor(kubernetes): remove ManifestForceCacheRefreshTask from manifest stages 

  Recent [changes](https://github.com/spinnaker/spinnaker/issues/5607#issuecomment-692717738) have refactored the operations for which it was important to keep Clouddriver's cache in a consistent state such that they now rely on live lookups against the Kubernetes cluster instead. We can now remove the task responsible for ensuring the cache is refreshed from all Kubernetes manifest-related stages.

* refactor(kubernetes): delete ManifestForceCacheRefreshTask 

  No longer used by any stages.

* refactor(kubernetes): remove unused ManifestForceCacheRefreshTask-related logic 

  ManifestAware.manifestsToRefresh is now uncalled, and the manifest refresh-related fields written to by PromoteManifestKatoOutputsTask are no longer read.


ezimanyi (approved):
🚫 📜 ➡️ 💰 💯 🥳 🎈 🥇

german-muzquiz (approved):
🏆

PR #3915, opened 2020-09-18
fix(igor): Handle timeouts kicking off builds

marchello2000:
This a companion PR to https://github.com/spinnaker/igor/pull/863

The issue here is that when kicking off a build takes a long time (e.g. when jenkins has a lot of job configs) orca might kick off a lot of builds instead of just one (and might still fail!)
Igor now adds support for "pending" operation so if the same request is received twice it will return 202. This change adds the corresponding support to orca, on 202, we just need to retry, because eventually igor will give us the build number

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartJenkinsJobTask.groovy
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartJenkinsJobTask.groovy
@@ -23,7 +23,9 @@ import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
 import com.netflix.spinnaker.orca.api.pipeline.Task
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
 import com.netflix.spinnaker.orca.api.pipeline.TaskResult
+import com.netflix.spinnaker.orca.exceptions.ExceptionHandler
 import com.netflix.spinnaker.orca.igor.BuildService
+import com.netflix.spinnaker.orca.retrofit.exceptions.RetrofitExceptionHandler
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
@@ -45,6 +47,9 @@ class StartJenkinsJobTask implements RetryableTask {
   @Autowired
   ObjectMapper objectMapper
 
+  @Autowired
+  RetrofitExceptionHandler retrofitExceptionHandler
+
   @Nonnull
   @Override
   TaskResult execute(@Nonnull StageExecution stage) {
@@ -55,7 +60,7 @@ class StartJenkinsJobTask implements RetryableTask {
       Response igorResponse = buildService.build(master, job, stage.context.parameters)
 
       if (igorResponse.getStatus() == HttpStatus.ACCEPTED.value()) {
-        log.info("build start for $master $job is pending, waiting")
+        log.info("build for job=$job on master=$master is pending, waiting for build to start")
         return TaskResult.RUNNING
       }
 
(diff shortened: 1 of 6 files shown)


marchello2000:
forgot to add a test, will add it in a bit

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartJenkinsJobTask.groovy
```diff
@@ -41,7 +50,41 @@ class StartJenkinsJobTask implements Task {
...
+      }
+
+      if (igorResponse.getStatus() == HttpStatus.OK.value()) {
+        String queuedBuild = igorResponse.body.in().text
+        return TaskResult
+            .builder(ExecutionStatus.SUCCEEDED)
+            .context([queuedBuild: queuedBuild])
+            .build()
+      }
+    }
```
srekapalli, line 69: I think RetrofitError will be handled by RetrofitException handler which decides whether Retry should happen or not. May be better to leave it to that and not handle Retrofit error here.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartJenkinsJobTask.groovy
```diff
@@ -41,7 +50,41 @@ class StartJenkinsJobTask implements Task {
...
+    catch (RetrofitError e) {
+      if (e.kind == RetrofitError.Kind.NETWORK) {
+        log.warn("Network failure communicating with igor to start a jenkins job, will retry", e)
+        return TaskResult.RUNNING
+      }
+      throw e
+    }
+
+    throw new SystemException("Failure starting script")
```
srekapalli, line 78:
```suggestion
    throw new SystemException("Failure starting jenkins task")
```

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartScriptTask.groovy
```diff
@@ -83,8 +93,40 @@ class StartScriptTask implements Task {
...
+      }
+
+      if (igorResponse.getStatus() == HttpStatus.OK.value()) {
+        String queuedBuild = igorResponse.body.in().text
+        return TaskResult
+            .builder(ExecutionStatus.SUCCEEDED)
+            .context([master: master, job: job, queuedBuild: queuedBuild, REPO_URL: repoUrl ?: 'default'])
+            .build()
+      }
+    }
```
srekapalli, line 111: Here also we can leave the retry to Retrofit exception handler.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/StartJenkinsJobTask.groovy
ajordens: Might want to flag this as a `RetryableTask` now?

marchello2000: > I think RetrofitError will be handled by RetrofitException handler which decides whether Retry should happen or not. May be better to leave it to that and not handle Retrofit error here.

It is true, it will. I am torn on this because since I know it can happen here and is retryable, i'd rather add a more specific error handler for clarity of logs/debugging. But I also recognize that this is code duplication... What do you think, @srekapalli ?

> Might want to flag this as a `RetryableTask` now?

I thought it is? Double checking

srekapalli: > > I think RetrofitError will be handled by RetrofitException handler which decides whether Retry should happen or not. May be better to leave it to that and not handle Retrofit error here.
> 
> It is true, it will. I am torn on this because since I know it can happen here and is retryable, i'd rather add a more specific error handler for clarity of logs/debugging. But I also recognize that this is code duplication... What do you think, @srekapalli ?

May be log and re-throw it and not do any task status updates here.

marchello2000: Looking more at this, I still prefer to keep it as is here. The reason being is that I don't want to rely on some  exception handler because, someone could (and probably should make this call a `POST` in which case it won't be retryable). I also considered wrapping and rethrowing a `SpinnakerException` with `isRetryable=true` but that creates a ton of noise in the logs (e.g. "failure running task, etc"). And i'd rather have the log clear and logic localized for when things do go wrong because it makes it easier to debug. What do you think?

srekapalli: Sounds good.

srekapalli (approved):

PR #3919, opened 2020-09-21
chore(tasks): Add logging for when tasks take a long time to run

marchello2000:
This change adds logging for when a tasks `execute` method takes a long time to run (over 30s).
Since tasks have a pretty sensitive ack timeout, having a long running taks might be indicative of a problem to come.

We have metrics for task invocation times as well, but once an operator knows that some task takes a long time, it's useful to find it in logs to see what it was doing for further investigation

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
@@ -86,6 +86,13 @@ class RunTaskHandler(
   private val dynamicConfigService: DynamicConfigService
 ) : OrcaMessageHandler<RunTask>, ExpressionAware, AuthenticationAware {
 
+  /**
+   *  If a task takes longer than this number of ms to run, we will print a warning.
+   *  This is an indication that the task might eventually hit the dreaded message ack timeout and might end
+   *  running multiple times cause unintended side-effects.
+   */
+  private val warningInvocationTimeMs = 30000
+
   override fun handle(message: RunTask) {
     message.withTask { origStage, taskModel, task ->
       var stage = origStage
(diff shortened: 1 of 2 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -203,6 +211,10 @@ class RunTaskHandler(
       ).forEach { name, tags ->
         registry.timer(name, tags).record(elapsedMillis, TimeUnit.MILLISECONDS)
       }
+
+      if (elapsedMillis >= warningInvocationTimeMs) {
+        log.info("Task invocation for task ${taskModel.implementingClass} in stage ${stage.type} with id ${stage.id} took over $warningInvocationTimeMs")
```
ajordens, line 216: Might be a little easier to parse this log statement (in sumo) if:

```
log.info("Task invocation took over {}ms (taskType: {}, stageType: {}, stageId: {})", ...)
```

Short of actually dropping them on MDC.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -86,6 +86,13 @@ class RunTaskHandler(
   private val dynamicConfigService: DynamicConfigService
 ) : OrcaMessageHandler<RunTask>, ExpressionAware, AuthenticationAware {
 
+  /**
+   *  If a task takes longer than this number of ms to run, we will print a warning.
+   *  This is an indication that the task might eventually hit the dreaded message ack timeout and might end
+   *  running multiple times cause unintended side-effects.
+   */
+  private val warningInvocationTimeMs = 30000
```
ajordens, line 94: Perhaps configurable with the default 30s.

ajordens (approved):
LGTM beyond a tweak to the log line.

PR #3920, opened 2020-09-22
fix(disable): Fixes issue when only using healthcheck provider

diasjorge:
* Google provider does not set server group flag as disabled causing
  the task to timeout since it will never finish.
  * Issue: spinnaker/spinnaker#6038


diasjorge:
Hi @dreynaud, please could you help me review this patch?

dreynaud:
> Hi @dreynaud, please could you help me review this patch?

Hi @diasjorge, I am concerned that this change will introduce side effects for providers that do set the `disabled` flag. If possible, I think it would be preferable for the Google provider to set the `disabled` flag (yay for consistency 🙌)

diasjorge:
I was waiting for some response but nothing yet. I think the google provider does not have access to the interestinghealthproviders used in this stage. With the patch we would only call instancesdowntask if the server group only cares about the cloud health check. This behavior is the same as before 1.22 but we’ll call the instancesdown task which should know how to handle things correctly as you changed in your patch. What do you think?

dreynaud:
@diasjorge we can't merge this change as-is, it will introduce a regression. Consider the case of an EC2 server group with `interestingHealthProviderNames == ["Amazon"]`. 

Before this change:
- we would first make sure that `disabled == true`
- then we would check that `waitForRequiredInstancesDownTask` succeeds, which it always does when only considering the cloud provider health

After this change:
- we would go straight to `waitForRequiredInstancesDownTask`, which always succeeds when only considering the cloud provider health
- `disabled` may be true or false, no guarantee anymore! This is a problem

diasjorge:
@dreynaud thanks for the explanation, it's just I don't know the internal to fix the google provider to do such a thing. I'm only a spinnaker user. As it stands, the shrink cluster stage doesn't work with google health checks after 1.22. I wish there was some process to escalate it to the right person.

PR #3924, opened 2020-09-24
Revert "fix(aws): do not filter scaling processes to resume/suspend"

cfieber:
Reverts spinnaker/orca#3910

Seeing that if a user manually selects a server group and issues a resize operation then we will always resume launch/terminate at the start of that and suspend launch/terminate at the end leaving the ASG in a bad state


cfieber:
closing this, I've just rolled our orca back for now

PR #3925, opened 2020-09-24
fix(aws): *do* filter scaling processes to resume/suspend

dreynaud:
This reverts commit b146bf638d3090b1542f860ec39b1cefac564f63 (#3910)


marchello2000 (approved):
😢

dreynaud:
1. no filtering means we now always resume (great!)
2. which means we now always have a scalingProcesses.${asgName} output from the resume scaling processes task (uh oh)
3. which means the suspend scaling processes task now always has something to pick up and suspend (not great :weary:)

Reverting this now to avoid spreading further badness. A longer term fix could be to:
1. take a snapshot of the target server group at the beginning of the resize stage
2. always resume scaling processes
3. turn "suspend scaling processes" into "restore scaling processes", i.e. only suspend things that were suspended in the snapshot. That way we avoid relying on the implicit signal ("if things have been resumed, they shall be suspended again")

PR #3927, opened 2020-09-25
feat(stages): Make pipeline stage skippable.

haleyw:
Update canManuallySkip to look at stage context and override it on the PipelineStage to enable pipeline stages to be skippable.


marchello2000 (approved):
LGTM, thanks for doing this!

PR #3933, opened 2020-09-29
feat(webhooks): Allow for webhook monitor tasks without JSON processing

hashbrowncipher:
This allows us to monitor webhook responses while allowing the recipient
server to hold the entire logic for determining status. This is possible
today with CreateWebhookTask, but CreateWebhookTask has a fixed timeout,
whereas MonitorWebhookTask has an overrideable timeout. Another option
would be to make CreateWebhookTask into an
OverrideableTimeoutRetryableTask.

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
@@ -68,9 +68,8 @@ class MonitorWebhookTask implements OverridableTimeoutRetryableTask {
   TaskResult execute(StageExecution stage) {
     WebhookStage.StageData stageData = stage.mapTo(WebhookStage.StageData)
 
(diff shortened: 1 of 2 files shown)


hashbrowncipher:
> We prefer small, well tested pull requests.

It's small but I can't claim it's well tested. I could use assistance from a member of the team in determining whether this is a desireable feature and the right way to approach the problem. Please consider this a feature request that happened to reach PR form.

marchello2000:
Looks good to me, thanks for putting this together.
If you don't mind, probably worth updating [this](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/help/help.contents.ts#L474) string to say something like 
```
"JSON path to the status information in the webhook's response JSON. (e.g. <samp>$.buildInfo.status</samp>)<br>
If left empty a 200 response from the status endpoint will be considered a success"
```

marchello2000 (approved):

hashbrowncipher:
PR opened: https://github.com/spinnaker/deck/pull/8608

Thanks for the pointer! I didn't know those docs existed.

PR #3939, opened 2020-10-02
fix(tasks): Only lookup task by type in RunTaskHandler; remove code introduced to support one-off case for now obsolete SimpleTask

jonsie:
Removes most of what was introduced here: https://github.com/spinnaker/orca/pull/3574

This is mostly a performance improvement so that we don't needlessly rebuild the task cache.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
@@ -232,14 +232,10 @@ class RunTaskHandler(
   private fun RunTask.withTask(block: (StageExecution, TaskExecution, Task) -> Unit) =
     withTask { stage, taskModel ->
       try {
-        taskResolver.getTask(taskModel.implementingClass)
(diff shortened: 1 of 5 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -232,14 +232,10 @@ class RunTaskHandler(
   private fun RunTask.withTask(block: (StageExecution, TaskExecution, Task) -> Unit) =
     withTask { stage, taskModel ->
       try {
-        taskResolver.getTask(taskModel.implementingClass)
```
jonsie, line 235: This call to lookup based on the implementing class name was introduced to support the now deleted `SimpleTask` pattern.  The problem with doing this is there cases wherein this will always fail (like when tasks are nested classes inside stages, the lookup by implementing class name will fail), thus needlessly re-triggering a rebuild of the task cache.

robzienert (commented):
Love it. Love deleted code.

robzienert (approved):

PR #3943, opened 2020-10-03
feat(disable): add an optional waitForDisabledServerGroup task

dreynaud:
When we disable a server group, we disable and also take instances down, but only wait for instances to become down. This means that it is possible to finish a DisableServerGroupStage successfully with down instances but a server group that does not yet appear disabled. This creates poor interactions, for instance if the stage that follows is a resize.

-- open for early feedback, the kotlin bits don't quite compile yet and I'm not sure why


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DisableServerGroupStage.groovy
```diff
@@ -33,13 +34,14 @@ import org.springframework.stereotype.Component
 @Component
 @CompileStatic
 class DisableServerGroupStage extends TargetServerGroupLinearStageSupport implements ForceCacheRefreshAware {
+  public static final String TOGGLE = "stages.disable-server-group.wait-for-disabled.enabled"
```
marchello2000, line 37: you don't want/need `enabled` if you are using `dynamicConfig.isEnabled()` as it will add append it.
There was some talk of using Spring's `Environment` instead of `DynamicConfigService` I think it removes that ambiguity..?

```suggestion
  public static final String TOGGLE = "stages.disable-server-group.wait-for-disabled"
```

dreynaud: I changed `dynamicConfig.isEnabled()` to not add it if it's already there because I strongly dislike that it adds it. I'm of the opinion that properties should be greppable

orca-clouddriver/orca-clouddriver.gradle
```diff
@@ -15,6 +15,7 @@
  */
 
 apply from: "$rootDir/gradle/groovy.gradle"
+apply from: "$rootDir/gradle/kotlin.gradle"
```
robzienert, line 18: You can't mix Kotlin into a module that has Groovy, which is why the kotlin bits don't work - this is why I've been doing refactors Groovy -> Java, instead of Groovy -> Kotlin.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ModifyAwsScalingProcessStage.groovy
```diff
@@ -101,7 +101,7 @@ class ModifyAwsScalingProcessStage extends TargetServerGroupLinearStageSupport i
         throw new IllegalStateException("No server group found (serverGroupName: ${stageData.region}:${stageData.serverGroupName})")
       }
 
-      def suspendedProcesses = getSuspendedProcesses(targetServerGroup.get())
+      def suspendedProcesses = targetServerGroup.get().getSuspendedProcesses()
```
robzienert, line 104: Risk of an NPE here.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DisableServerGroupStage.groovy
robzienert: You can always just use `Environment` instead. There's really no reason to use `DynamicConfigService`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/support/TargetServerGroup.groovy
```diff
@@ -48,6 +49,12 @@ class TargetServerGroup {
     serverGroup = new HashMap(serverGroupData).asImmutable()
   }
 
+  @CompileDynamic
```
robzienert, line 52: Since the class isn't `@CompileStatic`, adding `@CompileDynamic` is redundant.

orca-core/src/main/java/com/netflix/spinnaker/orca/clouddriver/utils/CloudProviderAware.java
```diff
@@ -17,9 +17,12 @@
 package com.netflix.spinnaker.orca.clouddriver.utils;
```
robzienert, line 17: nit: Surprised to see this here! This class, nor `RegionCollector` which uses this class, don't belong in `orca-core`, imo. I'd rather see these moved into `orca-clouddriver`.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/DisableServerGroupStage.groovy
dreynaud: alright then, `Environment` it is. Thanks!

orca-core/src/main/java/com/netflix/spinnaker/orca/clouddriver/utils/CloudProviderAware.java
dreynaud: eh, is it really worth it? That means making sure all of these have a dependency on `orca-clouddriver`:

```
orca-mine
orca-queue
orca-queue-tck
orca-integrations-cloudfoundry
orca-bakery
```

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/ModifyAwsScalingProcessStage.groovy
dreynaud: how so? There is a `targetServerGroup.present` check above

PR #3946, opened 2020-10-06
feat(queue): Add command to cleanup zombie executions

robzienert:


robzienert:
Ha. Whoops, forgot to delete a refactored-out version.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/ZombieExecutionService.kt
```diff
@@ -0,0 +1,118 @@
...
+  fun killZombies(minimumActivity: Duration = Duration.ofMinutes(30)) {
+    findAllZombies(minimumActivity).forEach {
+      killZombie(it)
+    }
+  }
+
+  /**
+   * Kill a single zombie pipeline execution.
+   *
+   * WARNING: This method is unprotected: It will set a pipeline to canceled without verifying that it is not a zombie.
```
marchello2000, line 93:
```suggestion
   * WARNING: This method is unprotected: It will set a pipeline to canceled without verifying that it is a zombie.
```

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/ZombieExecutionService.kt
```diff
@@ -0,0 +1,118 @@
...
+   * It is recommended to use [killZombies] instead.
+   */
+  fun killZombie(execution: PipelineExecution) {
+    log.warn("Killing zombie execution: ${execution.application}/${execution.id}")
```
marchello2000, line 97: nit:
```suggestion
    log.warn("Force cancelling zombie execution and all its stages: ${execution.application}/${execution.id}")
```

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/admin/web/QueueAdminController.kt
```diff
@@ -51,6 +55,17 @@ class QueueAdminController(
       )
     )
 
+  @PostMapping(value = ["/killZombies"])
```
marchello2000, line 58: i think we should support specifying a cutoff time
I am worried that 30m cutoff is too small. This will trip manual judgement stages that are running for a while and wait for execution window/wait stage

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/admin/web/QueueAdminController.kt
```diff
@@ -51,6 +55,17 @@ class QueueAdminController(
       )
     )
```
marchello2000, line 57: Would also like to see a `killZombie/{executionId}` so we can kill a specific execution (which i think is more common of a scenario [and safer])

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/admin/web/QueueAdminController.kt
robzienert: I bumped the default to 60 minutes, as well as opened it for configuration.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/admin/web/QueueAdminController.kt
robzienert: Added.

robzienert:
Ready for review.

marchello2000 (approved):
❤️

PR #3947, opened 2020-10-06
feat(pipeline executions/orca) : Added ability to add roles to manual…

sanopsmx:

### DIFF
--- a/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
+++ b/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
@@ -16,6 +16,9 @@
 
 package com.netflix.spinnaker.orca.echo.pipeline
 
+import com.netflix.spinnaker.fiat.model.Authorization
+import com.netflix.spinnaker.fiat.model.UserPermission
+import com.netflix.spinnaker.fiat.model.resources.Role
 import com.fasterxml.jackson.annotation.JsonAnyGetter
 import com.fasterxml.jackson.annotation.JsonAnySetter
 import com.fasterxml.jackson.annotation.JsonIgnore
@@ -35,6 +38,7 @@ import com.netflix.spinnaker.security.User
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator
 
 @Component
 class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage {
@@ -75,18 +79,37 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage 
     @Autowired(required = false)
     EchoService echoService
 
+    @Autowired(required = false)
+    private FiatPermissionEvaluator fiatPermissionEvaluator
+
     @Override
     TaskResult execute(StageExecution stage) {
       StageData stageData = stage.mapTo(StageData)
+      def stageAuthorized = stage.context.get('isAuthorized')
+      def stageRoles = stage.context.get('stageRoles')
+      def permissions = stage.context.get('permissions')
+      def username = stage.lastModified? stage.lastModified.user: ""
       String notificationState
       ExecutionStatus executionStatus
 
       switch (stageData.state) {
         case StageData.State.CONTINUE:
+          def flag = stageAuthorized || checkManualJudgmentAuthorizedGroups(stageRoles, permissions, username)
+          if(flag) {
+            stageAuthorized = true;
+          } else {
+            stageAuthorized = false;
+          }
           notificationState = "manualJudgmentContinue"
           executionStatus = ExecutionStatus.SUCCEEDED
           break
         case StageData.State.STOP:
+          def flag = stageAuthorized || checkManualJudgmentAuthorizedGroups(stageRoles, permissions, username)
+          if(flag) {
+            stageAuthorized = true;
+          } else {
+            stageAuthorized = false;
+          }
           notificationState = "manualJudgmentStop"
           executionStatus = ExecutionStatus.TERMINAL
           break
@@ -95,10 +118,61 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage 
           executionStatus = ExecutionStatus.RUNNING
           break
       }
-
+      if (!stageAuthorized) {
+        notificationState = "manualJudgment"
+        executionStatus = ExecutionStatus.RUNNING
+        stage.context.put("judgmentStatus", "")
+      }
       Map outputs = processNotifications(stage, stageData, notificationState)
 
       return TaskResult.builder(executionStatus).context(outputs).build()
+    }
+
+    boolean checkManualJudgmentAuthorizedGroups(def stageRoles, def permissions, def username) {
+
+      if (username) {
+        UserPermission.View permission = fiatPermissionEvaluator.getPermission(username);
+        if (permission == null) { // Should never happen?
+          return false;
+        }
+        // User has to have all the pipeline roles.
+        Set<Role.View> roleView = permission.getRoles()
+        def userRoles = []
+        roleView.each { it -> userRoles.add(it.getName().trim()) }
+        return checkAuthorizedGroups(userRoles, stageRoles, permissions)
+      } else {
+        return false
+      }
+    }
+
+    boolean checkAuthorizedGroups(def userRoles, def stageRoles, def permissions) {
+
+      def value = false
+      if (!stageRoles) {
+        return true
+      }
+      for (role in userRoles) {
+        if (stageRoles.contains(role)) {
+          for (perm in permissions) {
+            def permKey = perm.getKey()
+            List<String> strList = null
+            if (Authorization.CREATE.name().equals(permKey) ||
+                Authorization.EXECUTE.name().equals(permKey) ||
+                Authorization.WRITE.name().equals(permKey)) {
+              strList = perm.getValue()
+              if (strList && strList.contains(role)) {
+                return true
+              }
+            } else if (Authorization.READ.name().equals(permKey)) {
+              strList = perm.getValue()
+              if (strList && strList.contains(role)) {
+                value = false
+              }
+            }
+          }
+        }
+      }
+      return value
     }
 
     Map processNotifications(StageExecution stage, StageData stageData, String notificationState) {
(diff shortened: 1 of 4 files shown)


orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -75,18 +79,37 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
     @Autowired(required = false)
     EchoService echoService
 
+    @Autowired(required = false)
+    private FiatPermissionEvaluator fiatPermissionEvaluator
```
robzienert, line 83: While we're in here, I'd like to see these autowired properties be changed to a constructor instead:

```
private final EchoService echoService
private final FiatPermissionEvaluator fiatPermissionEvaluator

WaitForManualJudgmentTask(Optional<EchoService> echoService, Optional<FiatPermissionEvaluator> fpe) {
  this.echoService = echoService.orElse(null)
  this.fiatPermissionEvaluator = fpe.orElse(null)
}
```

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -75,18 +79,37 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
       ExecutionStatus executionStatus
 
       switch (stageData.state) {
         case StageData.State.CONTINUE:
+          def flag = stageAuthorized || checkManualJudgmentAuthorizedGroups(stageRoles, permissions, username)
+          if(flag) {
+            stageAuthorized = true;
+          } else {
+            stageAuthorized = false;
+          }
```
robzienert, line 102: This isn't written very concisely - `flag` isn't descriptive, nor necessary:

```
stageAuthorized = stageAuthorized || checkManualJudgmentAuthorizedGroups(stageRoles, permissions, username)
```

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -75,18 +79,37 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
           notificationState = "manualJudgmentContinue"
           executionStatus = ExecutionStatus.SUCCEEDED
           break
         case StageData.State.STOP:
+          def flag = stageAuthorized || checkManualJudgmentAuthorizedGroups(stageRoles, permissions, username)
+          if(flag) {
+            stageAuthorized = true;
+          } else {
+            stageAuthorized = false;
+          }
```
robzienert, line 112: Same goes for this. Can be tightened up a lot.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -95,12 +118,63 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
+        stage.context.put("judgmentStatus", "")
+      }
       Map outputs = processNotifications(stage, stageData, notificationState)
 
       return TaskResult.builder(executionStatus).context(outputs).build()
     }
 
+    boolean checkManualJudgmentAuthorizedGroups(def stageRoles, def permissions, def username) {
+
+      if (username) {
```
robzienert, line 133: Would prefer if this was more deliberate in what it's checking. `Strings.isNullOrEmpty(username)` seems to be what you want. Usernames like `false` or `null` (which would be silly names to have) would break this logic as-is.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -95,12 +118,63 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
+        UserPermission.View permission = fiatPermissionEvaluator.getPermission(username);
+        if (permission == null) { // Should never happen?
+          return false;
+        }
+        // User has to have all the pipeline roles.
+        Set<Role.View> roleView = permission.getRoles()
+        def userRoles = []
+        roleView.each { it -> userRoles.add(it.getName().trim()) }
+        return checkAuthorizedGroups(userRoles, stageRoles, permissions)
```
robzienert, line 142: We can avoid an unnecessary variable assignment / list object creation here:

```groovy
def userRole = permission.getRoles().collect { it.getName().trim() }
return checkAuthorizedGroups(userRoles, stageRoles, permissions)
```

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -95,12 +118,63 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
+      } else {
+        return false
+      }
+    }
+
+    boolean checkAuthorizedGroups(def userRoles, def stageRoles, def permissions) {
+
+      def value = false
```
robzienert, line 150: This variable is poorly named.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -95,12 +118,63 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
+                Authorization.WRITE.name().equals(permKey)) {
+              strList = perm.getValue()
+              if (strList && strList.contains(role)) {
+                return true
+              }
+            } else if (Authorization.READ.name().equals(permKey)) {
+              strList = perm.getValue()
+              if (strList && strList.contains(role)) {
+                value = false
+              }
```
robzienert, line 170: The use of `strList` in both of these conditional blocks is unnecessary. You can just do `perm.getValue().contains(role)` and avoid variable assignment altogether.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -204,6 +211,68 @@ class OperationsController {
     }
   }
 
+  private void addStageAuthorizedRoles(def request, Map pipeline) {
+
+    def applicationName = pipeline.application
+    if (applicationName) {
```
robzienert, line 217: A pipeline is always associated with an application - this check is unnecessary.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -204,6 +211,68 @@ class OperationsController {
...
+      Application application = front50Service.get(applicationName)
+      if (application) {
```
robzienert, line 219: This code isn't going to behave the way you expect. `Front50Service` will throw a `RetrofitError` if the application cannot be found.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -204,6 +211,68 @@ class OperationsController {
...
+            item.stageRoles = stageRoles
+            item.permissions = permissions
+          }
+        }
+      }
+    }
+  }
+
+  private boolean checkAuthorizedGroups(def userRoles, def stageRoles,
+                                        def permissions) {
```
robzienert, line 246: This entire method matches the logic found in `ManualJudgmentStage.checkAuthorizedGroups` - there's opportunity to pull this out into a utility class.

orca-echo/src/test/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStageSpec.groovy
```diff
@@ -42,11 +42,11 @@ class ManualJudgmentStageSpec extends Specification {
...
-    [judgmentStatus: "continue"] || ExecutionStatus.SUCCEEDED
-    [judgmentStatus: "Continue"] || ExecutionStatus.SUCCEEDED
-    [judgmentStatus: "stop"]     || ExecutionStatus.TERMINAL
-    [judgmentStatus: "STOP"]     || ExecutionStatus.TERMINAL
-    [judgmentStatus: "unknown"]  || ExecutionStatus.RUNNING
+    [judgmentStatus: "continue", isAuthorized: true] || ExecutionStatus.SUCCEEDED
+    [judgmentStatus: "Continue", isAuthorized: true] || ExecutionStatus.SUCCEEDED
+    [judgmentStatus: "stop", isAuthorized: true]     || ExecutionStatus.TERMINAL
+    [judgmentStatus: "STOP", isAuthorized: true]     || ExecutionStatus.TERMINAL
+    [judgmentStatus: "unknown", isAuthorized: true]  || ExecutionStatus.RUNNING
```
robzienert, line 49: Tests needed exercises the negative cases.

robzienert (changes requested):

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: done

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: done.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: done

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
sanopsmx: done

sanopsmx:
All the comments are addressed. 

Raised a new PR...

https://github.com/spinnaker/orca/pull/3983

robzienert:
No need to open a new PR - just fix the existing PR next time, as it keeps feedback for a change colocated to a single place.

PR #3953, opened 2020-10-07
chore(java11): Target Java 11

cfieber:


Dockerfile.compile
```diff
@@ -1,11 +1,9 @@
 FROM ubuntu:bionic
 RUN apt-get update && apt-get install -y \
-    openjdk-8-jdk \
     openjdk-11-jdk \
  && rm -rf /var/lib/apt/lists/*
 LABEL maintainer="sig-platform@spinnaker.io"
 ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk-amd64
```
plumpy, line 6: and remove `JAVA_HOME` here too

gradle.properties
```diff
@@ -1,10 +1,9 @@
-enablePublishing=false
 fiatVersion=1.26.0
 korkVersion=7.76.3
-keikoVersion=3.9.0
```
plumpy, line 4: Ooh, spinnaker/orca#3897! I hadn't seen that!

Is that what you're working on, having kayenta use the new keiko-integrated orca?

plumpy (approved):
Thanks!

gradle.properties
cfieber: yeah, that is the first trip to the yak barbershop ..

PR #3955, opened 2020-10-07
fix(cf): Use moniker/frigga utils to build moniker

german-muzquiz:
Instead of manually building `app-stack-detail` string to get a moniker, let's use the one already provided in the source asg or just use the Moniker builder instead.


zachsmith1 (approved):
LGTM! Thanks for cleaning this up

kevinawoo (approved):

PR #3959, opened 2020-10-08
feat(clouddriver): Look ahead region detection for "aws" can be disabled

ajordens:
Similar to the bake stage, `skipRegionDetection` can be used to determine
whether or not additional regions are inferred from downstream deploy stages.

The default value is `false` which preserves the existing behavior.

```
{
  ...
  "skipRegionDetection": true
}

```

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
@@ -92,6 +92,7 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
     List<String> zones
     List<String> namespaces
     Boolean onlyEnabled = true
+    Boolean skipRegionDetection = false
     Boolean resolveMissingLocations
     SelectionStrategy selectionStrategy = SelectionStrategy.NEWEST
     String imageNamePattern
@@ -124,7 +125,7 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
     Map<Location, String> imageIds = [:]
     Set<String> inferredRegions = new HashSet<>()
 
-    if (cloudProvider == 'aws') {
+    if (cloudProvider == 'aws' && !config.skipRegionDetection) {
       // Supplement config with regions from subsequent deploy/canary stages:
       def deployRegions = regionCollector.getRegionsFromChildStages(stage)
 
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -124,7 +125,8 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
     Map<Location, String> imageIds = [:]
     Set<String> inferredRegions = new HashSet<>()
 
-    if (cloudProvider == 'aws') {
+    Boolean skipRegionDetection = Boolean.TRUE == config.skipRegionDetection
+    if (cloudProvider == 'aws' && !skipRegionDetection) {
```
cfieber, line 129: any particular reason not to just use `!config.skipRegionDetection` here, I think null and false will evaluate as expected

cfieber (approved):
LGTM

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -124,7 +125,8 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
     Map<Location, String> imageIds = [:]
     Set<String> inferredRegions = new HashSet<>()
 
-    if (cloudProvider == 'aws') {
+    Boolean skipRegionDetection = Boolean.TRUE == config.skipRegionDetection
```
robzienert, line 128: Odd. Why not just use `true`?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
ajordens: Nah, none at all. I sadly just lifted that from the bake stage verbatim. Changed.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
ajordens: I was lazy and lifted this snippet. I've made it better now.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTaskSpec.groovy
```diff
@@ -182,23 +182,24 @@ class FindImageFromClusterTaskSpec extends Specification {
     ['foo-x86_64-201603232351']                                     | ['foo-x86_64-201603232351'] as Set
   }
 
-  @Unroll
+//  @Unroll
```
marchello2000, line 185: isn't this `@Unroll` needed here to get all the test cases in the `where` clause?

ajordens: Oops, ya. Well it doesn't stop the tests from running it just runs them back-to-back. 

Re-enabled this in a subsequent PR.

PR #3960, opened 2020-10-09
fix(bake): Log when bake request amiSuffix and stage context amiSuffix values are different and resolve the difference

jonsie:
We encountered an issue wherein the `amiSuffix` sent in the bake request was different than the `amiSuffix` value set in the stage context.  I have not found a logical reason for how this could happen, but if it does happen let's just set it to what is in the stage context and log it out.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
@@ -92,8 +92,14 @@ class CreateBakeTask implements RetryableTask {
     try {
       // If the user has specified a base OS that is unrecognized by Rosco, this method will
       // throw a Retrofit exception (HTTP 404 Not Found)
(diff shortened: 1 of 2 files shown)


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -92,8 +92,14 @@ class CreateBakeTask implements RetryableTask {
     try {
       // If the user has specified a base OS that is unrecognized by Rosco, this method will
       // throw a Retrofit exception (HTTP 404 Not Found)
-      def bake = bakeFromContext(stage, bakery)
+      BakeRequest bake = bakeFromContext(stage, bakery)
       String rebake = shouldRebake(stage) ? "1" : null
+
+      if (stage.context.amiSuffix != null && bake.amiSuffix != null && bake.amiSuffix != stage.context.amiSuffix) {
```
jonsie, line 98: No trust.

cfieber (approved):

PR #3963, opened 2020-10-09
fix(preconfigured job): Fix for https://github.com/spinnaker/spinnaker/issues/6052

marchello2000:
This should, hopefully, resolve https://github.com/spinnaker/spinnaker/issues/6052 which was a likely a regression of https://github.com/spinnaker/orca/pull/3760
If the mapping is empty, skip it.


marchello2000:
I don't really grok the cases where `mapping` is `null`, or just in general how this flow works. So someone with more manifest job understanding - would really appreciate your feedback

cfieber (approved):

PR #3965, opened 2020-10-13
fix(webhook): Pass stage configuration data to rest template provider…

srekapalli:
- To allow better customizations of `RestTemplateProvider`,  adding `StageData` as an additional arg that we can pass into provider impls.  This will allow us to take decisions using `StageData` instead of just url that we currently pass into these providers.
- Also allows us to extract some other stage configuration data provided and use for further processing as needed.
- Moved all the cancel webhook logic into `WebhookService`
- `WebhookService` now will  issue all calls related to `Webhook` stage

### DIFF
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/RestTemplateProvider.java
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/RestTemplateProvider.java
@@ -27,7 +27,10 @@ import org.springframework.web.client.RestTemplate;
  */
 public interface RestTemplateProvider<T extends WebhookStage.StageData> {
 
-  /** @return true if this {@code RestTemplateProvider} supports the given url */
+  /**
+   * @return true if this {@code RestTemplateProvider} supports the given url and stage
+   *     configuration
+   */
   boolean supports(String targetUrl, T stageData);
 
   /**
--- a/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/RestTemplateProvider.java
+++ b/orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/RestTemplateProvider.java
@@ -16,11 +16,19 @@
 
 package com.netflix.spinnaker.orca.webhook.service;
 
+import com.netflix.spinnaker.orca.webhook.pipeline.WebhookStage;
 import org.springframework.web.client.RestTemplate;
 
-public interface RestTemplateProvider {
+/**
+ * Concrete implementations will provide a rest template customized to call webhook related
+ * endpoints.
+ *
+ * @param <T> stage configuration data.
+ */
+public interface RestTemplateProvider<T extends WebhookStage.StageData> {
+
   /** @return true if this {@code RestTemplateProvider} supports the given url */
-  boolean supports(String targetUrl);
+  boolean supports(String targetUrl, T stageData);
 
   /**
    * Provides an opportunity for a {@code RestTemplateProvider} to modify any aspect of the target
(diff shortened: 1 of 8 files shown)


orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/RestTemplateProvider.java
```diff
@@ -16,11 +16,19 @@
...
-public interface RestTemplateProvider {
+/**
+ * Concrete implementations will provide a rest template customized to call webhook related
+ * endpoints.
+ *
+ * @param <T> stage configuration data.
+ */
+public interface RestTemplateProvider<T extends WebhookStage.StageData> {
+
   /** @return true if this {@code RestTemplateProvider} supports the given url */
```
marchello2000, line 30:
```suggestion
  /** @return true if this {@code RestTemplateProvider} supports the stage with given parameters */
```

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
```diff
@@ -47,30 +58,83 @@ class WebhookService {
...
 
-    URI validatedUri = userConfiguredUrlRestrictions.validateURI(
-      restTemplateProvider.getTargetUrl(url)
-    )
-    HttpHeaders headers = buildHttpHeaders(customHeaders)
-    HttpEntity<Object> httpEntity = new HttpEntity<>(headers)
-    return restTemplateProvider.getRestTemplate(url).exchange(validatedUri, HttpMethod.GET, httpEntity, Object)
+  }
+
+  private RestTemplateData getRestTemplateData(String taskType, StageExecution stageExecution) {
```
marchello2000, line 104: nit: maybe instead of a "magic" string for `taskType`, use an enum? or the actual task class (but i don't think you have it here) so I think enum would be nicer

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
```diff
@@ -47,30 +58,83 @@ class WebhookService {
...
+    for (RestTemplateProvider provider : restTemplateProviders) {
+      WebhookStage.StageData stageData = stageExecution.mapTo(provider.getStageDataType())
```
marchello2000, line 106: this seems weird, what am I missing?
you are "mapping" to a dynamic type (from provider) but the receiving type is always `WebhookStage.StageData` (won't it always be this anyway?)

marchello2000: And if you do that, it seems that you could take pretty much everything other than the ".supports" call out of the loop?

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
```diff
@@ -47,30 +58,83 @@ class WebhookService {
...
+          break
+      }
+
+      // Return on the first match
+      if (destinationUrl != null && !destinationUrl.isEmpty() && provider.supports(destinationUrl, stageData)) {
+        URI validatedUri = userConfiguredUrlRestrictions.validateURI(provider.getTargetUrl(destinationUrl, stageData))
+        RestTemplate restTemplate = provider.getRestTemplate(destinationUrl)
+        return new RestTemplateData(restTemplate, validatedUri, httpMethod, payloadEntity, stageData)
+      }
+    }
```
marchello2000, line 137: I think it's worth adding a log (in addition to the exception you already have when no rest template found) and then you can also log the url/stage params, etc. for better debuggability.
Also, explicitly return `null` to make it less groovy magic

orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/service/WebhookService.groovy
srekapalli: > this seems weird, what am I missing?
> you are "mapping" to a dynamic type (from provider) but the receiving type is always `WebhookStage.StageData` (won't it always be this anyway?)

the dynamic type is always a subclass of `WebhookStage.StageData` and here I'm just upcasting it so that subclass types won't bleed into here .

marchello2000: Ah, got it

marchello2000 (approved):
lgtm

jonsie (approved):

PR #3967, opened 2020-10-14
fix(docker): Support finding image from a docker registry

jonsie:
This is part 1 of essentially a three part change (one more PR to clouddriver and then another internal change).  This will allow us to fallback properly when looking up docker images from clusters wherein the cluster doesn't exist but we still want to gracefully find an image.

CloudDriver OSS change: https://github.com/spinnaker/clouddriver/pull/5019

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
@@ -324,6 +324,14 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
             mkDeploymentDetail((String) image.imageName, (String) image.amis[location.value][0], deploymentDetailTemplate, config)
           ]
         }
+        //Docker registry to deployment detail conversion
+        else if (imageSummaries[location] == null && image.repository != null && image.tag != null) {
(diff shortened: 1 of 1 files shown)


caseyhebebrand (approved):

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
```diff
@@ -324,6 +324,14 @@ class FindImageFromClusterTask extends AbstractCloudProviderAwareTask implements
             mkDeploymentDetail((String) image.imageName, (String) image.amis[location.value][0], deploymentDetailTemplate, config)
           ]
         }
+        //Docker registry to deployment detail conversion
```
marchello2000, line 327: should this be scoped to `titus`? Just making sure it's safe for other providers... (i think so, but not 100% sure)

marchello2000 (approved):
seems logical to me!

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/FindImageFromClusterTask.groovy
jonsie: I didn't really want to add any specific cloud provider logic here.  This wasn't working with any of the docker centric cloud providers before anyways, so I feel like if someone wants to make this work with k8s they can follow this pattern.

PR #3968, opened 2020-10-14
fix(stages): Add missing stages from stage provider

jonsie:
Original implementation was rushed and a bit sloppy - no real issues with it but just wanted to clean this up.


cfieber (approved):
nice cleanup

PR #3970, opened 2020-10-15
feat(disable): put the disabled flag enforcement behind a feature toggle

dreynaud:
We have received user reports that the Google cloud provider does not set the `disabled` flag, so the task can't succeed in this case. Let's make enforcement configurable as a workaround for this issue.

fixes spinnaker/issues/6038


dreynaud:
cc @diasjorge

cfieber (approved):

PR #3974, opened 2020-10-21
chore(queue): Delete unmaintained hydrate command

robzienert:
This code never truly worked. Apparently someone in OSS found it and tried to run it.


dbyron0:
I learned about it from https://spinnaker.io/guides/runbooks/orca-zombie-executions/#rehydrate-the-queue.

robzienert:
Thanks! I'll need to delete that section from the docs as well.

robzienert:
Related: https://github.com/spinnaker/spinnaker.github.io/pull/2075

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/ZombieExecutionCheckingAgent.kt
```diff
@@ -40,7 +40,7 @@ import org.springframework.stereotype.Component
 @ConditionalOnExpression(
   "\${queue.zombie-check.enabled:false}"
 )
-@ConditionalOnBean(MonitorableQueue::class)
+@ConditionalOnBean(ZombieExecutionService::class)
```
marchello2000, line 43: i think you still need the conditional on queue since it's used by this class?

marchello2000 (approved):

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/metrics/ZombieExecutionCheckingAgent.kt
robzienert: Is it?

robzienert:
Still trying to figure out why tests are failing in github, but not locally...

PR #3976, opened 2020-10-22
fix(webhook): Capture response headers in the stage output

srekapalli:


orca-webhook/src/main/groovy/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.groovy
```diff
@@ -104,6 +104,9 @@ class MonitorWebhookTask implements OverridableTimeoutRetryableTask {
       log.error(errorMessage, e)
       Map<String, ?> outputs = originalResponse
       outputs.webhook.monitor << [error: errorMessage]
+      if (!e.responseHeaders?.isEmpty()) {
```
jonsie, line 107: Could you write this in the non-Groovy way?  Just to make it read more like Java.

```
e.getResponseHeaders() != null && !e.getResponseHeaders().isEmpty()
```

jonsie (approved):

PR #3978, opened 2020-10-26
fix(keel): read configuration for keel from services

nimakaviani:
This is to make it compatible with igor and other services
that have dependencies on keel

This is most likely a breaking change for Netflix
@luispollo @robfletcher


robzienert (approved):

robzienert:
Merging - added @luispollo and @robfletcher as reviewers just for some additional visibility. One of you two will need to update Orca's configs.

PR #3983, opened 2020-10-29
feat(pipeline executions/orca) : Added ability to add roles to manual…

sanopsmx:

Added ManualJudgmentAuthzGroupsUtil.groovy

Added a utlity method to check authorized groups of the manual judgment stage.


robzienert (changes requested):

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -72,21 +75,40 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
+    private final EchoService echoService
+
+    private final FiatPermissionEvaluator fiatPermissionEvaluator
+
+    WaitForManualJudgmentTask() {
+    }
+
+    WaitForManualJudgmentTask(Optional<FiatPermissionEvaluator> fpe) {
+      this.fiatPermissionEvaluator = fpe.orElse(null)
+    }
```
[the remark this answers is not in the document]
sanopsmx: done

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthzGroupsUtil.groovy
```diff
@@ -0,0 +1,50 @@
+/*
```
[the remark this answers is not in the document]
sanopsmx: done

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthzGroupsUtil.groovy
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright 2020 Netflix, Inc.
```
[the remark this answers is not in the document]
sanopsmx: done

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -204,6 +206,37 @@ class OperationsController {
...
+
+    def applicationName = pipeline.application
+    if (applicationName) {
+      Application application = front50Service.get(applicationName)
+      if (application) {
+        def username = AuthenticatedRequest.getSpinnakerUser().orElse("")
+        if (application.getPermission().permissions && application.getPermission().permissions.permissions) {
+          def permissions = objectMapper.convertValue(application.getPermission().permissions.permissions,
+              new TypeReference<Map<String, Object>>() {})
+          UserPermission.View permission = fiatPermissionEvaluator.getPermission(username);
```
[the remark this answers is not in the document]
sanopsmx: done

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -204,6 +206,37 @@ class OperationsController {
     }
   }
 
+  private void addStageAuthorizedRoles(def request, Map pipeline) {
+
+    def applicationName = pipeline.application
+    if (applicationName) {
+      Application application = front50Service.get(applicationName)
+      if (application) {
```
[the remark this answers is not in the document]
sanopsmx: done

sanopsmx:
Addressed all the feedback and review.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -95,12 +110,31 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
 
       return TaskResult.builder(executionStatus).context(outputs).build()
     }
 
+    boolean checkManualJudgmentAuthorizedGroups(def stageRoles, def permissions, def username) {
+
+      if (!Strings.isNullOrEmpty(username)) {
+        UserPermission.View permission = fiatPermissionEvaluator.getPermission(username);
+        if (permission == null) { // Should never happen?
+          return false;
```
robzienert, line 128: If this should never happen, I presume we want to know when it does happen? Perhaps would be useful to have a warn-level log about it?

`log.warn("Attempted to get user permission for '$username' but none were found."` or something.

orca-echo/src/main/java/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthzGroupsUtil.java
```diff
@@ -0,0 +1,51 @@
...
+package com.netflix.spinnaker.orca.echo.util;
+
+import com.netflix.spinnaker.fiat.model.Authorization;
+import java.util.List;
+import java.util.Map;
+
+public class ManualJudgmentAuthzGroupsUtil {
+
+  public static boolean checkAuthorizedGroups(
+      List<String> userRoles, List<String> stageRoles, Map<String, Object> permissions) {
```
robzienert, line 26: Expected to see a test for this class.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -204,6 +206,53 @@ class OperationsController {
...
   }
 
+  private void addStageAuthorizedRoles(def request, Map pipeline) {
+
+    if (fiatStatus && fiatStatus.isEnabled()) {
+      def applicationName = pipeline.application
+      if (applicationName) {
+        Optional<Application> optApplication = getApplication(applicationName)
+        if (optApplication.isPresent()) {
+          Application application = optApplication.get();
```
robzienert, line 216: nit: Could be simplified to `getApplication(applicationName).ifPresent(application -> { /* ... */ });`

robzienert (commented):
Would like to see a test for the utility class, as well as some coverage for the controller logic that has been added. After that, we're looking pretty good, thanks!

orca-echo/src/main/java/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthzGroupsUtil.java
sanopsmx: 6 test cases are written/added in ManualJudgmentStageSpec.groovy only to test this logic. Please have a look at this test case.

@Unroll
  void "should return execution status based on authorizedGroups"() {

robzienert: Seems weird to have a test case for a different class (ManualJudgmentStage) specifically testing a different class, doesn't it?

sanopsmx: We are using the logic in the ManualJudgmentStage.groovy. So i have written 6 test cases to check the  ManualJudgmentAuthzGroupsUtil.java logic in the ManualJudgmentStageSpec.groovy.

robzienert: You're also using it in a controller, which means that the test cases don't belong where there are today.

I'd say the tests that are testing this code should live in `ManualJudgmentAuthzGroupsUtilSpec`, then test the implementation in `ManualJudgmentStageSpec` and `OperationsControllerSpec`, as you would with any other unit under test.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: done

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
sanopsmx: done

orca-echo/src/main/java/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthzGroupsUtil.java
sanopsmx: done.

sanopsmx:
Can you review the code

sanopsmx:
Raised a new PR.

https://github.com/spinnaker/orca/pull/3988

PR #3986, opened 2020-11-02
perf(sql): Store child pipelines as references

robzienert:
Changes pipeline triggers to be stored as a reference, rather than as the full execution.
This makes individual pipelines smaller to store, trading additional SQL queries to hydrate
the full `PipelineExecution`.


orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -830,10 +832,39 @@ class SqlExecutionRepository(
         stages.forEach { storeStageInternal(ctx, it, executionId) }
       }
     } finally {
+      // Restore original object state.
```
cfieber, line 835: the fact that we need to do this seems a bit wild to me - I wonder if it's worth just objectMapper copying the execution on the way in here so we are free to mutate it as we see fit..

unless the expectation is that the repository mutates to execution to set Ids on it or something...

anyway not super relevate to the overall change in this PR

cfieber:
the approach definitely makes sense to me

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/ExecutionMapper.kt
```diff
@@ -96,4 +102,31 @@ class ExecutionMapper(
...
+        .where(field("id").eq(trigger.parentExecutionId))
+        .fetchExecutions(mapper, 200, context)
+        .firstOrNull()
+
+      if (parentExecution == null) {
+        // If someone deletes the parent execution, we'll be unable to load the full, valid child pipeline. Rather than
+        // throw an exception, we'll continue to load the execution with [PipelineRefTrigger] and let downstream
+        // consumers throw exceptions if they need to. We don't want to throw here as it would break pipeline list
+        // operations, etc.
+        log.error("Attempted to load parent execution for '${execution.id}', but it no longer exists: ${trigger.parentExecutionId}")
```
marchello2000, line 125: nit: I would make this a `warn` so it doesn't show up sentry, etc.
Also, it's not really an error here, as you state. It's an error at a time when someone downstream needs it

marchello2000:
I remember someone mentioning that some users have (dozens/hundreds?!) parent pipelines. What happens in those cases, does the perf suffer a lot? Otherwise, i think it makes a ton of sense

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
robzienert: This was an optimization. Saving the execution is a high-throughput (as far as Orca is concerned) operation, and serialization of executions is very high cost.

robzienert:
@marchello2000 Perf suffers to the point until where MySQL's (or any SQL's) buffer space runs out -- in some pathological community deployments, people are seeing pipeline payloads going as high as 1GB because of the embedded pipeline trigger. When this buffer space runs out, it causes the query to fail (and as a result, anything in Orca failing, since its an irrecoverable error until someone increases the buffer size... which is not always possible).

ajordens:
This is sensible to me.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.JsonNode
+import com.netflix.spinnaker.kork.exceptions.SystemException
+import com.netflix.spinnaker.orca.api.pipeline.models.Trigger
+import com.netflix.spinnaker.orca.pipeline.model.support.CustomTriggerDeserializerSupplier
+import com.netflix.spinnaker.orca.pipeline.model.support.listValue
+import com.netflix.spinnaker.orca.pipeline.model.support.mapValue
+import com.netflix.spinnaker.orca.sql.pipeline.persistence.PipelineRefTrigger
+
+class PipelineRefTriggerDeserializerSupplier : CustomTriggerDeserializerSupplier {
```
robzienert, line 27: This needed to be made as a `CustomTriggerDeserializerSupplier` because `TriggerDeserializer` is in `orca-core`, whereas `PipelineRefTrigger` is an implementation detail of `orca-sql` only.

robzienert (commented):
This PR is now ready for review.

dilippai:
I would encourage the community to not consider use cases like nested pipelines as pathological or abnormal. As Spinnaker adoption grows, especially in the enterprise, you will find usage patterns that are different from yours -- an open source pipelining system needs to be at least somewhat flexible to pipeline patterns that users are naturally likely to gravitate towards.

This is an issue that has become existential for us (Salesforce). I'm not sure why the decision was made to store json as a blob in Orca, but this is leading to huge performance issues culminating in failures.

Nirmalyasen:
Since this is reverted back, what is the solution? Is there another PR to address the issue? 

This is becoming a big issue. Also, somewhere in 1.20, the no of artifacts produced out of a deployment manifest stage has increased many folds. So, that compounds the issue of child contexts in the pipelines.

PR #3988, opened 2020-11-03

sanopsmx:

If yes/no, set a isAuthorized flag to true/false by checking the userRoles, stageRoles and application Roles.

Added ManualJudgmentAuthzGroupsUtil.groovy

Added a utlity method to check authorized groups of the manual judgment stage.

Added ManualJudgmentAuthzGroupsUtilSpec.groovy

Added test cases check authorized groups method of ManualJudgmentAuthzGroupsUtil.groovy.


orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -95,12 +112,32 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
+        notificationState = "manualJudgment"
+        executionStatus = ExecutionStatus.RUNNING
+        stage.context.put("judgmentStatus", "")
+      }
       Map outputs = processNotifications(stage, stageData, notificationState)
 
       return TaskResult.builder(executionStatus).context(outputs).build()
     }
 
+    boolean checkManualJudgmentAuthorizedGroups(def stageRoles, def permissions, def username) {
```
ajordens, line 125: Please type these parameters (even if it's groovy)

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -72,21 +75,35 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
+      this.fiatPermissionEvaluator = fpe.orElse(null)
+    }
 
     @Override
     TaskResult execute(StageExecution stage) {
       StageData stageData = stage.mapTo(StageData)
+      def stageAuthorized = stage.context.get('isAuthorized')
+      def stageRoles = stage.context.get('stageRoles')
+      def permissions = stage.context.get('permissions')
+      def username = stage.lastModified? stage.lastModified.user: stage.context.get('username')
```
ajordens, line 93: I think you should be able to fetch the user from `AuthenticatedRequest.getSpinnakerUser()`.

Defaulting to `stage.context.get('username')` seems suspicious in this case.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -204,6 +205,52 @@ class OperationsController {
     }
   }
 
+  private void addStageAuthorizedRoles(def request, Map pipeline) {
```
ajordens, line 208: I don't understand what this is doing, or why it's doing it when we're starting a pipeline.

Is the idea not that a manual judgment stage can be configured with a set of roles needed to judge it?

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -72,21 +75,35 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
+
+    WaitForManualJudgmentTask(Optional<EchoService> echoService, Optional<FiatPermissionEvaluator> fpe) {
+      this.echoService = echoService.orElse(null)
+      this.fiatPermissionEvaluator = fpe.orElse(null)
+    }
 
     @Override
     TaskResult execute(StageExecution stage) {
       StageData stageData = stage.mapTo(StageData)
+      def stageAuthorized = stage.context.get('isAuthorized')
```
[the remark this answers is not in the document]
sanopsmx: done.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: done

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: I should have deleted it. Kept it for executing the test cases. Modified the test case.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
sanopsmx: This method will check for manual judgment stages and will find out if the user is authorized to continue. Will set the isAuthorized flag to true , if authorized,

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
ajordens: I still don't believe accessing `stage.lastModified` directly is correct, what does AuthenticatedRequest.getSpinnakerUser() return? 

I _believe_ it should return the correct user.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -72,21 +76,34 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
     final long backoffPeriod = 15000
     final long timeout = TimeUnit.DAYS.toMillis(3)
 
-    @Autowired(required = false)
-    EchoService echoService
+    private final EchoService echoService
+
+    private final FiatPermissionEvaluator fiatPermissionEvaluator
+
+    WaitForManualJudgmentTask(Optional<EchoService> echoService, Optional<FiatPermissionEvaluator> fpe) {
```
ajordens, line 83: No `@Autowired` annotation needed on this constructor?

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ajordens: It's still unclear why it needs to happen in the `OperationsController`. I'm confused as this code seems to be manipulating every stage in the pipeline?

Overall this implementation is hard to follow.

My expectation was that the manual judgment stage would have a set of required roles in its context, and would check against those roles when determining whether or not to continue. That shouldn't require changes anywhere other than the manual judgement stage/task.

As an aside, there is a small loophole here in that any stage context can be PATCH'd. See https://github.com/spinnaker/orca/blob/master/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy#L475.

orca-echo/src/test/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStageSpec.groovy
```diff
@@ -27,31 +30,67 @@ import static com.netflix.spinnaker.orca.echo.pipeline.ManualJudgmentStage.Notif
...
+    then:
+    1 * fpe.getPermission('abc@somedomain.io') >> {
+      new UserPermission().addResources([new Role('foo'), new Role('baz')]).view
+    }
+    result.status == expectedStatus
+
+    where:
+    context                      || expectedStatus
+    [judgmentStatus: "continue", isAuthorized: false, stageRoles: ['foo'],
+     permissions: [WRITE: ["foo"],READ: ["foo","baz"], EXECUTE: ["foo"]]] || ExecutionStatus.SUCCEEDED
```
ajordens, line 78: What is the difference between `permissions` and `stageRoles`? I thought the intention was you needed to have a particular role (or roles?). 

The reason for WRITE/READ/EXECUTE isn't clear.

sanopsmx: stageRoles --> Roles configured at  the manual judgment stage.
permissions --> roles configured during the application.creation.

Please refer the design document table for more clear picture.
https://github.com/spinnaker/governance/pull/80

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: There is no autowired annotation on this constructor. Please check the file again.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: stage.lastModified returns the user, who clicked on the 'continue' or 'stop' button of the manual judgment stage.

this should work. it returns the correct user. 

If you still insist, i can change the code.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
sanopsmx: It doesn't manipuate every stage. It only manipulates manual judgment stages.

Please refer the design document.
https://github.com/spinnaker/governance/pull/80

ajordens:
```
def stageList = pipeline.stages
            def stageRoles = []
            stageList.each { item ->
              stageRoles = item.selectedStageRoles
              item.isAuthorized = ManualJudgmentAuthzGroupsUtil.checkAuthorizedGroups(userRoles, stageRoles, permissions)
              item.stageRoles = stageRoles
              item.permissions = permissions
            }
```

What am I missing here, that's iterating across `pipeline.stages`.

sanopsmx: item.selectedStageRoles returns null for all the stages except manual judgment(if configured).
This decision to apply for all stages was taken in our design meet, so that in future, if we want to extend it to all stages, this will work.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
ajordens: I realize there isn't, but this class was changed from field injection to constructor injection and for consistency we usually add an `@Autowired` annotation to the constructor.

orca-echo/src/test/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStageSpec.groovy
ajordens: > Display all the groups the user has selected while creating application(permissions) in the authorized groups dropdown.

Sure. But that set of application permissions should just be looked up from the UI, it doesn't need to be stored in the stage context. 

I'd like this PR to really be centered around the managed delivery stage and required user roles needed to act on it.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
ajordens: If we wanted to apply similar logic to all stages, we would do so with a different implementation (likely a preprocessor that could run before task execution).

`selectedStageRoles` may be null, but you're also manipulating the context for every stage with `stageRoles` and `permissions`.

It also introduces another remote call in the flow to run a pipeline (to `front50`) which is even with this feature as currently implemented would be unnecessary to anyone who isn't using roles in manual judgment stages. As a general rule, we try to remove unnecessary rpc calls on critical paths.

I'd like this changed, and really this bit removed.

ajordens (changes requested):

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
ajordens: Yes, `lastModified` is an implementation detail. My preference would be to avoid referencing it directly.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
ajordens: Most of the project is going to have `@Autowired` on `@Component` constructors. I believe it works in this case because the class has only one constructor, but the norm as far as I'm concerned is to annotate the constructors even if there is only one.

This is a minor point on this PR.

orca-echo/src/test/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStageSpec.groovy
sanopsmx: I am unable to follow you here. Did not understand what you meant? Can you be more specific.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: Should i use @Autowired or not? Since, this is a minor point. I am dropping this feedback.

orca-echo/src/test/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStageSpec.groovy
ajordens: Sorry, for some odd reason I typed "managed delivery" when I meant "manual judgment". I blame the late hour, sorry :(

I think what would be clearer is if I did a quick prototype/pr of what is in my mind, and we can compare. I'll do that this morning.

ajordens:
Spent a few minutes this morning and put together https://github.com/spinnaker/orca/pull/3991.

How different are the intentions here than that? 

The `PATCH` call already verifies that the user has `EXECUTE` against the application, so I don't think any further validation is needed.

I did mention earlier that this only works in the happy path but that `PATCH` operation does allow for the set of required roles to be overwritten. I believe a more legitimate fix for that would be to split immutable inputs out of the stage context. No timeline on when that will happen.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
robzienert: I asked you to remove the `@Autowired` properties in favor of a [single constructor](https://github.com/spinnaker/orca/pull/3947/files#r513595076). My example did not include an `@Autowired` annotation, but that wasn't really intended to be copy/pasted, either. While `@Autowired` on a single constructor is optional, per @ajordens, precedent in the rest of the codebase does have us adding `@Autowired` on single constructors anyway, probably for explicitness.

This is a good example for why keeping a single PR for a change is valuable: I had to click through 4 different PRs to find the original thread of feedback about this. Had it been a single PR, context for other reviewers would've been easily discoverable.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: done...fetching user from AuthenticatedRequest.getSpinnakerUser().

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
sanopsmx: removed this piece of code completely now.

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: done..added @Autowired

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -73,20 +81,53 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
     final long timeout = TimeUnit.DAYS.toMillis(3)
 
     @Autowired(required = false)
-    EchoService echoService
+    private final EchoService echoService
+
+    @Autowired(required = false)
```
ajordens, line 86: This is still not correct.

The `@Autowired` should be on the constructor and not these attributes.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -104,6 +100,9 @@ class OperationsController {
   @Autowired(required = false)
   Front50Service front50Service
 
+  @Autowired(required = false)
+  private FiatPermissionEvaluator fiatPermissionEvaluator
```
ajordens, line 104: This does not seem needed.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/web/config/WebConfiguration.groovy
```diff
@@ -54,6 +56,11 @@ class WebConfiguration {
     return  new MetricsInterceptor(registry, "controller.invocations", ["application"], ["BasicErrorController"])
   }
 
+  @Bean
+  ManualJudgmentAuthzGroupsUtil mnlJudgUtil(Optional<Front50Service> front50Service) {
```
ajordens, line 60: Please do not truncate `mnl` and `Judg` when naming beans ... `s/mnlJudgUtl/ManualJugementAuthorizationUtil` (or something.

orca-echo/src/main/java/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthzGroupsUtil.java
```diff
@@ -0,0 +1,81 @@
...
+public class ManualJudgmentAuthzGroupsUtil {
+
+  @Autowired(required = false)
+  Front50Service front50Service;
+
+  public ManualJudgmentAuthzGroupsUtil(Optional<Front50Service> front50Service) {
+    this.front50Service = front50Service.orElse(null);
+  }
+
+  public static boolean checkAuthorizedGroups(
```
ajordens, line 41: You still have not added any inline comments here around what this function is intended to do.

I don't quite understand why we're checking against application permissions when the manual judgment could not have happened if the user didn't have EXECUTE.

```
@PreAuthorize("hasPermission(this.getPipeline(#id)?.application, 'APPLICATION', 'EXECUTE')")
@RequestMapping(value = "/pipelines/{id}/stages/{stageId}", method = RequestMethod.PATCH)
PipelineExecution updatePipelineStage(
```

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -73,20 +81,53 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
...
     @Override
     TaskResult execute(StageExecution stage) {
       StageData stageData = stage.mapTo(StageData)
+      def username = AuthenticatedRequest.getSpinnakerUser().orElse(stage.lastModified ? stage.lastModified.user : "")
+      boolean fiatEnabled = fiatStatus ? fiatStatus.isEnabled() : false
+      boolean isAuthorized = false
+      def appPermissions
+      def stageRoles
+      if (fiatEnabled) {
+        appPermissions = getApplicationPermissions(stage)
```
ajordens, line 117: I'd rather we not introduce a call to `front50` if there aren't any required roles here. The majority of use-cases (and certainly everything pre-existing) will not need this.

ajordens (changes requested):

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
sanopsmx: done...

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: I removed, but i find the @Autowired in many places. This should not be a concern.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/web/config/WebConfiguration.groovy
sanopsmx: done

orca-echo/src/main/java/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthzGroupsUtil.java
sanopsmx: Added inline comments.

Application permissions has to be 'CREATE, WRITE, EXECUTE' to proceed further in the manual judgment stage.

This method checks if the logged in user has role in the manual judgment stage authorized groups.
   We fetch the user roles and check if that role is authorized in the manual judgment stage role.
   if the user role exists , then we check with the application permission roles.
   If the application permission role has 'READ' then we return false(not authorized)
   If the application permission role has 'CREATE, EXECUTE, WRITE' then we return true(authorized)

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
sanopsmx: done..

ajordens (approved):
Going to merge this down and fast follow-up with a few stylistic tweaks.

I'll tag @sanopsmx on the follow-up PR.

ajordens:
I believe there are still some logical issues with this PR ... going to revert it and let's continue the effort over on https://github.com/spinnaker/orca/pull/4023.

sanopsmx:
@robzienert ,
@ajordens 

When is the spinnaker 1.25 release date? Is this fix going to be in 1.25 release?

sanopsmx:
> @robzienert ,
> @ajordens
> 
> When is the spinnaker 1.25 release date? Is this fix going to be in 1.25 release?

Even this deck fix should be in this release.
https://github.com/spinnaker/deck/pull/8779

PR #3989, opened 2020-11-05
feat(pipeline executions/orca): Add support for explicit opt-out from…

titirigaiulian:
… passing outputs to child executions

This PR introduces a feature flag `skipDownstreamOutput` that can be defined in `Run Pipeline` stage. By enabling it, clears the output of the parent pipeline, without affecting the child pipeline. 
This helps for stability issues when there are large deployment pipelines with multiple levels of parents. It can be set any parent level of the hierarchy.

Issue: https://github.com/spinnaker/spinnaker/issues/6159

### DIFF
--- a/orca-clouddriver/src/test/kotlin/com/netflix/spinnaker/orca/clouddriver/pipeline/KubernetesPreconfiguredJobSpec.kt
+++ b/orca-clouddriver/src/test/kotlin/com/netflix/spinnaker/orca/clouddriver/pipeline/KubernetesPreconfiguredJobSpec.kt
@@ -85,7 +85,7 @@ class KubernetesPreconfiguredJobSpec : JUnit5Minutests {
             .contains("\"ref\":\"/pipelines")
         }
 
-        verify(timeout = 1500) { katoRestService.requestOperations(any(), "kubernetes", match { it.toString().contains("alias=preconfiguredJob") }) }
+        verify(timeout = 2000) { katoRestService.requestOperations(any(), "kubernetes", match { it.toString().contains("alias=preconfiguredJob") }) }
       }
     }
   }
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
@@ -21,6 +21,7 @@ import static java.util.Collections.emptyMap;
 
 import com.netflix.spinnaker.orca.api.pipeline.CancellableStage;
 import com.netflix.spinnaker.orca.api.pipeline.graph.StageDefinitionBuilder;
+import com.netflix.spinnaker.orca.api.pipeline.graph.StageGraphBuilder;
 import com.netflix.spinnaker.orca.api.pipeline.graph.TaskNode;
 import com.netflix.spinnaker.orca.api.pipeline.models.PipelineExecution;
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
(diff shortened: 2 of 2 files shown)


anthonywoo:
+1 for this option. We have a pipeline orchestrating other pipelines which results in huge writes to our RDS database

 What do we need to do to get this feature pushed forward?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
```diff
@@ -71,6 +72,17 @@ public void taskGraph(@Nonnull StageExecution stage, @Nonnull TaskNode.Builder b
     }
   }
 
+  @Override
+  public void afterStages(@Nonnull StageExecution parent, @Nonnull StageGraphBuilder graph) {
+    if (parent
```
costimuraru, line 77: Nitpick: to make the `if` statement easier to understand, what do you think about extracting the condition in a dedicated method?
Something like:


```java
  @Override
  public void afterStages(@Nonnull StageExecution parent, @Nonnull StageGraphBuilder graph) {
    if (shouldSkipDownstreamOutput(parent)) {
      parent.setOutputs(emptyMap());
    }
  }

  private void shouldSkipDownstreamOutput(StageExecution parent) {
    return parent
        .getContext()
        .getOrDefault("skipDownstreamOutput", "false")
        .toString()
        .equals("true")
  }
```

Nirmalyasen:
spinnaker/spinnaker#6159 is affecting quite a few of us. This fix does reduce the payload and does not force anything on everybody. Users can decide if they want the outputs propagated or not. Can this be accepted to help Spinnaker scale?

emagana:
We are facing the same issue and this looks like a clean and easy fix. Maybe just waiting for @titirigaiulian  to make the changes suggested in the code review. Thanks!

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
```diff
@@ -71,6 +72,17 @@ public void taskGraph(@Nonnull StageExecution stage, @Nonnull TaskNode.Builder b
     }
   }
 
+  @Override
+  public void afterStages(@Nonnull StageExecution parent, @Nonnull StageGraphBuilder graph) {
```
Nirmalyasen, line 76: That parameter parent is really the current stage itself and not the parent. So, no calling it parent will reduce some confusion while reading the code.

titirigaiulian: done

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
titirigaiulian: done

orca-clouddriver/src/test/kotlin/com/netflix/spinnaker/orca/clouddriver/pipeline/KubernetesPreconfiguredJobSpec.kt
```diff
@@ -85,7 +85,7 @@ class KubernetesPreconfiguredJobSpec : JUnit5Minutests {
             .contains("\"ref\":\"/pipelines")
         }
 
-        verify(timeout = 1500) { katoRestService.requestOperations(any(), "kubernetes", match { it.toString().contains("alias=preconfiguredJob") }) }
+        verify(timeout = 2000) { katoRestService.requestOperations(any(), "kubernetes", match { it.toString().contains("alias=preconfiguredJob") }) }
```
titirigaiulian, line 88: this caused some build failures

anthonywoo:
Any updates on this?

We've been using our own version of orca (`1.22.2 `) which contains this PR. We haven't had any issues and our pipelines page is more stable now

dbrougham:
@costimuraru - Just wondering if there is anything outstanding that is preventing this from being merged in? We have noticed a similar issue in our prod Spinnaker instance with nested pipelines

robzienert (approved):
Sorry I missed this!

dbrougham:
Thanks @robzienert for taking a look 🙌

PR #3990, opened 2020-11-05
fix(waitforup): Don't use the min of snapshot capacity

marchello2000:
There are situations where we capture the snapshot capacity that is wrong. For example, if we resize a cluster `1->10`, then we do a force cache refresh and if that doesn't actually get the new state of `10`,
`WaitForUp` will wait for `1` instead of `10` instances (we defer to `WaitForUp` from `WaitForCapacityMatch`).
The other situation is when we do a rollback to a `0` size ASG, the first step of resizing it will resize it to correct rollback number, but since `WaitForUp` will succeed immediately we will tear down the old ASG and thus leave a rollback with no ASG ready to take traffic.
This is mostly an issue on Titus because `ForceRefresh` is a no-op on Titus.

It appears there already was an attempt to "fixing" this issue in https://github.com/spinnaker/orca/pull/2923, however it doesn't seem to cover all the cases.

This is an initial attempt that still needs figuring out that it doesn't break any other scenarios

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForUpInstancesTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForUpInstancesTask.groovy
@@ -127,15 +127,6 @@ class WaitForUpInstancesTask extends AbstractWaitingForInstancesTask {
       splainer.add("setting targetDesiredSize=${targetDesiredSize} from the desired size in current serverGroup capacity=${currentCapacity}")
     }
 
-    if (stage.context.capacitySnapshot) {
-      Integer snapshotCapacity = ((Map) stage.context.capacitySnapshot).desiredCapacity as Integer
-      // if the server group is being actively scaled down, this operation might never complete,
-      // so take the min of the latest capacity from the server group and the snapshot
-      def newTargetDesiredSize = Math.min(targetDesiredSize, snapshotCapacity)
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForUpInstancesTask.groovy
```diff
@@ -127,15 +127,6 @@ class WaitForUpInstancesTask extends AbstractWaitingForInstancesTask {
       splainer.add("setting targetDesiredSize=${targetDesiredSize} from the desired size in current serverGroup capacity=${currentCapacity}")
     }
 
-    if (stage.context.capacitySnapshot) {
```
marchello2000, line 130: This shouldn't be necessary anymore because we pin during deploys... but....

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/instance/WaitForUpInstancesTaskSpec.groovy
```diff
@@ -325,15 +323,14 @@ class WaitForUpInstancesTaskSpec extends Specification {
     )
 
     where:
-    result || useSource | healthy | configured | snapshot | asg | description
```
marchello2000, line 328: some of these tests fail - i started cleaning it up to understand what they actually test - `useSourceCapacity` isn't used anyway, so removing it

marchello2000:
@dreynaud @jonsie sorry I didn't get this to completion. Here is the very meager start.

dreynaud (approved):

jonsie (approved):

PR #3991, opened 2020-11-05
feat(echo): Support for a set of required roles in the Manual Judgment stage

ajordens:
The judging user _must_ have all specified roles in order for the
judgment to take.

If the user is missing one or more roles, the judgment status will be
cleared.

Supports spinnaker/spinnaker#4792


ajordens:
Relates to the conversation on https://github.com/spinnaker/orca/pull/3988.

PR #3993, opened 2020-11-05
fix(json): remove `pinned` from serialized Capacity objects

dreynaud:
It was meant as a helper function, but actually becomes a field in serialized server groups. It can make its way to deck and be confusing to users.


emjburns (approved):
This confused me in the past! Thank you.

aravindmd (approved):

ajordens (approved):

deverton:
This probably obsoletes #3923 which can be reverted.

PR #3995, opened 2020-11-11
fix(misc): Allow monitor only webhook tasks to default to webhook url…

srekapalli:
- Allow monitor only webhook tasks to default to webhook url if status endpoint is not provided


dreynaud (approved):
LGTM, thanks!

PR #3996, opened 2020-11-11
Add draining health state

dreynaud:
Clouddriver is now exposing that state in https://github.com/spinnaker/clouddriver/pull/5060

This models `draining` as:
- up-negative, i.e. it will prevent wait-for-up checks from succeeding (to prevent issues like https://github.com/spinnaker/spinnaker/issues/5867)
- down-negative, i.e. it will prevent wait-for-down checks from succeeding. If it did, we would end up terminating instances in the draining state even though they are still potentially doing work. This could lead to errors or data loss.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/model/HealthState.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/model/HealthState.java
@@ -27,7 +27,8 @@ public enum HealthState {
   Unknown,
   Starting,
   Succeeded,
-  Up;
+  Up,
+  Draining;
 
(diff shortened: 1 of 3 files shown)


jonsie (approved):
Appears to make sense.  Though I had to read these helper method names like 10 times.

aravindmd (approved):

cfieber (approved):

dreynaud:
> Appears to make sense, though I had to read these helper method names like 10 times.

Difficult indeed. I did not want to pile up with `someAreUpAndNoneAreDownOrStartingOrDraining` 😅 Eventually I'd like to move away from that `some are X and none are Y` model, but for the moment that's the world we live in

PR #3998, opened 2020-11-12
feat(core): Add SpEL helper to get all resolved stage artifacts

aufarg:
Hi, this PR adds a spell helper function called resolvedArtifacts which will return all the resolved stage artifacts. Further filtering of artifacts can be done with '.?' or other SpEL built-in features.

Closes: https://github.com/spinnaker/spinnaker/issues/5993


aufarg:
The change is a bit different from what was asked in the issue, because it'll just return all the resolved artifacts without any filters. I can change it to receive a name/type like `triggerResolvedArtifact(name)` and `triggerResolvedArtifactByType(type)` if you think that would be better.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ArtifactExpressionFunctionProvider.java
```diff
@@ -42,16 +45,21 @@ public Functions getFunctions() {
     return new Functions(
         new FunctionDefinition(
             "triggerResolvedArtifact",
-            "Looks up the an artifact in current execution given its name. If multiple artifacts are found, only 1 will be returned.",
+            "Looks up a resolved artifact in current execution given its name. If multiple artifacts are found, only 1 will be returned.",
```
dbyron0, line 48: To match the change below?
```suggestion
            "Looks up a resolved artifact in the current execution trigger given its name. If multiple artifacts are found, only 1 will be returned.",
```

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ArtifactExpressionFunctionProvider.java
```diff
@@ -42,16 +45,21 @@ public Functions getFunctions() {
...
             new FunctionParameter(
                 PipelineExecution.class, "execution", "The execution to search for artifacts"),
             new FunctionParameter(String.class, "name", "The name of the resolved artifact")),
         new FunctionDefinition(
             "triggerResolvedArtifactByType",
-            "Looks up the an artifact in current execution given its type. If multiple artifacts are found, only 1 will be returned.",
+            "Looks up a resolved artifact in current execution trigger given its type. If multiple artifacts are found, only 1 will be returned.",
```
dbyron0, line 54: nit
```suggestion
            "Looks up a resolved artifact in the current execution trigger given its type. If multiple artifacts are found, only 1 will be returned.",
```

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ArtifactExpressionFunctionProvider.java
```diff
@@ -42,16 +45,21 @@ public Functions getFunctions() {
...
             new FunctionParameter(
                 PipelineExecution.class, "execution", "The execution to search for artifacts"),
-            new FunctionParameter(String.class, "type", "The type of the resolved artifact")));
+            new FunctionParameter(String.class, "type", "The type of the resolved artifact")),
+        new FunctionDefinition(
+            "resolvedArtifacts",
+            "Looks up resolved artifacts in current execution.",
```
dbyron0, line 60: nit
```suggestion
            "Looks up resolved artifacts in the current execution.",
```

dbyron0:
Thanks for those changes!

robzienert (approved):
Nice.

PR #4000, opened 2020-11-14
fix(disable): skip disabled check for partially disabled server groups

dreynaud:
Rolling red black and monitored deployments create disable server group stages with a desiredPercentage, which only disables some instances and not the whole server group. In that case, we need to skip the check that the server group is disabled.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForDisabledServerGroupTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForDisabledServerGroupTask.java
@@ -79,10 +79,8 @@ public class WaitForDisabledServerGroupTask extends AbstractCloudProviderAwareTa
   }
 
   // RRB and Monitored Deployments do "partial disables", i.e. they run DisableServerGroupTask with
-  // a `desiredPercentage`
-  // which will only disable some instances, not the entire server group (so this won't set the
-  // `disabled` flag on the
-  // server group)
+  // a `desiredPercentage` which will only disable some instances, not the entire server group (so
+  // this won't set the `disabled` flag on the server group)
   private boolean isPartialDisable(TaskInput input) {
     return input.desiredPercentage != null && input.desiredPercentage < 100;
   }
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForDisabledServerGroupTask.java
```diff
@@ -64,6 +69,23 @@ public TaskResult execute(@NotNull StageExecution stage) {
...
+    if (desiredPercentage instanceof Integer) {
+      return ((Integer) desiredPercentage) < 100;
+    }
+
+    if (desiredPercentage instanceof String) {
+      return Integer.parseInt((String) desiredPercentage) < 100;
+    }
+
+    log.warn("Unexpected value in stage context, desiredPercentage={}", desiredPercentage);
+    return true;
```
ajordens, line 86: `stage.mapTo()` (and a try/catch) would probably save these conversations. 

Should the default be a partial disable if there's something wonky in `desiredPercentage`? That doesn't seem right.

dreynaud: ah, I always assumed `stage.mapTo()` was Groovy shenanigans

The default sounds wonky but it really means "when in doubt, don't enforce anything"

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForDisabledServerGroupTask.java
```diff
@@ -64,6 +69,23 @@ public TaskResult execute(@NotNull StageExecution stage) {
     }
   }
 
+  private boolean isPartialDisable(Object desiredPercentage) {
```
jonsie, line 72: Could you add a comment here such that we know this behavior is for rolling redblack and monitored deployment strategies?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForDisabledServerGroupTask.java
dreynaud: in f9a1dadbe717be6b35e9ac52d702d643bcfe6988

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForDisabledServerGroupTask.java
dreynaud: ya good point

dreynaud: in 13c608c5310f3dfc4c96cf11ad53907c4fbc4378

aravindmd (approved):

PR #4004, opened 2020-11-17
fix(script stage): Retry if igor's build endpoint returns an invalid identifier

julienduchesne:
When lots of jobs are launched in parallel, sometimes the identifier is either null or "null" (Don't know which, formatting either in a string will result in the same output)
Orca then fails when polling the queued job:
HTTP 400 http://spin-igor.spinnaker:8088/builds/queue/jenkins.dep/null (2ms)

In the code, the "execute" method will not directly fail if igor returns an empty queuedBuild id. This is an attempt to fix that

Let me know if that looks acceptable. As we have more and more Spinnaker pipelines calling our Jenkins instance, this is starting to happen more often


julienduchesne:
Fixed by this: https://github.com/spinnaker/igor/pull/921

PR #4006, opened 2020-11-18
feat(managed-delivery): Delete Managed Delivery data upon application deletion

luispollo:
Adds a call from the `DeleteApplicationTask` to Keel to delete any associated records for Managed Delivery.

Closes https://github.com/spinnaker/keel/issues/1567

### DIFF
--- a/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
+++ b/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
@@ -51,7 +51,10 @@ class DeleteApplicationTask extends AbstractFront50Task {
           return TaskResult.builder(ExecutionStatus.TERMINAL).outputs(outputs).build()
         }
         // delete Managed Delivery data
-        keelService.deleteDeliveryConfig(application.name)
+        if (keelService != null) {
+          log.debug("Deleting Managed Delivery data for {}", application.name)
+          keelService.deleteDeliveryConfig(application.name)
+        }
       }
     } catch (RetrofitError e) {
       if (e.response?.status == 404) {
--- a/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
+++ b/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
@@ -50,9 +50,9 @@ class DeleteApplicationTask extends AbstractFront50Task {
           log.error("Could not delete application permission: {}", re.toString(), re)
           return TaskResult.builder(ExecutionStatus.TERMINAL).outputs(outputs).build()
         }
+        // delete Managed Delivery data
+        keelService.deleteDeliveryConfig(application.name)
       }
-      // delete Managed Delivery data
-      keelService.deleteDeliveryConfig(application.name)
     } catch (RetrofitError e) {
       if (e.response?.status == 404) {
         return TaskResult.SUCCEEDED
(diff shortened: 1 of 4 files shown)


orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
```diff
@@ -42,15 +47,17 @@ class DeleteApplicationTask extends AbstractFront50Task {
           if (re.response?.status == 404) {
             return TaskResult.SUCCEEDED
           }
-          log.error("Could not create or update application permission", re)
+          log.error("Could not delete application permission: {}", re.toString(), re)
```
lorin, line 50: I thought the logger already dumps out a string representation of the exception (in addition to the stack trace) when you pass the exception as the last argument.

lorin (approved):
LGTM (with a minor question)

ajordens:
Is there anything on the keel side that would reconcile if any of these deletes fail? 
(or at least would give an indication of how many delivery configs exist without accompanying application metadata)

luispollo:
> Is there anything on the keel side that would reconcile if any of these deletes fail?
> (or at least would give an indication of how many delivery configs exist without accompanying application metadata)

Good question. Not at the moment. In fact, keel has largely operated independently of app metadata from front50. I've added a couple `TODOs` to the code over time in places where I thought we might want to check against the app existing in front50, or cross-checking against metadata from front50, but we don't have anything implemented at the moment.

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
luispollo: Yeah, looking over the docs and some examples again, I think you're right. I remember this talk/chat with @dreynaud a long time ago where he discussed best practices around exception logging, and was under the impression that was not the case, but I'm probably misremembering. Will undo.

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
```diff
@@ -42,15 +47,17 @@ class DeleteApplicationTask extends AbstractFront50Task {
           if (re.response?.status == 404) {
             return TaskResult.SUCCEEDED
           }
-          log.error("Could not create or update application permission", re)
+          log.error("Could not delete application permission", re)
           return TaskResult.builder(ExecutionStatus.TERMINAL).outputs(outputs).build()
         }
+        // delete Managed Delivery data
+        keelService.deleteDeliveryConfig(application.name)
```
ajordens, line 54: `keelService` may be null here.

ajordens:
> > Is there anything on the keel side that would reconcile if any of these deletes fail?
> > (or at least would give an indication of how many delivery configs exist without accompanying application metadata)
> 
> Good question. Not at the moment. In fact, keel has largely operated independently of app metadata from front50. I think we have a couple `TODOs` in places where we might want to check against the app existing in front50, or cross-checking against metadata from front50, but we don't have anything implemented at the moment, to the best of my knowledge.

👍 

Personally, this will likely work the majority of the time but I'd still lean towards adding some periodic reconciliation as a failsafe.

luispollo:
@ajordens Agreed. We could also add retries to this task. But something for another PR. 🙂

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
luispollo: So can `front50Service` in the existing code... 🙂 

TBH, I never understood why we follow this pattern of marking all beans not-required and then assuming they're available, but I didn't want to stray from the pattern for fear of there being context I just don't have. Let me add a check here.

ajordens: Halyard, I believe would launch a minimal dependencies `orca`. I'm not a fan tho! (as halyard is decommissioned these patterns should go away).

ajordens (approved):

PR #4007, opened 2020-11-18
feat(tasks): add a generic feature toggle for every task

dreynaud:
This would give us a consistent way to disable any task (moving straight to SKIPPED) without the need for each task implementation to reinvent the wheel.

💡 Opening this for early feedback on the general idea. Some questions:
- is this too dangerous?
- is StartTaskHandler the right place for this? 
- should we only do this for tasks that opt in? For instance, they could implement an interface like `DisableableTask`

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartTaskHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartTaskHandler.kt
@@ -16,20 +16,25 @@
 
 package com.netflix.spinnaker.orca.q.handler
 
+import com.google.common.base.CaseFormat
 import com.netflix.spinnaker.orca.TaskResolver
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus.RUNNING
+import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus.SKIPPED
 import com.netflix.spinnaker.orca.api.pipeline.models.TaskExecution
+import com.netflix.spinnaker.orca.events.TaskComplete
 import com.netflix.spinnaker.orca.events.TaskStarted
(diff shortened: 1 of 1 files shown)


jonsie:
Can you describe a quick use case for this feature?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartTaskHandler.kt
```diff
@@ -39,22 +44,41 @@ class StartTaskHandler(
...
+        queue.push(CompleteTask(message, SKIPPED))
+        publisher.publishEvent(TaskComplete(this, mergedContextStage, task))
+      }
     }
   }
 
+  fun getTaskToggle(task: TaskExecution): String {
+    val justTheClassName = task.implementingClass.substring(task.implementingClass.lastIndexOf('.') + 1)
+    val snakifiedClassName = CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_HYPHEN, justTheClassName)
+    return "tasks.$snakifiedClassName.enabled"
```
jonsie, line 76: Since the task is still being resolved from the `TaskResolver` I feel like we should be literal with the name here and call this something like `tasks.$snakifiedClassName.skip.enabled`.

ajordens: I don't think we need to snakeify things ... would be nice to at least stay consistent with existing configuration under `tasks`.

```
tasks:
  monitorFront50Task:
    successThreshold: 0
  monitorKatoTask:
    taskNotFoundTimeoutMs: 600000
    terminalRetries:
      enabled: true
```

dreynaud:
Closing in favor of https://github.com/spinnaker/orca/compare/master...dreynaud:generic-task-toggle-better?expand=1

PR #4008, opened 2020-11-19
feat(tasks): add a generic feature toggle for skippable tasks

dreynaud:
This would give us a consistent way to disable any task (moving straight to SKIPPED) without the need for each task implementation to reinvent the wheel.


orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/SkippableTask.java
```diff
@@ -0,0 +1,3 @@
+package com.netflix.spinnaker.orca.api.pipeline;
+
+public interface SkippableTask extends Task {}
```
robzienert, line 3: Would like to see docs on this.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SadStage.java
```diff
@@ -0,0 +1,18 @@
...
+
+@Component
+public class SadStage implements StageDefinitionBuilder {
+  public static String STAGE_TYPE = "sad";
+
+  @Override
+  public void taskGraph(@NotNull StageExecution stage, @NotNull TaskNode.Builder builder) {
+    builder.withTask("sadNoOp", SadNoOpTask.class);
+  }
+}
```
robzienert, line 18: This looks like a stage for testing? Not sure I want to ship this to everyone. Can we put this in -nflx if we need it?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/SadNoOpTask.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+public class SadNoOpTask implements SkippableTask {
+  @NotNull
+  @Override
+  public TaskResult execute(@NotNull StageExecution stage) {
+    return TaskResult.builder(ExecutionStatus.TERMINAL).build();
+  }
+}
```
robzienert, line 17: Same goes here, I'd prefer to see this in -nflx rather than polluting OSS.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartTaskHandler.kt
```diff
@@ -39,22 +46,44 @@ class StartTaskHandler(
...
+        queue.push(RunTask(message, task.id, task.type))
+        publisher.publishEvent(TaskStarted(this, mergedContextStage, task))
+      }
     }
   }
 
+  fun getTaskToggle(task: TaskExecution): String {
+    val justTheClassName = task.implementingClass.substring(task.implementingClass.lastIndexOf('.') + 1)
+    val snakifiedClassName = CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_HYPHEN, justTheClassName)
+    return "tasks.$snakifiedClassName.enabled"
```
robzienert, line 78: Snake case is new, and inconsistent. Is there a reason we're diverging from convention for `snake_case` instead of the usual `kebab-case`?

robzienert (approved):
Like the general idea, but needs a few things addressed.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartTaskHandler.kt
robzienert: Ah, or like @ajordens mentioned, use the task name?

dreynaud: `LOWER_HYPHEN` is kebab case, I misnamed it snake case. We can certainly support just the task name, it just seemed less consistent

ajordens: At least in our current configuration we're not using kebab case for `tasks.*` config. It'd be nice to not use kebab unless absolutely necessary.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/SadNoOpTask.java
dreynaud: yes, it's for testing purposes. It lives right next to the already existing `NoOpTask`, is that also polluting OSS? I figured it might be useful even when you're testing out a new pipeline, let's say you want to see how a stage failing on a branch impacts your ongoing deploy on a different branch

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartTaskHandler.kt
dreynaud: do you have thoughts on stripping vs keeping the package name?

ajordens: I would continue to use the bean name.

`MonitorKatoTask` -> `monitorKatoTask`

dreynaud: Trying to figure out the bean name from the class name seems like a can of worms to me. 

`MonitorKatoTask` becomes `monitorKatoTask` but `MDSomething` stays `MDSomething`, and the name could be overriden via annotations or provided by a method name or some custom naming strategy 😵

ajordens: What I really mean is lower camel case.

dreynaud: Standardized on `lowerCamelCase` in 11cd05eda, along with docs and a nicer interface

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/tasks/SadNoOpTask.java
dreynaud: removed in b98bf056e

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/SadStage.java
dreynaud: removed in b98bf056e

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/SkippableTask.java
dreynaud: docs added in in 11cd05e

PR #4012, opened 2020-11-30
feat(sql): Add ExecutionRepositoryListener

robzienert:
Little persist listener, needed for some skunkworks observability stuff I'm working on.


orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -88,7 +89,8 @@ class SqlExecutionRepository(
   private val batchReadSize: Int = 10,
   private val stageReadSize: Int = 200,
   private val poolName: String = "default",
-  private val interlink: Interlink? = null
+  private val interlink: Interlink? = null,
+  private val executionRepositoryListener: ExecutionRepositoryListener? = null
```
robfletcher, line 93: Does it make sense to make this a collection? That way you can invoke it in a `forEach` without needing a null check.

robfletcher (approved):

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
robzienert: Sure, that makes sense. I wasn't anticipating having multiples of these, but probably wouldn't hurt.

PR #4015, opened 2020-12-01
fix(application): Fix deletion of Managed Delivery data from keel

luispollo:
While testing deletion of a managed app, I ran into an authorization error in keel as the call was being made without the `X-SPINNAKER-USER` header and so was treated as anonymous. This PR fixes the configuration of the keel service object so that the `SpinnakerRequestInterceptor` is used (which automatically adds that and other useful headers).

### DIFF
--- a/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/KeelService.kt
+++ b/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/KeelService.kt
@@ -20,7 +20,6 @@ import com.netflix.spinnaker.orca.keel.model.DeliveryConfig
 import retrofit.client.Response
 import retrofit.http.Body
 import retrofit.http.DELETE
-import retrofit.http.Header
 import retrofit.http.Headers
 import retrofit.http.POST
 import retrofit.http.Path
@@ -28,7 +27,7 @@ import retrofit.http.Path
 interface KeelService {
   @POST("/delivery-configs/")
   @Headers("Accept: application/json")
-  fun publishDeliveryConfig(@Body deliveryConfig: DeliveryConfig, @Header(value = "X-SPINNAKER-USER") user: String): Response
+  fun publishDeliveryConfig(@Body deliveryConfig: DeliveryConfig): Response
 
   @DELETE("/application/{application}/config")
   fun deleteDeliveryConfig(@Path("application") application: String): Response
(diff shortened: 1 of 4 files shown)


ajordens:
Is the `keel` service client wired up differently than all the others where we don't need to be explicit about `X-SPINNAKER-*` header propagation?

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
```diff
@@ -33,7 +34,8 @@ class DeleteApplicationTask extends AbstractFront50Task {
   KeelService keelService
 
   @Override
-  TaskResult performRequest(Application application) {
+  TaskResult performRequest(Application application, StageExecution stage) {
+    String user = stage.execution.trigger.user ?: "anonymous"
```
ajordens, line 38: This shouldn't be necessary (and it should be `AuthenticatedRequest.getSpinnakerUser()` anyhow).

luispollo:
> Is the `keel` service client wired up differently than all the others where we don't need to be explicit about `X-SPINNAKER-*` header propagation?

That was the first question that popped into my question too... Then I noticed that both the `keelService.publishDeliveryConfig` method and other methods to different services called from within task execution contexts were explicitly specifying the user to fill in the header based on the information in the trigger. I concluded that the automagic authorization stuff only works in the context of a chained HTTP request, not in a background execution. Does that make sense?

orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
luispollo: See comment below. I don't think that works outside the context of... well... an authenticated request.

ajordens: Followed up on our Slack, it does work and the `keelService` just needs to be wired up with the correct interceptor.

luispollo: P.S. I noticed we're missing this line when configuring `keelService`:
```java
setRequestInterceptor(spinnakerRequestInterceptor)
```

Could that be it?

luispollo: Yeah, I think that's it. Kork's `SpinnakerRequestInterceptor` adds the headers, and the task handler sets the user in the context object it grabs values from.

luispollo: I've updated the PR with a fix.

luispollo:
@ajordens you were right. I updated the PR to simply add the Spinnaker request interceptor to `keelService`, which should take care of including the header in every call.

ajordens (approved):
Love it.

PR #4022, opened 2020-12-07
chore: Update Cancel Waiting Executions reason message

christopherthielen:


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/StartWaitingExecutionsHandler.kt
```diff
@@ -51,7 +51,7 @@ class StartWaitingExecutionsHandler(
                 queue.push(CancelExecution(
                   source = purgedMessage,
                   user = "spinnaker",
-                  reason = "Keep waiting pipelines configuration is disabled"
+                  reason = "This execution was queued but then canceled because a newer queued execution superceded it. This pipeline is configured to automatically cancel waiting executions."
```
jonsie, line 54: Well said.

jonsie (approved):

PR #4023, opened 2020-12-08
feat(echo): Support for restricting who can perform a manual judgment

ajordens:
This PR verifies the current user roles against `context.selectedStageRoles`.

The operation is _only_ allowed if there is at least one overlapping role.

This is a rewrite of what originally went in as https://github.com/spinnaker/orca/pull/3988.

### DIFF
--- a/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
+++ b/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
@@ -23,6 +23,7 @@ import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
 import com.netflix.spinnaker.orca.api.pipeline.OverridableTimeoutRetryableTask
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
 import com.netflix.spinnaker.orca.api.pipeline.TaskResult
+import com.netflix.spinnaker.orca.echo.util.ManualJudgmentAuthorization
 
 import javax.annotation.Nonnull
 import java.util.concurrent.TimeUnit
@@ -42,7 +43,7 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage 
   @Override
   void taskGraph(@Nonnull StageExecution stage, @Nonnull TaskNode.Builder builder) {
     builder
-      .withTask("waitForJudgment", WaitForManualJudgmentTask.class)
+        .withTask("waitForJudgment", WaitForManualJudgmentTask.class)
   }
 
   @Override
@@ -72,8 +73,15 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage 
     final long backoffPeriod = 15000
     final long timeout = TimeUnit.DAYS.toMillis(3)
 
-    @Autowired(required = false)
-    EchoService echoService
+    private final EchoService echoService
+    private final ManualJudgmentAuthorization manualJudgmentAuthorization
+
+    @Autowired
+    WaitForManualJudgmentTask(Optional<EchoService> echoService,
+                              ManualJudgmentAuthorization manualJudgmentAuthorization) {
+      this.echoService = echoService.orElse(null)
+      this.manualJudgmentAuthorization = manualJudgmentAuthorization
+    }
 
     @Override
     TaskResult execute(StageExecution stage) {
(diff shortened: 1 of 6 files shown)


orca-echo/src/main/java/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthorization.java
```diff
@@ -0,0 +1,172 @@
...
+    this.front50Service = front50Service.orElse(null);
+    this.fiatPermissionEvaluator = fiatPermissionEvaluator.orElse(null);
+
+    this.fiatStatus = fiatStatus;
+    this.objectMapper = objectMapper;
+  }
+
+  /**
+   * A manual judgment will be considered "authorized" if the current user has at least one of the
+   * required judgment roles _and_ read/write/execute permissions on the application being judged.
```
ajordens, line 70: I still question why manual judgment authorization should care about application permissions. 

This comment has been left as-is to reflect the intent of the original PR but I believe it should be revisited.

orca-echo/src/main/java/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthorization.java
```diff
@@ -0,0 +1,172 @@
...
+    //    boolean isAuthorizedGroup = false;
+    //
+    //    for (String role : currentUserRoles) {
+    //      if (requiredJudgmentRoles.contains(role)) {
+    //        for (Map.Entry<String, Object> entry :
+    //            applicationPermissions.entrySet()) { // get the application permission roles.
+    //          if (Authorization.CREATE.name().equals(entry.getKey())
+    //              || Authorization.EXECUTE.name().equals(entry.getKey())
+    //              || Authorization.WRITE.name().equals(entry.getKey())) {
+    //            // If the application permission roles has 'CREATE, EXECUTE, WRITE', then user is
```
ajordens, line 120: Commented out this logic as I believe it's incorrect.

The comment indicates something about CREATE/EXECUTE/WRITE but is really just checking that one of the required judgment roles has CREATE, EXECUTE or WRITE access to the application.

I don't know what the point of this check is.

IMO, this would be useful feature if we just checked user roles against the required judgment roles. This would allow someone to restrict who is able to perform a judgment, which I thought was the original intent of this feature.

ajordens (commented):
My preference would be to get agreement on the application permission checks and clean that up.

I'm happy to take the work on to get this cleaned up and merged. I believe it is a useful feature.

ajordens:
cc: @sanopsmx

orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
```diff
@@ -202,6 +154,7 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage
   static class StageData {
     String judgmentStatus = ""
     List<Notification> notifications = []
+    List<String> selectedStageRoles = []
```
ajordens, line 157: @sanopsmx Did the UI side of this go out already? Curious if we can change `selectedStageRoles` to something a little clearer. I was using `requiredJudgmentRoles` in `ManualJudgmentAuthorization`.

sanopsmx: UI code was merged

orca-echo/src/main/java/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthorization.java
```diff
@@ -0,0 +1,94 @@
...
+      return true;
+    }
+
+    if (currentUserRoles == null) {
+      currentUserRoles = new ArrayList<>();
+    }
+
+    return Sets.intersection(new HashSet<>(requiredJudgmentRoles), new HashSet<>(currentUserRoles))
+            .size()
+        > 0;
```
robzienert, line 92: nit: `!isEmpty()` instead. `size()` has a higher cost that we don't need to pay here.

robzienert (approved):

orca-echo/src/main/java/com/netflix/spinnaker/orca/echo/util/ManualJudgmentAuthorization.java
ajordens: Yessir!

PR #4030, opened 2020-12-15
feat(stages): Add SpEL expression for converting canary config name to id.

haleyw:
This PR adds a SpEL expression which will take a canary config name and convert it to an ID.

I have written unit tests for this, but haven't tested it end-to-end yet (mostly because I'm not sure how?). Please let me know if there is more you would recommend testing at this point.

### DIFF
--- a/orca-kayenta/src/main/java/com/netflix/spinnaker/orca/kayenta/pipeline/functions/ConfigExpressionFunctionProvider.java
+++ b/orca-kayenta/src/main/java/com/netflix/spinnaker/orca/kayenta/pipeline/functions/ConfigExpressionFunctionProvider.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.kayenta.pipeline.functions;
+
+import com.google.common.base.Strings;
+import com.netflix.spinnaker.kork.api.expressions.ExpressionFunctionProvider;
+import com.netflix.spinnaker.kork.expressions.SpelHelperFunctionException;
+import com.netflix.spinnaker.orca.kayenta.KayentaCanaryConfig;
+import com.netflix.spinnaker.orca.kayenta.KayentaService;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.Nullable;
+import org.jetbrains.annotations.NotNull;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ConfigExpressionFunctionProvider implements ExpressionFunctionProvider {
+
+  // Static because it's needed during expression eval (which is a static)
+  private static KayentaService kayentaService;
+
+  public ConfigExpressionFunctionProvider(KayentaService kayentaService) {
+    this.kayentaService = kayentaService;
+  }
+
+  @Nullable
+  @Override
+  public String getNamespace() {
+    return null;
+  }
+
+  @NotNull
+  @Override
+  public Functions getFunctions() {
+    return new Functions(
+        new FunctionDefinition(
+            "acaNameToId",
+            "Look up the ACA config ID for the given Name and App",
+            new FunctionParameter(String.class, "name", "The name of the config"),
+            new FunctionParameter(String.class, "app", "The name of the app")));
+  }
+
+  /**
+   * SpEL expression used to convert the name of an ACA config to the ID.
+   *
+   * @param name Name of the config.
+   * @param app Application which owns the config.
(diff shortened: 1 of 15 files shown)


orca-kayenta/src/main/java/com/netflix/spinnaker/orca/kayenta/pipeline/functions/ConfigExpressionFunctionProvider.java
```diff
@@ -0,0 +1,87 @@
...
+    if (Strings.isNullOrEmpty(app)) {
+      throw new SpelHelperFunctionException(
+          "App is a required field for the acaNameToId function.");
+    }
+
+    List<KayentaCanaryConfig> configs = kayentaService.getAllCanaryConfigs();
+    Optional<KayentaCanaryConfig> found =
+        configs.stream()
+            .filter(c -> c.getApplications().contains(app))
+            .filter(c -> c.getName().equals(name))
```
dwest-netflix, line 78: I wonder if it's worth it to make these case insensitive? Especially since I think the application is typed into json currently for a kayenta config

haleyw: We could...however, just FYI Kayenta will let you save 2 names with different cases...
* https://www.spinnaker.mgmt.netflix.net/#/applications/chap/canary/config/d622b507-ad42-42bb-b0fa-451e8fe8e661
* https://www.spinnaker.mgmt.netflix.net/#/applications/chap/canary/config/f3ed9dc4-1d1f-4134-9bdf-ea70da79f6d7

It will block you from saving 2 configs with the exact same name, but it is case insensitive...so perhaps we should fix that as well?

dwest-netflix: ugh

dwest-netflix (approved):

orca-kayenta/src/main/java/com/netflix/spinnaker/orca/kayenta/pipeline/functions/ConfigExpressionFunctionProvider.java
dwest-netflix: With that context, it seeeems reasonable to go with this for now. Going down the path of making configs them case insensitive could be a hassle... Since that might cause some odd problems if people out there using Kayenta have configs with the same name?

haleyw: Yeah, I think to fix that correctly, we would have to first check if anyone has >1 with the same name different cases, migrate those, then lock it down...

Let's start with this and we can file a ticket for the other work if you think it's worth pursuing.

orca-kayenta/src/main/java/com/netflix/spinnaker/orca/kayenta/pipeline/functions/ConfigExpressionFunctionProvider.java
```diff
@@ -0,0 +1,87 @@
...
+    return null;
+  }
+
+  @NotNull
+  @Override
+  public Functions getFunctions() {
+    return new Functions(
+        new FunctionDefinition(
+            "acaNameToId",
+            "Look up the ACA config ID for the given Name and App",
```
robzienert, line 52: "ACA" is not an acronym that OSS is familiar with. Can we choose something else (would `canaryNameToId` make sense?)

orca-kayenta/src/main/java/com/netflix/spinnaker/orca/kayenta/pipeline/functions/ConfigExpressionFunctionProvider.java
```diff
@@ -0,0 +1,87 @@
...
+import com.netflix.spinnaker.orca.kayenta.KayentaCanaryConfig;
+import com.netflix.spinnaker.orca.kayenta.KayentaService;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.Nullable;
+import org.jetbrains.annotations.NotNull;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ConfigExpressionFunctionProvider implements ExpressionFunctionProvider {
```
robzienert, line 31: nit: Might consider a rename here; `KayentaConfigExpressionFunctionProvider` perhaps?

robzienert (commented):
> haven't tested it end-to-end yet

Not necessary. The unit tests should be fine.

Looks good, aside from some naming things.

orca-kayenta/src/main/java/com/netflix/spinnaker/orca/kayenta/pipeline/functions/ConfigExpressionFunctionProvider.java
haleyw: Sure, can rename to canaryConfigNameToId -- thanks!

haleyw: Done!

orca-kayenta/src/main/java/com/netflix/spinnaker/orca/kayenta/pipeline/functions/ConfigExpressionFunctionProvider.java
haleyw: Done!

anotherchrisberry (approved):
LGTM

PR #4031, opened 2020-12-18
feat(plugins): Expose application events as api types

robzienert:
Depends on https://github.com/spinnaker/kork/pull/835

I didn't bother promoting `ExecutionListenerAdpater` or `StageListenerAdapter`, as I see them as internal extension points, rather than something we want the plugin community adopting. Probably will want something similar eventually.


link108:
stale PR, closing. May want to revisit this if/as we adopt CDEvents

PR #4032, opened 2020-12-22
feat(pipeline executions/orca): attribue to explicitly skip the outputs section in Deployment Manifest or Run Job stage

Nirmalyasen:
Most K8S deployments do not use the outputs in downstream stages. So, if it is not used
users can skip the output of the Deployment Manifest and Run Job (Manifest) stage by
specifiying the attribute `noOutput` set to false in the stage json.

This significantly reduces the execution context being saved in redis/databse or
sent to the browser.r.

This attempts to address the issues mentioned in
https://github.com/spinnaker/spinnaker/issues/6159
https://github.com/spinnaker/spinnaker/issues/6159

The issue is compounded when there is a nesting of pipelines and the manifests are duplicated many times over in the execution context. There is one PR (https://github.com/spinnaker/orca/pull/3989) that also attempts to address the issue. It tries to do it at the pipeline stage. This PR attempts to apply the same fix but in a much more narrower scope - in Deploy Manifest and Run Job (Manifest) stages. One advantage of this is that it not only reduces the payload of the pipeline stages, but also of Deploy Manifest and Run Job stages. Hence overall it further reduces the data saved in redis/database as well as the amount of data transferred to the browser.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
@@ -79,6 +79,7 @@ public class RunJobStage implements StageDefinitionBuilder, CancellableStage {
     }
   }
 
+  @Override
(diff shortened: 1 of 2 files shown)


austinthao5:
I forked your fix to my own repo to fix the PR Status Checks
https://github.com/spinnaker/orca/pull/4242

PR #4037, opened 2021-01-06
Fix(manifest): Fix manifests resolved by Spel

edgarulg:
**Issue Summary**
The way an array of manifests passed into a pipeline as YAML using `Spel` is getting mapped changed from v1.21.5 to v1.22.0.  In v1.21, the manifests resolved by a `Spel` function are flattened to have a list of manifests. In v1.22, the manifest are not flattened so it throws a `java.lang.IllegalArgumentException` exception.

The behavior changed because this fix: https://github.com/spinnaker/kork/pull/720

**Steps to reproduce**
switch spinnaker to 1.22 or higher, execute the following stage:
```json
{
  "account": "kubernetes",
  "cloudProvider": "kubernetes",
  "manifests": [
    "${#yamlFromUrl(\"https://raw.githubusercontent.com/dmrogers7/spinnaker-issues/main/manifests/array-of-manifests.yaml\")}"
  ],
  "moniker": {
    "app": "bugs-reproduced"
  },
  "name": "Deploy (Manifest)",
  "skipExpressionEvaluation": false,
  "source": "text",
  "trafficManagement": {
    "enabled": false,
    "options": {
      "enableTraffic": false,
      "services": []
    }
  },
  "type": "deployManifest"
}
```
**About the fix**
This issue only happen when the manifest uses `Spel` function to get the manifests and it return a list. Before the fix on kork, `Spel` flattening the list so it ensure that the `ResolveDeploySourceManifestTask` received only a list of manifest. Now the flattening logic is gone so when execute `#yamlFromUrl` it unmarshal the yaml and pass thru `ResolveDeploySourceManifestTask` as it is. This break the models because it expect a list of manifest and it's possible to get a nested list at this point. The following link has an example yaml that shows the below issue: https://github.com/dmrogers7/spinnaker-issues/blob/main/manifests/array-of-manifests.yaml

I included test to show the current behavior, then I just flattened the nested lists in the manifests before it is map to `DeployManifestContext` because this is something specific for the `ResolveDeploySourceManifestTask`

PR #4044, opened 2021-01-22
fix(manifest): Fix deployment of artifact only stages

gavinbunney:
When using an artifact only pipeline (such as Kustomize -> Deploy Manifest), the changes added in #4037 causes a NPE as the manifests object is null. This ensures that the manifests are only flattened when they exist.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ResolveDeploySourceManifestTask.java
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ResolveDeploySourceManifestTask.java
@@ -63,6 +63,10 @@ public final class ResolveDeploySourceManifestTask implements Task {
   // Spel is not flattening nested List anymore
   // We flatten them because its possible have a List of manifests resolved by spel
   private List<Object> flattenNestedLists(Object manifests) {
+    if (manifests == null) {
+      return null;
+    }
+
     List<Object> result = new ArrayList<>();
     List<Object> tmp = (List<Object>) manifests;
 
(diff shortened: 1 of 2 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ResolveDeploySourceManifestTask.java
```diff
@@ -63,6 +63,10 @@ public TaskResult execute(@Nonnull StageExecution stage) {
   // Spel is not flattening nested List anymore
   // We flatten them because its possible have a List of manifests resolved by spel
   private List<Object> flattenNestedLists(Object manifests) {
+    if (manifests == null) {
+      return null;
```
fieldju, line 67: Should this return an empty list rather than null?

gavinbunney: Prior to the `flattenNestedLists` function being added, it used to set the context field to null (on line 47), so kept that the same. Happy to change if you think it should

fieldju: I'd feel better if it returned `list.of()` rather than null, I need to get this backported to fix some NPEs.
I like collections being empty rather than null for better fluent API comparability

```java
collection.stream().filter(...).map(...).collect(...)
```
vs

```java
Optional.ofNullable(collection).ifPresent(c -> collection.stream().filter(...).map(...).collect(...)).orElseThrow(() -> new RuntimeException("foo"))
```

fieldju: Lets keep null, since that was the previous behavior

edgarulg: I like the idea of return a empty list from `flattenNestedLists` method but the only concern is that changing a the value of the `manifest` field from `null` to `emptyList` could cause a side effect in further tasks.

fieldju:
As soon as the checks pass, I will merge and back-port to 1.22, 1.23, and 1.24

edgarulg:
The build is failing with a failure in SqlQueueIntegrationTest, I pulled the branch and tried to reproduce the issue but I couldn't get any error when running test. Maybe re-run the build will help @fieldju @gavinbunney

fieldju:
@Mergifyio backport release-1.22.x release-1.23.x release-1.24.x

PR #4048, opened 2021-01-27
fix(preconfigured job): Fix for k8s manifest resources limit/request …

j-sandy:
Fix for #6257

Issue Summary:

Putting limits/request in a custom job through ~/.hal/default/profiles/orca-local.yml, the custom stage breaks with an error as shown below.
image

About the fix:
In order to format above Map suitable for kubectl, "resources.request/limit.memory/cpu" field must take single value. This can be achieved by intercepting the preconfiguredMap object from class com.netflix.spinnaker.orca.clouddriver.pipeline.job.PreconfiguredJobStage and update it with required "resource" field format before passing it further.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.groovy
@@ -65,6 +65,21 @@ class PreconfiguredJobStage extends RunJobStage {
     // to override the underlying job. this avoids that problem by giving us a fresh "copy"
     // to work wit
     Map<String, Object> preconfiguredMap = objectMapper.convertValue(preconfiguredJob, Map.class)
+    
+    // Solution patch to resolve spec.template.spec.container.resources.limit
+    // and spec.template.spec.container.resources.requests parsed by io.kubernetes.client library
+    Map<String, Object> manifestMap= preconfiguredMap.get("manifest")
+    List<Map<String, Object>> containers = ((Map<String, Object>)((Map<String, Object>)((Map<String, Object>)manifestMap.get("spec")).get("template")).get("spec")).get("containers")
+
+        containers.stream()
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/job/PreconfiguredJobStage.groovy
```diff
@@ -65,6 +65,25 @@ class PreconfiguredJobStage extends RunJobStage {
     // to override the underlying job. this avoids that problem by giving us a fresh "copy"
     // to work wit
     Map<String, Object> preconfiguredMap = objectMapper.convertValue(preconfiguredJob, Map.class)
+    
+    // Solution patch to resolve spec.template.spec.container.resources.limit
+    // and spec.template.spec.container.resources.requests parsed by io.kubernetes.client library
+    if (preconfiguredMap["manifest"] != null) {
```
german-muzquiz, line 71: `PreconfiguredJobStage` is used for kubernete and titus, so it shouldn't have any kubernetes specific code here. I think a good approach would be to replace the line 

```
Map<String, Object> preconfiguredMap = objectMapper.convertValue(preconfiguredJob, Map.class)
```

With a new method call implemented by `PreconfiguredJobStageProperties` that makes the actual conversion. In that way you can have one implementation for Titus in `TitusPreconfiguredJobProperties` and the other specific for kubernetes in `KubernetesPreconfiguredJobProperties`.

Please take a look at `io.kubernetes.client.util.Yaml` for converting between java objects and yaml for kubernetes, we're trying to avoid having special cases like this where we need to manually set fields as parts of transformations. Today is request/limits, but we don't know how many other edge cases there are where using snakeyaml/ObjectMapper produce the wrong results for serializing k8s manifests. Using the official Yaml converter of the kubernetes library could potentially avoid this.

j-sandy: I check the required changes and update the code.

lpcruz:
Hi - running into this exact issue, will start watching this changeset.

PR #4055, opened 2021-01-29
feat(jenkins): Automatically set a build description that links back to Spinnaker

ajordens:
Requires that `spinnaker.wwwHost` be set in your `orca` configuration.

```
spinnaker:
  wwwHost: https://www.spinnaker.net
```

image

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/MonitorQueuedJenkinsJobTask.groovy
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/MonitorQueuedJenkinsJobTask.groovy
@@ -17,6 +17,8 @@
 package com.netflix.spinnaker.orca.igor.tasks
 
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
+import com.netflix.spinnaker.orca.igor.IgorService
+import org.springframework.beans.factory.annotation.Value
 
 import java.util.concurrent.TimeUnit
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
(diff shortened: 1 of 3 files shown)


ajordens:
Ended up going with `spinnaker.wwwHost` as there's some prior art at Netflix around `spinnaker.apiHost`.

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/MonitorQueuedJenkinsJobTask.groovy
```diff
@@ -59,4 +64,48 @@ class MonitorQueuedJenkinsJobTask implements OverridableTimeoutRetryableTask {
...
+   * Update the description of a running Jenkins build and provide a deep link back to the particular execution in
+   * Spinnaker.
+   *
+   * Requires that `wwwHost` be specified _and_ that the build is running (vs. queued).
+   */
+  private void updateBuildDescription(StageExecution stageExecution,
+                                      String jenkinsController,
+                                      String jobName,
+                                      Integer buildNumber) {
+    if (wwwHost == null) {
```
robzienert, line 88: Do we want an `info` log here potentially to inform existing operators that there's a handy feature that could be unlocked with some additional config?

robzienert (approved):
Looks great as-is, one question.

jonsie (approved):

orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/MonitorQueuedJenkinsJobTask.groovy
ajordens: I was going to add something to https://spinnaker.io/setup/ci/jenkins/, a `log.info()` might not hurt as well.

luispollo:
> Ended up going with `spinnaker.wwwHost` as there's some prior art at Netflix around `spinnaker.apiHost`.

FWIW, we use `spinnaker.baseUrl` in Keel. Looks great to me!

ajordens:
@luispollo you've convinced me to change the existing usage, I'll align on something `baseUrl` focused. Thanks!

PR #4057, opened 2021-01-31
feat(yandex): Support provider Yandex.Cloud

vasiliy-briginets:
1.  Add new provider, Yandex
2.  Add server group creator
3.  Add update strategy (rolling update is the default for yandex cloud)
4.  Add image finder

Proposal - https://github.com/spinnaker/governance/issues/166


vasiliy-briginets:
Outdated

PR #4058, opened 2021-02-02
handle empty string manifest path when importing delivery config

robfletcher:
If a user enters a value in the manifest field in Deck, then blanks it out again, the stage config will contain `"manifest": ""` resulting in a failure to look up the delivery config object. This PR makes it so that we handle a blank string in that field the same way we handle a `null` -- i.e. we default to `spinnaker.yml`.


emjburns (approved):

luispollo (approved):
Good catch! Thanks for the fix!

PR #4059, opened 2021-02-03
feat(kubernetes): process jobStatus failure results and raise excepton when state == failed

apoorvmahajan:


apoorvmahajan:
closing in favor of #4177

PR #4062, opened 2021-02-08
fix(api): Rename TaskNode.Builder method to not conflict with class name

robzienert:
When using Kotlin, the method name `TaskNode.Builder(GraphType)` conflicts with the class `TaskNode.Builder`, since the constructor arg is the same as the method on `TaskNode`. This alters the method on `TaskNode` to be `newBuilder` so there is no conflict.


jonsie (approved):

robzienert:
Not sure why this is failing in CI and not locally. Closing, found a new way around this.

PR #4065, opened 2021-02-09
fix(jenkins): Addresses startup issue when `spinnaker.base-url.www` is not set

ajordens:
Fixes spinnaker/spinnaker#6332


ajordens:
FYI @neil-yechenwei

jonsie (approved):

PR #4076, opened 2021-03-06
perf(events): create execution events with a ref to the execution

dreynaud:
This avoid loading things for the repository twice, e.g.:
- on a StartTask message, the handler loads the entire execution
- it then emits a TaskStarted event, which will load the entire execution again
- ... and so on for TaskComplete, StageStarted, StageComplete, etc.

On simple executions, it decreases the amount of executions fetched by 40%, which should translate to less load on the database and lower memory pressure on orca.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/events/StageListenerAdapter.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/events/StageListenerAdapter.java
@@ -16,14 +16,9 @@
 
 package com.netflix.spinnaker.orca.events;
 
-import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus;
-import com.netflix.spinnaker.orca.api.pipeline.models.PipelineExecution;
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
-import com.netflix.spinnaker.orca.listeners.DefaultPersister;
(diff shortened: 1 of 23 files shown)


robzienert (approved):
This is great.

orca-core/src/main/java/com/netflix/spinnaker/orca/events/StageListenerAdapter.java
```diff
@@ -58,62 +48,19 @@ public void onApplicationEvent(ExecutionEvent event) {
...
-                        .get(),
-                    status,
-                    // TODO: not sure if status.isSuccessful covers all the weird cases here
-                    status.isSuccessful()));
-  }
-
-  private PipelineExecution retrieve(ExecutionEvent event) {
-    return repository.retrieve(event.getExecutionType(), event.getExecutionId());
+    StageExecution stage = event.getStage();
+    delegate.afterTask(stage, event.getTask(), stage.getStatus(), stage.getStatus().isSuccessful());
```
dreynaud, line 64: major 🤦‍♂️, this should be `task.getStatus()` and not `stage.getStatus()`

dbyron-sf: Fixed in https://github.com/spinnaker/orca/pull/4105

PR #4077, opened 2021-03-08
fix(aws/cloudformation): Add onDemandCache refresh logic

sergiopena:
Add onDemand cache polling logic for cloud formation stacks.

Right now orca asks for the cache refresh and finishes the task.
This is most likely the cause of this issue:
https://github.com/spinnaker/spinnaker/issues/6125

This enables orca to start polling for the status of the force cache refresh
when the status received from clouddriver indicates that the force cache
the operation did not finish and has been queued.

This PR is complemented with https://github.com/spinnaker/clouddriver/pull/5276 
which add the OnDemand cache logic on the clouddriver side


xavileon:
@spinnaker/oss-approvers PTAL

xavileon (approved):

ajordens (approved):

PR #4078, opened 2021-03-08
fix(webhook): fix NPE in webhook config

dreynaud:
Mapping a webhook config with no `waitForCompletion` in its context results in `stageData.waitForCompletion` being null

Let's default to `false` if the value is absent, which is probably how it was before translating from Groovy to Java.


jonsie (approved):

dreynaud:
cc @Bijnagte

PR #4080, opened 2021-03-09
feat(pipeline executions/orca): Add configurable default for skipDown…

titirigaiulian:
This PR adds the possibility to set a default value for the `skipDownstreamOutput` introduced in [this](https://github.com/spinnaker/orca/pull/3989) PR.
This way we can control the behaviour globally and decide which one is the exception.
In our case none of the child pipelines uses the output of a previous one.

Issue: https://github.com/spinnaker/spinnaker/issues/6159

### DIFF
--- a/orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
+++ b/orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
@@ -34,12 +34,16 @@ import javax.annotation.Nonnull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 @Component
 public class PipelineStage implements StageDefinitionBuilder, CancellableStage {
 
   private final Logger log = LoggerFactory.getLogger(getClass());
+
+  @Value("${stages.pipeline.defaultSkipDownstreamOutput:false}")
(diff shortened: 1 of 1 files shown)


titirigaiulian:
@robzienert @ajordens @marchello2000 @robfletcher can you please take a look?

orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
```diff
@@ -34,13 +34,17 @@
...
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 @Component
 public class PipelineStage implements StageDefinitionBuilder, CancellableStage {
 
   private final Logger log = LoggerFactory.getLogger(getClass());
 
+  @Value("${defaultSkipDownstreamOutput:false}")
```
dreynaud, line 45: makes sense in principle, but if possible I think it would be better to avoid introducing a new top-level config. Maybe it can be namespaced, e.g.:

```suggestion
  @Value("${stages.pipeline.defaultSkipDownstreamOutput:false}")
```

dreynaud (changes requested):

orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java
titirigaiulian: done

dreynaud (approved):
thanks @titirigaiulian 👍

PR #4083, opened 2021-03-10
feat(orca-front50): Reuse service accounts based on their content.

ajbw:
Per spinnaker/spinnaker#6319, due to the number of roles, users and
applications we have in our environment, `fiat`'s real-time authorize
calls are extremely slow (well exceeding one second with roughly 500
users with up to 500 roles per user, and 600 applications with one
pipeline per application; we use Okta, so role information comes in via
SAML assertion), as is `fiat`'s role sync (exceeding seven minutes).
Real-time auth calls exceeding one second cause user requests in the UI
to fail with permissions errors, whilst slow role syncs cause errors
when attempting to save new pipeline triggers.

In order to reduce the number of service accounts that `fiat` needs to
process during role sync and real-time auth calls, we have modified
`orca` and `front50` to generate one service account per unique
combination of roles (rather than one service account per pipeline).
This has reduced our service account count from ~4000 to a couple of
dozen, and `fiat` real-time auth calls are back down to ~20ms, whilst
allowing us to retain the flexibility of managed service accounts.

This is the first of two PRs (the other will be on `front50`,
introducing & modifying some migrations to allow people to switch to
(and from) shared managed service accounts).

In this PR, instead of creating one service account per pipeline (and
naming service accounts by their corresponding pipeline IDs), create one
service account per unique combination of roles, and name them with a
stable hash of the contents (in this case, by joining the sorted list of
roles with `\0`).  Managed service accounts generated in this way are
named with `@shared-managed-service-account` (rather than
`@managed-service-account`).

This new behaviour is controlled by a new configuration option,
`useSharedManagedServiceAccounts`, which defaults to `false` to maintain
backwards compatibilty.


ajordens:
Few conflicts preventing this from moving forward.

ajbw:
@ajordens conflicts resolved :) thanks.

wwalters12:
@ajbw @ajordens Is there anything preventing this PR from being merged? It seems like it should have been merged alongside https://github.com/spinnaker/front50/pull/1022

PR #4084, opened 2021-03-10
feat(webhook): Allow configuring retry handling per status code on webhook monitoring.

Bijnagte:


orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/pipeline/WebhookStage.java
```diff
@@ -84,7 +85,7 @@ public void onFailureStages(@Nonnull StageExecution stage, @Nonnull StageGraphBu
     public String url;
     public Object payload;
     public Map<String, Object> customHeaders;
-    public List<Integer> failFastStatusCodes;
+    public List<Integer> failFastStatusCodes = List.of(504);
```
Bijnagte, line 88: default to having a gateway timeout result in a fail fast by default

dbyron0: Perhaps https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpStatus.html#GATEWAY_TIMEOUT instead to make that more obvious?

orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.java
```diff
@@ -119,135 +126,143 @@ public TaskResult execute(StageExecution stage) {
...
   }
 
+  private boolean shouldRetry(HttpStatus statusCode, WebhookStage.StageData stageData) {
+    int status = statusCode.value();
+    var retries = stageData.getRetries();
+
+    if (retries != null && retries.containsKey(status)) {
+      // specific retry limit configured
+      var retryConfig = retries.get(status);
+      long pastStatuses =
```
jonsie, line 250: nit; `pastStatuses` reads like a list of statuses, maybe change to `attempts`?

jonsie (approved):
This looks like a nice cleanup and addition to me.

Are you thinking the retry config would be added to the webhook stage UI?

Bijnagte:
Yeah, the UI will need to be updated to allow configuring. That will be my next task.

orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/tasks/MonitorWebhookTask.java
Bijnagte: I updated this and extracted a method to make it read a little better

srekapalli:
Will we also apply the same retry config for the initial webhook calls or only fo monitor webhook status calls..?

Bijnagte:
I was not sure about the retry on the initial call and if it should be kept distinct since they are different methods, I could certainly reuse the same values though. I also didn't necessarily want to make the full set of changes that would involve that in one PR.

PR #4088, opened 2021-03-11
fix(events): update in-memory execution before firing events

dreynaud:
This fixes a regression introduced in f04f16ceac67580b4d0629c08361f885867ace4d, which caused e.g. trigger.parentExecution.endTime to be missing from child pipelines.

Before:
- a handler invoked executionRepository.updateStatus(type, id, status)
- `executionRepository` would add fields like `endTime`, `canceled` and persist them
- (before f04f16ceac67580b4d0629c08361f885867ace4d) the application events would load the execution from the db and fire the echo event
- (after f04f16ceac67580b4d0629c08361f885867ace4d) the application events would fire the echo event based on the in-memory execution, which was not updated with fields like `endTime` or `canceled`

After this change:
- handlers have to invoke execution.updateStatus(newStatus) to update the in-memory execution
- handlers then invoke executionRepository.updateStatus(execution, status) to persist the in-memory execution
- we fire echo events based on the in-memory execution, which has been updated with the required fields

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncher.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncher.java
@@ -191,7 +191,7 @@ public class ExecutionLauncher {
       log.error("Failed to start {} {}", execution.getType(), execution.getId(), failure);
     }
     execution.updateStatus(TERMINAL);
-    executionRepository.updateStatus(execution, TERMINAL);
+    executionRepository.updateStatus(execution);
     executionRepository.cancel(execution.getType(), execution.getId(), canceledBy, reason);
     return executionRepository.retrieve(execution.getType(), execution.getId());
   }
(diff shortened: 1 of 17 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncher.java
```diff
@@ -191,7 +190,8 @@ private PipelineExecution handleStartupFailure(PipelineExecution execution, Thro
     } else {
       log.error("Failed to start {} {}", execution.getType(), execution.getId(), failure);
     }
-    executionRepository.updateStatus(execution.getType(), execution.getId(), status);
+    execution.updateStatus(TERMINAL);
+    executionRepository.updateStatus(execution, TERMINAL);
```
Bijnagte, line 194: It is a little strange to have to chain these 2 calls together all the time. Are there known cases of people using Repositories beyond SQL or Redis where it is not possible to eliminate the old method and combine these in the repository implementation? That would require docs about the expectation also though.

Bijnagte (approved):

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncher.java
dreynaud: @Bijnagte you're right, it's a little awkward. I changed it in cc0dffc37 to make it more intuitive, it becomes:

```
execution.updateStatus(TERMINAL);
executionRepository.updateStatus(execution)
```

Removing the new status from the execution repository call disambiguates that the execution model is expected to have the new status.

PR #4090, opened 2021-03-11
refactor(java): Refactor KatoService to Java; remove autowire configuration

jonsie:
I'm working on a new interface for cloud operation work, and as a first step I'd like to refactor `KatoService` to Java and clean-up configuration a bit.

Note this is still in the `groovy` dir in `orca-clouddriver`.  Not sure why, but `orca-clouddriver` does not have a `java` dir even though a fair bit of it has been converted to Java.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoService.java
```diff
@@ -0,0 +1,122 @@
...
+            requestId(operation), cloudProvider, operation.getOperationType(), operation);
+
+    TaskId taskId;
+    try {
+      InputStream body = null;
+      try {
+        body = response.getBody().in();
+      } finally {
+        if (body != null) {
+          body.close();
```
Bijnagte, line 73: This is how it was done before but it seems weird and wrong, the InputStream is being closed before it is read by the objectMapper. Also it seems like the whole thing should be switched to use a try with resources.

Bijnagte:
Why not make a java dir and relocate what can be?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoService.java
jonsie: Yeah, this method is honestly only being used from a stage/task that is completely unused (`CloudOperationStage`) so I didn't spend anytime thinking about this (pretty sure eventually we would just delete `CloudOperationStage`).

May as well clean it up though I suppose.

PR #4092, opened 2021-03-12
feat(md): scrape git metadata out of trigger, if it exists

emjburns:
If the trigger has the info we need, scrape all the git metadata out of it and pass that to keel to be used in: https://github.com/spinnaker/keel/pull/1864

This is opt-in. The stage has to set `"sendGitInfo": "true"` in the context. If this actually works I'll transition everyone to this.

### DIFF
--- a/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/model/SubmitDeliveryConfigBody.kt
+++ b/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/model/SubmitDeliveryConfigBody.kt
@@ -1,0 +1,6 @@
+package com.netflix.spinnaker.orca.keel.model
+
+data class SubmitDeliveryConfigBody(
+  val config: DeliveryConfig,
+  val gitMetadata: GitMetadata? = null
+)
--- a/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/model/SubmitDeliveryConfigBody.kt
+++ b/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/model/SubmitDeliveryConfigBody.kt
@@ -1,6 +1,0 @@
-package com.netflix.spinnaker.orca.keel.model
-
-data class SubmitDeliveryConfigBody(
-  val config: DeliveryConfig,
-  val gitMetadata: GitMetadata? = null
-)
--- a/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
+++ b/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
@@ -28,10 +28,16 @@ import com.netflix.spinnaker.orca.api.pipeline.models.SourceCodeTrigger
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
 import com.netflix.spinnaker.orca.api.pipeline.models.Trigger
 import com.netflix.spinnaker.orca.igor.ScmService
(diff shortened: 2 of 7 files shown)


orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/model/SubmitDeliveryConfigBody.kt
```diff
@@ -0,0 +1,6 @@
+package com.netflix.spinnaker.orca.keel.model
+
+data class SubmitDeliveryConfigBody(
```
luispollo, line 3: This is no longer used since the data is going in the `metadata` field now, right?

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -60,9 +64,17 @@ constructor(
         context.repoType, context.projectKey, context.repositorySlug, context.directory, context.manifest, context.ref
       )
 
+      if (context.sendGitInfo) {
```
luispollo, line 67: How about just attempting to parse the data and falling back to the default behavior when we can't, instead of using a flag? Similar to what you did on the keel side. One fewer thing to cleanup later...

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
```diff
@@ -280,6 +283,28 @@ internal class ImportDeliveryConfigTaskTests : JUnit5Minutests {
...
+      context("parsing git metadata") {
+        test("parses correctly") {
+          execute(mutableMapOf("sendGitInfo" to true))
+          val submittedConfig = slot<DeliveryConfig>()
+          verify(exactly = 1) {
+            keelService.publishDeliveryConfig(capture(submittedConfig))
+          }
+          val m: Any = submittedConfig.captured.getOrDefault("metadata", emptyMap<String, Any?>())!!
+          val metadata: Map<String, Any?> = objectMapper.convertValue(m)
+          expectThat(metadata["gitMetadata"]).isNotEqualTo(null)
```
luispollo, line 303: I think this could be simplified like this (`metadata` defaults to an `emptyMap()`, right?):
```suggestion
          expectThat(submittedConfig.captured.metadata["gitMetadata"]).isNotNull()
```

emjburns: I'm doing this for type safety so I don't have to use `!!`. Even if metadata defaults to an empty map I need to handle the null key there and also converting it to a map if it's stored as a string.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/model/SubmitDeliveryConfigBody.kt
emjburns: arg I could have swore I deleted this! yep, removing.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
emjburns: I'm not 100% confident this won't blow up. So I'm going to keep it behind a flag to make sure, and then I'll remove the flag and do the behavior you said. I'd rather have more to clean up but 0 user impact.

PR #4094, opened 2021-03-12
chore(java): Move java source to its own src dir and enable groovy joint compilation only where necessary

Bijnagte:


orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/ServiceManifest.java
```diff
@@ -61,6 +62,7 @@ public Artifact getArtifact() {
   }
 
   @Data
+  @EqualsAndHashCode(callSuper = false)
```
Bijnagte, line 65: this was just a compiler warning and there are no fields on the superclass to consider

cfieber (approved):

jonsie (approved):

PR #4096, opened 2021-03-12
config(logging): don't log humongous tasks in full

dreynaud:
Does a couple things:
- logs the execution id along with submitted tasks (otherwise surprisingly tricky to go from a submitted task to the actual task)
- avoid dumping the whole body of humongous tasks in the logs. Arbitrarily set at 1MB for now.
- does not serialize the config to string if this particular logger is not enabled


orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
```diff
@@ -472,12 +467,28 @@ class OperationsController {
...
+    if (!log.isInfoEnabled()) {
+      return
+    }
+
+    def json = objectMapper.writeValueAsString(config)
+    if (json.length() < 1_000_000) {
+      return json
+    }
+
+    return json[0..50_000] + " (...) " + json[-50_000..-1]
```
Bijnagte, line 489: The only Issue I see with this approach is it will make it so it can't be parsed. Maybe some indication that it was truncated not 50,000 characters in would be nice.

Bijnagte (approved):

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy
dreynaud: Good point, I'll tune the log message. Although lifting a multi-MB json from the logs isn't exactly practical either, I think it's more important to log the execution id so we can retrieve the execution later

ajordens: I'm assuming we don't get a nice `Content-Length` header shipped across? That would be too convenient eh?

dreynaud: @ajordens that reminds me, the gate request logger reports `responseSize` but not `requestSize`, that would be nice to have for this use case!

dreynaud: added length and fixed silly mistake in bd350448a

PR #4098, opened 2021-03-13
feat(taskdefresolver): Dynamic task definition resolver

srekapalli:
-- New TaskImplementationResolver  which resolves a task node to a  task definition

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/DynamicTaskImplementationResolver.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/DynamicTaskImplementationResolver.kt
@@ -1,0 +1,86 @@
+package com.netflix.spinnaker.orca
+
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
+import com.netflix.spinnaker.orca.api.pipeline.Task
+import com.netflix.spinnaker.orca.api.pipeline.graph.TaskNode.TaskDefinition
+import com.netflix.spinnaker.orca.api.pipeline.graph.TaskNode.DefinedTask
+import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
+import com.netflix.spinnaker.orca.clouddriver.utils.CloudProviderAware
+import com.netflix.spinnaker.orca.config.TaskOverrideConfigurationProperties
+import com.netflix.spinnaker.orca.config.TaskOverrideConfigurationProperties.TaskOverrideDefinition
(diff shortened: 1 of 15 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/DefaultDynamicTaskDefinitionResolver.kt
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Default resolver with no customizations.
+ */
+class DefaultDynamicTaskDefinitionResolver : DynamicTaskDefinitionResolver {
+
+  override fun resolveTaskDefinitions(
+    stageExecution: StageExecution,
+    taskGraph: TaskNode.TaskGraph
+  ): TaskNode.TaskGraph {
+    return taskGraph
```
jonsie, line 15: I assume the dynamic bits will be implemented here?

orca-core/src/main/java/com/netflix/spinnaker/orca/DynamicTaskImplementationResolver.kt
```diff
@@ -0,0 +1,87 @@
...
+  ): TaskImplementationResolver, CloudProviderAware {
+
+  private val log = LoggerFactory.getLogger(javaClass)
+
+  override fun resolve(stage: StageExecution, taskNode: DefinedTask): TaskExecution {
+    val taskOverrideDefinition = taskOverrideConfigurationProperties
+      ?.overrideDefinitions
+      ?.firstOrNull {
+        it.stageName.equals(stage.type, true) &&
+          it.originalTaskName.equals(taskNode.name, true)
```
jonsie, line 30: We should replace based on the the original task implementation class name.  The reason for this is the same task can have multiple names (look at the various names for `MonitorKatoTask` for example).

orca-core/src/main/java/com/netflix/spinnaker/orca/config/TaskOverrideConfigurationProperties.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Original task name as given while building the stage task graph
+     */
+    public var originalTaskName: String,
+
+    /**
+     * New task name to use in the task definition.
+     */
+    public var newTaskName: String,
```
jonsie, line 39: I think we should leave the task names alone and just replace the implementation class name.

jonsie (approved):

PR #4100, opened 2021-03-17
feat(api): Add OperationsRunner interface with KatoOperationsRunner implementation

jonsie:
Implementations of OperationsRunner can be used to fork operations to different services (besides CloudDriver, for example).

Implemented in `DeleteScalingPolicyStage` and `UpsertScalingPolicyStage` stages, though we'll probably eventually sweep through and replace all instances of `katoService` in these stages with `operationsRunner`.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/scalingpolicy/UpsertScalingPolicyTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/scalingpolicy/UpsertScalingPolicyTask.groovy
@@ -36,10 +36,10 @@ class UpsertScalingPolicyTask extends AbstractCloudProviderAwareTask implements 
   long backoffPeriod = TimeUnit.SECONDS.toMillis(5)
   long timeout = TimeUnit.SECONDS.toMillis(100)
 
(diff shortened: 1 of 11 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/scalingpolicy/UpsertScalingPolicyTask.groovy
```diff
@@ -16,38 +16,48 @@
...
 
 @Component
 @Slf4j
 class UpsertScalingPolicyTask extends AbstractCloudProviderAwareTask implements RetryableTask {
 
-  @Autowired
-  KatoService kato
-
   long backoffPeriod = TimeUnit.SECONDS.toMillis(5)
   long timeout = TimeUnit.SECONDS.toMillis(100)
```
srekapalli, line 38: can we name the variable as `operationsRunner` ?

jonsie: Oh yep, missed this during refactors.

srekapalli (approved):
LGTM

PR #4102, opened 2021-03-17
perf(executions): remove ability to launch executions from a json string

dreynaud:
At a high level, all this does is that it replaces

  ExecutionLauncher.start(ExecutionType type, String configJson)

with

  ExecutionLauncher.start(ExecutionType type, Map<String, Object> config)

In general, the lifecycle of an execution is that it can live as:
- a json string, e.g. when it is submitted to the API
- a Map<String, Object> "config" that needs to be munged and massaged (with things like a user, origin, trigger...)
- a PipelineExecution model when it is fully formed

With this PR, we try to only move "up the chain" (string -> map -> model) and never back down (string -> map -> string -> map -> model).

In practice this means avoiding calls to objectMapper.writeValueAsString() followed by objectMapper.readValue(), which in turns simplifies some tests quite a bit.


cfieber (approved):
awesome

mattgogerly (approved):

dbyron-sf:
@dreynaud I took a leap and rebased, resolved conflicts and force-pushed...hope that's OK.

PR #4111, opened 2021-03-19
feat(run-task-handler): added hook in finally block that will be ran regardless of whether the task completed without exception.

fieldju:
feat(run-task-handler): added hook in finally block that will be run regardless of whether the task completed without exception.

The existing hook doesn't execute when a task execution throws an exception.
This hook allows for an extension project to clear the security context that is set in the before task hook.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/TaskExecutionInterceptor.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/TaskExecutionInterceptor.java
@@ -50,4 +50,7 @@ public interface TaskExecutionInterceptor {
   default TaskResult afterTaskExecution(Task task, StageExecution stage, TaskResult taskResult) {
     return taskResult;
   }
+
+  default void finallyAfterTaskExecution(
+      Task task, StageExecution stage, TaskResult taskResult, Exception e) {}
 }
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/TaskExecutionInterceptor.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/TaskExecutionInterceptor.java
@@ -51,6 +51,18 @@ public interface TaskExecutionInterceptor {
     return taskResult;
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -189,13 +190,15 @@ class RunTaskHandler(
                   log.error("Error running ${message.taskType.simpleName} for ${message.executionType}[${message.executionId}]", e)
                 }
               }
-
+              taskException = e;
```
cfieber, line 193: is there a reason not to set taskException for the other cases in the catch block ?

fieldju: My original thought was that those other blocks retry, but thinking about it more no, it should be set right away.

I'll move it up.

orca-core/src/main/java/com/netflix/spinnaker/orca/TaskExecutionInterceptor.java
```diff
@@ -50,4 +50,7 @@ default StageExecution beforeTaskExecution(Task task, StageExecution stage) {
   default TaskResult afterTaskExecution(Task task, StageExecution stage, TaskResult taskResult) {
     return taskResult;
   }
+
+  default void finallyAfterTaskExecution(
```
cfieber, line 54: can you add a doc comment here on the purpose of this hook point for implementers

cfieber (approved):
LGTM would just like to see javadoc on the new interface method

orca-core/src/main/java/com/netflix/spinnaker/orca/TaskExecutionInterceptor.java
fieldju: - [X] add docs

PR #4112, opened 2021-03-23
fix(travis): Should not fail stage if Travis properties are not returned

jervi:
Continuation of https://github.com/spinnaker/orca/pull/4089 - it should've been part of that PR really.
Orca will now always check for properties from Travis, so it shouldn't fail if none are found.

### DIFF
--- a/orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
+++ b/orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
@@ -61,7 +61,7 @@ public class GetBuildPropertiesTask extends RetryableIgorTask<CIStageDefinition>
             stageDefinition.getPropertyFile(),
             stageDefinition.getMaster(),
             stageDefinition.getJob());
-    if (properties.size() == 0) {
+    if (properties.isEmpty()) {
       if (stageDefinition.getMaster().startsWith("travis-")) {
         return TaskResult.builder(ExecutionStatus.SUCCEEDED).build();
       }
--- a/orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
+++ b/orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
@@ -62,6 +62,9 @@ public class GetBuildPropertiesTask extends RetryableIgorTask<CIStageDefinition>
             stageDefinition.getMaster(),
             stageDefinition.getJob());
     if (properties.size() == 0) {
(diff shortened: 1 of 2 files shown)


orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
```diff
@@ -62,6 +62,9 @@
             stageDefinition.getMaster(),
             stageDefinition.getJob());
     if (properties.size() == 0) {
+      if (stageDefinition.getMaster().startsWith("travis-")) {
```
dreynaud, line 65: any chance for `stageDefinition.getMaster()` to be null here?

jervi: No, it is the name of the buildmaster. It needs to be defined and is fetched from igor.

orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
```diff
@@ -62,6 +62,9 @@
             stageDefinition.getMaster(),
             stageDefinition.getJob());
     if (properties.size() == 0) {
```
dreynaud, line 64: while you're in here, do you mind converting that to this?

```suggestion     
    if (properties.isEmpty()) {
```

orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTask.java
jervi: If it was null an exception would have been thrown a few lines up as well.

dreynaud (approved):

orca-igor/src/test/groovy/com/netflix/spinnaker/orca/igor/tasks/GetBuildPropertiesTaskSpec.groovy
```diff
@@ -146,6 +146,20 @@ class GetBuildPropertiesTaskSpec extends Specification {
     e.message == "Expected properties file $PROPERTY_FILE but it was either missing, empty or contained invalid syntax"
   }
 
+  def "does not fail stage even if properties are not returned from travis and build passed"() {
+    given:
+    def stage = createStage(PROPERTY_FILE, "travis-$MASTER")
+
+    and:
+    buildService.getPropertyFile(BUILD_NUMBER, PROPERTY_FILE, MASTER, JOB) >> [:]
```
jervi, line 154: This should reference `travis-$MASTER` as well. The test works because Spock returns an empty list by default for mocks.

PR #4115, opened 2021-04-07
feat(yandex): Support provider Yandex.Cloud

st8f:
* Add new provider, Yandex
* Add server group creator
* Add update strategy (rolling update is the default for yandex cloud)
* Add image finder

Proposal - spinnaker/governance#166


st8f:
Closed in favor of https://github.com/spinnaker/orca/pull/4057

PR #4117, opened 2021-04-20
feat(ecs): Override Task Def Artifacts

paragbhingre:
Add a support to override task definitions.

Backend changes to support [#6374](https://github.com/spinnaker/spinnaker/issues/6374)

Related PRs - 
Deck - [#9092](https://github.com/spinnaker/deck/pull/9092)
Clouddriver - [#5214](https://github.com/spinnaker/clouddriver/pull/5314)

Screen Shot 2021-04-14 at 4 56 00 PM

Test Case 1 - When `Evaluate SpEL expressions in artifact` is disabled (default case)

Configurations for deployment of a pipeline- 

Task def Artifact from GitHub 

Screen Shot 2021-05-13 at 4 03 32 PM

Deck - 

Screen Shot 2021-05-13 at 4 20 23 PM

Screen Shot 2021-05-13 at 3 56 59 PM

Screen Shot 2021-05-13 at 3 57 30 PM

Outputs - 

Screen Shot 2021-05-13 at 3 57 18 PM

Screen Shot 2021-05-13 at 3 58 27 PM

Conclusion - When `Evaluate SpEL expressions in artifact` is disabled, task definition artifact will not be evaluated.

Test case 2 -  When `Evaluate SpEL expressions in artifact` is enabled

Configurations for deployment of a pipeline- 

Task def Artifact from GitHub 

Screen Shot 2021-05-13 at 4 03 32 PM

Deck - 

Screen Shot 2021-05-13 at 3 56 59 PM

Screen Shot 2021-05-13 at 4 10 38 PM

Screen Shot 2021-05-13 at 3 57 30 PM

Outputs - 

Screen Shot 2021-05-13 at 4 11 17 PM

Screen Shot 2021-05-13 at 4 12 11 PM

Conclusion - When `Evaluate SpEL expressions in artifact` is enabled task definition artifact will be evaluated and processed artifact will sent to Clouddriver.


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
```diff
@@ -86,6 +129,51 @@ class EcsServerGroupCreator implements ServerGroupCreator, DeploymentDetailsAwar
     return [[(ServerGroupCreator.OPERATION): operation]]
   }
 
+  private Supplier<Iterable<Object>> fetchAndParseArtifact(Artifact artifact) {
+    return { ->
+      Response artifactText = oortService.fetchArtifact(artifact)
+      try {
+        return yamlParser.get().loadAll(artifactText.getBody().in());
```
allisaurus, line 136: I'm conflicted about using a YAML parser on file that's expected as JSON. On one hand, YAML is ostensibly a superset of JSON, so it should work fine and would let customers use either format (though I'm not sure whether the current object mapper in clouddriver will work with a YAML artifact...). On the other hand, a dedicated JSON parser may be more performant and we'd have even more confidence of avoiding YAML/JSON spec incompatibilities (if there are any?).

Do we know of anywhere else in Orca where a JSON parser is being used instead of the YAML one? Also, can we test a YAML artifact with true/false `evaluateTaskDefinitionArtifact` values to confirm whether YAML is de facto supported now?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
```diff
@@ -86,6 +130,55 @@ class EcsServerGroupCreator implements ServerGroupCreator, DeploymentDetailsAwar
...
+          throw new ConfigurationException("Invalid artifact configuration or task definition artifact is null")
+        }
+      } catch (Exception e) {
+        throw new IllegalStateException(e);
+      }
+    }
+  }
+
+  private List<Map<Object, Object>> coerceArtifactToList(Object artifact) {
+    if (artifact instanceof List) {
```
allisaurus, line 149: when would a task definition artifact contain an instance of a list? wouldn't this indicate an error state, since it needs to be a [RegisterTaskDefinitionRequest](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RegisterTaskDefinition.html) object?

paragbhingre: I have changed the type of the evaluated artifact from List to Map, so now it resolves this issue as well.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
paragbhingre: Currently Clouddriver mapper doesn't support YAML parsing and ORCA does support it, which means if the evaluateTaskDefinitionArtifact flag is set as true, then YAML or JSON  artifact both will get evaluated and sent to clouddriver. If evaluateTaskDefinitionArtifact is set to false and YAML has been used as an artifact then it will fail in clouddriver. So best case would to add a support for YAML also in clouddriver and we can create a new issue to add this functionality. Does that make sense?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
```diff
@@ -57,6 +84,23 @@ class EcsServerGroupCreator implements ServerGroupCreator, DeploymentDetailsAwar
     if (operation.useTaskDefinitionArtifact) {
       if (operation.taskDefinitionArtifact) {
         operation.resolvedTaskDefinitionArtifact = getTaskDefArtifact(stage, operation.taskDefinitionArtifact)
+        if (operation.evaluateTaskDefinitionArtifactExpressions) {
+          Iterable<Object> rawArtifact =
+              retrySupport.retry(
+                  fetchAndParseArtifact(operation.resolvedTaskDefinitionArtifact), 10, Duration.ofMillis(200), true)
```
allisaurus, line 90: is there precedence for this retry config (`10, Duration.ofMillis(200), true`) elsewhere in Orca? what is it for similar manifest evaluations?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
```diff
@@ -57,6 +84,23 @@ class EcsServerGroupCreator implements ServerGroupCreator, DeploymentDetailsAwar
...
+              StreamSupport.stream(rawArtifact.spliterator(), false)
+                  .filter(Objects.&nonNull)
+                  .map(this.&coerceArtifactToList)
+                  .collect()
+                  .flatten()
+
+          Map<Object, Object> evaluatedArtifact = getSpelEvaluatedArtifact(unevaluatedArtifact, stage)
+
+          operation.spelProcessedTaskDefinitionArtifact = evaluatedArtifact
+          operation.evaluateTaskDefinitionArtifactExpressions = true
```
allisaurus, line 102: why are we setting this to true? shouldn't it already be set to true in the customer's pipeline?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
paragbhingre: Yes, it is similar for manifest evaluations.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
paragbhingre: extremely good catch 👍🏼. I will change this.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
allisaurus: Link? (i believe you, but for future reference)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
```diff
@@ -86,6 +129,47 @@ class EcsServerGroupCreator implements ServerGroupCreator, DeploymentDetailsAwar
...
+
+  private List<Map<Object, Object>> coerceArtifactToList(Object artifact) {
+    Map<Object, Object> singleArtifact =
+        objectMapper.convertValue(artifact, new TypeReference<Map<Object, Object>>() {})
+    return List.of(singleArtifact)
+  }
+
+  private Map<Object, Object> getSpelEvaluatedArtifact(
+      List<Map<Object, Object>> unevaluatedArtifact, StageExecution stage) {
+    Map<String, Object> processorInput = ImmutableMap.of("artifact", unevaluatedArtifact);
```
allisaurus, line 155: Is there a reason `unevaluatedArtifact` is still a list here? does `contextParameterProcessor.process()` need to take in a map of string:list?

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreatorSpec.groovy
```diff
@@ -185,4 +191,235 @@ class EcsServerGroupCreatorSpec extends Specification {
...
+
+    then:
+
+    1 * oortService.fetchArtifact(*_) >> new retrofit.client.Response('http://oort.com', 200, 'Okay', [], new TypedString(response))
+    0 * oortService._
+    operations[0].createServerGroup.spelProcessedTaskDefinitionArtifact.toString().equals(expected)
+
+    where:
+    response                                               | expected
+    '{"foo": "${ENV1}.b.${ENV2}"}'                         | "[foo:\${ENV1}.b.\${ENV2}]"
```
allisaurus, line 383: I see above you're defining a parameter `ENV1` but then you not referencing it w/ "parameters" syntax here so we're expecting the text as it appears to be resolved instead. Is this understanding of the test correct?

paragbhingre: So I am basically trying to show here that ENV1 and ENV2 are environment variables which shouldn't be evaluated as we send environment variables and command properties as is to AWS. So instead of having parameter ENV1 available there it is not evaluating that means code knows that ${ENV1} is a env variable and shouldn't be evaluated. Does that make sense?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
paragbhingre: https://github.com/spinnaker/orca/blob/76a9d1e4887ded5c5da952d459ec494111874cf8/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/ManifestEvaluator.java#L131

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/providers/ecs/EcsServerGroupCreator.groovy
allisaurus: discussed offline: `StreamSupport.stream(...)` returns a list and its usage here matches its use in similar logic for other providers.

allisaurus (approved):

PR #4134, opened 2021-06-04
fix(jobs): use stage.execution.application as default when looking job status

edgarulg:
The WaitOnJobCompletion task make http calls to clouddriver to get the job status for a Kubernetes PreconfiguredJob. By default, when the stage.context does not have a moniker or application field defined the name of the object is used but it is parsed to Frigga Names object.

When an end user execute a pipeline and the WaitOnJobCompletion task is executed, if the stage's context does not have the moniker or application field defined, the Application name will be null or the job's name.

Clouddriver uses the account & pipeline name to determinate permissions, when sending the job name as the application name we can get an Access Denied error.

gfs-job-read-only (1)

Instead of use the job name as default, I changed it to use the Application name in the PipelineExecution. This prevent to search permission in Applications that might not exist.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
@@ -18,7 +18,6 @@ package com.netflix.spinnaker.orca.clouddriver.tasks.job
 
 import com.fasterxml.jackson.core.type.TypeReference
 import com.fasterxml.jackson.databind.ObjectMapper
-import com.netflix.frigga.Names
 import com.netflix.spinnaker.kork.core.RetrySupport
 import com.netflix.spinnaker.kork.exceptions.ConfigurationException
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
@@ -120,8 +119,7 @@ public class WaitOnJobCompletion implements CloudProviderAware, OverridableTimeo
       }
 
       def name = names[0]
-      def parsedName = Names.parseName(name)
-      String appName = stage.context.moniker?.app ?: stage.context.application ?: parsedName.app
+      String appName = stage.context.moniker?.app ?: stage.context.application ?: stage.execution.application
 
       InputStream jobStream
       retrySupport.retry({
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
```diff
@@ -120,8 +119,7 @@ public class WaitOnJobCompletion implements CloudProviderAware, OverridableTimeo
       }
 
       def name = names[0]
-      def parsedName = Names.parseName(name)
-      String appName = stage.context.moniker?.app ?: stage.context.application ?: parsedName.app
+      String appName = stage.context.moniker?.app ?: stage.context.application ?: stage.execution.application
```
danielpeach, line 122: Should `parsedName` still be the last fallback here? It is strange to parse the job name as the application name, but I am worried that people may depend on this behavior and that changing it now would be a breaking change.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobUtils.java
```diff
@@ -57,14 +56,17 @@ public void cancelWait(StageExecution stage) {
...
 
       String name = names.get(0);
-      Names parsedName = Names.parseName(name);
       Moniker moniker = (Moniker) stage.getContext().get("moniker");
       String appName;
 
       if (moniker != null) {
         appName = moniker.getApp();
       } else {
-        appName = (String) stage.getContext().getOrDefault("application", parsedName.getApp());
```
danielpeach, line 67: Same here - I think `parsedName` should be the last fallback.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
edgarulg: that's right, the best way to fallback to the `parsedName` should be validate if there is an application with the `appName` if the application doesn't exist, then we will use the application name from `stage.execution.application`

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobUtils.java
edgarulg: I can validate that the application exists, the `stage.execution.application` won't be null so fallback to parseName won't be possible as long as I verify the application name.

danielpeach (approved):

link108:
@mergifyio backport release-1.26.x

PR #4135, opened 2021-06-15
feat(cloudfoundry/unbind): delete service binding stage

jorgebee65:
This work is to create a new stage in spinnaker for cloudfoundry. 
This stage specifically handles Unbinding services from applications.


orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryDeleteServiceBindingsTask.java
```diff
@@ -0,0 +1,91 @@
...
+    this.tsgResolver = tsgResolver;
+    this.mapper = mapper;
+  }
+
+  @Override
+  @Nonnull
+  public TaskResult execute(@Nonnull StageExecution stage) {
+    List<TargetServerGroup> tsgList = tsgResolver.resolve(stage);
+    if (!tsgList.isEmpty()) {
+      Optional.ofNullable(tsgList.get(0))
```
rvazquezglez, line 54: Do we care just about the first one?

zachsmith1: I've only seen it resolve to a single instance

zachsmith1 (approved):
LGTM

PR #4137, opened 2021-06-17
fix(webhooks): forcefully convert response.body() to String

morbid98:
This should fix https://github.com/spinnaker/spinnaker/issues/6454. 
Apparently, response.body(), which is returned from webhook is not a string, which breaks JsonPath parsing to bind an artifact. Forcing response.body() to be a string fixes that issue.


mattgogerly:
Any chance of a test case for this? @morbid98

morbid98:
> Any chance of a test case for this? @morbid98

I'm not familiar with Java too much, but will try :)

PR #4138, opened 2021-06-23
feat(cloudfoundry): Uses last operation state & description

rvazquezglez:
feat(cloudfoundry): Uses last operation state & description to surface that something went wrong.

Related PR: https://github.com/spinnaker/clouddriver/pull/5396

### DIFF
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryWaitForDeployServiceTask.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryWaitForDeployServiceTask.java
@@ -46,18 +46,24 @@ public class CloudFoundryWaitForDeployServiceTask extends AbstractWaitForService
         oortService.getServiceInstance(account, cloudProvider, region, serviceInstanceName);
 
     ExecutionStatus status = oortStatusToTaskStatus(serviceInstance);
-    String lastOperationDescription =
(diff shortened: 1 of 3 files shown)


zachsmith1 (approved):
LGTM

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/providers/cf/CloudFoundryWaitForDeployServiceTask.java
```diff
@@ -31,6 +34,41 @@ public CloudFoundryWaitForDeployServiceTask(OortService oortService) {
...
+
+    ExecutionStatus status = oortStatusToTaskStatus(serviceInstance);
+
+    TaskResult.TaskResultBuilder taskResultBuilder = TaskResult.builder(status);
+    Optional.ofNullable(serviceInstance)
+        .ifPresent(
+            (si) -> {
+              String lastOperationDescription =
+                  Optional.ofNullable(serviceInstance.get("lastOperationDescription"))
+                      .orElse("")
```
mattgogerly, line 56: Would it make sense to have a generic failure message if the last operation description is null?

rvazquezglez: Yes, makes sense. Added a default one.

zachsmith1 (approved):
LGTM

PR #4145, opened 2021-07-09
chore(build): add license information to POM file

mattgogerly:
Adds licensing information to generated POMs.

cc @link108 @dbyron-sf

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,10 +20,12 @@ plugins {
   id "org.jetbrains.kotlin.plugin.allopen" version "$kotlinVersion" apply false
   id "io.gitlab.arturbosch.detekt" version "1.7.4" apply false
   id "org.jetbrains.dokka" version "1.4.10" apply false
+  id "nebula.maven-apache-license" version "17.3.2" apply false
 }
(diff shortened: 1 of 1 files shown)


dbyron-sf:
Could you let us know the reason for this?  Also, can you post a before/after diff of the resulting pom?

dbyron-sf:
related PR: https://github.com/spinnaker/clouddriver/pull/5423

If this is going to end up everywhere, I suspect https://github.com/spinnaker/spinnaker-gradle-project/tree/master/spinnaker-project-plugin is the place to put it.

mattgogerly:
> Could you let us know the reason for this? Also, can you post a before/after diff of the resulting pom?

The reasoning is that corporate dependency procurement checks the POM file for Maven artifacts for the licensing information. If the license isn't found then manual intervention is required, meaning a load of tickets for every new version. By placing the license in the POM we bypass this.

Agree that would be the right place to put it, will open a PR there with a diff on Monday 👍

dbyron-sf:
Sounds good.  Thanks!

PR #4176, opened 2021-08-31
fix(artifacts): Do not further strip version for nuget packages

PerGon:
Nuget packages files names are written in the following format `MyPackage.1.3.3.7.nupkg` and the version separator is the `.` which is the same character used in the version Major/Minor/Patch separator (has opposed to `deb` and `rpm` which have distinct version separators).

Given this, we do not need to "further strip" in case the "version separator" is found in the version itself (since it will always happen).


mattgogerly (approved):

mattgogerly:
@PerGon mind checking the failing test?

PR #4177, opened 2021-09-01
feat(clouddriver): update how run job failures are processed.

apoorvmahajan:
## Changes

- add logic to provide additional metadata when run job failures occur. The user will no longer see `"No reason provided."` as the error message, but will see a lot of additional info such as which pod failed, which container failed, and last few log lines, if the pod was configured correctly. See: https://github.com/spinnaker/clouddriver/pull/5184#issue-552934199 for an example of how it was before vs how it is now.

- Always raise an exception when state == failed. This helps us to show the error message in the UI as seen above.

- do not fail on errors while getting property file contents if jobStatus == Failed as it masks the real error behind the job failure. 

- refactor WaitOnJobCompletion class to use constructor based dependency injection

- add tests for the above changes.

## Testing

- unit tests
- examples from runs in our clusters can be seen here: https://github.com/spinnaker/clouddriver/pull/5184#issue-552934199


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
```diff
@@ -48,20 +49,12 @@ public class WaitOnJobCompletion implements CloudProviderAware, OverridableTimeo
...
-  @Autowired
-  JobUtils jobUtils
-
-  @Autowired(required = false)
+  private final KatoRestService katoRestService
+  private final ObjectMapper objectMapper
+  private final RetrySupport retrySupport
+  private final JobUtils jobUtils
   Front50Service front50Service
+  private final ExecutionRepository repository
```
mattgogerly, line 57: Minor stylistic thing, can we keep final properties together?

mattgogerly (commented):
Thank you for this! Losing job failures once the pod is gone has been a huge annoyance for us.

Only one very minor stylistic comment from me :)

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
apoorvmahajan: sure thing

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
```diff
@@ -152,21 +160,35 @@ public class WaitOnJobCompletion implements CloudProviderAware, OverridableTimeo
...
-              }
+          try {
+            retrySupport.retry({
+              properties = katoRestService.getFileContents(appName, account, location, name, stage.context.propertyFile)
+            }, 6, 5000, false) // retry for 30 seconds
+          } catch (Exception e) {
+            if (status == ExecutionStatus.SUCCEEDED) {
+              throw new ConfigurationException("Property File: ${stage.context.propertyFile} contents could not be retrieved. Error: " + e)
+            }
+            log.warn("failed to get file contents for ${appName}, account: ${account}, namespace: ${location}, " +
```
mattgogerly, line 171: What would cause this log to be hit?

apoorvmahajan: If a jobStatus had `TERMINATED` status. My thinking was at that point, if there is an exception while getting property file contents, then we should not throw that since it will mask the real error regarding why the job was in terminated status. So we just log it and throw the real error a few lines below in the `processFailure(job)` call

dbyron-sf:
@mergifyio update

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
```diff
@@ -182,4 +204,39 @@ public class WaitOnJobCompletion implements CloudProviderAware, OverridableTimeo
...
+   */
+  void updateStageContext(StageExecution stage, Map<String, Object> outputs) {
+    if (stage.context) {
+      stage.context.putAll(outputs)
+      repository.storeStage(stage)
+    }
+  }
+
+  /**
+   * this function is called into action when the job has failed. It returns a {@link JobFailedException} with all the
```
dbyron-sf, line 222:
```suggestion
   * this function is called into action when the job has failed. It throws a {@link JobFailedException} with all the
```

dbyron-sf (approved):

PR #4193, opened 2021-11-16
feat(pipeline executions/orca): Add support for max concurrent pipeline executions

guido9j:
It adds support for max concurrent pipeline executions. If concurrent pipeline execution is enabled, pipelines will queue when the max concurrent pipeline executions is reached. Any queued pipelines will be allowed to run once the number of running pipeline executions drops below the max. If the max is set to 0, then pipelines will not queue.

This is the proposed change for https://github.com/spinnaker/spinnaker/issues/6558.

Changes to deck, front50 and echo are also needed in order to expose max concurrent pipeline executions as a pipeline configuration to Spinnaker users.

(Front50 pull request: spinnaker/front50#1080)
(Echo pull request: https://github.com/spinnaker/echo/pull/1134)
(Deck pull request: https://github.com/spinnaker/deck/pull/9777)

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
@@ -47,6 +47,7 @@ internal interface OrcaMessageHandler<M : Message> : MessageHandler<M> {
   companion object {
     val log: Logger = LoggerFactory.getLogger(this::class.java)
     val mapper: ObjectMapper = OrcaObjectMapper.getInstance()
+    val MIN_PAGE_SIZE = 2
   }
 
   val repository: ExecutionRepository
@@ -119,20 +120,20 @@ internal interface OrcaMessageHandler<M : Message> : MessageHandler<M> {
       configId == null -> false
       !isLimitConcurrent -> {
         return when {
-          getMaxConcurrentExecutions() <= 0 -> false
-          else -> {
-            val criteria = ExecutionCriteria().setPageSize(maxConcurrentExecutions+2).setStatuses(RUNNING)
+          maxConcurrentExecutions > 0 -> {
+            val criteria = ExecutionCriteria().setPageSize(maxConcurrentExecutions+MIN_PAGE_SIZE).setStatuses(RUNNING)
             repository
               .retrievePipelinesForPipelineConfigId(configId, criteria)
               .filter { it.id != id }
               .count()
               .toBlocking()
-              .first() >= getMaxConcurrentExecutions()
+              .first() >= maxConcurrentExecutions
           }
+          else -> false
         }
       }
       else -> {
-        val criteria = ExecutionCriteria().setPageSize(2).setStatuses(RUNNING)
+        val criteria = ExecutionCriteria().setPageSize(MIN_PAGE_SIZE).setStatuses(RUNNING)
         repository
           .retrievePipelinesForPipelineConfigId(configId, criteria)
           .filter { it.id != id }
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
@@ -116,8 +116,21 @@ internal interface OrcaMessageHandler<M : Message> : MessageHandler<M> {
   fun PipelineExecution.shouldQueue(): Boolean {
     val configId = pipelineConfigId
     return when {
-      !isLimitConcurrent -> false
       configId == null -> false
+      !isLimitConcurrent -> {
+        return when {
+          getMaxConcurrentExecutions() <= 0 -> false
+          else -> {
+            val criteria = ExecutionCriteria().setPageSize(maxConcurrentExecutions+2).setStatuses(RUNNING)
+            repository
+              .retrievePipelinesForPipelineConfigId(configId, criteria)
+              .filter { it.id != id }
+              .count()
+              .toBlocking()
+              .first() >= getMaxConcurrentExecutions()
+          }
+        }
+      }
       else -> {
         val criteria = ExecutionCriteria().setPageSize(2).setStatuses(RUNNING)
         repository
--- a/orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
+++ b/orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
@@ -858,6 +858,10 @@ public class RedisExecutionRepository implements ExecutionRepository {
       execution.setCanceledBy(map.get("canceledBy"));
       execution.setCancellationReason(map.get("cancellationReason"));
       execution.setLimitConcurrent(Boolean.parseBoolean(map.get("limitConcurrent")));
+      execution.setMaxConcurrentExecutions(
+          (map.containsKey("maxConcurrentExecutions"))
+              ? Integer.parseInt(map.get("maxConcurrentExecutions"))
+              : 0);
       execution.setBuildTime(NumberUtils.createLong(map.get("buildTime")));
       execution.setStartTime(NumberUtils.createLong(map.get("startTime")));
       if (map.get("startTimeExpiry") != null) {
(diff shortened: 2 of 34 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
```diff
@@ -116,8 +116,21 @@ internal interface OrcaMessageHandler<M : Message> : MessageHandler<M> {
   fun PipelineExecution.shouldQueue(): Boolean {
     val configId = pipelineConfigId
     return when {
-      !isLimitConcurrent -> false
       configId == null -> false
+      !isLimitConcurrent -> {
+        return when {
+          getMaxConcurrentExecutions() <= 0 -> false
+          else -> {
+            val criteria = ExecutionCriteria().setPageSize(maxConcurrentExecutions+2).setStatuses(RUNNING)
```
mattgogerly, line 124: What's this magic 2?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
```diff
@@ -116,8 +116,21 @@ internal interface OrcaMessageHandler<M : Message> : MessageHandler<M> {
   fun PipelineExecution.shouldQueue(): Boolean {
     val configId = pipelineConfigId
     return when {
-      !isLimitConcurrent -> false
       configId == null -> false
+      !isLimitConcurrent -> {
+        return when {
+          getMaxConcurrentExecutions() <= 0 -> false
```
mattgogerly, line 122: Why less than 0?

orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
```diff
@@ -858,6 +858,10 @@ protected PipelineExecution buildExecution(
       execution.setCanceledBy(map.get("canceledBy"));
       execution.setCancellationReason(map.get("cancellationReason"));
       execution.setLimitConcurrent(Boolean.parseBoolean(map.get("limitConcurrent")));
+      execution.setMaxConcurrentExecutions(
+          (map.containsKey("maxConcurrentExecutions"))
+              ? Integer.parseInt(map.get("maxConcurrentExecutions"))
+              : 0);
```
mattgogerly, line 864: Does this need to be supported for SQL backed Orca also?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
guido9j: I added 2, because of the code that checks for any **RUNNING** pipeline a few lines below starts with a page size of 2. Plus 1 may work, but to be in the safe side I added 2:
`val criteria = ExecutionCriteria().setPageSize(2).setStatuses(RUNNING)`

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
guido9j: If `isLimitConcurrent == false && maxConcurrentExecutions <= 0` then don't queue. This implements/keeps the current logic, where `isLimitConcurrent == false` then don't queue.

orca-redis/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/jedis/RedisExecutionRepository.java
guido9j: I could not find any changes needed for the SQL backed Orca.

guido9j:
@mattgogerly: I replied on your comments. Let me know what you think. Thanks

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
mattgogerly: Makes sense, might make sense to extract it out to a constant so we don't end up peppering numbers throughout

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
mattgogerly: Shouldn't it just return false always, if limitConcurrent is false? Why is the check for maxConcurrent necessary?

mattgogerly: Shouldn't it just return false always, if limitConcurrent is false? Why is the check for maxConcurrent necessary?

guido9j: It needs to return true if `isLimitConcurrent == false && maxConcurrentExecutions > 0 && RUNNING >= maxConcurrentExecutions`. If makes more readable, I can change to:

      !isLimitConcurrent -> {
        return when {
          getMaxConcurrentExecutions() > 0 -> {
            val criteria = ExecutionCriteria().setPageSize(maxConcurrentExecutions+2).setStatuses(RUNNING)
            repository
              .retrievePipelinesForPipelineConfigId(configId, criteria)
              .filter { it.id != id }
              .count()
              .toBlocking()
              .first() >= getMaxConcurrentExecutions()
          }
          else -> false
        }
      }

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
guido9j: I'll change the `2` into a constant set to 2. I can name it 'MIN_PAGE_SIZE`.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/OrcaMessageHandler.kt
mattgogerly: Ah my bad, I had the implication of limitConcurrent as "have an upper limit on concurrent" (which is obviously what you're implementing) rather than "prevent concurrent altogether".

Sorry!

mattgogerly:
Looks good to me, but I'll defer to somebody a bit more experienced.

@dbyron-sf @german-muzquiz

guido9j:
@mattgogerly , @german-muzquiz : I looked at the build failure over the past few days, and as far as I can tell this is caused by a commit to master on Dec 4th. The failure is intermittent. I ran the build on the master branch, and it failed 3 out of 5 times:
Screen Shot 2021-12-07 at 7 25 03 AM

guido9j:
Here is the output from test that is failing now:
[TEST-com.netflix.spinnaker.orca.controllers.AdminControllerTest.xml.log](https://github.com/spinnaker/orca/files/7668471/TEST-com.netflix.spinnaker.orca.controllers.AdminControllerTest.xml.log)

guido9j:
Last build on master failed with same error: https://github.com/spinnaker/orca/runs/4438645255?check_suite_focus=true

dbyron-sf:
> Last build on master failed with same error: https://github.com/spinnaker/orca/runs/4438645255?check_suite_focus=true

Yup, it's a flaky test...would love some help tracking down the issue.  In the meantime, my magic bullet is re-running failed jobs...

dbyron-sf:
@mergifyio update

guido9j:
@dbyron-sf, @mattgogerly, @german-muzquiz : can this be merged to master? Thanks

link108 (approved):

link108:
@guido9j Approved! Could you please update the branch and/or make it editable by maintainers? eg https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

Then I can update and merge the branch 👍

guido9j:
@link108: The option "Allow edit from maintainers" is not available for me, since this is an organization-owned fork (not sure if I just don't have the needed permissions in my org). Either way, I went ahead and merged the latest changes from master and build is in progress now. Thanks

link108:
Thank you!

PR #4195, opened 2021-11-24
feat(orchestration): provide a way to allow only certain configured ad-hoc operations to be performed

apoorvmahajan:
## Purpose

There are two types of tasks - pipeline initiated tasks, and ad-hoc orchestration tasks. Some of the ad-hoc tasks can be done through the UI - such as update application, delete pipeline, update pipeline etc. For security reasons, we want the capability to disable ad-hoc tasks for all users (except the few that are permitted to execute them) in production environments. 

This PR introduces a few config knobs that provide a more fine-grained control over who can perform the above actions. By default, everyone can perform all actions (which equates to the current behavior). Users can opt-in to this capability by configuring the ad-hoc tasks in the manner presented below:
```
executions:
  blockOrchestrationExecutions: true
  allowedOrchestrationExecutions: # only applicable when blockOrchestrationExecutions: false
    - name: updateApplication # eg. config to show how to set the name and allowAllUsers properties for an execution
      allowAllUsers: true
    - name: saveApplication # eg. config to show how to set just the name of an execution. It defaults to allowAllUsers: true
    - name: savePipeline # eg. config to show how to set the name and permittedUsers properties for an execution
      permittedUsers:
       - "permitted-user@abc.com"
````

### DIFF
--- a/orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncherTest.java
+++ b/orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncherTest.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2021 Salesforce.com, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncherTest.java [hunk 1]
```diff
@@ -22,62 +22,87 @@
...
+import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType;
 import com.netflix.spinnaker.orca.api.pipeline.models.PipelineExecution;
 import com.netflix.spinnaker.orca.config.ExecutionConfigurationProperties;
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository;
 import java.time.Clock;
 import java.util.Map;
 import java.util.Optional;
-import org.junit.jupiter.api.BeforeEach;
+import org.junit.Before;
```
dbyron-sf, line 33: Would be nice to stick to the jupiter APIs here.

PR #4199, opened 2021-12-06
fix(aws): allowing null or empty AWS EC2 ssh keyPair

linjmeyer:
I believe this should fix [Spinnaker issue 4291](https://github.com/spinnaker/spinnaker/issues/4291) which is blocking null `keyPair` in EC2 deployments.  AWS accepts a null keyPair when using services like Systems Manager and EC2 Instance Connect.  Without this fix you will get an error about a hardcoded Netflix key pair name which I tracked down to this line in Orca.

I haven't had a chance to fully test this as I don't have a full Spinnaker dev env setup locally.  I can try to find time later to do that if needed.

I did verify clouddriver will allow an empty keypair by putting null as `defaultKeyPair` and `defaultKeyPairTemplate` in the AWS config, so this issue only seems to exist in Orca.  Here is the clouddriver credentials output:

```
curl "http://localhost:7002/credentials/stage?expand=true" --stderr - -s
{...
defaultKeyPair": "",
...}
```

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 fiatVersion=1.28.0
-korkVersion=7.126.0
+korkVersion=7.127.0
 kotlinVersion=1.4.32
 org.gradle.parallel=true
 spinnakerGradleVersion=8.16.0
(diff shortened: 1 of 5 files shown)


mattgogerly (approved):

mattgogerly:
Could you fix the failing tests?

linjmeyer:
Yep @mattgogerly done.  Removed it from the test for required params, added it to the test for optional ones.

mattgogerly (approved):

PR #4214, opened 2022-02-04
feat(expressions): add option to disable retry for fromUrl expression function

cristhian-castaneda:
To allow better customizations of CloseableHttpClient, adding HttpClientProperties as an additional property in UserConfiguredUrlRestrictions that we can use in HttpClientUtils. This will allow us to enable/disable retry options and configure other HttpClient options.

Now, you will be able to configure `httpClientProperties`
e.g
```yaml
user-configured-url-restrictions:
  httpClientProperties:
    enableRetry: true
    maxRetryAttempts: 1
    retryInterval: 1000
    timeoutMillis: 10000
```

### DIFF
--- a/orca-core/orca-core.gradle
+++ b/orca-core/orca-core.gradle
@@ -59,6 +59,7 @@ dependencies {
 
   testImplementation(project(":orca-test"))
   testImplementation(project(":orca-test-groovy"))
+  testImplementation("com.github.tomakehurst:wiremock:2.15.0")
   testImplementation("org.junit.jupiter:junit-jupiter-api")
   testImplementation("org.junit.platform:junit-platform-runner")
   testImplementation("org.assertj:assertj-core")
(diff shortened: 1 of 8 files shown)


jasonmcintosh (approved):
I think this is a definite improvement :)

mattgogerly:
This is a breaking change right? As in anybody already using webhooks will now need to configure this for them to continue working?

jasonmcintosh:
Yes we need to make sure this is documented VERY well.  POTENTIALLY we could leave the
`allowedHostnamesRegex: ".*"` and JUST set the retries to 1 by default, while documenting "Don't set retries higher than 1 without being aware of the regex" or similar...

link108:
@cristhian-castaneda is it possible to make this backwards compatible? I took a quick look and it seems like this could benefit from a feature flag or something similar, so that users could opt-in vs dealing with breaking changes. Similar to the discussion in https://github.com/spinnaker/orca/pull/3627

cristhian-castaneda:
Yeah, I think the @jasonmcintosh approach makes sense, will set the `allowedHostnamesRegex` default for ".*" and retries to 1 by default, and let the users modify it to avoid a breaking change.

cristhian-castaneda:
@jasonmcintosh  @mattgogerly I've updated the PR, now retry is enabled by default, and max attempts set it to 1. 
We'll need to document that users now have the ability to update those configurations.

jasonmcintosh:
Yeah I still feel like the ".*" is a security hole - be good on release notes to recommend or call out "We recommend setting this to something like *.github.com" or similar ;) BUT I think this gets us past the the original issue...

dogonthehorizon (approved):

PR #4219, opened 2022-02-14
fix(kotlin): create kotlinModule using Builder instead constructor

edgarulg:
I found issues in Orca when starting up.

```
2022-02-14 05:08:43.009  WARN 1 --- [           main] ConfigServletWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'redisOrcaQueueConfiguration': Injection of autowired dependencies failed; nested exception is java.lang.NoSuchMethodError: 'void com.fasterxml.jackson.module.kotlin.KotlinModule.<init>(int, boolean, boolean, boolean, com.fasterxml.jackson.module.kotlin.SingletonSupport, int, kotlin.jvm.internal.DefaultConstructorMarker)'
```

This issues is related to the `jackson-module-kotlin` used. It's hard to know what version of jackson will end in the service. So I changed how the KotlinModule is created. Instead creating it using the constructor I used its Builder. This ensure that KotlinModule will be created without rely in what version of jackson we have.

There is a breaking change between Jackson 2.11.1 <-> 2.12.3. Using the builder will helps us deal with it.


link108 (approved):

edgarulg:
@Mergifyio backport release-1.27.x

PR #4224, opened 2022-02-28
Fix meter registry processor causing incompatibility with Armory's Observability plugin

massimo-pacher-tw:
This PR fixes https://github.com/armory-plugins/armory-observability-plugin/issues/32#issuecomment-1036400066.

It allows other `BeanPostProcessor` (e.g. Spring `MeterRegistryBeanPostProcessor`) to process registry implementations, restoring the exported metrics and aligning the behavior to the other spinnaker services.

I've retrofitted a test showing the missed processing of a dummy BeanPostProcessor, but more details are available on the issue.

We noticed this issue extends to _Kayenta_ too, given it pulls in `orca-core` modules and the affected classes, preventing the Observability plugin to export standard metrics (due to the binders - e.g. JvmMetrics - not being bound ).

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/telemetry/TaskSchedulerMetricsPostProcessor.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/telemetry/TaskSchedulerMetricsPostProcessor.java
@@ -23,6 +23,7 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Lazy;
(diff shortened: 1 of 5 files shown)


mattgogerly:
Thank you for fixing this!

mattgogerly (approved):

dbyron-sf:
Yes, thanks much @massimo-pacher-tw!  If kayenta picks up the fix by releasing a new version of orca, and doesn't have any place to fix directly within kayenta, then I think we're set.  We'll figure out a good time to tag orca and bump the version in kayenta.

dbyron-sf:
I tagged v8.21.0 of orca with this fix.  [Here's the kayenta PR](https://github.com/spinnaker/kayenta/pull/868) to pick up the change.

massimo-pacher-tw:
I  did the exact same fix in our fork, thanks for the bump 🙇‍♂️

PR #4245, opened 2022-03-30
fix(orca): fail the stage as soon as a task fails.

SusmithaGU:
When a stage marked with "continuePipeline==true" has one of the tasks(except last task) failed then the stage need not execute remaining tasks. This change has been added in this PR. The pipeline continues to execute next stage. 
This PR addresses the issue [#6615](https://github.com/spinnaker/spinnaker/issues/6615)

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteTaskHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteTaskHandler.kt
@@ -94,9 +94,9 @@ class CompleteTaskHandler(
       return true
     }
 
-    if (originalStatus == FAILED_CONTINUE) {
-      // the task explicitly returned FAILED_CONTINUE and _should_ run subsequent tasks
-      return false
+    if (status == FAILED_CONTINUE) {
+      // the task explicitly returned FAILED_CONTINUE and _should_ not run subsequent tasks
+      return true
     }
 
     // the task _should not_ run subsequent tasks


jgrumboe:
@SusmithaGU the bot meant the commit title, so you'll need to force-push your last commit with the fixed commit message

SusmithaGU:
will do it.Thank you

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteTaskHandler.kt
```diff
@@ -94,9 +94,9 @@ class CompleteTaskHandler(
       return true
     }
 
-    if (originalStatus == FAILED_CONTINUE) {
-      // the task explicitly returned FAILED_CONTINUE and _should_ run subsequent tasks
-      return false
+    if (status == FAILED_CONTINUE) {
+      // the task explicitly returned FAILED_CONTINUE and _should_ not run subsequent tasks
```
mattgogerly, line 98: Can we reword this comment to make it clearer what's going on?

Something like `the task failed but the stage explicitly set FAILED_CONTINUE, so we _should not_ run subsequent tasks`

dbyron-sf:
Other than the general strategy of adding tests to motivate code changes, this seems like a fundamental enough part of the code that we really would benefit from some tests.

apoorvmahajan:
What happens to tasks that explicitly set `ExecutionStatus == FAILED_CONTINUE`? Tasks like [MonitoredDeployBaseTask](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java#L127) and [AddServerGroupEntityTagsTask](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/AddServerGroupEntityTagsTask.groovy#L66) are just a few examples (SavePipelinesTask is another and so on) that explicitly set that status, assuming the stage would continue. It seems like this would negatively impact those tasks.

SusmithaGU:
Raised this PR in order to let the first failed task's exception to be shown to the user instead of the last failed task's one but agreeing with @apoorvmahajan on this and hence closing this PR.

PR #4255, opened 2022-04-19
feat(stageExecution): supporting extra custom tags for stage execution metrics

zrashwani:
Adding the ability to add extra tags to the metrics recorded on the stage execution level. 
This will add flexibility in instrumenting custom stages provided via Orca plugins as follows:

* Adding public methods to set/get metric tags in `StageExecution` interface
* Injecting those custom tags to the metrics recorded in related stage handlers (e.g: `StartStageHandler`, `CompleteStageHandler`)

### DIFF
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/StageExecution.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/StageExecution.java
@@ -123,6 +123,11 @@ public interface StageExecution {
 
   void setLastModified(@Nullable LastModifiedDetails lastModified);
 
+  @Nullable
+  Map<String, String> getMetricTags();
+
+  void setMetricTags(@Nullable Map<String, String> metricTags);
+
   // ------------- InternalStageExecution?
   // A lot of these methods are used in a single place somewhere in Orca. I don't know why we
   // decided to put a bunch
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/StageExecution.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/StageExecution.java
@@ -124,9 +124,9 @@ public interface StageExecution {
   void setLastModified(@Nullable LastModifiedDetails lastModified);
 
   @Nullable
-  Map<String, String> getMetricTags();
+  Map<String, String> getAdditionalMetricTags();
 
-  void setMetricTags(@Nullable Map<String, String> metricTags);
+  void setAdditionalMetricTags(Map<String, String> additionalMetricTags);
 
   // ------------- InternalStageExecution?
   // A lot of these methods are used in a single place somewhere in Orca. I don't know why we
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
@@ -343,6 +343,23 @@ public class StageExecutionImpl implements StageExecution, Serializable {
     this.lastModified = lastModified;
   }
 
+  /**
+   * Additional tags to be used with stage metrics. This is useful to add extra dimensions to the
+   * metrics recorded for built-in or custom stages.
+   */
+  private Map<String, String> metricTags;
+
+  @Nullable
+  @Override
+  public Map<String, String> getMetricTags() {
+    return this.metricTags;
+  }
+
+  @Override
+  public void setMetricTags(@Nullable Map<String, String> metricTags) {
+    this.metricTags = metricTags;
+  }
+
   @Override
   public boolean equals(Object o) {
     if (this == o) return true;
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
@@ -347,17 +347,17 @@ public class StageExecutionImpl implements StageExecution, Serializable {
    * Additional tags to be used with stage metrics. This is useful to add extra dimensions to the
    * metrics recorded for built-in or custom stages.
    */
-  private Map<String, String> metricTags;
+  private Map<String, String> additionalMetricTags;
 
   @Nullable
   @Override
-  public Map<String, String> getMetricTags() {
-    return this.metricTags;
+  public Map<String, String> getAdditionalMetricTags() {
+    return this.additionalMetricTags;
   }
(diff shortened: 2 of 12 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -179,6 +179,10 @@ class CompleteStageHandler(
         stage.context["cloudProvider"]?.let {
           id.withTag("cloudProvider", it.toString())
         } ?: id
+      }.let { id ->
+        stage.metricTags?.let {
+          id.withTags(stage.metricTags)
```
deverton, line 184: Does this merge or replace the existing tags?

zrashwani: it merge the tags provided as a parameter with existing ones (here is the [implementation of `withTags`](https://github.com/Netflix/spectator/blob/main/spectator-api/src/main/java/com/netflix/spectator/api/DefaultId.java#L71-L73) method)

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
```diff
@@ -343,6 +343,23 @@ public void setLastModified(@Nullable LastModifiedDetails lastModified) {
     this.lastModified = lastModified;
   }
 
+  /**
+   * Additional tags to be used with stage metrics. This is useful to add extra dimensions to the
+   * metrics recorded for built-in or custom stages.
+   */
+  private Map<String, String> metricTags;
```
dbyron-sf, line 350: a bit of a nit, but perhaps
```suggestion
  private Map<String, String> additionalMetricTags;
```
to match the description?

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/StageExecution.java
```diff
@@ -123,6 +123,11 @@ public interface StageExecution {
 
   void setLastModified(@Nullable LastModifiedDetails lastModified);
 
+  @Nullable
+  Map<String, String> getMetricTags();
+
+  void setMetricTags(@Nullable Map<String, String> metricTags);
```
dbyron-sf, line 129: We don't need to allow a null argument here, do we?

orca-queue/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
dbyron-sf, line 1: Sorry, I'm not familliar with this. Why do we need it?

zrashwani: this give us the ability to mock or spy final classes, in this case, I needed to `spy(NoopRegistry)` in the tests [here](https://github.com/spinnaker/orca/pull/4255/files#diff-5030240d38e2011d634b6cf4ab9ef2868655238b0cb1f254c30ba0e4e3f03618R108)

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/StageExecution.java
zrashwani: yes, removed nullable from  method parameter

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
zrashwani: done

orca-queue/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
dbyron-sf: Thanks!

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandlerTest.kt
```diff
@@ -101,7 +104,8 @@ object CompleteStageHandlerTest : SubjectSpek<CompleteStageHandler>({
   val publisher: ApplicationEventPublisher = mock()
   val exceptionHandler: ExceptionHandler = DefaultExceptionHandler()
   val clock = fixedClock()
-  val registry = NoopRegistry()
+  val metricId: Id = mock()
+  val registry = spy(NoopRegistry())
```
dbyron-sf, line 108: Actually, as I look at this more closely, would it be simpler to change this to DefaultRegistry?  I think there's no longer a need to mock metricId.  The assertions change to something like (untested):
```
    assertThat(registry.timers()).hasSize(1);
    Timer timer = registry.timers().findFirst().orElseThrow(AssertionError::new);
    assertThat(timer.id().name()).isEqualTo("stage.invocations.duration");
    assertThat(timer.id().tags())
        .contains(
            tag("tag1", "value1"),
            tag("tag2", "value2"));
```

zrashwani: Yea, using DefaultRegistry does simplify things. I have modified that and removed the metricId mock, however in a different way from the snippet above, as `timers` have more than one element, and we cannot assume it has size=1.
I modified to validate the timer using `argumentCaptor` as below:
```
            argumentCaptor<Id>().let {
              verify(registry, atLeastOnce()).timer(it.capture())
              val metricId = it.firstValue
              assertThat(metricId.name()).isEqualTo("stage.invocations.duration")
              assertThat(metricId.tags().any{ t -> t.key() == "tag1" && t.value() == "value1" }).isTrue
              assertThat(metricId.tags().any{ t -> t.key() == "tag2" && t.value() == "value2" }).isTrue
            }
```

orca-queue/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
dbyron-sf, line 1: I think this can disappear now.

zrashwani: no, it is still needed, as `spy(DefaultRegistry())` is used in the tests, and DefaultRegistry is also a final class

dbyron-sf: Aaah, ok.

dbyron-sf (approved):

PR #4259, opened 2022-04-26
fix(orca): display task exception messages

SusmithaGU:
This PR allows to tie `exceptionDetails` to the task whenever a task throws an exception and also display each of them in UI along with the task name.

This addresses the issue[ #6615](https://github.com/spinnaker/spinnaker/issues/6615)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -201,7 +201,7 @@ class RunTaskHandler(
                 }
               }
               val status = stage.failureStatus(default = TERMINAL)
-              stage.context["exception"] = exceptionDetails
+              taskModel.taskExceptionDetails["exception"]=exceptionDetails
```
mattgogerly, line 204: What is this change doing? Are parts of Deck not dependent on the exception being at the root of the context rather than under each task?

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandlerTest.kt
```diff
@@ -1028,6 +1029,57 @@ object CompleteStageHandlerTest : SubjectSpek<CompleteStageHandler>({
...
+
+        on("receiving the message") {
+          subject.handle(message)
+        }
+
+        it("updates the stage context to store the task exception") {
+          verify(repository).storeStage(
+            check {
+              assertThat(it.status).isEqualTo(taskStatus)
+              assertThat(((((it.context["exception"] as Map<*, *>)["details"]) as Map<*, *>)["errors"] as ArrayList<*>)[2]).isEqualTo((((it.tasks[0].taskExceptionDetails["exception"] as Map<*, *>)["details"] as Map<*, *>)["errors"] as ArrayList<*>)[0])
```
mattgogerly, line 1063: Can you split this over a couple of lines? It's very difficult to understand.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
```diff
@@ -589,10 +589,10 @@ object RunTaskHandlerTest : SubjectSpek<RunTaskHandler>({
             )
           }
 
-          it("attaches the exception to the stage context") {
+          it("attaches the exception to the taskExceptionDetails") {
             verify(repository).storeStage(
               check {
-                assertThat(it.context["exception"]).isEqualTo(exceptionDetails)
+                assertThat(it.tasks[0].taskExceptionDetails["exception"]).isEqualTo(exceptionDetails)
```
mattgogerly, line 595: Same question as above - does Deck not rely on this?

I'm also concerned about people who may be using SpEL on this field today which this would break.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
SusmithaGU: Restored the previous code

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandlerTest.kt
SusmithaGU: Had to remove this test case and add a new one

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
SusmithaGU: Made changes in `completeStageHandler` to add the task exception errors to the `stageContext` Exception errors in order to display in UI only to work in case where the status is FAILED_CONTINUE.The other fields in the `stageContext `exception remain untouched

dbyron-sf:
@mergifyio update

mattgogerly:
@dbyron-sf you good with this?

mattgogerly (approved):

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteStageHandler.kt
```diff
@@ -127,8 +127,17 @@ class CompleteStageHandler(
               return@withStage
             }
           }
-
           stage.status = status
+          if (status == FAILED_CONTINUE) {
```
dbyron-sf, line 131: I'm curious why only FAILED_CONTINUE here.  What happens if we do this for all statuses, or if not that, maybe also TERMINAL?

dbyron-sf: FAILED_CONTINUE seems sufficient to address https://github.com/spinnaker/spinnaker/issues/6615...does the same exception info show up via some other mechanism for e.g. TERMINAL?  If so, where?

SusmithaGundu: In case of TERMINAL status,the exact failed task exception itself is registered in the `stagecontext` exception and is shown.But for FAILED_CONTINUE this was not happening and hence had to restrict for FAILED_CONTINUE only.

SusmithaGU: In case of TERMINAL status,the exact failed task exception itself is registered in the stagecontext exception and is shown.But for FAILED_CONTINUE this was not happening and hence had to restrict for FAILED_CONTINUE only.

dbyron-sf: In this code below, right?
```
        } catch (e: Exception) {
          log.error("Failed to construct after stages for ${stage.name} ${stage.id}", e)

          val exceptionDetails = exceptionHandlers.shouldRetry(e, stage.name + ":ConstructAfterStages")
          stage.context["exception"] = exceptionDetails
          stage.status = TERMINAL
        }
```

SusmithaGU: The exception for failed task is tied to `stagecontext ` in `Runtaskhandler.kt` in case of TERMINAL status.The fix proposes to tie the exceptions after the completion of the stage for FAILED_CONTINUE status.

dbyron-sf (approved):

mflis:
@SusmithaGU  what's the status of this one? Looks like sth ready to merge.

SusmithaGU:
yes all set to merge

SusmithaGU:
can someone merge this pr

dbyron-sf:
@mergifyio update

PR #4269, opened 2022-05-11
refactor(groovy): migrating Clouddriver services to Java

mattgogerly:
Refactoring to Java ahead of making changes for https://github.com/spinnaker/spinnaker/issues/6651


orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/MortService.java
```diff
@@ -0,0 +1,234 @@
+/*
```
mattgogerly, line 1: This whole class is uh.. fun.

Unfortunately whereas in Groovy we have ?: to gracefully handle null checking for nested objects, such a thing would be too convenient for Java. Because pretty much everything in these services is just a `Map` rather than strongly typed it makes it basically impossible to do any proper checks.

I'd migrate these Maps to be strongly typed but I'm not familiar enough with what the payloads look like for all the different cloud providers that use them.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/securitygroup/SecurityGroupUpserter.groovy
```diff
@@ -33,7 +32,7 @@ interface SecurityGroupUpserter {
    * @return true when, according to the underlying cloud provider, the security group has been updated to match the
    * specified security group.
    */
-  boolean isSecurityGroupUpserted(MortService.SecurityGroup upsertedSecurityGroup, StageExecution stage)
+  boolean isSecurityGroupUpserted(com.netflix.spinnaker.orca.clouddriver.MortService.SecurityGroup upsertedSecurityGroup, StageExecution stage)
```
mattgogerly, line 35: I have absolutely no idea why, but without giving the fully qualified name this fails to compile. Doesn't show any errors in my IDE without the fully qualified name and I don't know enough about Groovy/Java transpiling to debug it.

mattgogerly:
@Merfigyio update

@spinnaker/oss-reviewers

mattgogerly:
@Mergifyio update

@spinnaker/oss-reviewers

mattgogerly:
@Mergifyio update

dbyron-sf:
@mergifyio update

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/securitygroup/SecurityGroupUpserter.groovy
dbyron-sf: FWIW, this happens to me as well.  With the fully qualified name, IntelliJ warns about an "Unnecessary qualified reference", but then when I replace it with
```
diff --git a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/securitygroup/SecurityGroupUpserter.groovy b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/>
index 8d44739ef..d66a00c42 100644
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/securitygroup/SecurityGroupUpserter.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/securitygroup/SecurityGroupUpserter.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.orca.clouddriver.tasks.securitygroup
 
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
+import com.netflix.spinnaker.orca.clouddriver.MortService
 import groovy.transform.Canonical
 
 interface SecurityGroupUpserter {
@@ -32,7 +33,7 @@ interface SecurityGroupUpserter {
    * @return true when, according to the underlying cloud provider, the security group has been updated to match the
    * specified security group.
    */
-  boolean isSecurityGroupUpserted(com.netflix.spinnaker.orca.clouddriver.MortService.SecurityGroup upsertedSecurityGroup, StageExecution stage)
+  boolean isSecurityGroupUpserted(MortService.SecurityGroup upsertedSecurityGroup, StageExecution stage)
 
   /**
    * @return The cloud provider type that this object supports.
```
I get:
```
$ ./gradlew build

> Task :orca-clouddriver:compileGroovy
/Users/dbyron/src/spinnaker/orca/orca-clouddriver/build/tmp/compileGroovy/groovy-java-stubs/com/netflix/spinnaker/orca/clouddriver/tasks/securitygroup/SecurityGroupUpserter.java:15: error: package MortService does not exist
 boolean isSecurityGroupUpserted(MortService.SecurityGroup upsertedSecurityGroup, com.netflix.spinnaker.orca.api.pipeline.models.StageExecution stage);
                                            ^
1 error
startup failed:
Compilation failed; see the compiler error output for details.

1 error


> Task :orca-clouddriver:compileGroovy FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':orca-clouddriver:compileGroovy'.
> Compilation failed; see the compiler error output for details.
```
Note that the compile error is in SecurityGroupUpserter.java -- a generated file.

My next move is to try converting SecurityGroupUpserter.groovy into java, which didn't turn out to be so bad....and works without the fully qualified name here.  If you like, I'll push a commit that does this.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/MortService.java
```diff
@@ -0,0 +1,234 @@
...
+
+    static VPC findForRegionAndAccount(
+        Collection<VPC> allVPCs, String sourceVpcIdOrName, String region, String account) {
+      VPC sourceVpc =
+          allVPCs.stream()
+              .filter(
+                  vpc ->
+                      vpc.id.equalsIgnoreCase(sourceVpcIdOrName)
+                          || (sourceVpcIdOrName.equalsIgnoreCase(vpc.name)
+                              && vpc.region.equalsIgnoreCase(region)
```
dbyron-sf, line 207: The groovy code for this is:
```
      def sourceVpc = allVPCs.find { it.id == sourceVpcIdOrName || (sourceVpcIdOrName.equalsIgnoreCase(it.name) && it.region == region && it.account == account) }
```
so aren't the comparisons for region and account case-sensitive...i.e.
```suggestion
                              && vpc.region.equals(region)
```

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/MortService.java
```diff
@@ -0,0 +1,234 @@
...
+                              && vpc.account.equalsIgnoreCase(account)))
```
dbyron-sf, line 208: and
```suggestion
                              && vpc.account.equals(account)))
```
?

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/MortService.java
dbyron-sf: Thanks for the heads-up here...and for wading through it.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
```diff
@@ -0,0 +1,224 @@
...
+import org.springframework.context.annotation.Import;
+import retrofit.RequestInterceptor;
+import retrofit.RestAdapter;
+import retrofit.converter.JacksonConverter;
+
+@Configuration
+@Import(RetrofitConfiguration.class)
+@ComponentScan({
+  "com.netflix.spinnaker.orca.clouddriver",
+  "com.netflix.spinnaker.orca.kato.pipeline",
```
dbyron-sf, line 52: The groovy code also has
```
  "com.netflix.spinnaker.orca.oort.pipeline",
  "com.netflix.spinnaker.orca.oort.tasks",
```
here.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
```diff
@@ -0,0 +1,224 @@
...
+      List<String> urls =
+          cloudDriverConfigurationProperties.getCloudDriverReadOnlyBaseUrls().stream()
+              .map(CloudDriverConfigurationProperties.BaseUrl::getBaseUrl)
+              .collect(Collectors.toList());
+
+      if (urls.isEmpty()
+          || urls.stream()
+              .allMatch(
+                  url ->
+                      url.equalsIgnoreCase(
```
dbyron-sf, line 131: The groovy code is:
```
      if (cloudDriverConfigurationProperties.cloudDriverReadOnlyBaseUrls*.baseUrl == [cloudDriverConfigurationProperties.cloudDriverBaseUrl]) {
```
and `==` is a case-sensitive comparison, so:
```suggestion
                      url.equals(
```
?

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
```diff
@@ -0,0 +1,224 @@
...
+                          cloudDriverConfigurationProperties.getCloudDriverBaseUrl()))) {
+        log.info(
+            "No readonly URLs configured for clouddriver, using default URL {} for {}",
```
dbyron-sf, line 134: If the logic is really identical to the way it was in groovy, let's keep the same log message, yes?
```
log.info("readonly URL not configured for clouddriver, using writeable clouddriver $cloudDriverConfigurationProperties.cloudDriverBaseUrl for $type.simpleName")
```

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
```diff
@@ -0,0 +1,224 @@
...
+                    selector =
+                        (ServiceSelector)
+                            selectorClass.getConstructors()[0].newInstance(
+                                selector.getService(), selector.getPriority(), url.getConfig());
+                  } catch (Exception e) {
+                    log.error(
+                        "Failed to create selector for class {}, using default service selector for URL {}",
+                        url.getConfig().get("selectorClass"),
+                        url.getBaseUrl());
+                  }
```
dbyron-sf, line 162: Would this exception have bubbled up in the groovy code?  I support adding the logging, but is it really OK to swallow the exception?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/securitygroup/SecurityGroupUpserter.groovy
mattgogerly: If you've got it working you're welcome to push it, or I can replicate this next week. Thanks!

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/MortService.java
mattgogerly: Yep, not sure how I messed that up (and the ones below). Thanks!

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationProperties.java
```diff
@@ -0,0 +1,86 @@
...
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@Data
+@ConfigurationProperties
+public class CloudDriverConfigurationProperties {
+
+  @Getter
+  @Setter
+  public static class BaseUrl {
```
dbyron-sf, line 32: The groovy code has `@Canonical`, so maybe `@Data` is a closer match?  Or add `@ToString` and  `@EqualsAndHashCode` since `@Canonical` does that?

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/securitygroup/SecurityGroupUpserter.groovy
dbyron-sf: Pushed https://github.com/spinnaker/orca/pull/4269/commits/f39c69daa72a23dba7a64df55fcbbcb4fac22aa6.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationProperties.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    return null;
+  }
+
+  List<BaseUrl> getCloudDriverReadOnlyBaseUrls() {
+    if (clouddriver != null
+        && clouddriver.readonly != null
+        && clouddriver.readonly.baseUrl != null) {
+      BaseUrl url = new BaseUrl();
```
dbyron-sf, line 73: It's mildly tempting to add a`BaseUrl(String baseUrl)` constructor to simplify this code (and similar below)...

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
mattgogerly: That's a good catch, thanks. Throws a `GroovyCastException` if an invalid selector is given. I think an `IllegalArgumentException` is ok here?

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
mattgogerly: It doesn't look like those packages exist anymore. I'm assuming they were removed when they were all refactored into Clouddriver?

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
dbyron-sf: Maybe add a test to be sure?  Are you suggesting to catch `IllegalArgumentException` instead of `Exception` (and perhaps re-throw it)?  The more specific the better I guess....though maybe the logging is overkill and whatever exception we get has enough info in it?  A test would tell us.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
dbyron-sf: Good point.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationPropertiesSpec.groovy
```diff
@@ -50,7 +50,11 @@ class CloudDriverConfigurationPropertiesSpec extends Specification {
 
     return new CloudDriverConfigurationProperties.MultiBaseUrl(
       baseUrl: null,
-      baseUrls: baseUrls.collect { new CloudDriverConfigurationProperties.BaseUrl(baseUrl: it) }
+      baseUrls: baseUrls.collect { it ->
```
dbyron-sf, line 53: revert this change now that there's `BaseUrl(String baseUrl)`?

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
mattgogerly: Sure, currently it would throw various different exceptions depending on what was wrong, so I was thinking just catch the Exception and rethrow an IllegalArgumentException with the content. Though I agree the log is redundant - I tested it with an invalid selector class and Orca just fails to start.

The specific exceptions the new code needs to handle are `InstantiationException`, `IllegalAccessException` and `InvocationTargetException`.

dbyron-sf (approved):

dbyron-sf:
@mergifyio update

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationProperties.java
```diff
@@ -0,0 +1,85 @@
...
+    private String baseUrl;
+    private MultiBaseUrl readonly;
+  }
+
+  private BaseUrl mort;
+  private BaseUrl oort;
+  private BaseUrl kato;
+  private CloudDriver clouddriver;
+
+  String getCloudDriverBaseUrl() {
```
jaredstehler, line 56: this changes the visibility of this method from the groovy default of public, was this intended? it appears to be causing an issue with the aws lambda deployment plugin: https://github.com/spinnaker-plugins/aws-lambda-deployment-plugin-spinnaker/issues/125

```
java.lang.IllegalAccessError: class com.amazon.aws.spinnaker.plugin.lambda.verify.LambdaCacheRefreshTask tried to access method 'java.lang.String com.netflix.spinnaker.orca.clouddriver.config.CloudDriverConfigurationProperties.getCloudDriverBaseUrl()' (com.amazon.aws.spinnaker.plugin.lambda.verify.LambdaCacheRefreshTask is in unnamed module of loader org.pf4j.PluginClassLoader @6a00b08c; com.netflix.spinnaker.orca.clouddriver.config.CloudDriverConfigurationProperties is in unnamed module of loader 'app')
	at com.amazon.aws.spinnaker.plugin.lambda.verify.LambdaCacheRefreshTask.forceCacheRefresh(LambdaCacheRefreshTask.java:93)
```

or is there a more accepted way of getting the cloud driver base url in plugins?

jaredstehler: it's used a lot in this plugin: https://github.com/spinnaker-plugins/aws-lambda-deployment-plugin-spinnaker/search?q=getCloudDriverBaseUrl

jaredstehler: https://github.com/spinnaker/orca/pull/4317

PR #4270, opened 2022-05-12
fix(cfn): Fix detection of empty CloudFormation changesets

jaredstehler:
It seems that the status message from AWS has changed or could potentially differ. Updating to check for what I'm getting as output.


jaredstehler:
@dbyron-sf

mattgogerly (approved):

PR #4271, opened 2022-05-25
fix(tasks): Fix MonitorPipelineTask regression

CharlieTLe:
The way that the monitor pipeline task worked before was that it only
stored the _status_ of the child pipeline when the child was running
(see file before the commit
[here](https://github.com/spinnaker/orca/pull/3902/files#diff-4c694b67271c7d9436d5af695f160f0b60c7ad7753a0ba1eb9aea76d7ba4e851R128-R130)).
After the commit was merged, the MonitorPipelineTask was storing the
context of the child pipeline on every update regardless of the child
pipeline’s status ([see line in
commit](https://github.com/spinnaker/orca/blob/20e2b9ea61037e7fc8eb59c817dc5d33f4718af3/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy#L96)).

This should address the issue seen where large nested pipelines
especially those child pipelines that generate a lot of outputs (Deploy
Manifest) negatively impact the performance of Spinnaker.

See: https://github.com/spinnaker/spinnaker/issues/6159

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy
@@ -125,7 +125,7 @@ class MonitorPipelineTask implements OverridableTimeoutRetryableTask {
       result.executionStatuses = pipelineStatuses
     }
 
-    if (pipelineIds.size() == 1) {
+    if (pipelineIds.size() == 1 && allPipelinesCompleted) {
       result.insertPipelineContext(firstPipeline.getContext())
     }
 
--- a/orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTaskSpec.groovy
+++ b/orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTaskSpec.groovy
@@ -222,6 +222,32 @@ class MonitorPipelineTaskSpec extends Specification {
     ]
   }
 
+  def "do not propagate running child pipeline outputs"() {
+    def pipeline = pipeline {
+      application = "orca"
+      name = "a pipeline"
+      stage {
+        type = PipelineStage.PIPELINE_CONFIG_TYPE
+        name = "running stage with outputs"
+        outputs = [
+            "myVar1": "myValue1",
+            "myVar2": "myValue2"
+        ]
+        status = ExecutionStatus.RUNNING
+      }
+      status = ExecutionStatus.RUNNING
+    }
+
+    repo.retrieve(*_) >> pipeline
+
+    when:
+    def result = task.execute(stage)
+    def outputs = result.outputs
+
+    then:
+    outputs == [:]
+  }
+
   @Unroll
   def "respect #behavior behavior when monitoring multiple pipelines"() {
     ObjectMapper objectMapper = new ObjectMapper()


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy
```diff
@@ -125,7 +125,7 @@ class MonitorPipelineTask implements OverridableTimeoutRetryableTask {
       result.executionStatuses = pipelineStatuses
     }
 
-    if (pipelineIds.size() == 1) {
+    if (pipelineIds.size() == 1 && allPipelinesSucceeded) {
```
dbyron-sf, line 128: Can you add an automated test that verifies the new behavior please?

dbyron-sf: Thanks!

dbyron-sf:
I like this change, and thanks for making it.  I'd like to try to clarify something.  This reduces the amount of information stored while nested pipelines are running, but doesn't change behavior after the pipelines have completed, right?  So, definitely a step in the right direction, but we have (potentially a lot of) work to do to resolve https://github.com/spinnaker/spinnaker/issues/6159, yes?

CharlieTLe:
@dbyron-sf Right, this is only changing when the MonitorPipelineTask stores the child outputs into its own outputs. This change should significantly improve the issue seen in [6159](https://github.com/spinnaker/spinnaker/issues/6159).

dbyron-sf:
> @dbyron-sf Right, this is only changing when the MonitorPipelineTask stores the child outputs into its own outputs. This change should significantly improve the issue seen in [6159](https://github.com/spinnaker/spinnaker/issues/6159). If the `noOutput` property is set in the deploy stage alongside with this change, it should drastically improve performance and reduce the sizes of the pipeline executions that gets saved to the execution repository.

I think I'm still not understanding.  Doesn't this change only make a difference while a child pipeline is running, but once it's complete, the result is the same?  The code for `insertPipelineContext` is:
```
    public void insertPipelineContext(Map<String, Object> pipelineContext) {
      monitoredPipelineContext.putAll(pipelineContext);
    }
```
Have you noticed that the running pipeline's context changes significantly during execution and then has more changes when complete, such that multiple calls to `insertPipelineContext` would end up adding more information instead of replacing info each time?

CharlieTLe:
@dbyron-sf Please allow me to clarify. 

The problem is that the parent is unnecessarily storing the child pipeline's outputs while the child pipeline is still running. This problem is exacerbated when the child pipeline has a lot of outputs.

This change reduces the amount of information that's persisted to the execution repository while the child pipeline is still running.

There is the case where if the child pipeline chooses to clear its outputs before completing, the parent pipeline will not have any outputs to store into its own context. This makes a difference when the child pipeline's output is different depending on the status of the child pipeline.

Perhaps a concrete example would help.

Say I have 3 pipelines: A, B, C.

Pipeline A calls pipeline B, and pipeline B calls pipeline C. (A -> B -> C)

Pipeline C deploys 1000 Kubernetes objects (Deployments, Services, ConfigMaps).

Let's say that pipeline C is waiting for the manifests to stabilize and it takes a couple of hours to do so. Pipeline C stores all of the manifests it's deploying to its own outputs while this is happening of course.

Concurrently, pipeline B is storing the outputs of pipeline C into its own outputs while pipeline C is still running. Furthermore, pipeline A is storing the outputs of pipeline B since it is still running as well.

So pipeline A is storing the outputs of pipeline B, and pipeline B is storing the outputs of pipeline C. Orca has to save all of this to the ExecutionRepository on every invocation of the MonitorPipeline task. If the ExecutionRepository is unable to accommodate the requests, Orca grinds to a halt especially is the work queue is the same as the ExecutionRepository.

Now once this change is introduced, the problem disappears since pipeline A and pipeline B does not store the output of pipeline C until after pipeline C completes. If pipeline C implements the `noOutput` property to have the outputs deleted after the stage completes, pipeline A and B will have no need to store any outputs at all, thus making the pipeline sizes smaller which translates to faster processing of the MonitorPipeline task.

CharlieTLe:
@dbyron-sf Did that explanation help? Happy to answer more questions.

@robzienert @marchello2000 Are you two still the right people to help with this?

robzienert:
Neither of us are involved in the project anymore, sorry!

dbyron-sf:
> @dbyron-sf Did that explanation help? Happy to answer more questions.

Yup, that helps, thank you!

PR #4274, opened 2022-06-07
fix(orca-core): Added unimplemeted methods of ApplicationEventMulticaster class

madhubkk:
There are two unimplemented methods as per latest release of spring framework from [5.3.5](https://docs.spring.io/spring-framework/docs/5.3.5/javadoc-api/org/springframework/context/event/ApplicationEventMulticaster.html) version onwards.

While building orca with spring boot upgrade, faced errors like:
```
> Task :orca-core:compileKotlin
e: /home/opsmx/spinnakerforkcodebase/orca/orca-core/src/main/java/com/netflix/spinnaker/orca/listeners/DelegatingApplicationEventMulticaster.kt: (31, 1): Class 'DelegatingApplicationEventMulticaster' is not abstract and does not implement abstract member public abstract fun removeApplicationListeners(p0: Predicate<ApplicationListener<*>!>!): Unit defined in org.springframework.context.event.ApplicationEventMulticaster
```


dbyron-sf:
included in https://github.com/spinnaker/orca/pull/4302

PR #4287, opened 2022-08-04
feat(config): allow configuration of writeable clouddriver endpoints by account and/or cloudProvider

mattgogerly:
Adds the `SelectableService` wrapper around the Kato and Cache services, allowing for endpoint selection based on criteria configured per https://spinnaker.io/docs/guides/runbooks/sharding-spinnaker/#clouddriver-read-only-shards.

This relies on https://github.com/spinnaker/kork/pull/967 to build.

- [x] JavaDoc
- [x] spinnaker.io PR

Closes https://github.com/spinnaker/spinnaker/issues/6651

### Rationale
Running deployments across network tiers can be challenging from a controls and regulatory standpoint. For instance, Spinnaker running on public cloud provider A deploying to on-premise cloud provider B may have to pass credentials across network tiers. This is especially an issue with Cloud Foundry, which can sometimes require basic auth via UAA. 

Additionally the implementation of the CF provider also downloads an artifact locally before re-uploading it to the foundation, which is painful from a bandwidth and latency standpoint when you're having to download and then re-upload across tiers.

The goal of this configuration is to be able to deploy instances of Clouddriver in a network tier that can handle a subset deployments (i.e. to a specific provider), orchestrated by Orca running somewhere else, whilst still being able to read from a co-located Clouddriver.

I don't envision this config would be used by many, but making it a plugin proved challenging because of how wired up these classes are. There should be no impact to current functionality to anybody who doesn't specifically add this configuration.

### DIFF
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
@@ -143,7 +143,7 @@ public class CloudDriverConfiguration {
      * @return a {@link SelectableService} configured with the {@link ServiceSelector} instances
      *     from the profile
      */
-    private <T> SelectableService buildWriteableService(Class<T> type) {
+    public <T> SelectableService buildWriteableService(Class<T> type) {
       List<String> urls =
           cloudDriverConfigurationProperties.getCloudDriverWriteOnlyBaseUrls().stream()
               .map(CloudDriverConfigurationProperties.BaseUrl::getBaseUrl)
@@ -197,7 +197,7 @@ public class CloudDriverConfiguration {
       return serviceSelectors;
     }
 
-    private <T> T buildService(Class<T> type, String url) {
+    public <T> T buildService(Class<T> type, String url) {
       return new RestAdapter.Builder()
           .setRequestInterceptor(spinnakerRequestInterceptor)
           .setEndpoint(newFixedEndpoint(url))
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
@@ -28,6 +28,7 @@ import com.netflix.spinnaker.kork.web.selector.SelectableService;
 import com.netflix.spinnaker.kork.web.selector.ServiceSelector;
 import com.netflix.spinnaker.orca.api.operations.OperationsRunner;
 import com.netflix.spinnaker.orca.clouddriver.*;
+import com.netflix.spinnaker.orca.clouddriver.config.CloudDriverConfigurationProperties.BaseUrl;
 import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper;
 import com.netflix.spinnaker.orca.retrofit.RetrofitConfiguration;
 import com.netflix.spinnaker.orca.retrofit.logging.RetrofitSlf4jLog;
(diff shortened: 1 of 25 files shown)


mattgogerly:
@Mergifyio update

mattgogerly:
@Mergifyio update

mattgogerly:
Finally got round to a spinnaker.io PR - https://github.com/spinnaker/spinnaker.io/pull/299

mattgogerly:
@Margifyio update

mattgogerly:
@Mergifyio update

mattgogerly:
@Mergifyio update

dbyron-sf:
LGTM.  How much real-world testing has this got so far @mattgogerly ?

mattgogerly:
> LGTM. How much real-world testing has this got so far @mattgogerly ?

I haven't had a chance to let this out into the wild yet, only some basic dev testing. I'll sanity check it again tomorrow and update here, but my view is that since this is explicitly opt-in it should be safe to merge.

This change will form a big part of our architecture going forward so we're committed to resolving issues with it.

dbyron-sf:
Sounds great @mattgogerly .  If I'm reading this right, this introduces the notion of a write-only clouddriver, so even if it's opt-in, it feels significant.

mattgogerly:
Good thing I checked again, because the Cloud Provider selector doesn't work.

https://github.com/spinnaker/kork/pull/1012 for fix.

mattgogerly:
@dbyron-sf added a test demonstrating the bug.

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationTest.java
```diff
@@ -93,9 +93,10 @@ public OkHttpClient.Builder get(@NotNull ServiceEndpoint service) {
   @Test
   public void testConstructingSelectorsWithYAMLList() {
     /*
-     * Because the configuration of a ServiceSelector is a Map<String, Object>, Jackson converts a List
-     * to a LinkedHashMap. This test validates this by checking no ClassCastException is thrown trying to
-     * cast a LinkedHashMap to a List.
+     * Because the configuration of a ServiceSelector is a Map<String, Object>, Jackson converts a YAML list
+     * to a LinkedHashMap. This caused issues in Kork where the ServiceSelector expected, and tried, to cast
+     * the config from the Object to a List. This test validates that Orca is proving what Kork expects
```
dbyron-sf, line 98:
```suggestion
     * the config from the Object to a List. This test validates that Orca is providing what Kork expects
```

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationTest.java
```diff
@@ -0,0 +1,104 @@
...
+            logLevel,
+            requestInterceptor,
+            cloudDriverConfigurationProperties);
+  }
+
+  @DisplayName("when selector config is a YAML list, successfully constructs a SelectableService")
+  @Test
+  public void testConstructingSelectorsWithYAMLList() {
+    /*
+     * Because the configuration of a ServiceSelector is a Map<String, Object>, Jackson converts a YAML list
```
dbyron-sf, line 96: [ServiceSelector is an interface](https://github.com/spinnaker/kork/blob/5c70628ddaca00015222264e89d24faaa3bd9fd3/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ServiceSelector.java) with no mention of `Map<String, Object>`.  Maybe better to reference ServiceSelector implementations, e.g. [ByExecutionTypeServiceSelector](https://github.com/spinnaker/kork/blob/5c70628ddaca00015222264e89d24faaa3bd9fd3/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByExecutionTypeServiceSelector.java#L29)?  But then also it seems like the basis of this is the type of the [config member of BaseUrl](https://github.com/spinnaker/orca/blob/d1e42deb42aae0aa9e9089a95cc2dfb298030362/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationProperties.java#L42) -- Map<String, Object>, and then that getting passed to ServiceSelector constructors [here](https://github.com/spinnaker/orca/pull/4287/files#diff-4e82bc2b7d7efc31ab7780eeb09035ed0f8993a897f66d3c040bdfbbd926f09cR185).

mattgogerly: Updated the comment to better reflect what's going on, and why this test is necessary. Let me know if it's any better.

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationTest.java
```diff
@@ -0,0 +1,108 @@
...
+            cloudDriverConfigurationProperties);
+  }
+
+  @DisplayName("when selector config is a YAML list, successfully constructs a SelectableService")
+  @Test
+  public void testConstructingSelectorsWithYAMLList() {
+    /*
+     * Because the configuration of a BaseUrl in Orca is loosely typed as a Map<String, Object>, to allow variations of
+     * config that require different types, Jackson doesn't have enough type information and converts a YAML list to a
+     * LinkedHashMap. This caused issues in Kork where the ServiceSelector constructors expected a Map<String, Object>
```
dbyron-sf, line 98: I think we're getting there....maybe the first paragraph of this comment could live the test yml file since that's where the expectation for a list comes from?

After staring at this in the debugger, I also think it would help to add some detail about what the LinkedHashMap built from the yaml list looks like.  But then...still trying to clarify that we're not talking about what orca passes as the `config` constructor parameter -- it's a Map<String, Object>.  We're talking about the type of the individual values.  So far I've got:

```suggestion
     * LinkedHashMap (where the keys are 0, 1, 2, 3, etc. and the values are the list elements). This test verifies that the ServiceSelector constructors treat the values of `Map<String, Object> config` properly -- as Maps.
```
an then maybe the next two lines can disappear?

And yes, the test is failing (as expected until https://github.com/spinnaker/kork/pull/1012), so that's good.

mattgogerly: Just looked at it again, and you're right. It's not the Map<String, Object> that's the issue, it's the Object within the map being a Map rather than a List.

I'll clarify that tomorrow.

mattgogerly: I think the comment does a better job explaining now.

dbyron-sf: Agreed, thank you!

dbyron-sf (approved):

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java
```diff
@@ -129,43 +122,94 @@ private <T> SelectableService buildReadOnlyService(Class<T> type) {
...
+      List<ServiceSelector> selectors =
+          getServiceSelectors(
+              cloudDriverConfigurationProperties.getCloudDriverWriteOnlyBaseUrls(), type);
       return new SelectableService(selectors);
     }
+
+    private <T> List<ServiceSelector> getServiceSelectors(List<BaseUrl> baseUrls, Class<T> type) {
+      List<ServiceSelector> serviceSelectors = new ArrayList<>();
+
+      baseUrls.forEach(
```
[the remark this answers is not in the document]
mattgogerly: I did think about that, but Gate also does the same thing (with a different config structure because.. reasons..). Agree it would be nice to shift this stuff out of here.

mattgogerly:
@Mergifyio update

PR #4289, opened 2022-08-10
fix(stageExecution): In MJ stages find the correct authenticated user…

OscarMichelH:
### Issue:
When a Manual Judgment Stage (aka MJ stage) is SKIPPED, and we have a previous MJ SUCCEED with 'Propagate Authentication' enabled, the next stages are ignoring the previous approved MJ with Propagate Authentication, and taking as Autheticated User the execution authenticated user (The user that started the pipeline execution).
### Requisites
  - AuthZ system enabled
  - Roles configured to restrict pipelines/infrastructure interactions

### Example
Permissions used:
```
  permissions:
    READ:    
      - admin
      - restricted
    WRITE:    
      - admin
    EXECUTE:    
      - admin
      - restricted
```
Roles:
```
users:  - username: OscarMichelH
    roles:      
    - admin
  - username: oscar4armory
    roles:      
    - restricted
```
In the next example we have 2 similar pipelines deploying the same instance, and the MJ will be approved for the same unauthorized user:

image

After the approval from the same user we have the next response:
image

So, as the issue is describing, when we have 2 MJ, the second, if is skipped resets the permissions. In the last image the pipeline continue deploying because is taking the permissions from the user that started the pipeline, not the permission from the user that approved the first MJ.

If we tried the opposite case, starting the pipelines with the unauthorized user and approving the MJ’s with the authorized user, we have the next results:

image

As we can see, now the first pipeline stopped for an unauthorized reason because is not taking the propagation option with the second MJ, and is using the permissions from the user that triggered the pipeline, the second pipeline is taking the propagation auth from the correct user because we only have a single MJ.

Solution:
Look for a possible candidate in the previous stages to retrieve the last propagated authentication used ignoring the skipped stages


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
```diff
@@ -56,4 +57,23 @@ interface AuthenticationAware {
       ExecutionContext.clear()
     }
   }
+  fun backtrackSkippedStages(stage: StageExecution): StageExecution {
+    if (stage.type == "manualJudgment" &&
```
chriskotula, line 61: I have an idea how we could approach such checks in a slightly different way - not better, probably not worse, just different. 
There's the Demeter law that says that a module should not have the knowledge on the inner details of the objects it manipulates. In this case we need to get access to internal details (like type or status) and then use them in checks like stage.type == "X". We could replace it with methods like stage.isType("X") or stage.isStatus("S"). It's very convenient to use and it is future-proof. If one day stage.type implementation changes (now type is a String, but in the future it might be a designated new class like Type), checks like stage.type == "X" will need to be updated everywhere to match stage.type new class. If we create a convenient method like stage.isType("X"), we can make adjustments to new type class only inside that method. 

Just a food for thought :) I am fine with your changes as they are now!

chriskotula (approved):
Well done 👏

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
OscarMichelH: Nice though @chriskotula I applied your recommendation, It will let us keep a maintainable code for possible future changes. Thanks!

sergio-quintero (approved):

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/ExecutionStatus.java
```diff
@@ -71,6 +71,10 @@ public final boolean isComplete() {
     return complete;
   }
 
+  public final boolean isSkipped() {
+    return this.toString() == "SKIPPED";
```
nemesisOsorio, line 75: I believe you can use something like `SKIPPED.equals(this)`

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
```diff
@@ -56,4 +57,24 @@ interface AuthenticationAware {
...
     }
   }
+
+  fun backtrackSkippedStages(stage: StageExecution): StageExecution {
+    if (stage.isManualJudgmentType &&
+      !stage.status.isSkipped &&
+      stage.withPropagateAuthentication()) {
+      return stage;
+    }
+    else {
```
nemesisOsorio, line 67: you can remove `else` to avoid nested statements

orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/ExecutionStatus.java
OscarMichelH: Excellent, checking the previous code I was looking by something similar

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
OscarMichelH: Done!, Thank you

nemesisOsorio (approved):
LGTM

dbyron-sf (changes requested):
please add tests for this

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
```diff
@@ -614,6 +614,16 @@ public <O> O mapTo(@Nullable String pointer, @Nonnull Class<O> type) {
...
   }
 
+  @Override
+  public boolean isManualJudgmentType() {
+    return this.type.equals("manualJudgment");
+  }
+
+  @Override
+  public boolean withPropagateAuthentication() {
+    return Boolean.parseBoolean(context.get("propagateAuthenticationContext").toString());
```
dbyron-sf, line 624: What if `context.get("propagateAuthenticationContext")` returns null?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
```diff
@@ -28,10 +28,11 @@ interface AuthenticationAware {
   val stageNavigator: StageNavigator
 
   fun StageExecution.withAuth(block: () -> Unit) {
-    val authenticatedUser = stageNavigator
+    val firstAncestorOrNull = stageNavigator
```
dbyron-sf, line 31: Curious about the motivation here if nothing else uses `firstAncestorOrNull` except to figure out `authenticatedUser`.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
```diff
@@ -56,4 +57,22 @@ interface AuthenticationAware {
...
   }
+
+  fun backtrackSkippedStages(stage: StageExecution): StageExecution {
+    if (stage.isManualJudgmentType &&
+      !stage.status.isSkipped &&
+      stage.withPropagateAuthentication()) {
+      return stage;
+    }
+    val previousStage = if (stage.ancestors().size > 1) stage.ancestors().get(1) else null
+    previousStage?: return stage
```
dbyron-sf, line 68: I could use some comments here explaining the logic.

dbyron-sf:
@mergifyio update

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
OscarMichelH: I think was part of the strategy to implement an AuthenticatedStage as interface, this interface allows an implementing StageDefinitionBuilder to override the default pipeline authentication context. In this case, was retrieving the authenticated user from the previous stage, but when a stage is skipped the chain of authenticated user permissions  is broken.

dbyron-sf: I think I'm asking a simpler question...like, there's a diff here to add a new variable, but there's nothing else that uses the new variable...so, do we need the new variable after all?

OscarMichelH: Oh sorry my error, I did not understand, now I know about your recommendation. Thank you, I added a simplification to prevent that "one-call variables"

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
dbyron-sf: Thanks for that.  I could still use some comments here though.  Or maybe a change to the code would make it easier for me to read/understand without comments.  I think what it's doing is:

* if there's no previous stage, return the current stage
* otherwise, return `backtrackSkippedStages(previousStage);`

For me that would be easier to see with:
```
    if (previousStage == null) {
      return stage;
    }
```
A ternary would be even clearer, at least for me...but since kotlin doesn't have them
```
return if (previousStage == null) stage else backtrackSkippedStages(previousStage);
```

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
dbyron-sf: I was actually thinking of something different, to minimize the diffs, but I like splitting it out to a separate method too.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
OscarMichelH: I agree, looks more elegant, anyway I added extra comments to clarify for future developers 😃

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
OscarMichelH: I added an extra validation to prevent convert a null toString. Thank you for the observation

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
```diff
@@ -1839,6 +1831,70 @@ object RunTaskHandlerTest : SubjectSpek<RunTaskHandler>({
       }
     }
   }
+
+  describe("when a stage had skipped manual judgment stage with auth propagated") {
```
dbyron-sf, line 1835: Would this be clearer as:
```suggestion
  describe("when a previous stage was a skipped manual judgment stage with auth propagated") {
```

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
```diff
@@ -1839,6 +1831,70 @@ object RunTaskHandlerTest : SubjectSpek<RunTaskHandler>({
...
+    given("a stage with a manual judgment type and auth propagated") {
+      val timeout = Duration.ofMinutes(5)
+      val lastModifiedUser = StageExecution.LastModifiedDetails()
+      lastModifiedUser.user = "user2"
+      val pipeline = pipeline {
+        stage {
+          refId = "1"
+          type = "manualJudgment"
+          status = SUCCEEDED
+          requisiteStageRefIds = setOf("1")
```
dbyron-sf, line 1845: This is a circular dependency, right?  I don't think it's necessary...

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
dbyron-sf: Plus also, I ran these tests locally with this diff:
```
diff --git a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
index 59c16b120..6e4ea8df2 100644
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
@@ -56,7 +56,8 @@ interface AuthenticationAware {
     return stageNavigator
       .ancestors(stage)
       .firstOrNull { it.stageBuilder is AuthenticatedStage } ?.let{
-        (it.stageBuilder as AuthenticatedStage).authenticatedUser(solveSkippedStages(it.stage)).orElse(null)
+      //        (it.stageBuilder as AuthenticatedStage).authenticatedUser(solveSkippedStages(it.stage)).orElse(null)
+        (it.stageBuilder as AuthenticatedStage).authenticatedUser(it.stage).orElse(null)
       }
   }
 ```
and I expected them to fail, but they still pass.

File: orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt (hunk 8 above)
Comment (OscarMichelH) [reply orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt]: Removed

File: orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt (hunk 7 above)
Comment (OscarMichelH) [reply orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt]: Solved, try run again, I was not mocking the ancestor call to return something, so always was returning null, but now is checking working as expected

Comment (dbyron-sf) [reply orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt]: It does fail now, thank you.  The failure is:
```
org.mockito.exceptions.verification.TooFewActualInvocations: 
stageNavigator.ancestors(
    Stage {id='01GBRAFJC4SSD9AC74E5EZBAV3', executionId='01GBRAFJAFQF2443HFYHJSTN1P'}
);
Wanted 2 times:
-> at com.netflix.spinnaker.orca.pipeline.util.StageNavigator.ancestors(StageNavigator.java:46)
But was 1 time:
-> at com.netflix.spinnaker.orca.q.handler.AuthenticationAware$DefaultImpls.retrieveAuthenticatedUser(AuthenticationAware.kt:57)

	at com.netflix.spinnaker.orca.pipeline.util.StageNavigator.ancestors(StageNavigator.java:46)
	at com.netflix.spinnaker.orca.q.handler.RunTaskHandlerTest$1$8$1$5.invoke(RunTaskHandlerTest.kt:1923)
	at com.netflix.spinnaker.orca.q.handler.RunTaskHandlerTest$1$8$1$5.invoke(RunTaskHandlerTest.kt:86)
```
from this assertion:
```
      it("verify stage backtracking times") {
        verify(stageNavigator, times(2)).ancestors(stage)
      }
```
Enough time has passed that I don't remember the details of the PR well enough.  Is there a way to structure the test so there's a more "direct" failure -- like a pipeline fails where it would otherwise succeed, or some kind of auth failure, or something?  Counting the number of times a method gets called seems a little fragile.

Comment (OscarMichelH) [reply orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt]: The gap here is the task context, because I am not able to send a ManualJudgment stage SKIPPED that will crash by authenticated issues, it never crash because by default will take the execution user (that is the current bug), so go into the ancestors more that 1 time means that a backtracking was applied (By the moment I changed for 'atLeast' for remove some fragility in the test)

Comment (dbyron-sf) [reply orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt]: Are you on spinnaker slack?  I'd love to chat about this.

Comment (OscarMichelH) [reply orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt]: I added my account to spinnaker slack now, let me send you a message :D

Comment (OscarMichelH) [reply orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt]: Hey @dbyron-sf from our yesterday talk, I applied some changes to test directly the class ManualJudgmentStage, and I cleaned the code to have a direct test case. Again, tank you to talked with me, your advices were very helpfull

File: orca-queue/orca-queue.gradle [hunk 9]
```diff
@@ -38,5 +38,7 @@ dependencies {
   testImplementation(project(":orca-test-kotlin"))
   testImplementation(project(":orca-queue-tck"))
   testImplementation(project(":orca-queue-redis"))
+  testImplementation(project(":orca-echo"))
   testImplementation("org.assertj:assertj-core")
+  testImplementation("org.codehaus.groovy:groovy-all:3.0.12")
```
dbyron-sf, line 43: I tried taking this out locally and the test still ran fine.  If there's something I missed and we really do need this, please use `org.codehaus.groovy:groovy-all:2.5.14` to match what the rest of orca (and kork) uses.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
dbyron-sf: Excellent!

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
```diff
@@ -46,17 +48,7 @@ import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository
...
-import com.netflix.spinnaker.orca.q.DummyTask
-import com.netflix.spinnaker.orca.q.DummyTimeoutOverrideTask
-import com.netflix.spinnaker.orca.q.InvalidTask
-import com.netflix.spinnaker.orca.q.InvalidTaskType
-import com.netflix.spinnaker.orca.q.PauseTask
-import com.netflix.spinnaker.orca.q.RunTask
-import com.netflix.spinnaker.orca.q.StageDefinitionBuildersProvider
-import com.netflix.spinnaker.orca.q.TasksProvider
-import com.netflix.spinnaker.orca.q.singleTaskStage
+import com.netflix.spinnaker.orca.q.*
```
dbyron-sf, line 51: I thought the linting would catch this.  In general we try to avoid star imports.

orca-queue/orca-queue.gradle
dbyron-sf: OK, after a few more iterations, it appears something like this really is required.  My new, improved suggestion is:
```suggestion
  testImplementation("org.codehaus.groovy:groovy")
```

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
```diff
@@ -114,7 +109,15 @@ object RunTaskHandlerTest : SubjectSpek<RunTaskHandler>({
   val contextParameterProcessor = ContextParameterProcessor()
   val dynamicConfigService: DynamicConfigService = mock()
   val taskExecutionInterceptors: List<TaskExecutionInterceptor> = listOf(mock())
-  val stageResolver = DefaultStageResolver(StageDefinitionBuildersProvider(emptyList()))
+  val stageResolver : DefaultStageResolver = mock()
+  val manualJudgmentStage : ManualJudgmentStage = ManualJudgmentStage()
+
+  whenever(stageResolver.getStageDefinitionBuilder(manualJudgmentStage.type, null)) doReturn manualJudgmentStage
+  whenever(stageResolver.getStageDefinitionBuilder(zeroTaskStage.type, null)) doReturn zeroTaskStage
```
dbyron-sf, line 116: Is this required?  My credibility is potentially shot, but I took out this line and the tests still work.

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
```diff
@@ -1839,6 +1842,74 @@ object RunTaskHandlerTest : SubjectSpek<RunTaskHandler>({
...
+      lastModifiedUser.user = "user"
+      lastModifiedUser.allowedAccounts = listOf("user")
+
+      val pipeline = pipeline {
+        stage {
+          refId = "1"
+          type = manualJudgmentStage.type
+          manualJudgmentStage.plan(this)
+          status = SUCCEEDED
+          context["manualSkip"] = true
```
dbyron-sf, line 1859: The tests still work without this line too.

dbyron-sf (commented):
Thanks so much for making the changes.  The tests seem much more direct now.

orca-queue/orca-queue.gradle
OscarMichelH: Removed

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
OscarMichelH: Changed for only used packages

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
OscarMichelH: Not required, was part for the previous test implementation, I removed it now

orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt
OscarMichelH: Removed

orca-queue/orca-queue.gradle
```diff
@@ -40,5 +40,4 @@ dependencies {
   testImplementation(project(":orca-queue-redis"))
   testImplementation(project(":orca-echo"))
   testImplementation("org.assertj:assertj-core")
-  testImplementation("org.codehaus.groovy:groovy-all:3.0.12")
```
dbyron-sf: Sorry for the thrashing here.  At least for me locally, I need something like
```
  testImplementation("org.codehaus.groovy:groovy")
```
otherwise I see:
```
> Task :orca-queue:compileTestKotlin
e: Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:
    class com.netflix.spinnaker.orca.echo.pipeline.ManualJudgmentStage, unresolved supertypes: groovy.lang.GroovyObject

e: /Users/dbyron/src/spinnaker/orca/orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt: (1860, 38): Cannot access 'groovy.lang.GroovyObject' which is a supertype of 'com.netflix.spinnaker.orca.echo.pipeline.ManualJudgmentStage'. Check your module classpath for missing or conflicting dependencies
e: /Users/dbyron/src/spinnaker/orca/orca-queue/src/test/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerTest.kt: (1871, 36): Cannot access 'groovy.lang.GroovyObject' which is a supertype of 'com.netflix.spinnaker.orca.echo.pipeline.ManualJudgmentStage'. Check your module classpath for missing or conflicting dependencies
```

OscarMichelH: I removed it at all, I ran the tests, and are passing without problems. I was using the dependency to experiment with ManualJudgmentStage(groovy object)

OscarMichelH:
@Mergifyio update

OscarMichelH:
@Mergifyio update

dbyron-sf (approved):

PR #4290, opened 2022-08-15
fix(clouddriver): fix property binding for Clouddriver

mattgogerly:
I have no idea why this is necessary, since `@Data` includes `@RequiredArgsConstructor`, which in this case is equivalent to `@NoArgsConstructor`. Regardless, it doesn't work using `@Data`. Must be some weird interaction between Lombok and nested classes.

Fixes https://github.com/spinnaker/spinnaker/issues/6704


mattgogerly:
@Mergifyio update

dbyron-sf (approved):

PR #4294, opened 2022-08-23
feat(igor): Stop Jenkins job when job name has slashes in the job name

ovidiupopa07:
When defining a Jenkins job inside folders, the name contains slashes. Because of that instead of matching the request to the IGOR STOP endpoint (/masters/{name}/jobs/{jobName}/stop/{queuedBuild}/{buildNumber}), Spring is matching the request to the BUILD one  (/masters/{name}/jobs/**)

The stop request is failing because it is trying to start a job that does not exist.

Feature flag added to call the existing endpoint (which accepts the job name as path variable) or the new one (which accepts  the job name as a query parameter)

Related Igor PR https://github.com/spinnaker/igor/pull/1038

### DIFF
--- a/orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/BuildService.java
+++ b/orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/BuildService.java
@@ -18,15 +18,22 @@ package com.netflix.spinnaker.orca.igor;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
 import java.util.List;
 import java.util.Map;
-import lombok.RequiredArgsConstructor;
+
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import org.springframework.web.util.UriUtils;
 import retrofit.client.Response;
 
 @Component
-@RequiredArgsConstructor
 public class BuildService {
   private final IgorService igorService;
+  private final boolean jobNameAsQueryParameter;
+
+  public BuildService(IgorService igorService, @Value("${jobName.queryParameter:false}") boolean jobNameAsQueryParameter) {
+    this.igorService = igorService;
+    this.jobNameAsQueryParameter = jobNameAsQueryParameter;
+  }
+
 
   private String encode(String uri) {
     return UriUtils.encodeFragment(uri, "UTF-8");
@@ -42,6 +49,9 @@ public class BuildService {
   }
 
   public String stop(String master, String jobName, String queuedBuild, Integer buildNumber) {
+    if(this.jobNameAsQueryParameter) {
+      return igorService.stopWithJobNameAsQueryParameter(master, jobName, queuedBuild, buildNumber, "");
+    }
     return igorService.stop(master, jobName, queuedBuild, buildNumber, "");
   }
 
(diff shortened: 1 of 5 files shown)


dogonthehorizon (approved):

orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/BuildService.java
```diff
@@ -18,15 +18,22 @@
...
 
 @Component
-@RequiredArgsConstructor
 public class BuildService {
   private final IgorService igorService;
+  private final boolean jobNameAsQueryParameter;
+
+  public BuildService(
+      IgorService igorService,
+      @Value("${jobName.queryParameter:false}") boolean jobNameAsQueryParameter) {
```
dbyron-sf, line 33: A separate configuration class is usually cleaner than sprinkling `@Value` annotations around.  See e.g. [this example](https://github.com/spinnaker/orca/blob/master/orca-web/src/main/java/com/netflix/spinnaker/config/EnhancedMonitoringConfigurationProperties.java), enabled [here](https://github.com/spinnaker/orca/blob/master/orca-web/src/main/java/com/netflix/spinnaker/config/EnhancedMonitoringConfiguration.java#L39).

ovidiupopa07: Okay, I will create a separate properties class and extract the bean definition to a configuration class

dbyron-sf: Thanks.

orca-igor/src/test/groovy/com/netflix/spinnaker/orca/igor/BuildServiceSpec.groovy
```diff
@@ -29,10 +29,11 @@ class BuildServiceSpec extends Specification {
   private static final JOB_NAME_ENCODED = "name/with/slashes%20and%20spaces"
   private static final PARAMS = ['key': 'value']
   private static final FILENAME = 'file.txt'
+  private static final QUEUE_BUILD = ''
```
dbyron-sf, line 32:
```suggestion
  private static final QUEUED_BUILD = ''
```
to match the method argument?  Also, could you make it a non-empty string to serve as an example?

orca-igor/src/test/groovy/com/netflix/spinnaker/orca/igor/BuildServiceSpec.groovy
```diff
@@ -29,10 +29,11 @@ class BuildServiceSpec extends Specification {
   private static final JOB_NAME_ENCODED = "name/with/slashes%20and%20spaces"
   private static final PARAMS = ['key': 'value']
   private static final FILENAME = 'file.txt'
+  private static final QUEUE_BUILD = '12'
```
dbyron-sf, line 32: Thanks for changing the value.  I'd still like to change the name to match the method argument name.
```suggestion
  private static final QUEUED_BUILD = '12'
```

dbyron-sf (approved):

dbyron-sf:
@mergifyio update

link108:
@mergifyio backport release-1.26.x release-1.27.x release-1.28.x

PR #4307, opened 2022-10-03
fix(stageExecution): In evaluable variable stage restart scenario variables are not cleaned properly (#16)

OscarMichelH:
Issue: Expression evaluates the previous data when upstream stage(Manual Judgement) is restarted (or other restar case).
image

Steps to reproduce:
1. Execute the pipeline and choose Continue at the Manual Judgement stage. Upon completion note that the status variable inside "Evaluate Variables" was evaluated to 'success'; this is correct. 

2. Execute the pipeline again and when it reaches the Manual Judgment, wait 1 min for it to timeout or stop manually the judgment. Upon timeout, note that the status variable inside "Evaluate Variables" was evaluated to 'failed'; again this is correct. 

3. Click on the failed Manual Judgment stage and choose Actions -> Restart Manual Judgement from the drop down.  
The pipeline will execute. Choose Continue for the Manual Judgement. Upon completion note that the status variable inside "Evaluate Variables" has still evaluated to 'failed'; this is incorrect as the deploy stage was successful. 

Solution:
EvaluateVariablesStage does not have logic created for restart scenarios, so the values for the expression previously executed are preserved. 

prepareStageForRestart function for EvaluateVariablesStage object needs clear the values for a new execution.

Scenario whit approach applied:

image

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/EvaluateVariablesStage.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/EvaluateVariablesStage.java
@@ -50,6 +50,22 @@ public class EvaluateVariablesStage extends ExpressionAwareStageDefinitionBuilde
   }
 
   @Override
+  public void prepareStageForRestart(@Nonnull StageExecution stage) {
+    stage.getOutputs().clear();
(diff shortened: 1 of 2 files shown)


dogonthehorizon (approved):

link108:
@mergifyio backport release-1.27.x release-1.28.x release-1.29.x

PR #4316, opened 2022-10-20
fix(pipelineTriggers): change buildNumber type from int to string

jaredstehler:
in order to support build numbers which are not strict integers (i.e. "123.1")

https://github.com/spinnaker/spinnaker/issues/6743


orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTaskSpec.groovy
```diff
@@ -118,7 +118,7 @@ class CreateBakeTaskSpec extends Specification {
 
   @Shared
   def buildInfo = new JenkinsBuildInfo(
-    "name", 0, "http://jenkins", "SUCCESS",
+    "name", null, "http://jenkins", "SUCCESS",
```
dbyron-sf, line 121: Wouldn't `"0"` be a more straightforward conversion?  Though I'm all for testing with null too.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/BuildInfo.kt
```diff
@@ -2,7 +2,7 @@ package com.netflix.spinnaker.orca.pipeline.model
 
 abstract class BuildInfo<A>(
   open val name: String?,
-  open val number: Int,
+  open val number: String?,
```
dbyron-sf, line 5: Why nullable when it wasn't before?

jaredstehler: i believe in this case the constant `-1` was being sent to indicate null value

dbyron-sf:
@mergifyio update

orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTaskSpec.groovy
jaredstehler: i could refactor to treat `"0"` as null/missing?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/BuildDetailExtractor.java
```diff
@@ -112,7 +112,9 @@ public boolean tryToExtractBuildDetails(BuildInfo<?> buildInfo, Map<String, Obje
       if (buildInfo == null || request == null) {
         return false;
       }
-      if (buildInfo.getUrl() != null && buildInfo.getName() != null && buildInfo.getNumber() > 0) {
```
dbyron-sf, line 115: Aaah, here's the place that was treating a number <= 0 as missing...

orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTaskSpec.groovy
dbyron-sf: I think it's fine to use null to indicate missing...but then [this comment](https://github.com/spinnaker/orca/pull/4316#discussion_r1005038701) says -1 used to mean missing....but from [here](https://github.com/spinnaker/orca/pull/4316/files#r1020612013) and specifically the `buildInfo.getNumber() > 0` part...it's not only -1, is everything <= 0 that's considered missing...so this is fine.  Thanks for reading :)

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/BuildInfo.kt
dbyron-sf: Found [this place](https://github.com/spinnaker/orca/pull/4316/files#r1020612013) that treats <= 0 as missing....`String?` works for me.

jaredstehler:
can this be merged yet?

PR #4317, opened 2022-10-21
fix(config): restore prior visibility of methods on CloudDriverConfigurationProperties class

jaredstehler:
when this class was refactored from groovy into java, the visibility of these two methods was (inadvertently?) changed, causing downstream plugins to break with illegal access.

https://github.com/spinnaker-plugins/aws-lambda-deployment-plugin-spinnaker/issues/125

### DIFF
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationProperties.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationProperties.java
@@ -65,7 +65,7 @@ public class CloudDriverConfigurationProperties {
   private BaseUrl kato;
   private CloudDriver clouddriver;
 
-  String getCloudDriverBaseUrl() {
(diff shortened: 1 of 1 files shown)


mattgogerly:
The change was unintentional, but looking at it, perhaps it should have been intentional.

The plugin should probably be using `KatoService` (e.g. https://github.com/spinnaker/orca/blob/5881377d6eac533eca049025a542c2b5f8f14bcc/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/KatoRestService.java#L35) rather than constructing the URL directly.

jaredstehler:
I can look into making that PR against the plugin, but still am of the mind that if we wished to make this a breaking change we should do so in a subsequent spinnaker release?

mattgogerly (approved):

mattgogerly:
@Mergifyio backport release-1.29.x

dbyron-sf:
Since https://github.com/spinnaker/orca/pull/4269 went into 1.29, I'm backporting this to 1.29.

dbyron-sf:

PR #4331, opened 2022-11-10
fix(config): add back public visibility for ClouddriverRetrofitBuilde…

srinath0301:
When this class was migrated from groovy to java the visibility of the class went from public to package-private, so adding back public access


srinath0301:
can this PR be backported to 1.28/1.29

dbyron-sf:
For the record, https://github.com/spinnaker/orca/pull/4269 has the change from groovy to java.  It seems totally reasonable to restore public visibility.  I am curious though, what code needs it?

srinath0301:
We had a plugin internally which uses that particular class. As the class went from public to package-private the plugin was failing to access that.

dbyron-sf (approved):

srinath0301:
@dbyron-sf can this PR be backported to 1.29?

dbyron-sf:

PR #4333, opened 2022-11-11
fix(webhook): Safer defaults and more config for webhook URLs

jcavanagh:
fix(webhook): Safer defaults and more config for webhook URLs

Exclude by default: single-word hostnames (`https://orca`, `https://spin-orca`), the `.spinnaker` domain (a common k8s deployment namespace), common internal-name suffixes (`.local`, `.internal`, `.localdomain`), and all verbatim IP addresses.

Add new configuration to specify a list of exclusion patterns.  This greatly simplifies configuration, as it is not easy to do complex filtering in a single allow expression.

Add new configuration to dynamically exclude domains based on the values of specified environment variables.  For example, this can always exclude the k8s namespace Spinnaker is currently running in, long as there is some variable set that specifies what that is.  `POD_NAMESPACE` is commonly set by providers, and is included by default along with `ISTIO_META_MESH_ID`, as names in that domain are also resolvable.

Also allows `localhost` in all cases if the `rejectLocalhost` flag is `false`, disregarding the name filter.  This avoids the need to change the name filter to include all forms of local names while developing.


orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -92,14 +155,16 @@ public UserConfiguredUrlRestrictions build() {
   private final boolean rejectLinkLocal;
   private final Set<String> rejectedIps;
   private final HttpClientProperties clientProperties;
+  private final List<Pattern> excludedPatterns;
 
-  public UserConfiguredUrlRestrictions(
+  protected UserConfiguredUrlRestrictions(
```
jcavanagh, line 160: The visibility of this constructor was reduced to mandate the use of the `Builder`

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -144,37 +237,42 @@ public URI validateURI(String url) throws IllegalArgumentException {
             "Allowed Hostnames are not set, external HTTP requests are not enabled. Please configure 'user-configured-url-restrictions.allowedHostnamesRegex' in your orca config.");
       }
 
-      if (!allowedHostnames.matcher(host).matches()) {
-        throw new IllegalArgumentException(
-            "Host not allowed " + host + ". Host much match " + allowedHostnames.toString() + ".");
```
jcavanagh, line 149: A couple of these messages were trimmed down, since there is no need to expose that configuration.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictionsSpec.groovy
```diff
@@ -5,16 +5,21 @@ import spock.lang.Unroll
 
 class UserConfiguredUrlRestrictionsSpec extends Specification {
 
+  // Don't try to actually resolve hosts, and control the result of determining whether a host is localhost or link local
+  def spyOn(UserConfiguredUrlRestrictions subject, isLocalhost = false, isLinkLocal = false) {
```
jcavanagh, line 9: Most of the builder arguments in these tests were to dodge code paths that attempted to resolve names, so that has been mocked out and the non-relevant builder arguments removed from the below tests.  Those code paths are now more exercised than before.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictionsSpec.groovy
```diff
@@ -30,22 +35,206 @@ class UserConfiguredUrlRestrictionsSpec extends Specification {
...
     then:
     thrown(IllegalArgumentException.class)
 
     where:
-    uri << ['https://www.test.com', 'https://201.152.178.212', 'https://www.test_underscore.com']
+    uri << [
+        'https://192.168.0.1',
+        'http://172.16.0.1',
+        'http://10.0.0.1',
+        'https://fd12:3456:789a:1::1',
```
jcavanagh, line 142: The ipv6 standard is pretty weird, since the `:` is rather overloaded - in any case, the code within can now exclude them in both forms.

jcavanagh (commented):
I don't expect any breakage from these default filters, since it would be pretty unusual to be using a name or address within these patterns.  

Feedback on additional default exclusions is welcome!

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -37,13 +42,31 @@
...
+            "10.0.0.0/8",
+            "172.16.0.0/12",
+            "192.168.0.0/16",
+            "fc00::/7",
+            "fd00::/8"); // can contain IP addresses and/or IP ranges (CIDR block)
+
+    // Blanket exclusion on certain domains
+    // This default pattern will exclude anything that is suffixed with the excluded domain
+    private String excludedDomainTemplate = "(?=.+\\.%s$).*\\..+";
+    private List<String> excludedDomains = List.of("spinnaker", "local", "internal");
```
jasonmcintosh, line 63: I could see adding... localdomain svc.local and some other common patterns, but a good start ;)

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -112,6 +177,29 @@ public UserConfiguredUrlRestrictions(
...
+  }
+
+  boolean isLocalhost(InetAddress addr) throws SocketException {
+    return addr.isLoopbackAddress()
+        || Optional.ofNullable(NetworkInterface.getByInetAddress(addr)).isPresent();
+  }
+
+  boolean isLinkLocal(InetAddress addr) {
+    return addr.isLinkLocalAddress();
+  }
```
jasonmcintosh, line 194: PERSONALLY not a fan of one line function wrappers like this... I'd rather have the link local direct, as it's a bit clearer to me...

orca-core/src/test/groovy/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictionsSpec.groovy
```diff
@@ -30,22 +35,206 @@ class UserConfiguredUrlRestrictionsSpec extends Specification {
...
+
+    then:
+    thrown(IllegalArgumentException.class)
+
+    where:
+    uri << [
+        'https://orca',
+        'https://clouddriver.svc.cluster.local',
+        'https://echo.internal',
+        'https://orca.spinnaker',
```
jasonmcintosh, line 68: Sould check on https://spin-orca/ as I believe that's a very common one of the patterns... just on principle (looks like the regex SHOULD catch it).

jasonmcintosh (approved):

jasonmcintosh:
Note assuming merge... we'll want to call out this change in release notes, as it'd cause any hook to an internal call in another namespace to likely break.  Can ALSO impact hooks to other private CIDRs like private VPCs on a TGW/peer... so LIKELY a breaking change for most.

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
jcavanagh: This allows the tests to mock this functionality more easily, since it doesn't need to capture calls to `InetAddress`, just itself

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
jcavanagh: `svc.local` is already covered by `local`.  I'm not familiar with anything that uses `localdomain`, though - which tooling typically uses that?

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
jasonmcintosh: ahh ok yeah missed that!

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
jasonmcintosh: Localdomain was an oldddd unix local reference. Still around I think PRETTY regularly. 
 https://bbs.archlinux.org/viewtopic.php?id=156064 for example.... swear I've seen it even in hosts recently

jcavanagh:
> as it'd cause any hook to an internal call in another namespace to likely break

Other k8s namespaces should work just fine, so long as they are not named `spinnaker`, `local` or `internal`.  

> Can ALSO impact hooks to other private CIDRs like private VPCs on a TGW/peer... so LIKELY a breaking change for most.

Unfortunately, specifying private IPs verbatim in the webhook bypasses domain filtering, and could allow resolution of internal Spinnaker services.  Alternatively, we could just disallow any literal IP address in the webhook configuration.

jasonmcintosh (changes requested):
SO after further looking - this breaks the IP restriction where if a "PASS" host lands in a "REJECT" Ip range, the request would fail.  This PR makes that no longer happen :(

jasonmcintosh:
@dbyron-sf or @kskewes-sf FYI on this one...

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
jcavanagh: I've added `localdomain` by default - doesn't seem to me like there is any downside.

orca-core/src/test/groovy/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictionsSpec.groovy
jcavanagh: I've added `spin-orca` to that test

jcavanagh:
@jasonmcintosh I've made the following changes to the implementation:

- Verbatim IP addresses are now rejected by default
- IP filtering is again considered against the resolved address
- No IP ranges are excluded by default

Trying to filter internal requests by IP is an unwinnable game of whack-a-mole.  I'll give some thought to how we might recharacterize user-initiated requests in general, so that they can be rejected without configuration.

dbyron-sf:
In general I support this.  I'm a little hesitant because I get the feeling it changes existing behavior.  Could you @jcavanagh perhaps write the release note for it so we can see how it affects users?  Are we releasing versions of spinnaker fast enough to do this in two stages -- one adds the knobs, but doesn't change behavior, and the second one changes the defaults?

orca-core/src/main/java/com/netflix/spinnaker/orca/config/UserConfiguredUrlRestrictions.java
```diff
@@ -37,14 +42,27 @@
...
 
+    // Blanket exclusion on certain domains
+    // This default pattern will exclude anything that is suffixed with the excluded domain
+    private String excludedDomainTemplate = "(?=.+\\.%s$).*\\..+";
+    private List<String> excludedDomains = List.of("spinnaker", "local", "localdomain", "internal");
+    // Generate exclusion patterns based on the values of environment variables
+    // Useful for dynamically excluding all webhook requests to the current k8s namespace, for
+    // example
+    private List<String> excludedDomainsFromEnvironment =
+        List.of("POD_NAMESPACE", "ISTIO_META_MESH_ID");
```
dbyron-sf, line 63: My reflex is to leave these defaults out...I guess they're no-ops for people outside of k8s / not using istio...but somebody somewhere is gonna have these set and we're gonna mess them up.  If they do stay in, I'd love links to docs about them in comments.

dbyron-sf:
The description implies this, but I want to go ahead and write it out -- the idea here is that someone making calls into spinnaker itself in a webhook stage is probably doing something unsavory, and so let's forbid it by default, right?

jcavanagh:
> The description implies this, but I want to go ahead and write it out -- the idea here is that someone making calls into spinnaker itself in a webhook stage is probably doing something unsavory, and so let's forbid it by default, right?

Yes, that's correct.  This is not a complete fix, but does enhance the ability of operators to mitigate the most probable vector (Webhook stages) specifically.  The main issues here are toning down the flexibility of k8s DNS, and disallowing exhaustive attacks within a generally-constrained IP space.  

> perhaps write the release note for it so we can see how it affects users? Are we releasing versions of spinnaker fast enough to do this in two stages -- one adds the knobs, but doesn't change behavior, and the second one changes the defaults?

I'm not opposed to that - we could move those env-var values to the release notes as recommendations instead.

Disallowing any IP address by default could certainly be a breaking change, but I don't think there's any real justification for such a configuration.  If an operator wishes, the option can be disabled.

jasonmcintosh:
ORIGINALLY the plan was "ok let's break this and make it a DENY ALL by default, then REQUIRE end users to whitelist any supported webhook targets".  AKA you'd want to manually add either a list or a regex.  BUT make it secure default vs. not.  

The current implemetnation allows calls to even internal API's which it really really shouldn't as this lets you bypass some restrictions.  Link-local blocks help a TON (aka blocking ec2 metadata endpoints), but... ideally should block cluster addresses as well for these, and only allow cluster-external calls.

jcavanagh:
> ORIGINALLY the plan was "ok let's break this and make it a DENY ALL by default, then REQUIRE end users to whitelist any supported webhook targets". AKA you'd want to manually add either a list or a regex. BUT make it secure default vs. not.

When is "originally"?  This change was never a deny-all implementation, nor is that something that I think operators would want.  I might be missing some context here.

> The current implemetnation allows calls to even internal API's which it really really shouldn't as this lets you bypass some restrictions. Link-local blocks help a TON (aka blocking ec2 metadata endpoints), but... ideally should block cluster addresses as well for these, and only allow cluster-external calls.

I think you may have missed the most recent changes - see my other comment [here](https://github.com/spinnaker/orca/pull/4333#issuecomment-1312237642).

dbyron-sf:
@mergifyio update

jasonmcintosh (approved):
SO conversing with the various teams, I don't see `rejectVerbatimIps` being remotely useful since you can use various domains to get any IP you want.  E.g. http://1u.ms/ kinda stuff.  That said, will go ahead let this go through to get at least a little better.

dbyron-sf:
@mergifyio update

PR #4335, opened 2022-11-15
Update createBakeTask with manage image.

DanielaS12:
Add support for handling the bake of managed and custom images for Azure.

Tested the feature manually. 

To be pushed after the rosco packer templates have been validated and the entire workflow was tested.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
@@ -153,7 +153,7 @@ class CreateBakeTask implements RetryableTask {
   private BakeRequest bakeFromContext(StageExecution stage, SelectedService<BakeryService> bakery) {
     PackageType packageType
     packageType = stage.context.baseOs != null ? getBaseOsPackageType(bakery, stage) : getCustomPackageType(stage)
-
+    stage.context.packageType = packageType
     List<Artifact> artifacts = artifactUtils.getAllArtifacts(stage.getExecution())
 
     PackageInfo packageInfo = new PackageInfo(stage,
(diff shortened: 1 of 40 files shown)


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -188,10 +182,40 @@ class CreateBakeTask implements RetryableTask {
       requestMap.remove("baseName")
     }
 
+    if (stage.context.baseOs == null) {
```
ovidiupopa07, line 185: can you please also add unit tests for the new functionality?

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -188,10 +182,40 @@ class CreateBakeTask implements RetryableTask {
...
       request.other().clear()
     }
+
     return request
   }
+
+  private static PackageType getCustomPackageType(StageExecution stage) {
+    PackageType type
+    if(stage.context.osType != "linux") {
+      type = PackageType.NUPKG
```
ovidiupopa07, line 205: I would return instead the type. by doing this you don't need the `else` clause. you can directly do 
```
type = stage.context.packageType as PackageType
      if (type == null) {
        type = new OperatingSystem(stage.context.managedImage as String).getPackageType()
      }
```
without the else clause. For readability but it's just a personal preference.

ovidiupopa07 (changes requested):
looks good but I would add unit tests for the new functionality. there are quite a few new if clauses added so I would test that the bake request it is what you expect

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
DanielaS12: great catch, will do the changes

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
DanielaS12: sure

ovidiupopa07 (approved):

DanielaS12:
Closing in favor of https://github.com/spinnaker/orca/pull/4336

PR #4348, opened 2022-11-23
fix: check manifests types

boonware:
# Issue
https://github.com/spinnaker/spinnaker/issues/6769

# Description
Allow a manifest file to contain more than one embedded manifest when running a Job. This supports use cases such as a Job manifest bundled with a ConfigMap manifest in the same file. Strictly speaking, it is a manifest file containing more than one _Job_ definition that is the problem. This is solved by checking the `kind` of each manifest in the file, and only throwing an exception if more than one `Job` kind is found.


boonware:
@dbyron-sf Could I get a workflow approval, please?

dbyron-sf:
No need to tag me on PRs.  I see them fly by.

I'm not sure I agree that this is a fix though.  Seems more like a feature/change in behavior, and one that could break folks who depend on this validation.  I think I'd prefer that additional work go into filtering so it's possible to leave run job stages alone and deploy the additional manifests (presumably without the job) in a deploy manifest stage.

PR #4353, opened 2022-11-25
feat(AWS): Get the rollback timeout value from stage data.

DanielaS12:
Get the rollback timeout value from stage data. The value is now configurable from the UI. If the value is not set in the UI the hardcoded value is used.

Testing:
The rollback failed after 20 min and 8 sec with a timeout of 20min.

image

### DIFF
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
@@ -107,8 +107,8 @@ public class DetermineRollbackCandidatesTask implements CloudProviderAware, Retr
   }
 
   public long getDynamicTimeout(StageExecution stage) {
-    if (stage.getContext().containsKey("timeout")) {
-      return TimeUnit.MINUTES.toMillis((int) stage.getContext().get("timeout"));
+    if (stage.getContext().containsKey("rollbackTimeout")) {
+      return TimeUnit.MINUTES.toMillis((int) stage.getContext().get("rollbackTimeout"));
     }
 
     return getTimeout();
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
@@ -73,6 +73,8 @@ import org.springframework.stereotype.Component;
  */
 @Component
 public class DetermineRollbackCandidatesTask implements CloudProviderAware, RetryableTask {
+  private Integer timeout;
+
   private static final Logger logger =
       LoggerFactory.getLogger(DetermineRollbackCandidatesTask.class);
 
@@ -103,13 +105,14 @@ public class DetermineRollbackCandidatesTask implements CloudProviderAware, Retr
 
   @Override
   public long getTimeout() {
-    return TimeUnit.MINUTES.toMillis(5);
+    return TimeUnit.MINUTES.toMillis(timeout == null ? 5 : timeout);
   }
 
   @Nonnull
   @Override
   public TaskResult execute(@Nonnull StageExecution stage) {
     StageData stageData = stage.mapTo(StageData.class);
+    timeout = stageData.timeout;
     Moniker moniker =
         populateMonikerWithServerGroupInfo(
             stageData.moniker,
@@ -527,6 +530,7 @@ public class DetermineRollbackCandidatesTask implements CloudProviderAware, Retr
     public String credentials;
     public String cloudProvider;
     public String serverGroup;
+    public Integer timeout;
 
     public Integer targetHealthyRollbackPercentage;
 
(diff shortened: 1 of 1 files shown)


orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -103,13 +105,14 @@ public long getBackoffPeriod() {
 
   @Override
   public long getTimeout() {
-    return TimeUnit.MINUTES.toMillis(5);
+    return TimeUnit.MINUTES.toMillis(timeout == null ? 5 : timeout);
```
ovidiupopa07, line 108: I would return 0 and override `getDynamicTimeout` instead. That one accepts a `StageExecution` as the input param. 

This way, you do not need to define the `private Integer timeout` at the class level.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -103,13 +105,14 @@ public long getBackoffPeriod() {
...
   }
 
   @Nonnull
   @Override
   public TaskResult execute(@Nonnull StageExecution stage) {
     StageData stageData = stage.mapTo(StageData.class);
+    timeout = stageData.timeout;
```
ovidiupopa07, line 115: can be removed after you override `getDynamicTimeout`

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -527,6 +530,7 @@ private static class StageData {
     public String credentials;
     public String cloudProvider;
     public String serverGroup;
+    public Integer timeout;
```
ovidiupopa07, line 533: no need to define timeout here if you override `getDynamicTimeout`

ovidiupopa07 (changes requested):

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
DanielaS12: addressed

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
DanielaS12: addressed

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
DanielaS12: addressed

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
```diff
@@ -106,6 +106,14 @@ public long getTimeout() {
     return TimeUnit.MINUTES.toMillis(5);
   }
 
+  public long getDynamicTimeout(StageExecution stage) {
+    if (stage.getContext().containsKey("timeout")) {
```
ovidiupopa07, line 110: it would be nice to reuse (if possible) `public static final String STAGE_TIMEOUT_OVERRIDE_KEY = "stageTimeoutMs";` which already read in the `StageExecution`  - code snipped below from `StageExecutionImpl`

```

public Optional<Long> getTimeout() {
  Object timeout = getContext().get(STAGE_TIMEOUT_OVERRIDE_KEY);
  if (timeout instanceof Integer) {
    return Optional.of((Integer) timeout).map(Integer::longValue);
  } else if (timeout instanceof Long) {
    return Optional.of((Long) timeout);
  } else if (timeout instanceof Double) {
    return Optional.of((Double) timeout).map(Double::longValue);
  }
  return Optional.empty();
}
```

ovidiupopa07 (changes requested):

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cluster/DetermineRollbackCandidatesTask.java
DanielaS12: I tested using this value, but it is used in other places too and I don't think it should be update from the ui. As per Slack chat, I changed 'timeout' -> 'rollbackTimeout' for better understanding.

ovidiupopa07 (approved):
LGTM

dogonthehorizon (approved):

PR #4356, opened 2022-12-01
fix(waiting-executions) : Waiting executions doesn't follow FIFO

rjalander:
This PR has changes to fix the issue https://github.com/spinnaker/spinnaker/issues/6373

** Root Cause: **
Currently `StartWaitingExecutions` is getting pushed to the queue every time it receives `CompleteExecution` for all the execution status. 
But this is causing an issue when pipeline execution with branches is running with Disabled concurrent executions and trying to `StartWaitingExecutions`, which leads to remove the oldest execution from the waiting queue to start and re-insert the execution at the top of the queue as the current execution is still running.

** Fix: **
Changes to push `StartWaitingExecutions` to the queue only when execution status is not running and disabled concurrent executions, in all other cases no need to push `StartWaitingExecutions` to the queue.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
@@ -84,11 +84,21 @@ class CompleteExecutionHandler(
           }
         }
       }
-      execution.pipelineConfigId?.let {
-        queue.push(StartWaitingExecutions(it, purgeQueue = !execution.isKeepWaitingPipelines))
+      log.info("Execution ${execution.id} is with ${execution.status} status and  Disabled concurrent executions is ${execution.isLimitConcurrent}")
+        if (execution.isLimitConcurrent){
+          if (execution.status != RUNNING) {
+            execution.pipelineConfigId?.let {
+              queue.push(StartWaitingExecutions(it, purgeQueue = !execution.isKeepWaitingPipelines))
+            }
+          }else {
+            log.info("Not starting waiting executions as execution ${execution.id} is currently RUNNING with Disabled concurrent executions.")
+          }
+        } else{
+            log.info("Execution ${execution.id} is not Disabled for concurrent executions, no need to run waiting executions")
+        }
       }
     }
-  }
+
 
   private fun CompleteExecution.determineFinalStatus(
     execution: PipelineExecution,


mattgogerly:
Hi @rjalander, could you change your commit messages to adhere to conventional commits?

rjalander:
Hi @mattgogerly, thank you for reviewing this. 
I tried to squash multiple commits into one with the commit message. Not sure whether it solved the problem of commit conventions.

rjalander:
Hi @mattgogerly, 
Please share If you have any other comments to be addressed to get this PR approved.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
```diff
@@ -84,11 +84,21 @@ class CompleteExecutionHandler(
...
       }
-      execution.pipelineConfigId?.let {
-        queue.push(StartWaitingExecutions(it, purgeQueue = !execution.isKeepWaitingPipelines))
+      log.info("Execution ${execution.id} is with ${execution.status} status and  Disabled concurrent executions is ${execution.isLimitConcurrent}")
+        if (execution.isLimitConcurrent){
+          if (execution.status != RUNNING) {
+            execution.pipelineConfigId?.let {
+              queue.push(StartWaitingExecutions(it, purgeQueue = !execution.isKeepWaitingPipelines))
+            }
+          }else {
```
mattgogerly, line 93: nit: Can you fix the formatting here? spaces between { } and on L96 too

mattgogerly:
Sorry, I started a review and then apparently didn't click add comments.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
```diff
@@ -84,11 +84,21 @@ class CompleteExecutionHandler(
           }
         }
       }
-      execution.pipelineConfigId?.let {
-        queue.push(StartWaitingExecutions(it, purgeQueue = !execution.isKeepWaitingPipelines))
+      log.info("Execution ${execution.id} is with ${execution.status} status and  Disabled concurrent executions is ${execution.isLimitConcurrent}")
```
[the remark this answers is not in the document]
rjalander: yeah debug suits in that case, changed to debug.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
rjalander: fixed the formatting

jimmycasey:
Thanks @mattgogerly for your help reviewing this. It has been effecting our development velocity so will be great for us (@ericsson) once merged & released. 👍

jimmycasey:
Hi @mattgogerly,
Any further updates required here or is it good to merge?
Thanks
Jimmy

jasonmcintosh (commented):
SO I think the one thing I see missing is an order verification test.  E.g. we're changing the order logic but not testing that change in logic.  I like the fix - just would like a test for it ;)

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
```diff
@@ -84,11 +84,21 @@ class CompleteExecutionHandler(
...
+        if (execution.isLimitConcurrent) {
+          if (execution.status != RUNNING) {
+            execution.pipelineConfigId?.let {
+              queue.push(StartWaitingExecutions(it, purgeQueue = !execution.isKeepWaitingPipelines))
+            }
+          } else {
+            log.info("Not starting waiting executions as execution ${execution.id} is currently RUNNING with Disabled concurrent executions.")
+          }
+        } else {
+            log.info("Execution ${execution.id} is not Disabled for concurrent executions, no need to run waiting executions")
```
mattgogerly, line 97: Sorry for the slow updates here - I've been trying to find some time to sit down and read through the context of this change. I'm having a really hard time understanding how this fixes FIFO. 

1. If the pipeline is limiting concurrent executions and this execution isn't RUNNING, push a message to `StartWaitingExecutions` for this pipeline ID.
2. If this execution is RUNNING then don't do anything
3. If this pipeline isn't limiting concurrent executions then don't do anything

I guess my question is why would we have a `CompleteExecution` message at all if the execution is still `RUNNING`?

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
```diff
@@ -84,11 +84,21 @@ class CompleteExecutionHandler(
...
-      execution.pipelineConfigId?.let {
-        queue.push(StartWaitingExecutions(it, purgeQueue = !execution.isKeepWaitingPipelines))
+      log.debug("Execution ${execution.id} is with ${execution.status} status and  Disabled concurrent executions is ${execution.isLimitConcurrent}")
+        if (execution.isLimitConcurrent) {
+          if (execution.status != RUNNING) {
+            execution.pipelineConfigId?.let {
+              queue.push(StartWaitingExecutions(it, purgeQueue = !execution.isKeepWaitingPipelines))
+            }
+          } else {
+            log.info("Not starting waiting executions as execution ${execution.id} is currently RUNNING with Disabled concurrent executions.")
```
mattgogerly, line 94: Same comment as above, this should probably be debug. Same for the log below.

jimmycasey:
Thanks @mattgogerly & @jasonmcintosh for comments, I've reached out to Jalander to have a look.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/CompleteExecutionHandler.kt
rjalander: for case 2, even the entire pipeline execution is running and If the pipeline has multiple branches, as per the current behavior  the `CompleteExecution` message is sent on every pipeline stage completion.
Here is the problem once it receives the `CompleteExecution`, it assumes that the execution is complete and start the next waiting execution from the queue.

**Not sure about other impacts If we really stop sending `CompleteExecution` message at all after every stage with branch completion.**

for 3, yes If concurrent execution is enabled, we don't need to run `StartWaitingExecutions`, because it will start immediately and there is no pipelines waiting in queue.

mattgogerly: Does the first part of the if statement need to push the message too?

mattgogerly:
Ok, I think I'm happy with this as a fix. @rjalander could you fix the indentation of the code below the first log call?

And +1 to Jason's comment, if we can get a test that'd be fantastic.

rjalander:
Having some issues running tests for this specific scenario with order verification, as in this case multiple handlers involved for this order to reflect. Trying to add this scenario in `CompleteExecutionHandlerTest` but couldn't success. Any pointer where exactly this kind of test can fit in.

mattgogerly (approved):

jasonmcintosh (approved):

jasonmcintosh:
@mergify backport release-1.28.x release-1.29.x  release-1.27.x

jasonmcintosh:
I've triggered the backports.  BUT will discuss if we actually want to merge the backports.  @dbyron-sf Not sure... this almost isn't a "fix".

jasonmcintosh:
@rjalander so backports, let's skip on for now and discuss more.  Can we get release notes on this change updated please?  This is a nice call out we should make ;) https://github.com/spinnaker/spinnaker.io/pull/306 for an example.

jimmycasey:
Hi @mattgogerly @jasonmcintosh, would you have a rough idea of when 1.30 would be released? Thanks.

JamieChe:
Hi @rjalander , we reviewed your code change, found it would cause the waiting queue blocked because of the feature **Add support for max concurrent pipeline executions** https://github.com/spinnaker/orca/pull/4193 released in 1.28.0. 
When the **Maximum concurrent pipeline executions** is set >0, the executions in waiting queue will never get started. 
Please check from you side.

rjalander:
> Hi @rjalander , we reviewed your code change, found it would cause the waiting queue blocked because of the feature **Add support for max concurrent pipeline executions** #4193 released in 1.28.0. When the **Maximum concurrent pipeline executions** is set >0, the executions in waiting queue will never get started. Please check from you side.

I looked at the code for this issue, as part of #4356 fix waiting executions are running only for the pipelines for which  concurrent executions is **Disabled**.

With the feature added **Add support for max concurrent pipeline executions** #4193, the same issue exist for this scenario too, as executions doesn't follow FIFO order. 

So, the same fix #4356 is applicable for the pipelines with concurrent execution is Enabled also.

I will make the necessary change to add this fix for both of the scenarios to start waiting executions in FIFO order.

rjalander:
@mattgogerly @jasonmcintosh, Opened a new PR to fix the issue https://github.com/spinnaker/orca/pull/4415
Can you please review the same.

PR #4347, opened 2022-11-23
chore(front50): Make Monitor Pipeline Task timeout overridable

sergio-quintero:
Closes: https://github.com/spinnaker/spinnaker/issues/6739

PR #4367, opened 2022-12-16
feat(bakery): add tasks.monitor-bake.timeout-millis configuration property

dbyron-sf:
for MonitorBakeTask timeout

The default value is unchanged -- 3600000 milliseconds, or 1 hour.

See https://github.com/spinnaker/orca/pull/4347 which does a similar thing for MonitorPipelineTask.


dbyron-sf:
And yes, I'm not a fan of using `@Value`, but I don't have the stamina to try to organize these properties into a `@ConfigurationProperties` at the moment.

ovidiupopa07 (approved):

PR #4368, opened 2022-12-19
fix(stageExecution): Extend MJ auth propagate logic for exhaustive cases

OscarMichelH:
When Manual Judgment without propagate authentication enabled appear in a pipeline, next stages ignore the last user interaction in propagated stages, and is retrieving the executor user by default when the expected case is consider the last MJ with propagate authentication enabled.

Similar fix was applied in https://github.com/spinnaker/orca/pull/4289 but only considering skipped stages, now we extend the logic for sub cases

Requirements:
-AuthZ system enabled
-Roles properly configured to restrict and allow desired users

Ex:
Roles used:

-----------
    permissions:
      CREATE:
      - "admin" # group members that CAN create applications if enabled
      READ:
      - "admin"
      - "restricted"
      WRITE:
      - "admin"
      EXECUTE:
      - "admin"
      - "restricted"
Users:
OscarMichelH as `admin`
oscar4armory as `restricted`

The wrong behaviour is the next:
image
image

The restricted user executed the pipeline, but before to deploy in the first stage we have an approval with authentication propagate enabled and the last stage, the MJ was judged by the authorized user, and deployment stage looking by the Authenticated user, is just ignoring the MJ and retrieving the executor user by default, and the expected behaviour is take in consideration the previous MJ with propagate authentication.

With the fix applied the last stage is taking the previous judge from a propagated auth MJ stage even if the propagate authentication is disabled, and acting as expected
image
image

### DIFF
--- a/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
+++ b/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/pipeline/ManualJudgmentStage.groovy
@@ -56,7 +56,7 @@ class ManualJudgmentStage implements StageDefinitionBuilder, AuthenticatedStage 
   @Override
   Optional<PipelineExecution.AuthenticationDetails> authenticatedUser(StageExecution stage) {
     def stageData = stage.mapTo(StageData)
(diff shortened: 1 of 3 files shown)


ovidiupopa07 (approved):

kkotula (approved):
👏🏻 LGTM!

dogonthehorizon:

dogonthehorizon (approved):

PR #4372, opened 2022-12-27
fix(sql): No offset for retrievePipelinesForPipelineConfigId and retr…

xiazuojie:
Currently, param page is ignored by retrieve and retrievePipelinesForPipelineConfigId methods (although they are used by other methods), resulting in always returning the content of the first page.

This PR add page support for these 2 methods.


dbyron-sf:
What's involved in adding automated tests for this?

PR #4397, opened 2023-02-17
fix(artifacts): Be more lenient when filtering expected artifacts

jervi:
It is currently hard/impossible to resolve artifacts in pipelines that doesn't have any triggers defined by themselves, and are triggered by a pipeline stage in another pipeline. Previously it was possible to get it to work by editing the json of the pipeline or in the UI by adding artifacts in a temp trigger and then removing it (this would keep the expected artifacts around). When I introduced trigger specific artifact constraints in #4322, I made it a lot harder (if not impossible) to do this because no triggers are used and thus all expected artifacts are filtered out.

This PR will fix this by copying expected artifacts from the parent execution (only if triggered by a pipeline stage), thus avoiding the whole situation altogether. It will also not filter away any expected artifacts where `useDefaultArtifact` or `usePriorArtifact` is set to `true`, and it will bind artifacts defined inline in stages (so that you can match artifacts using regex and not only SpEL).

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
@@ -28,6 +28,7 @@ import com.google.common.collect.ImmutableSet;
 import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
 import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact;
+import com.netflix.spinnaker.kork.web.exceptions.InvalidRequestException;
 import com.netflix.spinnaker.orca.api.pipeline.models.PipelineExecution;
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
 import com.netflix.spinnaker.orca.pipeline.model.StageContext;
@@ -35,13 +36,20 @@ import com.netflix.spinnaker.orca.pipeline.model.StageExecutionImpl;
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository;
 import com.netflix.spinnaker.orca.pipeline.persistence.ExecutionRepository.ExecutionCriteria;
 import java.io.IOException;
-import java.util.*;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


jervi:
@dbyron-sf Do you think you can have a look at this? A little follow up on #4322 👍

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
```diff
@@ -139,7 +141,30 @@ private List<Artifact> getAllArtifacts(
...
+    }
+    try {
+      ExpectedArtifact expectedArtifact =
+          ExpectedArtifact.builder().matchArtifact(artifact).build();
+      return ArtifactResolver.getInstance(execution.getTrigger().getArtifacts(), true)
+          .resolveExpectedArtifacts(List.of(expectedArtifact))
+          .getResolvedArtifacts()
+          .stream()
+          .findFirst();
+    } catch (InvalidRequestException e) {
```
dbyron-sf, line 164: What is throwing InvalidRequestException that we might catch here?

dbyron-sf:
I could use an example of a pipeline + trigger that doesn't work before this PR, but does work after.  Especially if it's also one that worked before https://github.com/spinnaker/orca/pull/4322.  And then, is there a place to put that pipeline in a test?  Maybe the first commit in this PR could be adding that test?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
jervi: [`ArtifactResolver.resolveExpectedArtifacts()`](https://github.com/spinnaker/orca/blob/a4a82084a48cfc323e3b0b04b670eb17125a3ece/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java#L127) throws this if it's unable to resolve the artifact. It's not something we care about, as most often it will probably not be matched.

jervi:
I have rebased and split the commits so that I add the tests in the first commit and the implementation code in the second. 

> I could use an example of a pipeline + trigger

The whole point is to fix pipelines _without_ any defined triggers, that are triggered by some other pipeline (using a pipeline stage). They will get an implicit trigger when being executed, but the pipeline config is (optionally) defined without any triggers at all.

jervi:
To elaborate - You have a regular pipeline with a trigger and some expected artifacts. This pipeline has a bunch of pipeline stages that starts different pipelines in order. In one of these child pipelines (that doesn't have any triggers defined), you have a deploy manifest stage, with an inline artifact defined.
image
Previously, this would not work - it would try to download the file `paas.*.yml` literally.
With this PR the example in the image will work. It will compare the inline artifact with artifacts from the parent pipeline and bind it successfully, without needing to resort to SPeL expressions.

dbyron-sf:
Thanks for the detail, and splitting the commits (nit, the first commit is now `test(artifacts)` not `fix(artifacts)`).  Would you mind confirming one more thing -- whether the pipeline you describe [here](https://github.com/spinnaker/orca/pull/4397#issuecomment-1438392175) worked before https://github.com/spinnaker/orca/pull/4322?

jervi:
Not directly - defined inline it would not have worked earlier. But what you could do before was to define the expected artifact(s) on the child pipeline (as a workaround), and then use that artifact instead of defining it inline. But that is kind of hackish - you need to add a trigger to be able to define the artifact, and then you can remove the trigger again. After #4322 any expected artifacts defined using this workaround would be ignored, since Orca now only considers artifacts on the trigger actually used in the execution.
The workaround made more sense with the legacy/old artifact support - then expected artifacts weren't connected to a trigger.

dbyron-sf:
Thanks much @jervi .  Think any of this warrants a release note?

dbyron-sf (approved):

jervi:
@dbyron-sf I made a PR for some release notes: https://github.com/spinnaker/spinnaker.io/pull/308

PR #4400, opened 2023-02-24
fix(tasks): removing context from TaskResult in WaitForManualJudgment…

kkotula:
This PR fixes an issue of ManualJudgement stage data overwriting, that happens when two threads concurrently process:
1. PATCH request to update the stage in CompoundExecutionOperator,
2. RunTask event by RunTaskHandler,

The diagram below explains what happens
image

If we don't pass the context, while building the ManualJudgement TaskResult, RunTaskHandler will not make any update calls to the database, hence it will not overwrite the data set in PATCH request.


ovidiupopa07 (approved):
LGTM 🚀 I wonder if this has some implications we cannot foresee though

PR #4403, opened 2023-03-01
fix(manual judgement): Fixed concurrent stage updates

kkotula:
In this PR I am trying to solve a problem of concurrent alterations to Manual Judgement stage.
The diagram below presents exactly the behaviour that results in data inconsistency.
bug diagram

The solution depends on distributed locking mechanism that prevents multiple threads from updating the same resource.
fix CompoundOperatorExecution acquires lock


mattgogerly:
I'll try and have a look at the code tomorrow; is there a reason this PR is targeting release-1.27 and not master?

PR #4404, opened 2023-03-02
fix(artifacts): Stop copying expectedArtifactIds to child pipelines

jervi:
This commit partially reverts #4397. It broke some pipelines for a team that had a CI stage that created more artifacts with the same name as the expected artifact from the original trigger, and then had a pipeline stage. The result was that the child pipeline was triggered with multiple artifacts with the same name (but different version numbers), and then failed to start because the artifact resolver matched multiple artifacts instead of exactly one. Turns out the changes in DependentPipelineStarter wasn't really needed to fix the issue that #4397 tried to solve, so I'm reverting them.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
@@ -84,11 +84,6 @@ class DependentPipelineStarter implements ApplicationContextAware {
     } collectMany {
       it.expectedArtifactIds ?: []
     }
-    if (!expectedArtifactIds && parentPipelineStageId) {
-      expectedArtifactIds = parentPipeline.trigger.resolvedExpectedArtifacts.collect {
-        it.id
-      }
-    }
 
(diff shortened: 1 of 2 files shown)


jervi:
@dbyron-sf This is getting ridiculous, but I have another fix I need you to look at! Sorry 😅

dbyron-sf:
This does make me wonder if there's a different kind of test that we could use to verify this behavior.  I haven't looked to see if anything like this exists...do you know?  Somewhere to define a pipeline that executes another pipeline, or something?

dbyron-sf (approved):

jervi:
> This does make me wonder if there's a different kind of test that we could use to verify this behavior. I haven't looked to see if anything like this exists...do you know? Somewhere to define a pipeline that executes another pipeline, or something?

`DependentPipelineStarterSpec` more or less tests this. The issue was more a use case I didn't imagine. When I changed the test [here](https://github.com/spinnaker/orca/pull/4404/files#diff-5756f1a669ee61f4178c9484c756d5dbff915689a89b6f1dde90f34c45a42831R527) to add a stage that produced an artifact, I was able to reproduce it.
But you are maybe thinking of some integration tests or something that tests this?

PR #4410, opened 2023-03-09
fix(judgements): Fix/manual judgment concurrent execution

kkotula:
This PR aims to address the issue of concurrent changes to execution stages. We have discovered a rather intriguing bug that arises from concurrent read/writes of stage data to storage. This bug is not limited to a particular storage solution, as it occurs with both Redis and SQL alternatives. Therefore, the solution must be implemented on the application code level.

Here's a brief summary of what's happening. Both CompoundExecutionOperator and RunTaskHandler read and write stage data to storage. However, if one thread executes the logic of CompoundExecutionOperator while another thread executes RunTaskHandler, there may be data inconsistency. This inconsistency can result in unexpected behavior.

To address this issue, we need to ensure that the application code is equipped to handle concurrent changes to execution stages. One possible solution is to implement locking mechanisms that prevent multiple threads from accessing the same stage data at the same time. This will ensure that data consistency is maintained, and the application runs as expected.

Below is a diagram that illustrates the problem.
bug diagram

The approach I've taken in order to ensure the reliability of the distributed locking pattern is to leverage externalised storage. By doing so, we can ensure that the locking mechanism is properly implemented and that the system is able to handle multiple requests at once without failing. In this regard, we have implemented two different solutions that are both effective in their own right.

The first solution is the Redis locking mechanism. This solution utilizes the LockManager interface and implementation that is already present in the kork library.  The second solution is the SQL-based approach that is based on the spring shedlock. This solution is also effective in ensuring the reliability of the distributed locking pattern.
Below is a diagram that illustrates the solution.
Distribiuted locking

Locking is disabled by default. It can be enabled by setting flags redis.external-lock.enabled=true or sql.external-lock.enabled=true. If none of the flags is set, application will fallback to default NoOp locking that doesn’t do any locking and executes Runnable/Callable instances.


armory-abedonik:
Hi Chris, I would like to draw your attention to the fact that both MySQL and PostgreSQL offer support for transactions. As such, it is worth considering whether these options may be more advantageous than utilizing the use of a shared lock.

ovidiupopa07 (approved):
lgtm 🚀

orca-core/orca-core.gradle
```diff
@@ -54,6 +54,8 @@ dependencies {
   implementation("com.jayway.jsonpath:json-path:2.2.0")
   implementation("org.yaml:snakeyaml")
   implementation("org.codehaus.groovy:groovy")
+  implementation("net.javacrumbs.shedlock:shedlock-spring:2.2.0")
```
dbyron-sf, line 57: Could you please provide an overview of this dependency, and why it's a good choice, other alternatives, etc.

orca-core/src/main/java/com/netflix/spinnaker/orca/lock/ExternalLock.kt
```diff
@@ -0,0 +1,173 @@
...
+import net.javacrumbs.shedlock.core.LockProvider
+import net.javacrumbs.shedlock.core.SimpleLock
+import org.slf4j.LoggerFactory
+import java.time.Duration
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import java.util.*
+import java.util.concurrent.Callable
+
+interface RunOnLockAcquired {
```
dbyron-sf, line 31: javadoc please

mrampton (approved):

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -20,32 +20,18 @@ import com.netflix.spectator.api.BasicTag
...
-import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus.RUNNING
-import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus.SKIPPED
-import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus.STOPPED
-import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus.SUCCEEDED
-import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus.TERMINAL
-import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType
-import com.netflix.spinnaker.orca.api.pipeline.models.PipelineExecution
-import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
-import com.netflix.spinnaker.orca.api.pipeline.models.TaskExecution
+import com.netflix.spinnaker.orca.api.pipeline.*
```
jasonmcintosh, line 24: Wildcard vs. not?

jasonmcintosh: (nitpick not too worried personally)

jasonmcintosh:
SO biggest long term issue for me is ... right now there's a couple different lock manager type implementations crossing kork & igor.  They're not great implementations today however.  This sounds like a good start (assuming answers to the lock library), eventually extract to kork, and then get implemented consistently across the services.  The challenge is of course are a few of the libs which do NOT use SQL (and should... looking at you Igor).  

ONLY other concern, and I can see why - is Primary beans cause exclusive locks preventing replacement by alternative implementations.  But I'm ok with it for now.

Ignoring THAT... this has some GREAT details, GREAT Explanation of what's going on/why/where.  THANK YOU!

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
dbyron-sf: Yes please to no star imports.

orca-core/orca-core.gradle
jasonmcintosh: The alternative I know for distributed services locks:
* Raft consensus protocols
* jgroups
* Quartz (which is ... full CRON scheduler)
* This lib :) which is pure SQL it seems

I don't know that we really want raft (aka https://ratis.apache.org/) or jgroups (which.. keycloak uses.. and is kinda a geast).  Quartz is a bit overkill.  I'd be open to another distributed lock lib - ideally though this is one that replaces any OTHER distributed lock needs LIKE Igor's current system for polling locks.

orca-core/src/main/java/com/netflix/spinnaker/orca/lock/ExternalLock.kt
kkotula: Done!

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
kkotula: Fixed! Thank you guys!

kkotula:
> SO biggest long term issue for me is ... right now there's a couple different lock manager type implementations crossing kork & igor. They're not great implementations today however. This sounds like a good start (assuming answers to the lock library), eventually extract to kork, and then get implemented consistently across the services. The challenge is of course are a few of the libs which do NOT use SQL (and should... looking at you Igor).
> 
> ONLY other concern, and I can see why - is Primary beans cause exclusive locks preventing replacement by alternative implementations. But I'm ok with it for now.
> 
> Ignoring THAT... this has some GREAT details, GREAT Explanation of what's going on/why/where. THANK YOU!

Hi @jasonmcintosh ,

I totally understand where you're coming from. It would definitely be better to have fewer locking mechanisms and store them in the kork library. However, while working on this feature, I noticed that some of the locking options aren't compatible with each other, and it's kind of tricky to merge them all into one solution.

For example, Redis uses keyName to create a lock in the cache, but this library [sql library](https://github.com/spinnaker/kork/blob/ec9ab5f31b7e479a1f48496b7e91f55d9e632492/kork-sql/kork-sql.gradle#L15) creates a lock per session. They aren't compatible with each other and they accomplish different goals.

orca-core/orca-core.gradle
kkotula: Hey there!

@jasonmcintosh 
Thanks for sharing alternative options! All of them would get the job done!

I was looking for a simple library that focuses on providing a locking mechanism without any additional scheduling features, which I didn't need. After some research, I found a library that perfectly fits my needs: it is lightweight and easy to use, and it provides a reliable locking mechanism that can be integrated with Redis (maybe it can replace LockingManager?) and other storage options. This library has the potential to become the foundation of a new implementation that could replace the current messy locking situation. In the future, by centralizing the locking options within kork, we can achieve better code organization and reduce the risk of errors. Currently, the locking options are spread across the kork and apps repositories, which makes it hard to manage and maintain them. By using this library, we can simplify the locking process.

If you have any questions or concerns, please don't hesitate to reach out. I am happy to help or explain more!

jasonmcintosh: ONLY other question is the 2.2.0 version.  Which I'd expect to match spring maybe?  Which seems like it's not quite matching?  At that point, it might need a kork change :(

kkotula: Bumped the version to 4.44. Cannot bump it to 5.2, because it requires Java 17.

dbyron-sf:
How far back to backport this?  At the moment the oldest active branch is release-1.29.x.

dbyron-sf:
Still curious how far to backport this.   But also, I can't quite see what this has to do with manual judgment.

ovidiupopa07:
@Mergifyio backport release-1.29.x release-1.28.x

kkotula:
> Still curious how far to backport this. But also, I can't quite see what this has to do with manual judgment.

Hey @dbyron-sf,

@ovidiupopa07 already opened backport PRs:
1. https://github.com/spinnaker/orca/pull/4493
2. https://github.com/spinnaker/orca/pull/4494

There are conflicts though. I fixed them and created two more PRs:
1. https://github.com/spinnaker/orca/pull/4496
2. https://github.com/spinnaker/orca/pull/4497

dbyron-sf:
Do you have permissions to push the fixes directly to the backport PRs?  Doesn't it make sense to backport to release-1.30.x as well?

Still curious what this has to do with manual judgment.

PR #4412, opened 2023-03-15
feat(igor): Default the feature flag (to send the job name as quey parameter) to on

ovidiupopa07:


kkotula (approved):
LGTM 🚀

dogonthehorizon (approved):
LGTM, do we also need to update the feature preview for 1.30 to highlight that this defaults to 'on' now?

PR #4414, opened 2023-03-21
Fix/blue green deploy

kkotula:
This PR addresses an interesting bug observed while using Blue/Green (Red/Black) deployment.

If the first replica set deployment fails (e.g. due to an incorrect image), a consecutive deployment with a fixed image name would also fail. This was due to one of the new deployment stages disabling the previous deployment. The DisableManifestStage waits for the manifest to stabilize, which never happens since the original pods were not running in the first place.

In this PR, I attempt to tackle the problem by verifying the old manifest status. When it is not stable but not failed, we are dealing with the situation described above.

I also made slight refactoring to the existing code. The primary focus of the refactoring was to extract new helper classes that encapsulate some parts of the existing logic.

### DIFF
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStage.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStage.java
@@ -99,7 +99,7 @@ public class DeployManifestStage extends ExpressionAwareStageDefinitionBuilder {
   }
 
   private void addStagesForOldManifests(
-      Map<String, Object> parentContext, StageGraphBuilder graph, String stageType) {
+      Map<String, Object> parentContext, StageGraphBuilder graph, String defaultStageType) {
     List<Map<String, ?>> deployedManifests = getNewManifests(parentContext);
     String account = (String) parentContext.get("account");
     Map manifestMoniker = (Map) parentContext.get("moniker");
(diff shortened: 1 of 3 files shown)


orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStage.java
```diff
@@ -90,81 +104,214 @@ public void afterStages(@Nonnull StageExecution stage, @Nonnull StageGraphBuilde
...
+     * During a B/G deployment, if the blue deployment fails to create pods (due to issues such as
+     * an incorrect image name), the deployment will not fail, but will wait indefinitely to achieve
+     * stability. This is indicated by status.failed=false and status.stable=false. This method
+     * checks for such a situation.
+     *
+     * @param account used to run deployment
+     * @param name of the manifest
+     * @return true, if manifest was not deployed correctly and waits to get stable, false otherwise
+     */
+    boolean previousDeploymentNeitherStableNorFailed(String account, String name) {
```
kkotula, line 249: This method is the actual fix to the bug. We need to very previous manifest status.

mattgogerly (approved):

armory-abedonik (approved):

ovidiupopa07 (approved):
Lgtm

ovidiupopa07:
@mergify backport release-1.27.x release-1.28.x release-1.29.x

PR #4383, opened 2023-01-31
fix(timeout): Added feature flag for rollback timeout ui input.

DanielaS12:
Feature was tested in previous pr. Adding feature flag.

PR #4416, opened 2023-03-24
chore(ff): Set default for dynamicRollbackTimeoutEnabled to true.

DanielaS12:
Defaulting the feature flag to true.

The feature flag enables the timeout value to be configurable from the UI. If the value is not set in the UI the hardcoded value is used.

The implementation was tested on a previous pr.


dbyron-sf (changes requested):
Can we please stop doing this?  This flag went in in https://github.com/spinnaker/orca/pull/4383 which hasn't been released in 1.30.0 yet.  Until it is has, it's not ok to enable the feature by default.

cc: @jasonmcintosh

jasonmcintosh:
Intentionally closing this since we don't have this feature in a release yet.  We can re-open in the future after evaluation

PR #4453, opened 2023-05-11
fix(deployment): fixed missing namespace while fetching manifest deta…

kkotula:
This fix addresses a problem of missing “namespace” attribute in a HTTP call, that’s being sent while fetching manifest details from clouddriver. Without it clouddriver cannot find correct manifest and returns 404.

### DIFF
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStage.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStage.java
@@ -158,7 +158,7 @@ public class DeployManifestStage extends ExpressionAwareStageDefinitionBuilder {
           (name, manifest) -> {
             var oldManifestIsUnstable =
                 this.manifestOperationsHelper.previousDeploymentNeitherStableNorFailed(
-                    manifest.getAccount(), name);
+                    manifest.getAccount(), manifest.getNamespace(), name);
             var nextStageType =
                 oldManifestIsUnstable
                     ? DeleteManifestStage.PIPELINE_CONFIG_TYPE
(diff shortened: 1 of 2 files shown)


dbyron-sf:
This seems like a fundamental enough change that it's hard to believe we haven't run into it yet / that it doesn't fail for "everyone."  Do you have a specific pipeline that fails?  I'd also like to see a test that demonstrates the bug, though I realize that might be hard given that it involves both orca and clouddriver.  Can you at least link to the code in clouddriver that generates the 404?

I'm trying to figure out what's at risk of breaking with this change.  Can you help clarify that?

mattgogerly:
> This seems like a fundamental enough change that it's hard to believe we haven't run into it yet / that it doesn't fail for "everyone." Do you have a specific pipeline that fails? I'd also like to see a test that demonstrates the bug, though I realize that might be hard given that it involves both orca and clouddriver. Can you at least link to the code in clouddriver that generates the 404?
> 
> I'm trying to figure out what's at risk of breaking with this change. Can you help clarify that?

I think the context is here https://github.com/armory/docs/pull/1965. I'm not clear how it broke though.

dbyron-sf:
> I think the context is here [armory/docs#1965](https://github.com/armory/docs/pull/1965). I'm not clear how it broke though.

That helps.  The text there says:
> Version 2.28.6 introduced a bug which affects the Blue/Green(Red/Black) deployment strategy in Kubernetes. 

Can someone point to the offending commit?

kkotula:
Here's the fix in action while doing deployment with namespace overridden:

https://github.com/spinnaker/orca/assets/25036832/37eba574-2969-42da-bc58-87fa1968f7d7

jasonmcintosh:
I'm inclined to approve, based upon the videos & RL testing.  Would feel more comfortable with better integration tests on this kinda stuff longer term...

jasonmcintosh:
@mergify backport release-1.28.x release-1.29.x release-1.30.x

PR #4460, opened 2023-05-26
feat(orca-bakery/manfests): add helmfile support

thameezb:
Related to https://github.com/spinnaker/spinnaker/issues/6837

Adds in support for [helmfile](https://helmfile.readthedocs.io/en/latest/) as a "Render Engine" in the "Bake (Manifest)" stage.

Relies on https://github.com/spinnaker/rosco/pull/986


jasonmcintosh:
SO minor, but some tests please would be nice :) I'd not expect huge tests, just a few to verify some of these

thameezb:
> SO minor, but some tests please would be nice :) I'd not expect huge tests, just a few to verify some of these

Hey @jasonmcintosh I have added in some tests which covers most of `CreateBakeManifestTask.java`. Please advise if this is what you were looking for

orca-bakery/src/main/java/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java
```diff
@@ -111,6 +112,10 @@ public TaskResult execute(@Nonnull StageExecution stage) {
     }
 
     String outputArtifactName = expectedArtifacts.get(0).getMatchArtifact().getName();
+    if (context.getOutputName() == null || !context.getOutputName().equals(outputArtifactName)) {
```
jasonmcintosh, line 115: ONLY concern - since this impacts other bake types, I'd expect this to have more of an impact than JUST helmfiles. 
 That said not opposed to this, I just didn't recall the output name HAVING to match previously.  ID yes not necessarily name :) Relatively minor thing I'll dig a bit..

thameezb: I added it originally as a extra verification that the matched artifact is actually the artifact one wants (based on the cloudfoundary bake task). 

However upon further testing this morning, existing Bake steps do not actually require one to set the name under "Match Artifacts" and therefore enabling this might break some user's pipelines (had they not been explicit in their match artifact). 

I will remove this check and corresponding test

thameezb: Removed and updated branch with latest master

jasonmcintosh (approved):

thameezb:
@jasonmcintosh big thanks 🚀

PR #4462, opened 2023-06-02
feat(mpt-v1): Support for jinja expressions in stages

jervi:
Adds support for using jinja expressions at the top level in the stages block. Because the template file had to be both well-formed YAML and adhering to the schema defined in `PipelineTemplate`, there was no way to use expressions at the top level here, as can be seen in the `dynamic-stages-001.yml` file. With this patch Orca will first load the yaml file as a `Map` instead of directly as a `PipelineTemplate`. Then it will parse any variables to create a context for the Jinja parser, and if the stages section is a single string, it will parse any Jinja expressions in the stages section. Finally it will load the result as `PipelineTemplate`, just as before, and proceed with processing the rest of the handler chain.

This lets you do stuff like this:

```yaml
schema: "1"
id: simpleTemplate
variables:
  - name: myList
    type: list
    defaultValue:
      - item-1
      - item-2
stages: |
  {%- for item in myList %}
  - id: wait-{{ item }}
    name: Wait {{ item }}
    type: wait
    config:
      waitTime: 5
  {%- endfor %}
```

And it will render like this:
```yaml
schema: "1"
id: simpleTemplate
variables:
  - name: myList
    type: list
    defaultValue:
      - item-1
      - item-2
stages:
  - id: wait-item-1
    name: Wait item-1
    type: wait
    config:
      waitTime: 5
  - id: wait-item-2
    name: Wait item-2
    type: wait
    config:
      waitTime: 5
```

### DIFF
--- a/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/loader/TemplateLoader.java
+++ b/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/loader/TemplateLoader.java
@@ -18,58 +18,153 @@ package com.netflix.spinnaker.orca.pipelinetemplate.loader;
 
 import static java.lang.String.format;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.netflix.spinnaker.orca.pipelinetemplate.exceptions.TemplateLoaderException;
+import com.netflix.spinnaker.orca.pipelinetemplate.exceptions.TemplateRenderException;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.TemplateMerge;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.PipelineTemplate;
 import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.model.TemplateConfiguration;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.render.RenderContext;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.render.RenderUtil;
+import com.netflix.spinnaker.orca.pipelinetemplate.v1schema.render.Renderer;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 @Component
+@Slf4j
 public class TemplateLoader {
   private Collection<TemplateSchemeLoader> schemeLoaders;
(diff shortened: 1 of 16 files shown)


orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessorSpec.groovy
```diff
@@ -334,6 +338,16 @@ class PipelineTemplatePipelinePreprocessorSpec extends Specification {
     result.limitConcurrent == false
   }
 
+  def "should render stages dynamically with jinja"() {
```
dbyron-sf, line 341: What else could stages be but a String?  Possible to add a test for a non-String case?

orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/pipelinetemplate/loader/TemplateLoader.java
```diff
@@ -18,58 +18,154 @@
...
+  private Renderer renderer;
 
   @Autowired
-  public TemplateLoader(Collection<TemplateSchemeLoader> schemeLoaders) {
+  public TemplateLoader(
+      Collection<TemplateSchemeLoader> schemeLoaders,
+      ObjectMapper objectMapper,
+      Renderer renderer) {
     this.schemeLoaders = schemeLoaders;
+    this.objectMapper =
```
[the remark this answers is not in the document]
jervi: The incoming one is configured for json, we need one for yaml. I just copied the approach that was done in another place, but I have no strong feelings about this. I can remove the config part.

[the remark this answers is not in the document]
dbyron-sf: I don't have strong feels either.  This seems fine.

orca-pipelinetemplate/src/test/groovy/com/netflix/spinnaker/orca/pipelinetemplate/PipelineTemplatePipelinePreprocessorSpec.groovy
jervi: It's not much else it can be... If it's not a String, then it will be an object (specifically Map), and all the other tests are already verifying this.

dbyron-sf: The PR description says
> if the stages section is a single string

which is why I was thinking about it.

jervi: Ah, right! The test _is_ actually testing that: https://github.com/spinnaker/orca/pull/4462/files#diff-8a720d0cfc5eace01931edc129f9ad66d30d584c561234cbfce8cd08bef6f308R10
(Note the pipe at the end of the line, which makes the entire block a single, multiline string)

dbyron-sf: Aaah, ok.  Perhaps update the name of the test to mention this?

jervi: I've updated the test description and added a comment in the yaml file that is being tested

dbyron-sf (approved):

PR #4469, opened 2023-06-15
fix(queue): Manual Judgment propagation

jcavanagh:
Manual Judgment was always propagated if the MJ stage was the first one in the pipeline, regardless of the value of `propagateAuthenticationContext`.

The code for backtracking through stages to find an `AuthenticatedStage` provider would improperly return any stage that did not have an ancestor, ignoring criteria for selecting a stage that should be propagating authentication.

Additionally, code in `AuthenticationAware` would mutate a `SKIPPED` MJ stage, copying the `LastModifiedDetails` value from the "real" MJ stage ancestor (or the root stage, as described above).  This mutation is not necessary, and `AuthenticationAware` can just ignore it as it traverses ancestors of the current stage.

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
@@ -60,33 +60,16 @@ interface AuthenticationAware {
   fun retrieveAuthenticatedUser(stage: StageExecution) : PipelineExecution.AuthenticationDetails? {
     return stageNavigator
       .ancestors(stage)
-      .firstOrNull { it.stageBuilder is AuthenticatedStage } ?.let{
-        (it.stageBuilder as AuthenticatedStage).authenticatedUser(solveAuthStages(it.stage)).orElse(null)
+      .firstOrNull {
+        it.stageBuilder is AuthenticatedStage &&
+        it.stage.isManualJudgmentType &&
+        !it.stage.status.isSkipped &&
+        it.stage.withPropagateAuthentication()
+      }?.let{
+        val authStage = it.stage
+        return if(authStage != null) {
+          (it.stageBuilder as AuthenticatedStage).authenticatedUser(authStage).orElse(null)
+        } else null
       }
-  }
-
-
-  // When a first valid candidate is found in the ancestors chain is returned
-  // until the ancestor chain was iterated completely at the pipeline beginning
-  fun backtrackStages(stage: StageExecution): StageExecution {
-    if (stage.isManualJudgmentType &&
-      !stage.status.isSkipped &&
-      stage.withPropagateAuthentication()) {
-      return stage;
-    }
-    val previousStage = if (stageNavigator.ancestors(stage).size > 1) stageNavigator.ancestors(stage).get(1).stage else null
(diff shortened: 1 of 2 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
```diff
@@ -60,33 +60,16 @@ interface AuthenticationAware {
...
+        return if(authStage != null) {
+          (it.stageBuilder as AuthenticatedStage).authenticatedUser(authStage).orElse(null)
+        } else null
       }
   }
-
-
-  // When a first valid candidate is found in the ancestors chain is returned
-  // until the ancestor chain was iterated completely at the pipeline beginning
-  fun backtrackStages(stage: StageExecution): StageExecution {
```
jcavanagh, line 71: None of this recursion should be needed - `stageNavigator.ancestors()` provides an ordered list of ancestor stages back to the root, and iterating over that according to the above criteria will find the proper stage to import authentication context from.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
```diff
@@ -60,33 +60,16 @@ interface AuthenticationAware {
...
-    if (stage.isManualJudgmentType &&
-      !stage.status.isSkipped &&
-      stage.withPropagateAuthentication()) {
-      return stage;
-    }
-    val previousStage = if (stageNavigator.ancestors(stage).size > 1) stageNavigator.ancestors(stage).get(1).stage else null
-    return if (previousStage == null) stage else backtrackStages(previousStage)
```
jcavanagh, line 78: This line is the core issue - returning stage with the only criteria being that it has no ancestors. This results in any Manual Judgment stage that happens to be the first stage in a pipeline being used to propagate auth, regardless of its state or configuration.

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/AuthenticationAware.kt
```diff
@@ -60,33 +60,16 @@ interface AuthenticationAware {
...
-  }
-
-  //Next method will look by a possible stage with authentication propagated in case that previous
-  //stage brokes the auth chain, iterating the stage ancestors. By the moment only MJ stages approved with
-  //auth propagated are considerated as candidates
-  fun solveAuthStages(stage: StageExecution): StageExecution {
-    if (stage.isManualJudgmentType()) {
-      val result = backtrackStages(stage)
-      stage.lastModified = result.lastModified
```
jcavanagh, line 87: This is mutating other stages in the process of finding an authentication source, and is unnecessary.

There was a test that verified this mutation was happening in RunTaskHandlerTest, but was deleted by another PR a couple months back:
https://github.com/spinnaker/orca/pull/4410/files#diff-3178adb0b74ce922d8fee4ad7b299cc95bc9c7c28f08bdb7779e446f60226648L1852

mattgogerly (approved):
LGTM, thanks for the test!

jasonmcintosh (approved):

jasonmcintosh:
@mergify backport release-1.29.x release-1.30.x

PR #4473, opened 2023-06-16
fix(queue): fix ability to cancel a zombied execution

mattgogerly:
Tentative fix for https://github.com/spinnaker/spinnaker/issues/6224.

### DIFF
--- a/orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/ExecutionLatch.kt
+++ b/orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/ExecutionLatch.kt
@@ -46,6 +46,18 @@ class ExecutionLatch(private val predicate: Predicate<ExecutionComplete>) :
   fun await() = latch.await(10, TimeUnit.SECONDS)
 }
 
+fun ConfigurableApplicationContext.run(execution: PipelineExecution, launcher: (PipelineExecution) -> Unit) {
+  val latch = ExecutionLatch(
+    Predicate {
+      it.executionId == execution.id
+    }
+  )
+  addApplicationListener(latch)
+  launcher.invoke(execution)
+
+  Thread.sleep(500)
+}
+
 fun ConfigurableApplicationContext.runToCompletion(execution: PipelineExecution, launcher: (PipelineExecution) -> Unit, repository: ExecutionRepository) {
   val latch = ExecutionLatch(
     Predicate {
@@ -54,7 +66,6 @@ fun ConfigurableApplicationContext.runToCompletion(execution: PipelineExecution,
   )
   addApplicationListener(latch)
   launcher.invoke(execution)
-  assert(latch.await()) { "Pipeline did not complete" }
 
   repository.waitForAllStagesToComplete(execution)
 }
--- a/keiko-core/src/main/kotlin/com/netflix/spinnaker/q/Queue.kt
+++ b/keiko-core/src/main/kotlin/com/netflix/spinnaker/q/Queue.kt
@@ -88,6 +88,11 @@ interface Queue {
   fun retry() {}
 
   /**
+   * Used for testing zombie executions. Wipes all messages from the queue.
+   */
+  fun clear() {}
+
+  /**
    * The expired time after which un-acknowledged messages will be retried.
    */
   val ackTimeout: TemporalAmount
--- a/keiko-mem/src/main/kotlin/com/netflix/spinnaker/q/memory/InMemoryQueue.kt
+++ b/keiko-mem/src/main/kotlin/com/netflix/spinnaker/q/memory/InMemoryQueue.kt
@@ -131,6 +131,10 @@ class InMemoryQueue(
     }
   }
 
+  override fun clear() {
+    queue.removeAll { true }
+  }
+
   override fun readState() =
     QueueState(
       depth = queue.size,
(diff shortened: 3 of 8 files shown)


mattgogerly:
@Mergifyio update

orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/ExecutionLatch.kt
```diff
@@ -46,6 +46,18 @@ class ExecutionLatch(private val predicate: Predicate<ExecutionComplete>) :
   fun await() = latch.await(10, TimeUnit.SECONDS)
 }
 
+fun ConfigurableApplicationContext.run(execution: PipelineExecution, launcher: (PipelineExecution) -> Unit) {
```
nemesisOsorio, line 49: dumb question: what's the purpose of this function?

mattgogerly: `runUntilCompletion()` starts the pipeline and then waits 10 seconds for it to finish, which doesn't work for the purposes of this test when we need to mutate the queue after starting the execution.

This method is the same as `runUntilCompletion` without the part that waits. The sleep is to allow the messages for starting the execution to be processed before we continue on to deleting the queue.

nemesisOsorio (approved):

dbyron-sf (approved):
LGTM, and thank you for fixing this!  I do have some questions though.  Apologies in advance for no good deed going unpunished.  Now that you've figured this out, I'd love to try to get all the info in your brain written down.
* https://github.com/spinnaker/spinnaker/issues/6224 talks about not cancelling pipelines.  This PR seems to fix cancelling of zombies...is it clear that 6224 is really about zombies?
* Any chance this fixes the "Failed to evaluate" messages, or the "hundreds of calls to clouddriver" from https://github.com/spinnaker/spinnaker/issues/6224#issuecomment-904277249?
* Does this fix cancelling of `NOT_STARTED` stages (e.g. https://github.com/spinnaker/spinnaker/issues/6224#issuecomment-908520505)?
* Is it worth logging a warning in [RedisQueue.reschedule](https://github.com/spinnaker/orca/blob/82942412a67738a0717f7701c0f8ac0dc4f6f534/keiko-redis/src/main/kotlin/com/netflix/spinnaker/q/redis/RedisQueue.kt#L159), similar to [SqlQueue.reschedule](https://github.com/spinnaker/orca/blob/82942412a67738a0717f7701c0f8ac0dc4f6f534/keiko-sql/src/main/kotlin/com/netflix/spinnaker/q/sql/SqlQueue.kt#L534-L536)?  Maybe an error is more appropriate.  Updating [the javadoc](https://github.com/spinnaker/orca/blob/82942412a67738a0717f7701c0f8ac0dc4f6f534/keiko-core/src/main/kotlin/com/netflix/spinnaker/q/Queue.kt#L67) to mention what happens in this case seems like a good idea too.
* Anything to update in https://spinnaker.io/docs/guides/runbooks/orca-zombie-executions/ to help people who might struggle with this?  Maybe something about the kind of failure-to-cancel that could occur before spinnaker versions X.Y.Z (i.e. before this PR is merged/backported)?

mattgogerly:
> https://github.com/spinnaker/spinnaker/issues/6224 talks about not cancelling pipelines. This PR seems to fix cancelling of zombies...is it clear that 6224 is really about zombies?

I'd like to get people in spinnaker/spinnaker#6224 to confirm if they still run into this once this is merged.

It's possible there are other issues that could prevent cancelling executions. I'm interested in @karlskewes account of running into this with the SQL queue as it's not clear to me how you could get zombies with SQL (assuming it's not running on Kubernetes..)

I can't see anything obviously wrong with the actual cancel logic. It just (fairly not not always realistically) assumes there's a message to reschedule.

> Does this fix cancelling of NOT_STARTED stages (e.g. https://github.com/spinnaker/spinnaker/issues/6224#issuecomment-908520505)?

If zombies are the issue, kinda. That issue is the zombie lookup agent only considers stages that are running.

This change should allow users to just cancel themselves rather than relying on the agent.

> Is it worth logging a warning in [RedisQueue.reschedule](https://github.com/spinnaker/orca/blob/82942412a67738a0717f7701c0f8ac0dc4f6f534/keiko-redis/src/main/kotlin/com/netflix/spinnaker/q/redis/RedisQueue.kt#L159), similar to [SqlQueue.reschedule](https://github.com/spinnaker/orca/blob/82942412a67738a0717f7701c0f8ac0dc4f6f534/keiko-sql/src/main/kotlin/com/netflix/spinnaker/q/sql/SqlQueue.kt#L534-L536)? Maybe an error is more appropriate. Updating [the javadoc](https://github.com/spinnaker/orca/blob/82942412a67738a0717f7701c0f8ac0dc4f6f534/keiko-core/src/main/kotlin/com/netflix/spinnaker/q/Queue.kt#L67) to mention what happens in this case seems like a good idea too.

Probably. I'll add one.

> Anything to update in https://spinnaker.io/docs/guides/runbooks/orca-zombie-executions/ to help people who might struggle with this? Maybe something about the kind of failure-to-cancel that could occur before spinnaker versions X.Y.Z (i.e. before this PR is merged/backported)?

Sure. PR incoming.

karlskewes:
> It's possible there are other issues that could prevent cancelling executions. I'm interested in @karlskewes account of running into this with the SQL queue as it's not clear to me how you could get zombies with SQL (assuming it's not running on Kubernetes..)

Thank you but sorry, I don't have access to the environment to validate.
Orca (all incl queue) was backed by AWS Aurora MySQL 2.09.x at the time.
IIRC, it was more likely to happen when Orca pods were rolled mid pipeline execution, but it was a couple of years ago.
Might have to see what others in community say. Definitely not a blocker for this fix.

PR #4476, opened 2023-06-19
fix(config): Exception in fresh development environment (LocalGit)

gustavotemple:
After following the guide below in a fresh development environment (LocalGit):
https://spinnaker.io/docs/community/contributing/code/developer-guides/dev-env/getting-set-up/

I got:
`BeanInstantiationException: Failed to instantiate [com.netflix.spinnaker.fiat.shared.FiatService]: Factory method 'fiatService' threw exception`

When I run `./gradlew run`

### DIFF
--- a/orca-web/config/orca.yml
+++ b/orca-web/config/orca.yml
@@ -32,6 +32,11 @@ kayenta:
   enabled: ${services.kayenta.enabled:false}
   baseUrl: ${services.kayenta.baseUrl:http://localhost:8090}
 
+services:
+  fiat:
+    enabled: false
+    baseUrl: http://localhost:7003
+
 redis:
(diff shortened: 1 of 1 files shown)


orca-web/config/orca.yml
```diff
@@ -32,6 +32,10 @@ kayenta:
   enabled: ${services.kayenta.enabled:false}
   baseUrl: ${services.kayenta.baseUrl:http://localhost:8090}
 
+services:
+  fiat:
+    baseUrl: http://localhost:7003
```
mattgogerly, line 37: Thanks for this!

I'd suggest adding `services.fiat.enabled: false` as the default too, to ensure Orca doesn't fail to start if Fiat isn't also running locally.

nemesisOsorio (approved):

PR #4449, opened 2023-04-27
feat(lambda): migrate stages from AWS Lambda plugin to OSS

mattgogerly:
Moving the Orca component of https://github.com/spinnaker-plugins/aws-lambda-deployment-plugin-spinnaker to OSS.

I've cleaned up a whole load of low hanging fruit code smells in the process (stuff like inlining returns, type inference for collections etc). There's a whole load of other cleanup (constructor injection over field injection as an example) that's still needed.

Note that the plugin is pretty much built around calling Clouddriver directly rather than using existing services like `KatoService`. The bulk of this is in `LambdaCloudDriverUtils`, which creates its own `OkHttpClient` rather than a static one. Because of this the plugin has separate network configuration to the rest of Orca.

In the interests of keeping this migration as safe and as easy for users as possible I'm saving refactoring of this for a future PR.

To follow:
- Deck PR
- Breaking change release note

PR #4492, opened 2023-07-27
chore(lambda): rewrite to use Kato/Oort

mattgogerly:
As promised, cleaning up the Lambda stages to use Kato/Oort rather than directly calling Clouddriver. Couple of other bits of cleanup too.

- Use `Kato` to request Lambda ops and check status
- Use `Oort` to get Lambdas from cache
- Use existing `stage.mapTo()` to map context to classes rather than custom util


kyledmitchell:
Hey @mattgogerly.  Thanks for taking this on!  Is your overall intent with this PR to move capabilities currently in the [Lambda Plugin](https://github.com/spinnaker-plugins/aws-lambda-deployment-plugin-spinnaker) natively into OSS so we can deprecate the plugin?  

Is there anything in the plugin that you are not planning on replacing?  

Do you think additional related PRs will need to be created in addition to this one?

mattgogerly:
> Hey @mattgogerly. Thanks for taking this on! Is your overall intent with this PR to move capabilities currently in the [Lambda Plugin](https://github.com/spinnaker-plugins/aws-lambda-deployment-plugin-spinnaker) natively into OSS so we can deprecate the plugin?
> 
> Is there anything in the plugin that you are not planning on replacing?
> 
> Do you think additional related PRs will need to be created in addition to this one?

The plugin code was moved into Orca in https://github.com/spinnaker/orca/pull/4449.

This PR rewrites that code to use Kato/Oort.

dbyron-sf:
@mergifyio update

PR #4498, opened 2023-08-08
fix(redirect): Allow for redirects instead of throwing exception

ciurescuraul:
When using SpEL functions like `*FromUrl` when the request requires auth and causes a redirect response, we get the following "Attempted redirect from xxx to xxx which is not supported " error message. 

This [PR](https://github.com/spinnaker/orca/pull/2978/files) was to change failed redirects with an explicit error message instead of failing with an error message that didn't explain why redirects failed.

Make it allow redirects instead of throwing _RetryRequestException_ under the retry-enabled mechanism using a more lenient approach. 
This can be useful in scenarios where we want to treat all redirects as GET requests to ensure smoother redirection in cases where some servers might not handle method-specific redirects correctly.

As a side note, there is also an update for the WireMock artifact version due to reported CVE vulnerabilities [CVE-2018-9116](https://devhub.checkmarx.com/cve-details/CVE-2018-9116); [CVE-2018-9117](https://devhub.checkmarx.com/cve-details/CVE-2018-9117) in v2.15.0

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
@@ -19,13 +19,11 @@ package com.netflix.spinnaker.orca.pipeline.util;
 import com.google.common.io.CharStreams;
 import com.google.common.util.concurrent.Uninterruptibles;
 import com.netflix.spinnaker.orca.config.UserConfiguredUrlRestrictions;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.Reader;
-import java.util.Arrays;
-import java.util.List;
-import java.util.concurrent.TimeUnit;
-import org.apache.http.*;
+import org.apache.http.HttpHost;
+import org.apache.http.HttpRequest;
+import org.apache.http.HttpResponse;
+import org.apache.http.HttpStatus;
+import org.apache.http.ProtocolException;
 import org.apache.http.client.RedirectStrategy;
 import org.apache.http.client.ServiceUnavailableRetryStrategy;
 import org.apache.http.client.config.RequestConfig;
@@ -42,10 +40,18 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
 @Component
 public class HttpClientUtils {
 
   private CloseableHttpClient httpClient;
+  private final UserConfiguredUrlRestrictions urlRestrictions;
   private static final Logger LOGGER = LoggerFactory.getLogger(HttpClientUtils.class);
   private static final String JVM_HTTP_PROXY_HOST = "http.proxyHost";
   private static final String JVM_HTTP_PROXY_PORT = "http.proxyPort";
@@ -58,6 +64,7 @@ public class HttpClientUtils {
 
   public HttpClientUtils(UserConfiguredUrlRestrictions userConfiguredUrlRestrictions) {
     this.httpClient = create(userConfiguredUrlRestrictions.getHttpClientProperties());
+    this.urlRestrictions = userConfiguredUrlRestrictions;
   }
 
   private CloseableHttpClient create(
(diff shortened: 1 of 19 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -160,9 +162,32 @@ public String httpGetAsString(String url) throws IOException {
     }
   }
 
+  private int getProxyPort() {
```
kkotula, line 165: I like that you decided to improve the existing code. It's much safer than it was before, thanks!

kkotula (approved):
LGTM 🚀

orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtilsTest.java
```diff
@@ -32,6 +33,7 @@ public class HttpClientUtilsTest {
   private final String host = "localhost";
   private final int port = 8080;
   private final String resource = "/v1/text";
+  private final String redirectUrl = "https://armory.jfrog.io/artifactory/";
```
jasonmcintosh, line 36: Ideally this is some OTHER url.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -160,9 +162,32 @@ public String httpGetAsString(String url) throws IOException {
...
+
   class RetryRequestException extends RuntimeException {
     RetryRequestException(String cause) {
       super(cause);
     }
   }
+
+  class CustomRedirect {
+    public RedirectStrategy getLaxRedirectStrategy(){
+      return new LaxRedirectStrategy();
```
jasonmcintosh, line 190: I'm not familiar with the lax redirect strategy, but ideally failures should be a SpinnakerException vs. one of the other exceptions on failure to do the redirect.  This SHOULD bubble up the errors more effectively from an end user messaging perspective.

ciurescuraul: I reverted to safer logic, so I left the default redirection strategy as it was before and applied a more relaxed one in case the retry mechanism is enabled and handled the exception separately. Added more tests for this.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/HttpClientUtils.java
```diff
@@ -160,9 +175,44 @@ public String httpGetAsString(String url) throws IOException {
...
+
+  class CustomRedirectStrategy {
+    public RedirectStrategy getLaxRedirectStrategy() {
+      return new LaxRedirectStrategy() {
+        @Override
+        public boolean isRedirected(
+            HttpRequest request, HttpResponse response, HttpContext context) {
+
+          int statusCode = response.getStatusLine().getStatusCode();
+          if ((statusCode >= 300) && (statusCode <= 399)) {
```
jasonmcintosh, line 209: I'd not run the full range, as each status code is supposed to be handled SOMEWHAT differently.
* 301 & 302 - follow & go to the requested URL using "GET" method
* 307 & 308 - this one is trickier.  This is "make a request using the SAME requesting method and PAYLOAD to the new URL".  NORMALLY the requirement is ALSO to prompt the user to ENSURE they really want this.  I'd be hesitant given the situation with spinnaker to support 307/308 redirects in any shape, and ONLY support 301/302s (which should cover MOST use cases)

jasonmcintosh: https://www.rfc-editor.org/rfc/rfc9110.html#name-redirection-3xx FYI.  `Automatic redirection needs to be done with care for methods not known to be [safe](https://www.rfc-editor.org/rfc/rfc9110.html#safe.methods), as defined in [Section 9.2.1](https://www.rfc-editor.org/rfc/rfc9110.html#safe.methods), since the user might not wish to redirect an unsafe request.` PARTICULARLY given this can expose URLs that are NOT safe... there's definitely a "caution" flag here.  E.g. you get into link local redirection follows for pulling metadata WHICH SHOULD NOT be allowed.  


ANOTHER THING NEEDED:
UserConfiguredUrlRestrictions.validateURI on the redirect URL MUST BE CALLED.  Or you're opening up link local attacks.  Since I think this is recreating the http client as part of retry operations...

jasonmcintosh:
Last thing
```
* What went wrong:
Execution failed for task ':orca-core:spotlessJavaCheck'.
> The following files had format violations:
```

Need to do a spotless apply but overall I think this is good!

jasonmcintosh:
approved.  ONLY thing i'm... iffy on is if this is truly a "fix" or an enhancement.  It fixes broken redirects (aka artifactory's s3 storage options).  I'm inclined towards "fix" b/c the current functionality of redirects WAS there, then removed b/c sometimes it breaks and this is just re-enabling it in a safer manner.

PR #4508, opened 2023-08-18
fix(expressions): fetch labels from actually deployed manfiest

mattgogerly:
We have a use case where we want to make a stage conditional based on the value of the sequence label. This label isn't present in the `manifests` field of the context, because it's only applied during deployment by Clouddriver.

The `outputs.manfiests` field contains the full Kubernetes manifests that were deployed by Clouddriver, including labels added by Spinnaker during deployment. This should be a backwards compatible change, as the labels in the `manifests` field are obviously also included in the `outputs.manfiests` field.

I would have preferred to make this change optional using a default function argument of false, but it seems the `FunctionParameter` [class](https://github.com/spinnaker/kork/blob/1d5d80b2b9618bc801e012bf5bf29c97041ac8ec/kork-api/src/main/java/com/netflix/spinnaker/kork/api/expressions/ExpressionFunctionProvider.java#L154) has no concept of a default value.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ManifestLabelValueExpressionFunctionProvider.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ManifestLabelValueExpressionFunctionProvider.java
@@ -87,7 +87,11 @@ public class ManifestLabelValueExpressionFunctionProvider implements ExpressionF
           "A valid Deploy Manifest stage name is required for this function");
     }
 
-    List<Map> manifests = (List<Map>) stage.get().getContext().get("manifests");
+    // using outputs.manifests to give access to labels added by Spinnaker itself during the
+    // deployment
+    // this is safe as we assert above that we're only using successful deploy stages, so this key
+    // should always exist
+    List<Map> manifests = (List<Map>) stage.get().getContext().get("outputs.manifests");
 
     if (manifests == null || manifests.size() == 0) {
       throw new SpelHelperFunctionException(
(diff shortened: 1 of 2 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/expressions/functions/ManifestLabelValueExpressionFunctionProvider.java
```diff
@@ -87,7 +87,11 @@ public static String manifestLabelValue(
           "A valid Deploy Manifest stage name is required for this function");
     }
 
-    List<Map> manifests = (List<Map>) stage.get().getContext().get("manifests");
+    // using outputs.manifests to give access to labels added by Spinnaker itself during the
+    // deployment
+    // this is safe as we assert above that we're only using successful deploy stages, so this key
+    // should always exist
+    List<Map> manifests = (List<Map>) stage.get().getContext().get("outputs.manifests");
```
dbyron-sf, line 94: Is this safe timing-wise?  Like, could this run when manifests exists, but output.manifests doesn't?

mattgogerly: I'm reasonably confident `outputs.manifests` should always exist once a deploy stage is succeeded - and this expression only considers stages in a succeeded state.

dbyron-sf: Aaah, I see it [a few lines up](https://github.com/spinnaker/orca/pull/4508/files#diff-275c694ae6c05141ecb04851930851ddefd1d55195336b0f3c5ac1e46639f758R82).

dbyron-sf (approved):
Seems like this is worth a release note so people know to be on the lookout for a change in behavior...even if we don't expect one.

mattgogerly:
@Mergifyio update

PR #4517, opened 2023-08-31
fix(orca): tying exceptions to tasks

SusmithaGU:
To address this issue:
https://github.com/spinnaker/spinnaker/issues/6867


mattgogerly:
Could you add a test for this to ensure it doesn't get broken again?

SusmithaGU:
> Could you add a test for this to ensure it doesn't get broken again?

done

mattgogerly (approved):

PR #4522, opened 2023-09-14
feat(lambda): Configure post timeouts for clouddriver.  Use built-in okhttp vs. recreating http client

jasonmcintosh:
https://github.com/spinnaker-plugins/aws-lambda-deployment-plugin-spinnaker/pull/134/files basically copy of this


dbyron-sf:
nit: perhaps `feat(lambda)` instead of `feat(timeouts)` ?

christosarvanitis (approved):

PR #4524, opened 2023-09-14
feat(retrofit): add SpinnakerServerExceptionHandler

dbyron-sf:
Not used yet, since orca doesn't use SpinnakerRetrofitErrorHandler, and doesn't explicitly throw SpinnakerServerException, nor its children.  Once we do use it, we'll get similar information present in the execution context if a SpinnakerServerException happens as we get today if a RetrofitError happens, and the same retry behavior.

### DIFF
--- a/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/BaseRetrofitExceptionHandler.groovy
+++ b/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/BaseRetrofitExceptionHandler.groovy
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2023 Salesforce, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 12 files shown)


orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/BaseRetrofitExceptionHandler.groovy
```diff
@@ -0,0 +1,93 @@
...
+  private boolean isMalformedRequest(String kind, String exceptionMessage) {
+    // We never want to retry errors like "Path parameter "blah" value must not be null.
+    return "UNEXPECTED".equals(kind) && exceptionMessage?.contains("Path parameter")
+  }
+
+  private static boolean isIdempotentRequest(Exception e) {
+    findHttpMethodAnnotation(e) in ["GET", "HEAD", "DELETE", "PUT"]
+  }
+
+  private static String findHttpMethodAnnotation(Exception exception) {
```
jasonmcintosh, line 62: Only worry, just at a glance here... there's a fair bit of reflection on this that I've seen cause major perf issues.  This particular code doesn't look bad off hand ... just one of those triggering a "caution" flag.  I'd wonder about using AnnotationUtils or similar spring libs which as i recall do caching of reflection calls so the forName/method search is only done once.

jasonmcintosh: That said... it IS a "move" soo.... ignore for now, just one of those caught my eye

jasonmcintosh (approved):

orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/BaseRetrofitExceptionHandler.groovy
dbyron-sf: Yup, it's a drag, but it's no more reflection than we were already doing.  I don't think we need it once we move to retrofit2.

PR #4526, opened 2023-09-15
fix(artifacts): Resolving is broken in multiple places

xibz:
8df68b79cf1 broke custom triggers to Spinnaker integration. Prior to this commit if `getBoundArtifactForStage` failed to find a bound artifact, resolution would happen much later by utilizing the `triggers` field in stages.  However, `getBoundArtifactForStage` now throws an exception, which causes custom triggers to fail prematurely. Also if a custom trigger calls orchestrate directly and any pipeline contains `getBoundArtifactForStage`, this will throw an exception and prevent any resolution.

@jervi: please take a look at this as I noticed a couple of your PRs made some changes to artifact resolving, but I believe they only partially implemented the needed solution.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
@@ -241,20 +241,13 @@ public class ArtifactUtils {
     Set<String> expectedArtifactIdsListConcat =
         new HashSet<>(getExpectedArtifactIdsFromMap(trigger));
 
-    // Due to 8df68b79cf1 getBoundArtifactForStage now does resolution which can
-    // potentially return an empty list or null back, which will throw an
-    // exception. Before this commit, when getBoundArtifactForStage was called,
-    // the method would just retrieve the bound artifact from the stage context,
-    // and return it as an ExpectedArtifact type instead of a map. "Resolution"
-    // would happen later in the DependentPipelineStarter class which handles
-    // the overall pipeline triggers case which we have now added here.
-    //
-    // Another case is when any custom trigger occurs, and it runs the
-    // orchestrate endpoint and if any pipeline stage uses
-    // getBoundArtifactForStage then this will throw an exception, instead of
-    // returning the bound artifact which will be later be added as an resolved
-    // artifact in the context. A good example of this is the
-    // CreateBakeManifestTask.
+    // Due to 8df68b79cf1 getBoundArtifactForStage now does resolution which
+    // can potentially return null artifact back, which will throw an exception
+    // for stages that expect a non-null value. Before this commit, when
+    // getBoundArtifactForStage was called, the method would just retrieve the
+    // bound artifact from the stage context, and return the appropriate
+    // artifact back. This change prevents tasks like CreateBakeManifestTask
+    // from working properly, if null is returned.
     //
     // reference: https://github.com/spinnaker/orca/pull/4397
     triggers.stream()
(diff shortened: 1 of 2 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
```diff
@@ -227,10 +230,45 @@ public List<Artifact> getArtifactsForPipelineIdWithoutStageRef(
         .orElse(Collections.emptyList());
   }
 
+  private List<String> getExpectedArtifactIdsFromMap(Map<String, Object> trigger) {
+    return Optional.ofNullable((List<String>) trigger.get("expectedArtifactIds"))
+        .orElse(emptyList());
```
nemesisOsorio, line 235: in simple cases like this, I prefer to use the ternary conditional operator `?:`, we avoid the creation of another object(`Optional`)

xibz: Okay, I generally avoid ternaries as they can make readability more difficult, but that's a philosophy haha. But yea, let me go ahead and update

nemesisOsorio (approved):

jervi (approved):
I _think_ this makes sense 😅 Thanks for fixing @xibz, and I'm sorry for introducing any breaking changes... The artifact resolving code is complicated - it's hard to introduce/fix anything there without inadvertently ruining someone else's workflows 🥴 (some of those workflows are being dependent on what I would call bugs in earlier Orca versions though)

xibz:
@jervi awesome thank you! And no worries for the breaking changes; it happens :p especially as something as complex as the artifact resolving logic.

One thing I plan on doing is documenting all my findings and knowledge in a README to help explain the whole complicated process, which may help facilitate discussion on the architecture. I do think it's overly complicated and there is room for improving the logic where we can simplify it, but that can be a separate discussion. So with that said, I'll ping you as well so we can have a group effort on updating this future README so we get everyone who's touched artifact resolving ensure that the knowledge gets dumped.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
```diff
@@ -227,10 +230,45 @@ public List<Artifact> getArtifactsForPipelineIdWithoutStageRef(
...
-        (List<?>) trigger.getOrDefault("expectedArtifactIds", emptyList());
+    List<?> triggers = Optional.ofNullable((List<?>) pipeline.get("triggers")).orElse(emptyList());
+    Set<String> expectedArtifactIdsListConcat =
+        new HashSet<>(getExpectedArtifactIdsFromMap(trigger));
+
+    // Due to 8df68b79cf1 getBoundArtifactForStage now does resolution which can
+    // potentially return an empty list or null back, which will throw an
+    // exception. Before this commit, when getBoundArtifactForStage was called,
+    // the method would just retrieve the bound artifact from the stage context,
+    // and return it as an ExpectedArtifact type instead of a map. "Resolution"
```
dbyron-sf, line 248: Not sure I understand the `instead of a map` part.    Looking at https://github.com/spinnaker/orca/commit/8df68b79cf103e954f305c342fbc33733f26b8f2#diff-b25c50ecaceee2f8b0ae1ddc77fc92d48d16fb7ff9d15d27612685f1985b2a1eL136-R144, didn't both the "before" and "after" code return an Artifact?

This means I don't understand how getBoundArtifactForStage can return an empty list since the return type is `Artifact`.

I'm also a little unclear why there's a comment here, in `resolveArtifacts` about the behavior of `getBoundArtifactForStage`.  I could use a bit more connecting of the dots.

dbyron-sf (approved):

PR #4528, opened 2023-09-17
feat(clouddriver): use SpinnakerRetrofitErrorHandler for retrofit client beans in CloudDriverConfiguration

dbyron-sf:
to improve the behavior of RetrySupport, to improve error messages in the UI, and to pave
the way to move to retrofit2 by reducing dependencies on retrofit1 classes
(e.g. RetrofitError).  The main consequence of this is that code that previously checked
for RetrofitError now checks for Spinnaker*Exception.

Specifically, the beans for these classes now use SpinnakerRetrofitErrorHandler:

* KatoRestService
* FeaturesRestService
* MortService
* CloudDriverCacheStatusService
* OortService
* CloudDriverTaskStatusService

and then as well

* KatoService since it wraps KatoRestService
* CloudDriverService since it wraps OortService

There are a bunch of refactor commits that could be in their own PR, but they're connected enough with the exception handling changes that I kept them here.  All the individual commits to keep up with the addition of SpinnakerRetrofitErrorHandler could all be together in one, but it seemed easier to make incremental progress as I was doing it to commit each change.

### DIFF
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
@@ -20,6 +20,7 @@ import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.kork.annotations.VisibleForTesting
 import com.netflix.spinnaker.kork.core.RetrySupport
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
 import com.netflix.spinnaker.orca.api.pipeline.RetryableTask
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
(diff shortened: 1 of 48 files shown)


orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
```diff
@@ -108,8 +108,8 @@ class MonitorKatoTask implements RetryableTask, CloudProviderAware {
         katoTask = kato.lookupTask(taskId.id, false)
       }, MAX_HTTP_INTERNAL_RETRIES, Duration.ofMillis(100), false)
       outputs['kato.task.notFoundRetryCount'] = 0
-    } catch (RetrofitError re) {
-      if (re.kind == RetrofitError.Kind.HTTP && re.response.status == HttpURLConnection.HTTP_NOT_FOUND) {
+    } catch (SpinnakerHttpException re) {
+      if (re.getResponseCode() == HttpURLConnection.HTTP_NOT_FOUND) {
```
mattgogerly, line 112: There's a bit of mix and matching going on between magic numbers and enums. Does it make sense to settle on one approach?

mattgogerly (approved):

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/MonitorKatoTask.groovy
dbyron-sf: It does.  It's a level of hygiene I wasn't focused on while doing this unfortunately.  Madly trying to get rid of retrofit1.

PR #4534, opened 2023-09-22
fix(vpc): add data annotation to vpc

dmart:
Annotate VPC as a data class to address VPC attributes always being null.


dbyron-sf:
Can you tell us more about what fails without this?  Possible to add a test that demonstrates it?

dmart:
Calls to `findForRegionAndAccount` throw an NPE, without a constructor or getters/setters Jackson is unable to set the properties. This worked previously before being ported to Java.

dbyron-sf:
OK, thanks.  For the record, that conversion to java happened in https://github.com/spinnaker/orca/pull/4269.  I'm all for fixing this, but would love a test to keep us from breaking this again if you're up for it.

dmart:
Added a simple test.

dbyron-sf (approved):
Thank you!

dbyron-sf:
https://github.com/spinnaker/orca/pull/4269 first went in to 1.29, which is old enough that we're not releasing fixes for it.  Gonna backport the fix there, and to the other live branches.

dbyron-sf:
@mergifyio backport release-1.29.x release-1.30.x release-1.31.x release-1.32.x

PR #4554, opened 2023-10-09
fix(front50): filtering null execution ids for MonitorPipelineTask

nemesisOsorio:
We have a plugin that implements [ExecutionPreprocessor](https://github.com/spinnaker/orca/blob/master/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/ExecutionPreprocessor.java), the implementation throws an exception given some rules to avoid the execution of [StartPipelineTask](https://github.com/spinnaker/orca/blob/master/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/StartPipelineTask.groovy), if you configure your [PipelineStage](https://github.com/spinnaker/orca/blob/master/orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/pipeline/PipelineStage.java) to `ignore the failure` it will continue the execution of [MonitorPipelineTask](https://github.com/spinnaker/orca/blob/master/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy), that scenario throws an error in [executionRepository.retrieve](https://github.com/spinnaker/orca/blob/cd9781f13438fc32aa6149c6adc04feddace3489/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy#L70) because the execution Id does not exist.

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy
@@ -59,8 +59,14 @@ class MonitorPipelineTask implements OverridableTimeoutRetryableTask {
     if (stage.type == MonitorPipelineStage.PIPELINE_CONFIG_TYPE) {
       pipelineIds = stageData.executionIds
     } else {
-      pipelineIds = Collections.singletonList(stageData.executionId)
+      pipelineIds = [stageData.executionId]
       isLegacyStage = true
+    }
+
+    pipelineIds.removeAll { it == null }
+
+    if (pipelineIds.isEmpty()) {
+      return TaskResult.builder(ExecutionStatus.TERMINAL).context([error: "no execution Ids were provided"]).build()
     }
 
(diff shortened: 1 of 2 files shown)


dbyron-sf:
Sorry to not have made this earlier.  https://github.com/spinnaker/orca/pull/4555 is an alternate implementation we've had running at Salesforce for awhile.  How's it look?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/MonitorPipelineTask.groovy
```diff
@@ -59,10 +59,16 @@ class MonitorPipelineTask implements OverridableTimeoutRetryableTask {
     if (stage.type == MonitorPipelineStage.PIPELINE_CONFIG_TYPE) {
       pipelineIds = stageData.executionIds
     } else {
-      pipelineIds = Collections.singletonList(stageData.executionId)
+      pipelineIds = [stageData.executionId]
```
nemesisOsorio, line 62: `Collections.singletonList` does not support remove

nemesisOsorio:
Closing in favor of https://github.com/spinnaker/orca/pull/4555

PR #4489, opened 2023-07-19
fix(artifacts): consider requiredArtifactIds in expected artifacts when trigger is pipeline type

nemesisOsorio:
fixes https://github.com/spinnaker/spinnaker/issues/6850

Expected artifacts can be used in automated triggers and stages,
as in the following example:

artifacts

if we trigger the previous pipeline with a pipeline stage like the next image

trigger-pipeline-using-stage

you will get the following error even if the parent pipeline provides the artifact

trigger-pipeline

in the example, I'm using a `Find Artifacts From Resource (Manifest)` stage 
and producing a docker image as an artifact, I believe this is a similar case for building
the docker image using Jenkins like is mentioned in https://github.com/spinnaker/spinnaker/issues/6850

Before the following PRs (i.e. in spinnaker < 1.30) `Deploy (Manifest)` stage was working
https://github.com/spinnaker/orca/pull/4322
https://github.com/spinnaker/orca/pull/4322
https://github.com/spinnaker/orca/pull/4322

the intent of this PR is to restore the previous behavior

Here are the pipelines

[pr-pipelines.zip](https://github.com/spinnaker/orca/files/12115106/pr-pipelines.zip)

PR #4568, opened 2023-10-17
fix(artifacts): Consider artifacts from stages when triggering from a pipeline stage

sergio-quintero:
Follow up to https://github.com/spinnaker/orca/pull/4489

Triggering a pipeline from Pipeline A using the [pipeline stage](https://spinnaker.io/docs/reference/pipeline/stages/#pipeline) where Pipeline B uses BakeManifestStage or DeployManifestStage which rely on expectedArtifacts is broken. [ArtifactUtils](https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java#L264-L275) is filtering the expectedArtifacts needed in stages. 

In 1.28, you were able to define expectedArtifacts in Pipeline B, and when you triggered from Pipeline A, Pipeline B was able to resolve the artifacts from Pipeline A. [Passing artifacts between pipelines](https://spinnaker.io/docs/reference/ref-artifacts/in-pipelines/#passing-artifacts-between-pipelines)

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
@@ -85,12 +85,24 @@ class DependentPipelineStarter implements ApplicationContextAware {
       it.expectedArtifactIds ?: []
     }
 
-    // we are following a similar approach as triggers above
+    // following a similar approach as triggers above
     // expectedArtifacts can be used in triggers and stages
-    // for now we identified DeployManifestStage
-    // in ResolveDeploySourceManifestTask using ManifestEvaluator.getRequiredArtifacts
-    def requiredArtifactIds = pipelineConfig.get("stages", []).collectMany {
-      it.requiredArtifactIds ?: []
+    def requiredArtifactIds = pipelineConfig.get("stages", []).collectMany { stage ->
+      def artifactIds = []
+
+      //DeployManifestStage in ResolveDeploySourceManifestTask using ManifestEvaluator.getManifestArtifact
+      artifactIds += stage.manifestArtifactId ?: []
+
+      // DeployManifestStage in ResolveDeploySourceManifestTask using ManifestEvaluator.getRequiredArtifacts
+      artifactIds += stage.requiredArtifactIds ?: []
+
+      // BakeManifestStage in CreateBakeManifestTask using BakeManifestContext for Kustomize
+      artifactIds += stage.inputArtifact?.id ?: []
+
+      // BakeManifestStage in CreateBakeManifestTask using BakeManifestContext for Helm
+      artifactIds += stage.inputArtifacts?.collect { it.id } ?: []
+
+      return artifactIds
     }
     expectedArtifactIds.addAll(requiredArtifactIds)
 
--- a/orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarterSpec.groovy
+++ b/orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarterSpec.groovy
@@ -22,6 +22,7 @@ import com.netflix.spectator.api.NoopRegistry
 import com.netflix.spinnaker.kork.artifacts.ArtifactTypes
 import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact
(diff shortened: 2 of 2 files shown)


jervi (approved):
Looks good to me! Tagging @xibz as well, since you also showed some interest in this area

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -85,12 +85,24 @@ class DependentPipelineStarter implements ApplicationContextAware {
...
+    // following a similar approach as triggers above
     // expectedArtifacts can be used in triggers and stages
-    // for now we identified DeployManifestStage
-    // in ResolveDeploySourceManifestTask using ManifestEvaluator.getRequiredArtifacts
-    def requiredArtifactIds = pipelineConfig.get("stages", []).collectMany {
-      it.requiredArtifactIds ?: []
+    def requiredArtifactIds = pipelineConfig.get("stages", []).collectMany { stage ->
+      def artifactIds = []
+
+      //DeployManifestStage in ResolveDeploySourceManifestTask using ManifestEvaluator.getManifestArtifact
```
xibz, line 93:
```suggestion
      // DeployManifestStage in ResolveDeploySourceManifestTask using ManifestEvaluator.getManifestArtifact
```
nit

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -85,12 +85,24 @@ class DependentPipelineStarter implements ApplicationContextAware {
...
+      artifactIds += stage.manifestArtifactId ?: []
+
+      // DeployManifestStage in ResolveDeploySourceManifestTask using ManifestEvaluator.getRequiredArtifacts
+      artifactIds += stage.requiredArtifactIds ?: []
+
+      // BakeManifestStage in CreateBakeManifestTask using BakeManifestContext for Kustomize
+      artifactIds += stage.inputArtifact?.id ?: []
+
+      // BakeManifestStage in CreateBakeManifestTask using BakeManifestContext for Helm
```
xibz, line 102: Are these comments accurate? What I mean is I understand that `Helm` bakes with `inputArtifacts` but other stages like `BakeCloudFoundry` also utilize this?

Can we make this more generic and say something more like

```
// Some bake manifest stages may populate the inputArtifacts array. So we need to look to for those ids as well
```

And we may be able to get rid of the other comments as they seem to be very specific but don't capture the whole picture

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -85,12 +85,24 @@ class DependentPipelineStarter implements ApplicationContextAware {
...
     }
 
-    // we are following a similar approach as triggers above
+    // following a similar approach as triggers above
     // expectedArtifacts can be used in triggers and stages
-    // for now we identified DeployManifestStage
-    // in ResolveDeploySourceManifestTask using ManifestEvaluator.getRequiredArtifacts
-    def requiredArtifactIds = pipelineConfig.get("stages", []).collectMany {
-      it.requiredArtifactIds ?: []
+    def requiredArtifactIds = pipelineConfig.get("stages", []).collectMany { stage ->
```
xibz, line 90: Do we need this variable? Can't we just use `expectedArtifactIds` here?

xibz (changes requested):
Looks good! I have some very minor nit comments

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
xibz: Maybe even explain why we are looking for these ids as opposed to the expected artifact. Would be good to just document here saying we dont know how consuming pipelines will utilize artifacts from the parent pipeline, so we rely on any input artifact as opposed to the expected artifact

edgarulg (approved):
LGTM! but before merging it let's address all comments :)

sergio-quintero:
Closing this PR because we found more stages with different field names that were not considered in this PR. Adding a list of fields in `DependentPipelineStarter` is not maintainable. Also, if someone creates a plugin with a new Stage that utilizes expectedArtifacts, they can define any field name which will not be included in this class. 

We'll send a new PR with a better fix for this issue.

PR #4570, opened 2023-10-24
fix(artifacts): separate resolve artifacts for trigger of type pipeline

nemesisOsorio:
Follow up to
* https://github.com/spinnaker/orca/pull/4489
* https://github.com/spinnaker/orca/pull/4568

Triggering a pipeline from Pipeline A using the [pipeline stage](https://spinnaker.io/docs/reference/pipeline/stages/#pipeline) where Pipeline B uses BakeManifestStage or DeployManifestStage which rely on expectedArtifacts is broken. [ArtifactUtils](https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java#L264-L275) is filtering the `expectedArtifacts` needed in stages.

In 1.28, you were able to define expectedArtifacts in Pipeline B, and when you triggered from Pipeline A, Pipeline B was able to resolve the artifacts from Pipeline A. [Passing artifacts between pipelines](https://spinnaker.io/docs/reference/ref-artifacts/in-pipelines/#passing-artifacts-between-pipelines)

We are partially reverting some changes from:
* https://github.com/spinnaker/orca/pull/4397
* https://github.com/spinnaker/orca/pull/4489
* https://github.com/spinnaker/orca/pull/4526

because they are not needed anymore. 

Fixes https://github.com/spinnaker/spinnaker/issues/6897

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
@@ -17,7 +17,6 @@
 package com.netflix.spinnaker.orca.pipeline.util;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
-import static java.lang.String.format;
 
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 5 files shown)


nemesisOsorio:
Closing in favor of https://github.com/spinnaker/orca/pull/4579

PR #4575, opened 2023-10-26
fix(artifacts): Parent and child pipeline artifact resolution

xibz:
This fix allows us to revert https://github.com/spinnaker/orca/pull/4397
since that makes retrieving bound artifacts actually more strict for
most use cases. So this commit addresses the root cause that that PR was
trying to implement.

When a parent pipeline triggers a child pipeline, the child pipeline should be able to
resolve incoming artifacts properly. Right now that does not happen
because `expectedArtifactIds` is needed in the trigger section,
otherwise no resolution will occur.

This change fixes the parent payload to include `expectedArtifactIds` by
viewing the child pipeline's top level expected artifacts and finding
the intersection between expected artifacts and artifacts to be sent.
With that intersection we can grab the expected artifact ids and
properly set the the necessary field to allow for child pipelines to
ingest the artifact.

@nemesisOsorio

Also, we can probably still utilize some of https://github.com/spinnaker/orca/pull/4570 and still revert the changes that were introduced due to not properly handling the root cause

### DIFF
--- a/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
+++ b/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
@@ -19,6 +19,8 @@ package com.netflix.spinnaker.orca.front50
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spectator.api.Id
 import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.kork.artifacts.model.Artifact
+import com.netflix.spinnaker.kork.artifacts.model.ExpectedArtifact
 import com.netflix.spinnaker.kork.exceptions.ConfigurationException
 import com.netflix.spinnaker.orca.api.pipeline.models.PipelineExecution
 import com.netflix.spinnaker.orca.api.pipeline.models.Trigger
@@ -102,7 +104,6 @@ class DependentPipelineStarter implements ApplicationContextAware {
       parameters           : [:],
       strategy             : suppliedParameters.strategy == true,
       correlationId        : "${parentPipeline.id}_${parentPipelineStageId}_${pipelineConfig.id}_${parentPipeline.startTime}".toString(),
-      expectedArtifactIds  : expectedArtifactIds.toSet().toList()
     ]
     /* correlationId is added so that two pipelines aren't triggered when a pipeline is canceled.
      * parentPipelineStageId is added so that a child pipeline (via pipeline stage)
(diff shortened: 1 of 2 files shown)


orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -133,6 +136,25 @@ class DependentPipelineStarter implements ApplicationContextAware {
...
+        // artifacts. So if the expected artifact does not match on any incoming
+        // artifact, we will simply not add the expected artifact ID to the
+        // array. This is very similar to how echo works.
+        def artifact = objectMapper.convertValue(it, Artifact)
+        return expectedArtifact.matches(artifact)
+      }.collect {
+        return expectedArtifact.id
+      }
+    } as String[])
+    trigger.expectedArtifactIds = expectedArtifactIds
```
sergio-quintero, line 156: Would it make sense to wrap this code under the condition when the trigger is from a pipeline stage? Maybe add it under line 130 where it checks `if (parentPipelineStageId != null)` ?

sergio-quintero: After reviewing this, we shouldn't move this code because if an expected artifact is added to the pipeline json without an artifact constraint in a trigger, it will not be able to resolve it.

sergio-quintero (approved):
LGTM

nemesisOsorio (approved):
can we please backport this all the way down to `1.30`?

nemesisOsorio:
@jervi can you please take a look?

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/DependentPipelineStarter.groovy
```diff
@@ -133,6 +136,25 @@ class DependentPipelineStarter implements ApplicationContextAware {
     // This is required for template source with jinja expressions
     trigger.artifacts = pipelineConfig.receivedArtifacts
 
+    // expectedArtifacts for triggers are defined at the top level of the
```
dbyron-sf, line 139: Trying to follow this, I started above with:
```
    def expectedArtifactIds = pipelineConfig.get("triggers", []).findAll {
      it.type == "pipeline" && it.pipeline == parentPipeline.pipelineConfigId
    } collectMany {
      it.expectedArtifactIds ?: []
    }
```
I'd love some help connecting what that block does to here.

dbyron-sf (approved):

jervi:
Sorry, I'm late to the game! I think this looks good as part of a much overdue refresh of the artifact handling in Spinnaker. Thank you for spending time on this, @xibz and @nemesisOsorio 🚀

link108:
@mergifyio backport release-1.30.x release-1.31.x release-1.32.x

PR #4579, opened 2023-10-30
fix(artifacts): Automated triggers with artifact constraints are broken if you have 2 or more of the same type

nemesisOsorio:
Follow up to
* https://github.com/spinnaker/orca/pull/4575

We are partially reverting some changes from:
* https://github.com/spinnaker/orca/pull/4397
* https://github.com/spinnaker/orca/pull/4489
* https://github.com/spinnaker/orca/pull/4526

because they are not needed anymore. 

Fixes https://github.com/spinnaker/spinnaker/issues/6897


xibz (changes requested):
Overall looks good.

The two boolean checks

```
|| artifact.isUseDefaultArtifact()
|| artifact.isUsePriorArtifact())
```

That was the only kept from the PR and doesn't have any tests. Can we add two tests for both of these booleans, please?

nemesisOsorio:
@jervi can you please help us to review this?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
```diff
@@ -230,38 +201,10 @@ public List<Artifact> getArtifactsForPipelineIdWithoutStageRef(
...
-    // Due to 8df68b79cf1 getBoundArtifactForStage now does resolution which
-    // can potentially return null artifact back, which will throw an exception
-    // for stages that expect a non-null value. Before this commit, when
-    // getBoundArtifactForStage was called, the method would just retrieve the
-    // bound artifact from the stage context, and return the appropriate
-    // artifact back. This change prevents tasks like CreateBakeManifestTask
-    // from working properly, if null is returned.
-    //
-    // reference: https://github.com/spinnaker/orca/pull/4397
-    triggers.stream()
```
xibz, line 253: So this PR addresses an earlier PR, but we can also revert it, because the earlier PR that was submitted only implemented a partial solution. I've talked some about this offline to @nemesisOsorio, but we can revert this now since it does break multiple triggers of the same time. We will need to implement a proper fix

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
```diff
@@ -230,38 +201,10 @@ public List<Artifact> getArtifactsForPipelineIdWithoutStageRef(
...
-        // expectedArtifactIds unless it is the expected trigger type that was
-        // triggered
-        //
-        // reference: https://github.com/spinnaker/orca/pull/4322
-        .filter(it -> trigger.getOrDefault("type", "").equals(it.get("type")))
-        .map(this::getExpectedArtifactIdsFromMap)
-        .forEach(expectedArtifactIdsListConcat::addAll);
-
-    final List<String> expectedArtifactIds = new ArrayList<>(expectedArtifactIdsListConcat);
+    List<?> expectedArtifactIds =
```
xibz, line 206: This concept is not what we should be doing. This now makes triggers of any type have a required parameter, `expectedArtifactIds` when this is just an echo concept. I will have a follow up that makes this cleaner and simpler since this is due to an earlier PR

xibz (approved):
LGTM! I mentioned some comments as there are still more work needing to be done around artifact resolution.

Mostly design considerations that were made need to be re-thought of.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java
jervi: Is the plan to create the proper fix in another PR?

xibz: Yea, Ill have a PR that should address the issues with artifact resolution. I think Im going to write a proposal first, since some of the ideas I have require some refactors. So want to make sure we get that write in design first :)

dbyron-sf (approved):

link108:
@mergifyio backport release-1.30.x release-1.31.x release-1.32.x

PR #4584, opened 2023-11-02
refactor(bakery): fix incompatible generic type issue faced during upgrade to springboot 2.5.15

j-sandy:
While upgrading springboot 2.5.15, encounter below error in orca-bakery module:
```
startup failed:
/orca/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy: 177: [Static type checking] - Incompatible generic argument types. Cannot assign java.util.List <java.lang.Object> to: java.util.List <String>
 @ line 177, column 43.
   <String> distinctImageNames = relatedBak
                                 ^

1 error

> Task :orca-bakery:compileGroovy FAILED
```
To fix this issue use explicit object reference of `List<StageExecution>` to hold and operate on the object stream, and `List` to capture the filtered generic objects.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
@@ -155,7 +155,7 @@ class BakeStage implements StageDefinitionBuilder {
 
     @Nonnull
     TaskResult execute(@Nonnull StageExecution stage) {
-      def relatedBakeStages = stage.execution.stages.findAll {
+      List<StageExecution> relatedBakeStages = stage.execution.stages.findAll {
         it.type == PIPELINE_CONFIG_TYPE && stage.id == it.parentStageId
       }
 
@@ -174,7 +174,7 @@ class BakeStage implements StageDefinitionBuilder {
 
       if (failOnImageNameMismatchEnabled()) {
         // find distinct image names in bake stages that are actually related to the stage passed into the task
-        List<String> distinctImageNames = relatedBakeStages
+        List<Object> distinctImageNames = relatedBakeStages
           .findAll { childStage -> childStage.parentStageId == stage.id && childStage.context.imageName }
           .stream()
           .map { childStage -> childStage.context.imageName }


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -174,7 +174,7 @@ class BakeStage implements StageDefinitionBuilder {
 
       if (failOnImageNameMismatchEnabled()) {
         // find distinct image names in bake stages that are actually related to the stage passed into the task
-        List<String> distinctImageNames = relatedBakeStages
+        List distinctImageNames = relatedBakeStages
```
dbyron-sf, line 177: Can't this be
```suggestion
        List<StageExecution> distinctImageNames = relatedBakeStages
```
or at least `List<Object>` ?  I'd like to avoid using a raw type.

j-sandy: Since this return list of object, I will make it as `List<Object>`.

j-sandy: Done

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy
```diff
@@ -174,7 +174,7 @@ class BakeStage implements StageDefinitionBuilder {
 
       if (failOnImageNameMismatchEnabled()) {
         // find distinct image names in bake stages that are actually related to the stage passed into the task
-        List<String> distinctImageNames = relatedBakeStages
+        List<Object> distinctImageNames = relatedBakeStages
           .findAll { childStage -> childStage.parentStageId == stage.id && childStage.context.imageName }
           .stream()
           .map { childStage -> childStage.context.imageName }
```
dbyron-sf, line 180: Aaah, ok.  Looks like we're running into the snag where [StageExecution.getContext()](https://github.com/spinnaker/orca/blob/2e8ebf811073b3c2100712e16a2cf68fb675ec00/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/StageExecution.java#L85) returns a Map<String, Object>, so unless we cast `childStage.context.imageName` to a String, Object is the best we can do.

dbyron-sf (approved):

PR #4593, opened 2023-11-13
refactor(igor): OortService - Cleaned up the RetrofitError catch block in GitCommitsTask as it is no more relevant

Pranav-b-7:
As part of this PR , the target is to cleanup the usage of RetrofitError in GitCommitsTask. 

- The Retrofit configuration in ClouddriverConfiguration class is already set with the error handler : SpinnakerRetrofitErrorHandler for the OortService client as part of this commit : https://github.com/spinnaker/orca/blob/7e1e74f90b5d9ab786d2a5efd18100a4a9a794c7/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java#L212
- As part of the commit : https://github.com/spinnaker/orca/pull/4528/commits/781ad44998b4da43139f807ca5e4f53229f49345 the OortService client exception handling is implemented in GitCommitsTask class and will use SpinnakerRetrofitErrorHandler if any exceptions happen.

The catch block with RetrofitError in GitCommitsTask class will now be irrelevant from the above 2 commits , hence its cleaned up.

### DIFF
--- a/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
+++ b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
@@ -34,7 +34,6 @@ import com.netflix.spinnaker.orca.pipeline.model.PipelineTrigger
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
-import retrofit.RetrofitError
 
 @Slf4j
 @Component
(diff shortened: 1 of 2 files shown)


orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy
```diff
@@ -122,32 +121,13 @@ class GetCommitsTask implements DiffTask {
       }
 
       return TaskResult.builder(ExecutionStatus.SUCCEEDED).context(outputs).build()
-    } catch (RetrofitError e) {
```
dbyron-sf, line 125: There's a call to [getRepoInfo](https://github.com/spinnaker/orca/blob/b013bc8ec4de2356e0337f6ee57adacbf9776c71/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/GetCommitsTask.groovy#L245) inside this try/catch block.  getRepoInfo uses a Front50Service bean to make a retrofit call.. Currently, that [Front50Service bean](https://github.com/spinnaker/orca/blob/b013bc8ec4de2356e0337f6ee57adacbf9776c71/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/config/Front50Configuration.groovy#L73) doesn't use SpinnakerRetrofitErrorHandler, so I believe the code to catch RetrofitErrors needs to stay.

Pranav-b-7: I too have verified the code.. this Front50Service bean doesnt use SpinnakerRetrofitErrorHandler , so this PR is no more relevant and hence I am closing this PR

PR #4594, opened 2023-11-13
fix(databaseChangeLog): fix for 'addAfterColumn is not allowed on postgresql'

kirangodishala:
Liquibase 4.13.0 introduced a change which breaks processing orca's changelog files. Liquibase 4.24.0 has a workaround for this issue.

Check the following kork PR for the background: https://github.com/spinnaker/kork/pull/1117 - upgrade liquibase to 4.24.0

This PR requires above kork PR to be merged first.

The following snippet is needed under `databaseChangeLog` for liquibase to ignore the `afterColumn` construct for postgresql which doesn't support the feature. 
```yaml
- removeChangeSetProperty:
    change: addColumn
    dbms: postgresql
    remove: afterColumn
```


dbyron-sf:
included in https://github.com/spinnaker/orca/pull/4601.

PR #4597, opened 2023-11-16
fix(clouddriver): change exception handling in ClouddriverClearAltTablespaceTask

Pranav-b-7:
change exception handling in ClouddriverClearAltTablespaceTask to check for Spinnaker*Exception since, as of https://github.com/spinnaker/orca/pull/4528, CloudDriverCacheService uses SpinnakerRetrofitErrorHandler.

### DIFF
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTask.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTask.java
@@ -2,6 +2,7 @@ package com.netflix.spinnaker.orca.clouddriver.tasks.cache;
 
 import static java.util.Collections.emptyList;
 
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException;
 import com.netflix.spinnaker.orca.api.pipeline.Task;
 import com.netflix.spinnaker.orca.api.pipeline.TaskResult;
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus;
@@ -17,8 +18,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
-import retrofit.RetrofitError;
-import retrofit.mime.TypedByteArray;
 
 @Component
 public class ClouddriverClearAltTablespaceTask implements Task {
@@ -48,17 +47,11 @@ public class ClouddriverClearAltTablespaceTask implements Task {
           result.getOrDefault("tables", emptyList()));
 
       return TaskResult.builder(ExecutionStatus.SUCCEEDED).context(result).build();
-    } catch (RetrofitError e) {
+    } catch (SpinnakerServerException e) {
       Map<String, Object> output = new HashMap<>();
       List<String> errors = new ArrayList<>();
 
-      if (e.getResponse() != null && e.getResponse().getBody() != null) {
-        String error = new String(((TypedByteArray) e.getResponse().getBody()).getBytes());
-        log.error("Failed clearing clouddriver table namespace: {}", error, e);
-        errors.add(error);
-      } else {
-        errors.add(e.getMessage());
-      }
+      errors.add(e.getMessage());
       output.put("errors", errors);
       return TaskResult.builder(ExecutionStatus.TERMINAL).context(output).build();
     }
--- a/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTaskTest.java
+++ b/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTaskTest.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2023 OpsMx, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTask.java
```diff
@@ -48,18 +47,21 @@ public TaskResult execute(@Nonnull StageExecution stage) {
...
+        Map<String, Object> errorBody = e.getResponseBody();
+        log.error("Failed clearing clouddriver table namespace: {}", errorBody, e);
+        output.put("errors", errorBody);
       } else {
-        errors.add(e.getMessage());
+        output.put("errors", e.getMessage());
       }
-      output.put("errors", errors);
+      return TaskResult.builder(ExecutionStatus.TERMINAL).context(output).build();
+    } catch (SpinnakerServerException e) {
```
dbyron-sf, line 61: instanceof is never great, but in this case having one `catch (SpinnakerServerException e) {` block and handling the SpinnakerHttpException case within it is I think overall simpler.  No need to declare outputs above, and we reduce some duplication.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTask.java
```diff
@@ -48,18 +47,21 @@ public TaskResult execute(@Nonnull StageExecution stage) {
...
-      List<String> errors = new ArrayList<>();
+    } catch (SpinnakerHttpException e) {
+      output = new HashMap<>();
 
-      if (e.getResponse() != null && e.getResponse().getBody() != null) {
-        String error = new String(((TypedByteArray) e.getResponse().getBody()).getBytes());
-        log.error("Failed clearing clouddriver table namespace: {}", error, e);
-        errors.add(error);
+      if (e.getResponseBody() != null) {
+        Map<String, Object> errorBody = e.getResponseBody();
```
dbyron-sf, line 54: nit: assign this above the if to only call `e.getResponseBody()` once.

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTaskTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+  @Test
+  void testSpinnakerHttpExceptionHandlingWithErrorBody() throws JsonProcessingException {
+
+    String namespace = "test-namespace";
+    HashMap<String, String> errorJsonResponse = new HashMap<>();
+    StageExecution stageExecution = new StageExecutionImpl();
+    Map<String, Object> context = new HashMap<>();
+
+    errorJsonResponse.put("error", "Failure, as the JSON response is bad");
```
dbyron-sf, line 67: Thanks for adding tests!

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTask.java
```diff
@@ -48,18 +47,21 @@ public TaskResult execute(@Nonnull StageExecution stage) {
...
+        log.error("Failed clearing clouddriver table namespace: {}", errorBody, e);
+        output.put("errors", errorBody);
```
dbyron-sf, line 56: Before this PR the errors property contained the response body as a String.  The way the code is now, the errors property is either a String or a Map.  I think this makes things tough on consumers.  We're stuck making some kind of behavior change since we no longer have access to the entire response body.  Seems like it's sufficient/less disruptive to use the exception message.  At least the type is consistently a String, which is what it was before.

As I look at the [underlying clouddriver endpoint](https://github.com/spinnaker/clouddriver/blob/44124e218892abce09a0782b826913e62385a7d9/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/controllers/CatsSqlAdminController.kt#L36) I don't see any special exception handling, so I expect all the interesting info is in the exception message.

Pranav-b-7: FYI - Before this PR the errors property contained either a String or List of Strings , it was not consistent even before.

Pranav-b-7: FYI - Before this PR the errors property contained List of Strings. However , I got the context . 
Will do the required changes to make the property values consistent

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTask.java
Pranav-b-7: Removed the catch block which handles SpinnakerHttpException and now it has just one catch block to handle the SpinnakerServerException which will now simply populates the Task context property with the exception message.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTask.java
Pranav-b-7: This suggestion is no more relevant as the response body will no more be considered wrt the latest commit

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/cache/ClouddriverClearAltTablespaceTask.java
Pranav-b-7: To avoid the breakage in consumers of the pipeline execution context, I have modified the errors property to hold List of Strings which retains the state like it was before.

dbyron-sf (approved):

PR #4608, opened 2023-11-22
feat(deploymentmonitor): add SpinnakerRetrofitErrorHandler to DeploymentMonitorService

Pranav-b-7:
add SpinnakerRetrofitErrorHandler to DeploymentMonitorService and convert corresponding RetrofitError exception handling to use Spinnaker*Exception

Note that there's a behaviour change here.  Before this PR the entire body of error responses, status and headers appears in a log message.  With Spinnaker*Exception, the status is separated and will be printed as part of the stack trace itself.

In case of any failure while reading the error response details , the log would simply print empty String values for headers and response body.

This behavioural changes are demonstrated using the below test cases : 

- https://github.com/spinnaker/orca/blob/e21a5cc06971cddde34d9b3ac59264b7aa31ea86/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTaskTest.java#L78C3-L81C34  .
- https://github.com/spinnaker/orca/blob/e21a5cc06971cddde34d9b3ac59264b7aa31ea86/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTaskTest.java#L123C1-L126C17 .


orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -137,12 +133,12 @@ public long getDynamicTimeout(StageExecution stage) {
...
     try {
       return executeInternal(stage, monitorDefinition);
-    } catch (RetrofitError e) {
+    } catch (SpinnakerServerException e) {
       log.warn(
           "HTTP Error encountered while talking to {}->{}, {}}",
           monitorDefinition,
           e.getUrl(),
-          getRetrofitLogMessage(e.getResponse()),
+          getRetrofitLogMessage(e),
```
dbyron-sf, line 141: It's pretty tempting to use `e.getMessage()` here, or to ditch it completely since we're already logging the entire exception.  This looks dynamic enough (i.e. we're not talking about an endpoint elsewhere in spinnaker) that the headers could actually be helpful.  getRetrofitLogMessage collects status and reason that are already in e.getMessage.  For SpinnakerHttpException, the url is there too.  Perhaps there's some work to do to include the URL in e.getMessage() for SpinnakerServerException.

After staring at this for a while, I'd like to:
* rename getRetrofitLogMessage to something that doesn't have retrofit in the name
* change this method to only return body and headers since status/reason are already handled.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -267,22 +263,21 @@ private MonitoredDeployStageData getStageContext(StageExecution stage) {
...
+  private String getRetrofitLogMessage(SpinnakerServerException spinnakerException) {
+    if (!(spinnakerException instanceof SpinnakerHttpException)) {
       return "<NO RESPONSE>";
     }
 
+    SpinnakerHttpException httpException = (SpinnakerHttpException) spinnakerException;
     String body = "";
     String status = "";
     String headers = "";
+    ObjectMapper objectMapper = new ObjectMapper();
```
dbyron-sf, line 275: Let's only allocate one of these instead of making a new one every time this method is called.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -267,22 +263,21 @@ private MonitoredDeployStageData getStageContext(StageExecution stage) {
...
-          CharStreams.toString(
-              new InputStreamReader(response.getBody().in(), StandardCharsets.UTF_8));
-      headers =
-          response.getHeaders().stream().map(Header::toString).collect(Collectors.joining("\n"));
+      status = String.format("%d (%s)", httpException.getResponseCode(), httpException.getReason());
+      body = objectMapper.writeValueAsString(httpException.getResponseBody());
+      headers = httpException.getHeaders().toString();
     } catch (Exception e) {
       log.error(
           "Failed to fully parse retrofit error while reading response from deployment monitor", e);
```
dbyron-sf, line 283:
```suggestion
          "Failed to fully parse http response while reading response from deployment monitor", e);
```
and, is it correct to only wrap the objectMapper.writeValueAsString in the try/catch?

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -267,22 +263,21 @@ private MonitoredDeployStageData getStageContext(StageExecution stage) {
...
+    ObjectMapper objectMapper = new ObjectMapper();
 
     try {
-      status = String.format("%d (%s)", response.getStatus(), response.getReason());
-      body =
-          CharStreams.toString(
-              new InputStreamReader(response.getBody().in(), StandardCharsets.UTF_8));
-      headers =
-          response.getHeaders().stream().map(Header::toString).collect(Collectors.joining("\n"));
+      status = String.format("%d (%s)", httpException.getResponseCode(), httpException.getReason());
```
dbyron-sf, line 278: let's ditch this since it's already in the exception message.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -267,22 +263,21 @@ private MonitoredDeployStageData getStageContext(StageExecution stage) {
...
+      body = objectMapper.writeValueAsString(httpException.getResponseBody());
```
dbyron-sf, line 279: Need a null check for httpException.getResponseBody.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: - Renamed the method getRetrofitLogMessage() to getErrorMessage().
- Modified the method getErrorMessage() to return the concatenated String values of e.getMessage(), headers and body.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: - Changed the variable initialisation of ObjectMapper to instance variable of the class.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: Reading the status code from the exception instance is now removed and now we are leftover with 2 lines, reading the body and headers, So wrapping the line : httpException.getHeaders().toString(); in try/catch is anyway not gonna harm us and the statements outside the try/catch seems to less risky. 

hence keeping it the same way like it was before, and just going with the change in error log

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: Added null check with the help of Optional<> class in java

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -267,28 +264,24 @@ private MonitoredDeployStageData getStageContext(StageExecution stage) {
...
+              .orElse("");
+      return String.format("headers: %s\nresponse body: %s", httpException.getHeaders(), body);
     } catch (Exception e) {
       log.error(
-          "Failed to fully parse retrofit error while reading response from deployment monitor", e);
+          "Failed to fully parse http response while reading response from deployment monitor", e);
     }
 
-    return String.format("status: %s\nheaders: %s\nresponse body: %s", status, headers, body);
+    return "headers: \nresponse body: ";
```
dbyron-sf, line 284: I think we still need String.format here, to include the actual headers and response body.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
dbyron-sf: I'm trying to make sure the error message is correct.  If we get an exception dealing with the headers, then the message isn't correct.  It's not really about reading the response either at this point...it's about serializing the response body....or something...

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -267,28 +264,24 @@ private MonitoredDeployStageData getStageContext(StageExecution stage) {
...
-      headers =
-          response.getHeaders().stream().map(Header::toString).collect(Collectors.joining("\n"));
+      String body =
+          Optional.ofNullable(objectMapper.writeValueAsString(httpException.getResponseBody()))
+              .orElse("");
+      return String.format("headers: %s\nresponse body: %s", httpException.getHeaders(), body);
     } catch (Exception e) {
       log.error(
-          "Failed to fully parse retrofit error while reading response from deployment monitor", e);
+          "Failed to fully parse http response while reading response from deployment monitor", e);
```
Pranav-b-7, line 281:
```suggestion
          "Exception occurred while parsing the http error response details", e);
```

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: I dont understand what is the need to format the String here. If you see the code before the PR, 
We had the initialisation like below:
```
    String body = "";
    String headers = "";
```

In case of any exceptions while parsing the response details the default empty values are returned, which now I have simplified and doing the same thing.

The actual headers and body are parsed and returned inside the try block itself in case of success scenario.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: Please refer to this suggeston - https://github.com/spinnaker/orca/pull/4608/files/4c608efdc4f4783fd240be38142f4e1d3e374955#r1403955017

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
dbyron-sf: Aaah, I see.  I think it'd be better to indicate some kind of error than possibly misleading people into thinking there were no headers/body.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/EvaluateDeploymentHealthTaskSpec.groovy
```diff
@@ -45,7 +52,7 @@ class EvaluateDeploymentHealthTaskSpec extends Specification {
     given:
     def monitorServiceStub = Stub(DeploymentMonitorService) {
       evaluateHealth(_) >> {
-        throw RetrofitError.networkError("url", new IOException())
+        throw new SpinnakerServerException(RetrofitError.networkError("url", new IOException()))
```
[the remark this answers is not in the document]
dbyron-sf: It's possible, but while the code is still using retrofit1, wrapping RetrofitError is the most similar to what happens when the code runs for real.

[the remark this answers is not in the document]
dbyron-sf: Agreed.  At least we can remove the references from e.g. src/main / pave the way to switch to retrofit2.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: The indication about the error will be printed in the log statement in the above catch block : https://github.com/spinnaker/orca/blob/e21a5cc06971cddde34d9b3ac59264b7aa31ea86/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java#L279C7-L281C6

This should indicate the users on why the headers and response bodies are empty

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: I have modified the log message to : https://github.com/spinnaker/orca/blob/e21a5cc06971cddde34d9b3ac59264b7aa31ea86/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java#L280C6-L280C91 . 
Replaced parse with read

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTaskTest.java
```diff
@@ -0,0 +1,186 @@
...
+        new DeploymentMonitorServiceProvider(
+            okClient, retrofitLogLevel, requestInterceptor, deploymentMonitorDefinitions);
+    monitoredDeployBaseTask =
+        new MonitoredDeployBaseTask(deploymentMonitorServiceProvider, new NoopRegistry());
+  }
+
+  @Test
+  @DisplayName(
+      "should parse/read the error response and return the http status, http headers and http error body when retrofit httpError/conversionError has occurred")
+  public void test1() {
```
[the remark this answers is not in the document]
dbyron-sf: Looks like it could help to split this into two methods -- one with httpError and another with conversionError.

kirangodishala (commented):
@Pranav-b-7 - Since this PR is all about adding tests, to make the title simple to understand, I would prefer something like this: `test(MonitoredDeployBaseTask): add tests to verify log message for different types of RetrofitError`

orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTaskTest.java
```diff
@@ -0,0 +1,211 @@
...
+    RetrofitError httpError =
+        RetrofitError.httpError(
+            "https://foo.com/deployment/evaluateHealth", response, new JacksonConverter(), null);
+
+    String logMessageOnHttpError =
+        monitoredDeployBaseTask.getRetrofitLogMessage(httpError.getResponse());
+    String status = HttpStatus.BAD_REQUEST.value() + " (" + HttpStatus.BAD_REQUEST.name() + ")";
+    String body = "{\"error\":\"404 - Resource not found\"}";
+
+    assertEquals(
```
dbyron-sf, line 101: I'd like to start using `org.assertj.core.api.Assertions.assertThat` instead of the jupiter assertions.  It not in this PR, then let's do it going forward.

dbyron-sf (approved):

PR #4614, opened 2023-12-08
refactor(deploymentmonitor): Configure SpinnakerRetrofitErrorHandler for DeploymentMonitorService

Pranav-b-7:
### Changes Made

- 	Configured the error handler for the Retrofit client, DeploymentMonitorService, with SpinnakerRetrofitErrorHandler.
- 	Modified catch blocks with RetrofitError to use Spinnaker*Exception, aligning with the upcoming Retrofit version upgrade to retrofit2.x.

### Purpose
This PR represents foundational work for upgrading the Retrofit version to retrofit2.x, paving the way for future improvements and compatibility.
### Behavioural Changes

- The significant modification involves updating catch blocks, resulting in a change in the format of logger messages. This aligns with the anticipated modifications discussed in the related PR: https://github.com/spinnaker/orca/pull/4608 , which talks about the broader context related to this change.
- Additionally, when the error response body is null, an appropriate error message will be printed in the logger.

### DIFF
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
@@ -16,10 +16,12 @@
 
 package com.netflix.spinnaker.orca.clouddriver.tasks.monitoreddeploy;
 
-import com.google.common.io.CharStreams;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.config.DeploymentMonitorDefinition;
 import com.netflix.spinnaker.kork.annotations.VisibleForTesting;
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException;
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException;
 import com.netflix.spinnaker.orca.api.pipeline.RetryableTask;
 import com.netflix.spinnaker.orca.api.pipeline.TaskResult;
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus;
@@ -31,19 +33,13 @@ import com.netflix.spinnaker.orca.deploymentmonitor.models.DeploymentStep;
 import com.netflix.spinnaker.orca.deploymentmonitor.models.EvaluateHealthResponse;
 import com.netflix.spinnaker.orca.deploymentmonitor.models.MonitoredDeployInternalStageData;
 import com.netflix.spinnaker.orca.deploymentmonitor.models.StatusExplanation;
-import java.io.InputStreamReader;
-import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import retrofit.RetrofitError;
-import retrofit.client.Header;
-import retrofit.client.Response;
 
 public class MonitoredDeployBaseTask implements RetryableTask {
   static final int MAX_RETRY_COUNT = 3;
@@ -52,6 +48,7 @@ public class MonitoredDeployBaseTask implements RetryableTask {
   private DeploymentMonitorServiceProvider deploymentMonitorServiceProvider;
   private final Map<EvaluateHealthResponse.NextStepDirective, String> summaryMapping =
       new HashMap<>();
+  ObjectMapper objectMapper = new ObjectMapper();
 
   MonitoredDeployBaseTask(
       DeploymentMonitorServiceProvider deploymentMonitorServiceProvider, Registry registry) {
@@ -138,12 +135,12 @@ public class MonitoredDeployBaseTask implements RetryableTask {
 
     try {
       return executeInternal(stage, monitorDefinition);
-    } catch (RetrofitError e) {
+    } catch (SpinnakerServerException e) {
       log.warn(
           "HTTP Error encountered while talking to {}->{}, {}}",
           monitorDefinition,
           e.getUrl(),
-          getRetrofitLogMessage(e.getResponse()),
+          getErrorMessage(e),
           e);
 
       return handleError(stage, e, true, monitorDefinition);
@@ -269,28 +266,26 @@ public class MonitoredDeployBaseTask implements RetryableTask {
   }
 
   @VisibleForTesting
-  String getRetrofitLogMessage(Response response) {
-    if (response == null) {
+  String getErrorMessage(SpinnakerServerException spinnakerException) {
+    if (!(spinnakerException instanceof SpinnakerHttpException)) {
       return "<NO RESPONSE>";
     }
 
-    String body = "";
-    String status = "";
-    String headers = "";
+    SpinnakerHttpException httpException = (SpinnakerHttpException) spinnakerException;
 
     try {
-      status = String.format("%d (%s)", response.getStatus(), response.getReason());
-      body =
-          CharStreams.toString(
-              new InputStreamReader(response.getBody().in(), StandardCharsets.UTF_8));
-      headers =
-          response.getHeaders().stream().map(Header::toString).collect(Collectors.joining("\n"));
+      String body = "";
+      if (httpException.getResponseBody() != null) {
+        body = objectMapper.writeValueAsString(httpException.getResponseBody());
+      } else {
+        body = "Failed to serialize the error response body";
+      }
+      return String.format("headers: %s\nresponse body: %s", httpException.getHeaders(), body);
     } catch (Exception e) {
-      log.error(
-          "Failed to fully parse retrofit error while reading response from deployment monitor", e);
+      log.error("Failed to fully serialize http error response details", e);
     }
 
-    return String.format("status: %s\nheaders: %s\nresponse body: %s", status, headers, body);
+    return "headers: \nresponse body: ";
   }
 
   private boolean shouldFailOnError(StageExecution stage, DeploymentMonitorDefinition definition) {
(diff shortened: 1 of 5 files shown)


orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/EvaluateDeploymentHealthTaskSpec.groovy
```diff
@@ -45,7 +52,7 @@ class EvaluateDeploymentHealthTaskSpec extends Specification {
     given:
```
kirangodishala, line 52: though underlying exceptions at present are retrofit, wouldn't it be meaningful to change  "retrofit errors" to SpinnakerServerExceptions in the test name?

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -269,28 +266,24 @@ private MonitoredDeployStageData getStageContext(StageExecution stage) {
...
 
     try {
-      status = String.format("%d (%s)", response.getStatus(), response.getReason());
-      body =
-          CharStreams.toString(
-              new InputStreamReader(response.getBody().in(), StandardCharsets.UTF_8));
-      headers =
-          response.getHeaders().stream().map(Header::toString).collect(Collectors.joining("\n"));
+      String body =
+          Optional.ofNullable(objectMapper.writeValueAsString(httpException.getResponseBody()))
```
dbyron-sf, line 278: The null check doesn't quite seem in the right place here.  It's `httpException.getResponseBody()` that could be null here.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -269,28 +266,24 @@ private MonitoredDeployStageData getStageContext(StageExecution stage) {
...
+              .orElse("");
+      return String.format("headers: %s\nresponse body: %s", httpException.getHeaders(), body);
```
dbyron-sf, line 280: I'm somehow nervous about including an empty body when there might not be one.  If the response was present, but wasn't json, I'd like to make that clear somehow.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -269,28 +266,24 @@ private MonitoredDeployStageData getStageContext(StageExecution stage) {
...
     } catch (Exception e) {
       log.error(
-          "Failed to fully parse retrofit error while reading response from deployment monitor", e);
+          "Failed to fully parse http response while reading response from deployment monitor", e);
```
dbyron-sf, line 283: I don't think there's any parsing here, nor any response reading.  We're serializing some json.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/EvaluateDeploymentHealthTaskSpec.groovy
```diff
@@ -198,6 +205,50 @@ class EvaluateDeploymentHealthTaskSpec extends Specification {
     false              | null             || ExecutionStatus.FAILED_CONTINUE
   }
 
+  def "should return status as RUNNING when SpinnakerHttpException is thrown"() {
```
dbyron-sf, line 208: Possible to add this test before the code change so we can see any change in behavior?

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: Addressed the null check for `httpException.getResponseBody()` .

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: Modified the log statement to **Failed to fully read http error response details**

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: Even before this PR, the code would still return empty body headers and status, in case of any parse/read failures.

If we are planning to change this behaviour, I think we have to add a test case before this modifications to demonstrate the behavioural change.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/EvaluateDeploymentHealthTaskSpec.groovy
Pranav-b-7: I think there is no change in the behaviour here. This test case is to verify if the execute() method returns status as RUNNING when SpinnakerHttpException is thrown. This was the same case even with RetrofitError as well.

dbyron-sf: Without having the test before the code change, it's hard to see that.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
```diff
@@ -269,28 +266,24 @@ private MonitoredDeployStageData getStageContext(StageExecution stage) {
...
-          response.getHeaders().stream().map(Header::toString).collect(Collectors.joining("\n"));
+      String body = "";
+      if (httpException.getResponseBody() != null) {
+        body = objectMapper.writeValueAsString(httpException.getResponseBody());
+      }
+      return String.format("headers: %s\nresponse body: %s", httpException.getHeaders(), body);
     } catch (Exception e) {
-      log.error(
-          "Failed to fully parse retrofit error while reading response from deployment monitor", e);
+      log.error("Failed to fully read http error response details", e);
```
dbyron-sf, line 283:
```suggestion
      log.error("Failed to serialize http error response details", e);
```

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: A new specific exception handler is introduced to manage JsonProcessing during the serialization of the HTTP error body

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/EvaluateDeploymentHealthTaskSpec.groovy
Pranav-b-7: With reference to this comment : https://github.com/spinnaker/orca/pull/4617#discussion_r1434434134 , moved this test case to a separate commit for a better understanding .

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: The error log is now is now reword to : "Failed to serialize http error response details"

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
dbyron-sf: I'm talking about something different.  With SpinnakerHttpException, getResponseBody is null if the response isn't json whereas RetrofitError provided access to the response body in this case.

orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/EvaluateDeploymentHealthTaskSpec.groovy
```diff
@@ -198,6 +203,50 @@ class EvaluateDeploymentHealthTaskSpec extends Specification {
     false              | null             || ExecutionStatus.FAILED_CONTINUE
   }
 
+  def "should return status as RUNNING when Retrofit http error is thrown"() {
```
dbyron-sf: Since the code changes are happening in MonitoredDeployBaseTask, is it possible to add additional coverage to MonitoredDeployBaseTaskTest?  Is there anything specific to EvaluateDeploymentHealthTaskSpec that's important? 
 As well, MonitoredDeployBaseTaskTest is java and I'd much rather add more java code than groovy.

Pranav-b-7: The EvaluateDeploymentHealthTask is one of the child class of MonitoredDeployBaseTask . 
The executeInternal() method is overridden in this child class , and the there is also a call to evaluateHealth API in this method.
This API call is what we are mocking in the test case. So, I think this test case is more specific to EvaluateDeploymentHealthTask. 
Moving this to MonitoredDeployBaseTaskTest would make it more complex , because it requires lots of mocking

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/monitoreddeploy/MonitoredDeployBaseTask.java
Pranav-b-7: Changes added : when the error response body is null, an appropriate error message will be printed in the logger.

PR #4615, opened 2023-12-11
feat(webhooks): Hide sensitive headers in custom webhook stages

vjda:
This partially fixes https://github.com/spinnaker/spinnaker/issues/2787

Allow Spinnaker administrators to set up custom headers in preconfigured webhook stages to be hidden in pipeline context and task/stages messages passed through echo service.

This is useful for invoking endpoints that require some headers to authenticate, and we don't want to expose those credentials in the pipeline context to prevent any user with permissions on those pipeline executions from being able to make use of them.

Add a new config key `sensitiveHeaders` to exclude headers from being passed through pipeline context and task/stage messages.

For instance:

```yaml
webhook:
  preconfigured:
    - label: Some Webhook
      description: This is a webhook stage, but it appears as a native stage in Spinnaker
      type: customWebhook
      url: https://my.webhook.com/event
      customHeaders:
        Accept:
          - application/json
        Content-Type:
          - application/json
      sensitiveHeaders: # Sensitive headers will be hidden in the pipeline context and logs
        Authorization:
          - Basic dXNlcm5hbWU6cGFzc3dvcmQ=
      method: GET
```


vjda:
could anyone from @spinnaker/reviewers check this PR?

PR #4618, opened 2023-12-14
feat(jenkins): Enable Jenkins job triggers for jobs in sub-folders

ciurescuraul:
### Description
When defining a Jenkins job in a sub-folder, the path contains slashes. Consequently, the Spring framework is not capable of accurately matching the request.

### Problem
The trigger request fails because it is attempting to initiate a job that is not found. 

### Solution
A feature flag has been added. This flag influences if we should call the existing endpoint (which accepts the job name as a path variable) or the new one (which accepts the job name as a query parameter).

### Before
image

### After
image

### Related PR's
Echo: https://github.com/spinnaker/echo/pull/1373
Igor: https://github.com/spinnaker/igor/pull/1204


ovidiupopa07 (approved):
LGTM

orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/IgorService.java
```diff
@@ -66,6 +66,12 @@ Map<String, Object> getBuild(
       @Path("master") String master,
       @Path(encode = false, value = "job") String job);
 
+  @GET("/builds/status/{buildNumber}/{master}")
+  Map<String, Object> getBuildWithJobAsQueryParam(
+      @Path("buildNumber") Integer buildNumber,
+      @Path("master") String master,
+      @Query(encodeValue = false, value = "job") String job);
```
jasonmcintosh, line 73: One concern here is on the encodeValue... if there are any kinda characters in the job name, I could see this not working UNLESS they're encoded.  Wouldn't think "job" but... it's a caution flag.

jasonmcintosh (approved):

ovidiupopa07:
@Mergifyio backport release-1.28.x release-1.30.x release-1.31.x release-1.32.x release-1.33.x

PR #4620, opened 2023-12-21
feat(sql): conditionally compress large pipeline execution bodies

dbyron-sf:
with the following new configuration properties and their defaults.
```
execution-repository:
  sql:
    compression:
      enabled: false
      compressionMode: read-write # read-only also valid
      bodyCompressionThreshold: 1024
      compressionType: "ZLIB"
```
Perhaps obvious, but set execution-repository.sql.compression.enabled to true to enable compression.

We've been running this with `bodyCompressionThreshold: 1048576` for quite some time now.

Without this, some of our pipelines were big enough to exceed the max column width.

This at least partially addresses https://github.com/spinnaker/spinnaker/issues/6159 and https://github.com/spinnaker/spinnaker/issues/6006.


orca-sql/src/main/kotlin/com/netflix/spinnaker/config/ExecutionCompressionProperties.kt
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2018 Netflix, Inc.
```
mattgogerly, line 2: Salesforce?

mattgogerly (approved):

orca-sql/src/main/kotlin/com/netflix/spinnaker/config/ExecutionCompressionProperties.kt
dbyron-sf: Whoops, thanks. https://github.com/spinnaker/orca/pull/4620/commits/3b22a918cffdf6fb1491eb00c3110a570befcb34

jasonmcintosh (approved):
Long term refactor goals
* Fix the debug statements to not do string concat on every operation
* Change the lib to use commons-compression stuff 
but future enhancements, looks good as a good start!  REMINDER for those on this, optionally there are things you can do at the driver layer to enable compression, doesn't help storage on the database though which THIS does.

jasonmcintosh:
https://spinnakerteam.slack.com/archives/C0DPVDMQE/p1703184749271979 slack conversation if those curious, but overall looks good.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -903,11 +910,117 @@ class SqlExecutionRepository(
...
+          field("id") to id,
+          field("compressed_body") to compressedBody,
+          field("compression_type") to compressionProperties.compressionType.type
+        )
+        isBodyCompressed = true
+      }
+    }
+
+    log.info("Upserting execution into the ${table.name} table with id $id")
+    upsert(ctx,
```
xibz, line 1000: Out of curiosity, I noticed that this becomes two upserts to two tables. What was the reasoning for separating out the upserts instead of just re-using the `body` field, and having a flag on whether or not it was compressed as the column?

So old executions, would be fetched, look for the `compression_type`, and look for either `NULL`, `""` or `NONE`, which ever option, and instead always just set the `compression_type` field and have it be a single upsert? Rather than needing a separate table.

The benefits I see with the separate table is the blast radius, but does this affect any service outside of orca? Since we use liquidbase, adding a new column isn't too difficult as well

xibz (commented):
Thanks David! This looks great. I had one question around the design and wanted to get your thoughts behind the second table choice rather than re-using the `body` field. Let me know your thoughts.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
jasonmcintosh: Definitely concerned with more queries.  I see one potential advantage from a DB perspective - a separate table is a separate disk operation, separate indexes for updating, etc.  I'd rather like numbers on this as I do see some definitely risk for slower operations. I could see the optimizer going either faster OR much slower, and extra operations in particular if pools are saturated could be problematic.  BUT with the selection on this... just would love to see some numbers of separate table vs. same table.

xibz: Yea, I think definitely having numbers will help in terms of performance, but it just seems odd to add a separate table unless there's a really good reason for it which isn't too clear to me.

My biggest concern is race conditions as well. You could have two saves come in one finishing an upsert 1, but put back in the thread queue, another comes in and upserts at upsert 1 but also finishes upsert 2, however, this allows the sleeping thread to come back in and finish upsert 2. It really depends on the thread scheduler, but I can definitely see this occurring.

So you have a partial save on request A on upsert 2, but another partial save on request B on upsert 1.

dbyron-sf: I'd love to have the data from trying this as a separate table vs. separate column, but unfortunately I don't.  What I can say is that we've been running this for coming up on three years now so that gives me pretty good confidence in it.  I do see the concerns with race conditions and would love to be able to put them to rest, but I don't think I can properly do that either.  I could see these races manifesting as zombie pipelines, which we do occasionally see, but there are many many reasons for zombie pipelines.

None of this is particularly satisfying.  When pipelines are "just too big" this does provide a way forward.  I wish I could address all the quality-of-implementation issues, but I don't see it happening.  I don't like asking for a shortcut, but in this case I think it's worth merging, as it paves the way for me to push other internal changes upstream, like the changes for using read replicas, and more.

jcavanagh: >I see one potential advantage from a DB perspective - a separate table is a separate disk operation, separate indexes for updating, etc.

Postgres already [offloads](https://www.postgresql.org/docs/current/storage-toast.html) large `TEXT` values from the main table storage by default, so I don't think there would be any performance gain by separating the large values into a new table.  AFAIK InnoDB does something similar in MySQL, but I'm not familiar on the details.  

I don't expect a significant performance impact from a simple join on two indexed PK columns either - that wouldn't be typical.

> My biggest concern is race conditions as well. You could have two saves ...

This is true, and these two queries should certainly be part of a transaction.  Seems like a simple fix to me.  However, there is also no impact if execution compression is not enabled, since the second query is never executed.  

> None of this is particularly satisfying. When pipelines are "just too big" this does provide a way forward.

The way I see it, this solves a real problem well enough for a large Spinnaker deployment, and does no harm to those not using the feature.  

@dbyron-sf @xibz @jasonmcintosh 
Given the low risk of the change, perhaps we can reach a compromise?  Either addressing the transactional consistency issue now or filing an issue for later would suffice to me.  Again, it only impacts users who have chosen to enable an off-by-default feature.  

Once merged, we can measure performance if we have a desire to, and make further adjustments and improvements.   

We can also collapse the two tables later if we want to - moving a few columns from one table to another is a simple (but perhaps lengthy, in this case) migration.  If an operator has not enabled this feature, that migration becomes trivial.

jasonmcintosh: I lean towards adding transaction support & merge but... reminder:  that even with the feature off - this STILL does the join call.  Granted should be minor and fast and as you said, indexed call kinda thing.  Good call out on the TEXT values being disk separated though I'm more thinking of indexes which are separated out by table and how the optimizer might work with those being better though given the way the join operates I'd expect limited if any impact.  But.. really not dug through that near enough on the different DB's to know THAT well how they'd optimize without a lot more digging & those metrics/numbers.  I still lean towards merge as overall think it's a good step forward and we can fix the downsides as we go.

I WILL caution ont he "migration" - we did that in a few places and saw it break startup with liquibase which required a LOT more manual handholding when it tries to update LARGE amounts of data.

xibz: I think the transaction is important, and should be added sooner rather than later. Migration is definitely an interesting subject, but having users migrate because of a poor choice we made in the past is definitely not a good user experience. We can always revisit that **if** that needs to happen.

As per the joins always happening, I don't know enough about databases. Im fairly certain none of us have also worked on SQL databases, so Im hoping that joins are fast on an empty table. I'd sleep better at night with numbers. However, this shouldn't be a blocker for this PR.

dbyron-sf: I just pushed up a few commits.  After staring at the code, it turns out that upserts to the two tables happened in a transaction even before the commits.  In other words, all the calls to upsert already happen inside transaction block.

I did adjust the code to only use the compression-related fields when compression is enabled, and added a read-only mode so it's possible to turn off new compression, but keep existing compressed executions working.

I still don't have numbers on separate table vs. additional column, but with any luck there's enough here that it's ok to merge.

xibz (approved):
Changes look good! Talked with David about the transaction cause Im a little confused how they could be under the same transaction, but Im completely ignorant on the sql implementation, and I trust David has done the research :)

orca-sql/src/main/resources/db/changelog-master.yml
```diff
@@ -53,6 +53,12 @@ databaseChangeLog:
 - include:
     file: changelog/20200603-deleted-executions-table-not-mysql.yml
     relativeToChangelogFile: true
+- include
```
jervi, line 56: I've made a PR to fix this typo: https://github.com/spinnaker/orca/pull/4639

christosarvanitis:
@dbyron-sf since this is behind a feature flag wondering if you are ok with backporting it to 1.33.x and 1.32.x?

dbyron-sf:
> @dbyron-sf since this is behind a feature flag wondering if you are ok with backporting it to 1.33.x and 1.32.x?

Because it involves a database migration, I'd rather not.  I suppose I could be convinced, but I don't think 1.34 is too far away.

PR #4621, opened 2023-12-22
Optimize RestartStageHandler

Sitkevich88:
[Issue #6906](https://github.com/spinnaker/spinnaker/issues/6906)

Restarting algorithm was ineffective, if stages had multiple dependents. Instead of using recursion of downstreamStages method I opted for already existing allDownstreamStages method.  allDownstreamStages method could add same stages to queue multiple times, If they had not been visited yet, but were already in the queue. I fixed that with `!queue.contains(s)` check.


Sitkevich88:
@jasonmcintosh , this PR solves the issue I mentioned in the description. Could you please view the Pull Request

ovidiupopa07 (approved):
It looks good to me 👍

PR #4623, opened 2023-12-29
refactor(front50): use SpinnakerRetrofitErrorHandler with Front50Service

Pranav-b-7:
This PR lays the foundational work for upgrading the retrofit version to 2.x, specifically focusing on refactoring the exception handling for Front50Service.

There are no behaviour changes detected with these code changes and hence no additional tests added to demonstrate the functionalities.
Few existing tests are modified to align it with the new changes.

### DIFF
--- a/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
+++ b/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import com.netflix.spinnaker.kork.core.RetrySupport
 import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
 import com.netflix.spinnaker.orca.api.pipeline.RetryableTask
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
@@ -41,7 +42,6 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
-import retrofit.RetrofitError
 
 
 import static com.netflix.spinnaker.kork.web.selector.v2.SelectableService.*
@@ -86,7 +86,7 @@ class CreateBakeTask implements RetryableTask {
           stage.context.user = user
         }
       }
-    } catch (RetrofitError e) {
+    } catch (SpinnakerServerException e) {
       // ignore exception, we will just use the owner passed to us
       if (!e.message.contains("404")) {
         log.warn("Error retrieving application {} from front50, ignoring.", stage.execution.application, e)
--- a/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
+++ b/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
@@ -34,7 +34,6 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
 import org.springframework.stereotype.Component
-import retrofit.RetrofitError
 
 import javax.annotation.Nonnull
 import javax.annotation.Nullable
@@ -198,11 +197,7 @@ public class WaitOnJobCompletion implements CloudProviderAware, OverridableTimeo
     if (appName == null || front50Service == null) {
       return false
     }
-    try {
-      return front50Service.get(appName) != null
-    } catch (RetrofitError e) {
-      throw e
-    }
+    return front50Service.get(appName) != null
   }
 
   /**
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobUtils.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobUtils.java
@@ -30,7 +30,6 @@ import java.util.Map;
 import java.util.Optional;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
-import retrofit.RetrofitError;
 
 @Component
 public class JobUtils implements CloudProviderAware {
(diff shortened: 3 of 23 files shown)


orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
```diff
@@ -86,7 +86,7 @@ class CreateBakeTask implements RetryableTask {
           stage.context.user = user
         }
       }
-    } catch (RetrofitError e) {
+    } catch (SpinnakerServerException e) {
       // ignore exception, we will just use the owner passed to us
       if (!e.message.contains("404")) {
```
dbyron-sf, line 91: Something doesn't seem right about this...like, why not check the response code...and why not re-throw the exception if it's not ignored?  Let's close our eyes in this PR.

orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/job/WaitOnJobCompletion.groovy
```diff
@@ -200,7 +200,7 @@ public class WaitOnJobCompletion implements CloudProviderAware, OverridableTimeo
     }
     try {
       return front50Service.get(appName) != null
-    } catch (RetrofitError e) {
+    } catch (SpinnakerServerException e) {
```
dbyron-sf, line 203: If all the catch block does is re-throw the exception, let's remove it.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/job/JobUtils.java
```diff
@@ -95,7 +95,7 @@ private Boolean applicationExists(String appName) {
     }
     try {
       return front50Service.get(appName) != null;
-    } catch (RetrofitError e) {
+    } catch (SpinnakerServerException e) {
```
dbyron-sf, line 98: Another no-op try/catch to remove please.

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
```diff
@@ -466,10 +466,9 @@ public boolean hasDisableLock(Moniker clusterMoniker, String account, Location l
     Application application;
     try {
       application = front50Service.get(clusterMoniker.getApp());
-    } catch (RetrofitError e) {
+    } catch (SpinnakerHttpException e) {
```
dbyron-sf, line 469: I see a test in TrafficGuardSpec that doesn't strictly need to change because of what's happening in this PR.  It does look like a [broken test](https://github.com/spinnaker/orca/blob/7e923cc05af0f66b952245722aa69c667f685c5c/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuardSpec.groovy#L454) though...and there's a unnecessary null check of application a few lines below here.  Because [Front50Service.get](https://github.com/spinnaker/orca/blob/7e923cc05af0f66b952245722aa69c667f685c5c/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/Front50Service.groovy#L33) returns a "real" type (i.e. not a Response), it never returns null.  So it doesn't make sense to have a test that mocks a null return value.  Can you clean this up please?

dbyron-sf: Looks like a similar thing happens in [UpsertApplicationTaskSpec](https://github.com/spinnaker/orca/blob/7e923cc05af0f66b952245722aa69c667f685c5c/orca-applications/src/test/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTaskSpec.groovy#L60) ...

orca-web/src/main/java/com/netflix/spinnaker/orca/controllers/ExecutionsImportController.java
```diff
@@ -70,10 +71,12 @@ ExecutionImportResponse createExecution(@RequestBody PipelineExecution execution
     try {
       application = front50Service.get(execution.getApplication());
       log.info("Importing application with name: {}", application.name);
-    } catch (RetrofitError e) {
-      if (e.getKind() == RetrofitError.Kind.HTTP && e.getResponse().getStatus() != 404) {
+    } catch (SpinnakerHttpException e) {
```
dbyron-sf, line 74: What you have here matches the [original logic](https://github.com/spinnaker/orca/blob/7e923cc05af0f66b952245722aa69c667f685c5c/orca-web/src/main/java/com/netflix/spinnaker/orca/controllers/ExecutionsImportController.java#L73-L77).  I have to wonder if it's really correct though.  If the idea is to be silent about 404s, but warn on other things, wouldn't we want to have one `catch (SpinnakerServerException)` block that logs a warning for anything else?

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java
dbyron-sf: Thanks for [updating TrafficGuardSpec](https://github.com/spinnaker/orca/pull/4623/commits/0c986730d3918caf7b9d8509dea45cb426d9d81a).  The code below still has a null check that doesn't need to be there.

Pranav-b-7: > Thanks for [updating TrafficGuardSpec](https://github.com/spinnaker/orca/pull/4623/commits/0c986730d3918caf7b9d8509dea45cb426d9d81a). The code below still has a null check that doesn't need to be there.

Are you talking about this line : https://github.com/Pranav-b-7/orca/blob/0c986730d3918caf7b9d8509dea45cb426d9d81a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/utils/TrafficGuard.java#L477 ? 

If yes, this check is still relevant, because in the above catch block when the response code is 404 or 403, application variable is assigned to null.

orca-web/src/main/java/com/netflix/spinnaker/orca/controllers/ExecutionsImportController.java
dbyron-sf: The point of this PR is to move to Spinnaker*Exception, so let's leave the logic as it is for now.

orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CreateBakeTask.groovy
Pranav-b-7: Ok. Lets not cover this in this PR

dbyron-sf (approved):

PR #4636, opened 2024-01-22
refactor(keel): use SpinnakerRetrofitErrorHandler with KeelService

Pranav-b-7:
This PR lays the foundational work for upgrading the retrofit version to 2.x, specifically focusing on refactoring the exception handling for KeelService.

Note, there's a behaviour change on the Task Results error message format when KeelService API throws any 4xx/5xx http errors with empty error body.

- On any 4xx http errors with empty error body:

  
  before:

  11:56:19.324 [Test worker] ERROR com.netflix.spinnaker.orca.keel.task.ImportDeliveryConfigTask - {message=Non-retryable HTTP response 400 received from downstream service: HTTP 400 http://localhost:62130/delivery-configs/: 400 Bad Request}

  after:

  12:00:02.018 [Test worker] ERROR com.netflix.spinnaker.orca.keel.task.ImportDeliveryConfigTask - {message=Non-retryable HTTP response 400 received from downstream service: HTTP 400 http://localhost:62275/delivery-configs/: Status: 400, URL: http://localhost:62275/delivery-configs/, Message: Bad Request}

- On any 5xx http errors with empty error body:

  before:
  
  TaskResult(status=RUNNING, context={repoType=stash, projectKey=SPKR, repositorySlug=keeldemo, directory=., manifest=spinnaker.yml, ref=refs/heads/master, attempt=2, maxRetries=5, errorFromLastAttempt=Retryable HTTP response 500 received from downstream  service: HTTP 500 http://localhost:65311/delivery-configs/: 500 Server Error}, outputs={})

  after:

  TaskResult(status=RUNNING, context={repoType=stash, projectKey=SPKR, repositorySlug=keeldemo, directory=., manifest=spinnaker.yml, ref=refs/heads/master, attempt=1, maxRetries=5, errorFromLastAttempt=Retryable HTTP response 500 received from downstream  service: HTTP 500 http://localhost:49862/delivery-configs/: Status: 500, URL: http://localhost:49862/delivery-configs/, Message: Server Error}, outputs={})

Note : The RetrofitError catch block will still be relevant even after the code changes in the below classes, due to the reasons mentioned inline:

- ImportDeliveryConfigTask.kt - (IgorService retrofit client is not configured with SpinnakerRetrofitErrorHandler , so the APIs still throw RetrofitError)
- DeleteApplicationTask.groovy - (Front50Service retrofit client is not configured with SpinnakerRetrofitErrorHandler, so the APIs still throw RetrofitError)

### DIFF
--- a/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
+++ b/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
@@ -18,6 +18,8 @@ package com.netflix.spinnaker.orca.applications.tasks
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException
 import com.netflix.spinnaker.orca.api.pipeline.TaskResult
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
 import com.netflix.spinnaker.orca.front50.Front50Service
@@ -92,6 +94,15 @@ class DeleteApplicationTask extends AbstractFront50Task {
         return TaskResult.SUCCEEDED
       }
       log.error("Could not delete application", e)
+      return TaskResult.builder(ExecutionStatus.TERMINAL).outputs(outputs).build()
+    } catch (SpinnakerHttpException httpException){
+      if (httpException.responseCode == 404) {
+        return TaskResult.SUCCEEDED
+      }
+      log.error("Could not delete application", httpException)
+      return TaskResult.builder(ExecutionStatus.TERMINAL).outputs(outputs).build()
+    } catch (SpinnakerServerException serverException) {
+      log.error("Could not delete application", serverException)
       return TaskResult.builder(ExecutionStatus.TERMINAL).outputs(outputs).build()
     }
     return TaskResult.builder(ExecutionStatus.SUCCEEDED).outputs(outputs).build()
--- a/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
+++ b/orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
@@ -19,6 +19,9 @@ package com.netflix.spinnaker.orca.keel.task
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.convertValue
 import com.fasterxml.jackson.module.kotlin.readValue
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerNetworkException
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException
 import com.netflix.spinnaker.kork.web.exceptions.InvalidRequestException
 import com.netflix.spinnaker.orca.KeelService
 import com.netflix.spinnaker.orca.api.pipeline.RetryableTask
@@ -76,6 +79,8 @@ constructor(
 
       TaskResult.builder(ExecutionStatus.SUCCEEDED).context(emptyMap<String, Any?>()).build()
     } catch (e: RetrofitError) {
+      handleRetryableFailures(e, context)
+    } catch (e: SpinnakerServerException) {
       handleRetryableFailures(e, context)
     } catch (e: Exception) {
       log.error("Unexpected exception while executing {}, aborting.", javaClass.simpleName, e)
@@ -151,6 +156,69 @@ constructor(
     // this is just a friend URI-like string to refer to the delivery config location in logs
     return "${context.repoType}://${context.projectKey}/${context.repositorySlug}/<manifestBaseDir>/${context.directory
       ?: ""}/${context.manifest}@${context.ref}"
+  }
+
+  /*
+  * Handle (potentially) all Spinnaker*Exception. Smart casts to the respective type on Http error and/or Network error.
+  * @return default error message on non-http and non-network errors.
+  * */
+  private fun handleRetryableFailures(error: SpinnakerServerException, context: ImportDeliveryConfigContext): TaskResult {
+    return when {
+      error is SpinnakerNetworkException -> {
+        // retry if unable to connect
+        buildRetry(
+          context,
+          "Network error talking to downstream service, attempt ${context.attempt} of ${context.maxRetries}: ${error.networkErrorMessage}"
+        )
+      }
+      error is SpinnakerHttpException -> {
+        handleRetryableFailures(error, context)
+      } else -> {
+        buildRetry(
+          context,
+          "Server error talking to downstream service, attempt ${context.attempt} of ${context.maxRetries}: ${error.serverErrorMessage}"
+        )
+      }
+    }
+  }
+
+  /**
+   * Handle (potentially) retryable failures by looking at the HTTP status code. A few 4xx errors
+   * are handled as special cases to provide more friendly error messages to the UI.
+   */
+  private fun handleRetryableFailures(httpException: SpinnakerHttpException, context: ImportDeliveryConfigContext): TaskResult{
+    return when {
+      httpException.responseCode in 400..499 -> {
+        val responseBody = httpException.responseBody
+        // just give up on 4xx errors, which are unlikely to resolve with retries, but give users a hint about 401
+        // errors from igor/scm, and attempt to parse keel errors (which are typically more informative)
+        buildError(
+          if (httpException.fromIgor && httpException.responseCode == 401) {
+            UNAUTHORIZED_SCM_ACCESS_MESSAGE
+          } else if (httpException.fromKeel && responseBody!=null && responseBody.isNotEmpty()) {
+            // keel's errors should use the standard Spring format
+            try {
+              if (responseBody["timestamp"] !=null) {
+                SpringHttpError(responseBody["error"] as String, responseBody["status"] as Int, responseBody["message"] as? String, Instant.ofEpochMilli(responseBody["timestamp"] as Long), responseBody["details"] as? Map<String, Any?>)
+              } else {
+                SpringHttpError(error = responseBody["error"] as String, status = responseBody["status"] as Int, message = responseBody["message"] as? String, details = responseBody["details"] as? Map<String, Any?>)
+              }
+            } catch (_: Exception) {
+              "Non-retryable HTTP response ${httpException.responseCode} received from downstream service: ${httpException.httpErrorMessage}"
+            }
+          } else {
+            "Non-retryable HTTP response ${httpException.responseCode} received from downstream service: ${httpException.httpErrorMessage}"
+          }
+        )
+      }
+      else -> {
+        // retry on other status codes
+        buildRetry(
+          context,
+          "Retryable HTTP response ${httpException.responseCode} received from downstream service: ${httpException.httpErrorMessage}"
+        )
+      }
+    }
   }
 
   /**
@@ -240,13 +308,40 @@ constructor(
       "$message: ${cause?.message ?: ""}"
     }
 
+  val SpinnakerHttpException.httpErrorMessage: String
+    get() {
+      return "HTTP ${responseCode} ${url}: ${cause?.message ?: message}"
+    }
+
+  val SpinnakerNetworkException.networkErrorMessage: String
+    get() {
+      return "$message: ${cause?.message ?: ""}"
+    }
+
+  val SpinnakerServerException.serverErrorMessage: String
+    get() {
+      return "$message"
+    }
+
   val RetrofitError.fromIgor: Boolean
     get() {
       val parsedUrl = URL(url)
       return parsedUrl.host.contains("igor") || parsedUrl.port == 8085
     }
 
+  val SpinnakerServerException.fromIgor: Boolean
+    get() {
+      val parsedUrl = URL(url)
+      return parsedUrl.host.contains("igor") || parsedUrl.port == 8085
+    }
+
   val RetrofitError.fromKeel: Boolean
+    get() {
+      val parsedUrl = URL(url)
+      return parsedUrl.host.contains("keel") || parsedUrl.port == 8087
+    }
+
+  val SpinnakerServerException.fromKeel: Boolean
     get() {
       val parsedUrl = URL(url)
       return parsedUrl.host.contains("keel") || parsedUrl.port == 8087
@@ -271,7 +366,7 @@ constructor(
     val error: String,
     val status: Int,
     val message: String? = error,
-    val timestamp: Instant = Instant.now(),
+    val timestamp: Instant? = Instant.now(),
     val details: Map<String, Any?>? = null // this is keel-specific
   )
 
--- a/orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
+++ b/orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
@@ -35,6 +35,7 @@ import com.github.tomakehurst.wiremock.http.Fault;
 import com.github.tomakehurst.wiremock.http.HttpHeaders;
 import com.github.tomakehurst.wiremock.junit5.WireMockExtension;
 import com.github.tomakehurst.wiremock.junit5.WireMockRuntimeInfo;
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerRetrofitErrorHandler;
 import com.netflix.spinnaker.okhttp.SpinnakerRequestInterceptor;
 import com.netflix.spinnaker.orca.KeelService;
 import com.netflix.spinnaker.orca.api.pipeline.TaskResult;
@@ -95,6 +96,7 @@ public class ImportDeliveryConfigTaskTest {
             .setRequestInterceptor(new SpinnakerRequestInterceptor(true))
             .setEndpoint(wmRuntimeInfo.getHttpBaseUrl())
             .setClient(okClient)
+            .setErrorHandler(SpinnakerRetrofitErrorHandler.getInstance())
             .setLogLevel(retrofitLogLevel)
             .setConverter(new JacksonConverter(objectMapper))
             .build()
@@ -250,7 +252,11 @@ public class ImportDeliveryConfigTaskTest {
         String.format(
             "Non-retryable HTTP response %s received from downstream service: %s",
             HttpStatus.BAD_REQUEST.value(),
-            "HTTP 400 " + wireMock.baseUrl() + "/delivery-configs/: 400 Bad Request");
+            "HTTP 400 "
+                + wireMock.baseUrl()
+                + "/delivery-configs/: Status: 400, URL: "
+                + wireMock.baseUrl()
+                + "/delivery-configs/, Message: Bad Request");
 
     var errorMap = new HashMap<>();
     errorMap.put("message", expectedMessage);
@@ -279,7 +285,9 @@ public class ImportDeliveryConfigTaskTest {
         "errorFromLastAttempt",
         "Retryable HTTP response 500 received from downstream service: HTTP 500 "
             + wireMock.baseUrl()
-            + "/delivery-configs/: 500 Server Error");
+            + "/delivery-configs/: Status: 500, URL: "
+            + wireMock.baseUrl()
+            + "/delivery-configs/, Message: Server Error");
 
     TaskResult running = TaskResult.builder(ExecutionStatus.RUNNING).context(contextMap).build();
 
--- a/orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
+++ b/orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
@@ -48,6 +48,7 @@ import com.netflix.spinnaker.orca.pipeline.model.StageExecutionImpl;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.stream.Stream;
@@ -240,6 +241,83 @@ public class ImportDeliveryConfigTaskTest {
 
     // The timestamp field will have the current time, and hence only the instance type is verified
     assertThat(actualHttpErrorBody.getTimestamp()).isInstanceOf(Instant.class);
+  }
+
+  @Test
+  public void testTaskResultWhenErrorBodyIsEmpty() {
+
+    String expectedMessage =
+        String.format(
+            "Non-retryable HTTP response %s received from downstream service: %s",
+            HttpStatus.BAD_REQUEST.value(),
+            "HTTP 400 " + wireMock.baseUrl() + "/delivery-configs/: 400 Bad Request");
+
+    var errorMap = new HashMap<>();
+    errorMap.put("message", expectedMessage);
+
+    TaskResult terminal =
+        TaskResult.builder(ExecutionStatus.TERMINAL).context(Map.of("error", errorMap)).build();
+
+    // Simulate any 4xx http error with empty error response body
+    String emptyBody = "";
+    simulateFault("/delivery-configs/", emptyBody, HttpStatus.BAD_REQUEST);
+
+    getDeliveryConfigManifest();
+
+    var result = importDeliveryConfigTask.execute(stage);
+
+    verifyGetDeliveryConfigManifestInvocations();
+
+    assertThat(result).isEqualTo(terminal);
+  }
+
+  @Test
+  public void testTaskResultWhenHttp5xxErrorIsThrown() {
+
+    contextMap.put("attempt", (Integer) contextMap.get("attempt") + 1);
+    contextMap.put(
+        "errorFromLastAttempt",
+        "Retryable HTTP response 500 received from downstream service: HTTP 500 "
+            + wireMock.baseUrl()
+            + "/delivery-configs/: 500 Server Error");
+
+    TaskResult running = TaskResult.builder(ExecutionStatus.RUNNING).context(contextMap).build();
+
+    // Simulate any 5xx http error with empty error response body
+    String emptyBody = "";
+    simulateFault("/delivery-configs/", emptyBody, HttpStatus.INTERNAL_SERVER_ERROR);
+
+    getDeliveryConfigManifest();
+
+    var result = importDeliveryConfigTask.execute(stage);
+
+    verifyGetDeliveryConfigManifestInvocations();
+
+    assertThat(result).isEqualTo(running);
+  }
+
+  @Test
+  public void testTaskResultWhenAPIFailsWithNetworkError() {
+
+    contextMap.put("attempt", (Integer) contextMap.get("attempt") + 1);
+    contextMap.put(
+        "errorFromLastAttempt",
+        String.format(
+            "Network error talking to downstream service, attempt 1 of %s: Connection reset: Connection reset",
+            contextMap.get("maxRetries")));
+
+    TaskResult running = TaskResult.builder(ExecutionStatus.RUNNING).context(contextMap).build();
+
+    // Simulate network failure
+    simulateFault("/delivery-configs/", Fault.CONNECTION_RESET_BY_PEER);
+
+    getDeliveryConfigManifest();
+
+    var result = importDeliveryConfigTask.execute(stage);
+
+    verifyGetDeliveryConfigManifestInvocations();
+
+    assertThat(result).isEqualTo(running);
   }
 
   private static Stream<Arguments> parameterizePositiveHttpErrorScenario() {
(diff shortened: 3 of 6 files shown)


orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/DeleteApplicationTask.groovy
```diff
@@ -93,6 +95,15 @@ class DeleteApplicationTask extends AbstractFront50Task {
       }
       log.error("Could not delete application", e)
       return TaskResult.builder(ExecutionStatus.TERMINAL).outputs(outputs).build()
+    } catch (SpinnakerHttpException httpException){
```
kirangodishala, line 98: Why does the RetrofitError catch block still present?

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -77,6 +80,10 @@ constructor(
       TaskResult.builder(ExecutionStatus.SUCCEEDED).context(emptyMap<String, Any?>()).build()
     } catch (e: RetrofitError) {
       handleRetryableFailures(e, context)
+    } catch (networkException: SpinnakerNetworkException) {
```
kirangodishala, line 83: here as well. Shouldn't RetrofitError catch block be removed?

dbyron-sf, line 83: What about SpinnakerConversionException?  Why not catch SpinnakerServerException here and write one new handleRetryableFailures method that takes SpinnakerServerException?

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -153,6 +160,52 @@ constructor(
...
+    return when {
+      httpException.responseCode in 400..499 -> {
+        val responseBody = httpException.responseBody
+        // just give up on 4xx errors, which are unlikely to resolve with retries, but give users a hint about 401
+        // errors from igor/scm, and attempt to parse keel errors (which are typically more informative)
+        buildError(
+          if (httpException.fromIgor && httpException.responseCode == 401) {
+            UNAUTHORIZED_SCM_ACCESS_MESSAGE
+          } else if (httpException.fromKeel && responseBody!=null && responseBody.isNotEmpty()) {
+            // keel's errors should use the standard Spring format, so we try to parse them
```
dbyron-sf, line 188: SpinnakerHttpException has already parsed the response, so at least the comment needs to change here.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
dbyron-sf: We definitely need test coverage for whatever new method(s) we end up with here.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -153,6 +160,52 @@ constructor(
...
+            try {
+              objectMapper.convertValue<SpringHttpError>(responseBody);
```
dbyron-sf, line 190: I guess tests will tell us if this works.  It would be easier for me to follow to call a SpringHttpError constructor directly instead of getting jackson involved.

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
```diff
@@ -0,0 +1,197 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.springframework.http.HttpStatus;
+import retrofit.RestAdapter;
+import retrofit.client.OkClient;
+import retrofit.converter.JacksonConverter;
+
+public class ImportDeliveryConfigTaskTest {
```
dbyron-sf, line 54: I just found [ImportDeliveryConfigTaskTests](https://github.com/spinnaker/orca/blob/c3528667ae353baa79e8cf75df0d3f66376aff88/orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt).    Can you explain how these are different in javadoc?

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
Pranav-b-7: Added a catch block to handle the conversion errors. With this change, a behaviour change is expected . 
To demonstrate the same, I have made a specific commit on it - 099b09ea78bbc23fbf97f586747e406105995b40

RetrofitError.conversionError contains status code in it, whereas the SpinnakerConversionException does not consist this, and hence a behaviour change is expected

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
Pranav-b-7: The comments are removed now its only "_keel's errors should use the standard Spring format_"

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
Pranav-b-7: Added javadoc to describe the need of writing ImportDeliveryConfigTaskTest class

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
dbyron-sf: Conversion errors only happen on 2xx responses, so leaving it out doesn't seem so bad.  It does also seem straightforward to add a response code member to SpinnakerConversionException.  We could even add a HasResponseCode interface that both SpinnakerConversionException and SpinnakerHttpException implement if that helps.

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
dbyron-sf: I don't get it...Why not add the conversion exception test here?

It seems to me that the benefit of using wiremock is that tests don't depend on RetrofitError, nor specific exception classes nor whether the code uses SpinnakerRetrofitErrorHandler.

But then if we're ending up with duplicate coverage, let's remove the tests from ImportDeliveryConfigTaskTest that this class handles.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
Pranav-b-7: Removed the catch block to handle the SpinnakerConversionException. As the return type of the API : keelService.publishDeliveryConfig is retrofit.client.Response. The conversion error never happens

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
Pranav-b-7: Removed the usage ObjectMapper here and modified it by reading the response body map and injecting this values to SpringHttpError constructor.

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
Pranav-b-7: The Conversion error is no more relevant in this case and hence I have not added any tests related to it.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -153,6 +160,62 @@ constructor(
       ?: ""}/${context.manifest}@${context.ref}"
   }
 
+  /**
+   * Handle (potentially) retryable failures for SpinnakerServerException.
+   */
+  private fun handleRetryableFailures(serverException: SpinnakerServerException, context: ImportDeliveryConfigContext): TaskResult{
```
dbyron-sf, line 166: I don't see this method getting called, though I think things would be simpler and more correct if it did.  Like, instead of catching SpinnakerNetworkException and SpinnakerHttpException above, catch SpinnakerServerException instead and call this method....and then either combine the network and http cases here, or call those methods based on the class of exception.

Pranav-b-7: Simplified the catch block with the parent exception class SpinnakerServerException. 
Introduced a new method which would handle all Spinnaker*Exception and smart casts on http error and/or network error, and also returns a default error message on other Spinnaker Exceptions.

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
```diff
@@ -377,26 +379,32 @@ internal class ImportDeliveryConfigTaskTests : JUnit5Minutests {
                 manifestLocation.manifest,
                 manifestLocation.ref
               )
-            } throws RetrofitError.httpError(
+            } returns manifest
+          }
+          with(keelService){
+            every {
+              publishDeliveryConfig(manifest)
+            } throws  SpinnakerHttpException(RetrofitError.httpError(
```
dbyron-sf, line 387:
```suggestion
            } throws SpinnakerHttpException(RetrofitError.httpError(
```

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
```diff
@@ -367,6 +368,7 @@ internal class ImportDeliveryConfigTaskTests : JUnit5Minutests {
 
       context("keel access denied error") {
         modifyFixture {
+          val converter =  JacksonConverter(objectMapper)
```
dbyron-sf, line 371:
```suggestion
          val converter = JacksonConverter(objectMapper)
```

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
Pranav-b-7: Removed the extra space

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
Pranav-b-7: Removed extra space

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
```diff
@@ -377,26 +379,32 @@ internal class ImportDeliveryConfigTaskTests : JUnit5Minutests {
...
+              converter, null
+            ))
           }
         }
 
         test("task fails and includes the error details returned by keel") {
           val result = execute(manifestLocation.toMap())
           expectThat(result.status).isEqualTo(ExecutionStatus.TERMINAL)
-          expectThat(result.context["error"]).isA<SpringHttpError>().isEqualTo(accessDeniedError)
+          expectThat(result.context["error"]).isA<SpringHttpError>()
```
dbyron-sf, line 401: Can we restore an assertion on the value of the error here?  It's a shame to remove it and weaken the test.  I tried
```suggestion
          **expectThat(result.context["error"]).isA<SpringHttpError>().isEqualTo(accessDeniedError)
```
but it fails with:
```
      ✗ is equal to SpringHttpError(error=Forbidden, status=403, message=Access denied, timestamp=2024-02-15T16:50:02.144Z, details=null)
              found SpringHttpError(error=Forbidden, status=403, message=Access denied, timestamp=2024-02-15T16:50:02.165917Z, details=null)
```
is there something different we can do with the timestamp so that passes?

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
```diff
@@ -377,26 +379,32 @@ internal class ImportDeliveryConfigTaskTests : JUnit5Minutests {
...
         }
       }
 
       context("delivery config parsing error") {
         modifyFixture {
+          val converter =  JacksonConverter(objectMapper)
```
dbyron-sf, line 407: Simpler to have one of these at the class level?  Otherwise
```suggestion
          val converter = JacksonConverter(objectMapper)
```

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
```diff
@@ -407,21 +415,26 @@ internal class ImportDeliveryConfigTaskTests : JUnit5Minutests {
...
+              converter, null
+            ))
           }
         }
 
         test("task fails and includes the error details returned by keel") {
           val result = execute(manifestLocation.toMap())
           expectThat(result.status).isEqualTo(ExecutionStatus.TERMINAL)
-          expectThat(result.context["error"]).isA<SpringHttpError>().isEqualTo(parsingError)
+          expectThat(result.context["error"]).isA<SpringHttpError>()
```
dbyron-sf, line 437: Same deal here with restoring the equality check
```
      ✗ is equal to SpringHttpError(error=Bad Request, status=400, message=Parsing error, timestamp=2024-02-15T16:53:45.817Z, details={message=Parsing error, path=[{type=SomeClass, field=someField}], pathExpression=.someField})
              found SpringHttpError(error=Bad Request, status=400, message=Parsing error, timestamp=2024-02-15T16:53:45.823999Z, details={message=Parsing error, path=[{type=SomeClass, field=someField}], pathExpression=.someField})
```

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
kirangodishala, line 401:
```suggestion
          expectThat(result.status).isEqualTo(ExecutionStatus.TERMINAL)
          expectThat(result.context["error"]).isA<SpringHttpError>()
          val keelError: SpringHttpError = result.context["error"] as SpringHttpError
          //before comparing, copy timestamp as they don't match even when all the other details do
          accessDeniedError = accessDeniedError.copy(timestamp = keelError.timestamp)
          expectThat(result.context["error"]).isA<SpringHttpError>().isEqualTo(accessDeniedError)
```

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
kirangodishala, line 437:
```suggestion
          expectThat(result.context["error"]).isA<SpringHttpError>()
          val keelError: SpringHttpError = result.context["error"] as SpringHttpError
          //before comparing, copy timestamp as they don't match even when all the other details do
          parsingError = parsingError.copy(timestamp = keelError.timestamp)
          expectThat(result.context["error"]).isA<SpringHttpError>().isEqualTo(parsingError)
```

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
Pranav-b-7: I have moved this part of the test to java file , which uses wiremock. This is covered in the commit : 7a48b311ff9454124496432a02fe12e1db9aae8e

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
Pranav-b-7: These tests are now deleted to avoid deletion . Deleted in the commit : 39aed3f92f17ce65893746a5da757e2f0bf6bed4

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
Pranav-b-7: I have moved this test scenario to java file which uses wiremock. This is covered in the commit : https://github.com/spinnaker/orca/commit/7a48b311ff9454124496432a02fe12e1db9aae8e

orca-keel/src/test/kotlin/com/netflix/spinnaker/orca/keel/ImportDeliveryConfigTaskTests.kt
Pranav-b-7: I have moved this test scenario to java file which uses wiremock. This is covered in the commit : https://github.com/spinnaker/orca/commit/7a48b311ff9454124496432a02fe12e1db9aae8e

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
```diff
@@ -0,0 +1,385 @@
...
+            (String) contextMap.get("manifest"),
+            (String) contextMap.get("ref")))
+        .thenReturn(mockResponseBody);
+
+    var result = importDeliveryConfigTask.execute(stage);
+
+    assertThat(result).isEqualTo(running);
+  }
+
+  private ImportDeliveryConfigTask.SpringHttpError mockSpringHttpError(
```
dbyron-sf, line 375: There's no mocking here, so let's rename the method.
```suggestion
  private ImportDeliveryConfigTask.SpringHttpError makeSpringHttpError(
```

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
```diff
@@ -0,0 +1,385 @@
...
+            (String) contextMap.get("repoType"),
+            (String) contextMap.get("projectKey"),
+            (String) contextMap.get("repositorySlug"),
+            (String) contextMap.get("directory"),
+            (String) contextMap.get("manifest"),
+            (String) contextMap.get("ref")))
+        .thenReturn(mockResponseBody);
+
+    var result = importDeliveryConfigTask.execute(stage);
+
```
dbyron-sf, line 371: Please verify that scmService.getDeliveryConfigManifest was called as expected.

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
```diff
@@ -0,0 +1,385 @@
...
+   * all the fields.
+   *
+   * <p>Other positive scenarios when the timestamp results in accurate value, are when the units in
+   * : {@link ChronoUnit#SECONDS} {@link ChronoUnit#DAYS} {@link ChronoUnit#HOURS} {@link
+   * ChronoUnit#HALF_DAYS} {@link ChronoUnit#MINUTES}
+   */
+  @Test
+  public void verifyTaskReturnsSpringHttpError() throws JsonProcessingException {
+
+    var mockResponseBody =
```
dbyron-sf, line 147:
```suggestion
    var scmResponseBody =
```

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
```diff
@@ -0,0 +1,385 @@
...
+   * timestamp.
+   *
+   * <p>Other cases where the timestamp will result in incorrect value, are when the units in :
+   * {@link ChronoUnit#MICROS}
+   */
+  @Test
+  public void verifyTaskResultReturnsSpringHttpErrorWhenTimestampIsInNanos()
+      throws JsonProcessingException {
+
+    var mockResponseBody =
```
dbyron-sf, line 199:
```suggestion
    var scmResponseBody =
```

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
```diff
@@ -0,0 +1,251 @@
...
+    var result = importDeliveryConfigTask.execute(stage);
+
+    var httpErrorBody = (ImportDeliveryConfigTask.SpringHttpError) result.getContext().get("error");
+
+    // assert all the values in the http error body are true except the timestamp
+    assertThat(httpErrorBody.getStatus()).isEqualTo(httpStatus.value());
+    assertThat(httpErrorBody.getError()).isEqualTo(httpStatus.getReasonPhrase());
+    assertThat(httpErrorBody.getMessage()).isEqualTo(httpStatus.name());
+    assertThat(httpErrorBody.getDetails()).isEqualTo(Map.of("exception", "Http Error occured"));
+    assertThat(httpErrorBody.getTimestamp()).isNotEqualTo(timestamp);
```
dbyron-sf, line 241: Is there something more we know about httpErrorBody.getTimestamp()?  It's not equal to timestamp, but what is it?

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
```diff
@@ -0,0 +1,251 @@
...
+   * timestamp in {@link ChronoUnit#NANOS}, which WILL NOT be parsed to exact timestamp in the
+   * method @see {@link ImportDeliveryConfigTask#handleRetryableFailures(RetrofitError,
+   * ImportDeliveryConfigTask.ImportDeliveryConfigContext)} and results in will contain incorrect
+   * timestamp.
+   *
+   * <p>Other cases where the timestamp will result in incorrect value, are when the units in :
+   * {@link ChronoUnit#MICROS}
+   */
+  @Test
+  public void verifyTaskResultReturnsSpringHttpErrorWhenTimestampIsInNanos()
```
dbyron-sf, line 196: There's a fair amount of duplication here. Can you parameterize this test across the various ChronoUnits you're testing?

dbyron-sf: I'd also like to see a test where the response from keel includes no timestamp at all.

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
Pranav-b-7: The method name is modified to makeSpringHttpError

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
Pranav-b-7: Added verification step to verify the method invocation of scmService.getDeliveryConfigManifest and extracted the same to a private method, because its being used through out all the tests.

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
Pranav-b-7: This part is extracted to a private method : getDeliveryConfigManifest() and the response body is directly injected to the expected response instead of assigning to a variable.

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
Pranav-b-7: This part is extracted to a private method : getDeliveryConfigManifest() and the response body is directly injected to the expected response instead of assigning to a variable.

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
Pranav-b-7: Writing a test on this would result in NullPointerException, and the test will not get executed.

dbyron-sf: Or maybe a better thing to ask is, where does the NullPointerException come from?

Pranav-b-7: Here is how the SpringHttpError class is defined

```
data class SpringHttpError(
    val error: String,
    val status: Int,
    val message: String? = error,
    val timestamp: Instant = Instant.now(),
    val details: Map<String, Any?>? = null // this is keel-specific
  )
```
  The parameter without '?' symbol are @NotNull annotated and the parameters with '?' symbol are @Nullable .
  
  So , when we pass null values to the @NotNull parameters , the  NullPointerException is thrown at the init stage itself. 
  
  Here is the exact exception thrown : 
  
 ` java.lang.NullPointerException: Parameter specified as non-null is null: method com.netflix.spinnaker.orca.keel.task.ImportDeliveryConfigTask$SpringHttpError.<init>, parameter timestamp`

dbyron-sf: So, change the code to check for null before calling the constructor.

Pranav-b-7: Logic added to handle the case when the timestamp in SpringHttpError is null. The constructor will initialize  this field with the default value "Instant.now()"/ current time.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
```diff
@@ -153,6 +158,69 @@ constructor(
...
+        val responseBody = httpException.responseBody
+        // just give up on 4xx errors, which are unlikely to resolve with retries, but give users a hint about 401
+        // errors from igor/scm, and attempt to parse keel errors (which are typically more informative)
+        buildError(
+          if (httpException.fromIgor && httpException.responseCode == 401) {
+            UNAUTHORIZED_SCM_ACCESS_MESSAGE
+          } else if (httpException.fromKeel && responseBody!=null && responseBody.isNotEmpty()) {
+            // keel's errors should use the standard Spring format
+            try {
+              if (responseBody.get("timestamp")!=null) {
```
dbyron-sf, line 201: There's a fair amount of duplication here.  Bummer that SpringHttpError doesn't have a builder class.  IntelliJ complains about using `get`, so how 'bout:
```suggestion
              if (responseBody["timestamp"] != null) {
```
and similarly below?

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
```diff
@@ -0,0 +1,442 @@
...
+   * Test to verify when the timestamp is null , field will be initialized with default value {@link
+   * Instant#now}
+   */
+  @Test
+  public void testTaskResultWhenTimestampIsNullForSpringHttpError() throws JsonProcessingException {
+
+    var httpStatus = HttpStatus.BAD_REQUEST;
+
+    // SpringHttpError is initialized with no timestamp
+    var httpError = makeSpringHttpError(httpStatus, null);
```
dbyron-sf, line 308: It would be nice for this test to appear in the "before" commit as well, so we can see how it behaves.

But also, this way of building a response body is a little indirect.  The point isn't that timestamp is null in SpringHttpError, the point is that the response body doesn't have a timestamp in it.

orca-keel/src/main/kotlin/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTask.kt
Pranav-b-7: Replaced the get() call with the indexing operator

orca-keel/src/test/java/com/netflix/spinnaker/orca/keel/task/ImportDeliveryConfigTaskTest.java
Pranav-b-7: New test added in the first commit : 5d711ab18a2c26c3420c55ffcd89af0c1ed67d4f to verify the case when the error response body doesn't contain timestamp field.

dbyron-sf (approved):

PR #4644, opened 2024-02-05
feat(kubernetes): identify and retry if the deploy manifest k8s task owner died while orca is monitoring the task

dbyron-sf:
Before this, DeployManifestStage uses MonitorKatoTask to wait for a DeployManifestTask operation to complete. If the task isn't complete, MonitorKatoTask does have logic to call kato.resumeTask (POST task/{id}:resume, clouddriver's OperationsController.resumeTask), but that only applies when the task is retryable. Before this change only tasks in sagas are potentially retryable, so then it's up to orca to re-execute the MonitorKatoTask, and all that's happening is waiting for clouddriver to do its work.

This PR introduces a MonitorDeployManifestTask that starts a new DeployManifestTask if it detects that the clouddriver pod executing the current DeployManifestTask is no longer present. This is only relevant if clouddriver itself is running in kubernetes, and that clouddriver is configured with a kubernetes account for "itself".

Kubernetes DeployManifestTask tasks in orca correspond to KubernetesDeployManifestOperations in clouddriver. That operation is responsible for kubectl apply (or similar based on annotations in the manifest(s) being deployed). It's not typically a long operation, but if there are multiple (or large) manifests, or unreliable communications with the target cluster, it's certainly possible for the clouddriver pod to die before it completes.

The following configuration flags (and their defaults) control the behavior of this new feature:
```
tasks.monitor-kato-task.kubernetes.deploy-manifest:
  retry-task.enabled: false
  maximum-period-inactivity-ms: 300000L
  maximum-forced-retries: 3
  account: "" # the name of the account where clouddriver runs, required if enabled is true
  namespace: spinnaker
```
The execution context of kubernetes deploy manifest stages contain the following information to track the progress of additional attempts:
```
kato.task.forcedRetryAttempts: this is more for debugging....if forcedRetryAttempts < forcedRetries then forceRetryFatalError is true
kato.task.forcedRetries: the number of retries
kato.task.forceRetryFatalError: true if there was an error retrying
```
It's a little unfortunate that restarting the DeployManifestTask means creating a whole new task. Clouddriver's task repository doesn't store its inputs, at least for non-saga tasks, so it's not possible to restart an existing one. See OperationsService.collectAtomicOperationsFromSagas for more.

depends on https://github.com/spinnaker/clouddriver/pull/5833

### DIFF
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
@@ -54,7 +54,9 @@ public class RunJobStage implements StageDefinitionBuilder, CancellableStage {
 
   @Override
   public void taskGraph(@Nonnull StageExecution stage, @Nonnull TaskNode.Builder builder) {
-    builder.withTask("runJob", RunJobTask.class).withTask("monitorDeploy", MonitorJobTask.class);
+    builder
+        .withTask(RunJobTask.TASK_NAME, RunJobTask.class)
+        .withTask(MonitorJobTask.TASK_NAME, MonitorJobTask.class);
 
     getCloudProviderDecorator(stage).ifPresent(it -> it.afterRunJobTaskGraph(stage, builder));
 
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/MonitorDeployManifestTask.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/MonitorDeployManifestTask.java
@@ -1,0 +1,286 @@
+/*
+ * Copyright 2021 Salesforce, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca.clouddriver.tasks.manifest;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.kork.annotations.VisibleForTesting;
+import com.netflix.spinnaker.kork.core.RetrySupport;
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService;
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException;
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException;
+import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
+import com.netflix.spinnaker.orca.clouddriver.KatoService;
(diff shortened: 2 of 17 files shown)


orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/pipeline/job/RunJobStage.java
```diff
@@ -54,7 +54,9 @@ public RunJobStage(
 
   @Override
   public void taskGraph(@Nonnull StageExecution stage, @Nonnull TaskNode.Builder builder) {
-    builder.withTask("runJob", RunJobTask.class).withTask("monitorDeploy", MonitorJobTask.class);
+    builder
+        .withTask(RunJobTask.TASK_NAME, RunJobTask.class)
```
xibz, line 58: +1 on this!

orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/MonitorDeployManifestTask.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.Map;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+public class MonitorDeployManifestTask extends MonitorKatoTask {
```
xibz, line 47: One thing to note, is it seems like this task is only possible if dynamic configuration is used?

If that is the case, can we add comments or some documentation outlining that?

dbyron-sf: I don't really like the DynamicConfigService, but I think it's always available...or at least orca wouldn't start without it, even before this PR.

xibz: It always available, but this line https://github.com/spinnaker/orca/pull/4644/files/1984caaf0cb4ed7e5dd282f1687202d19e55c74f#diff-2ba79f1fbda91e9376143202ef17da92baa1bae4f6c520e8146360de93bf7c64R104 will always return false. Since the default dynamic configuration just return the default value. I think that means this feature will **only** work if it is present and not returning false?

dbyron-sf: The whole term in the `if` expression is
```
!getDynamicConfigService()
            .isEnabled("tasks.monitor-kato-task.kubernetes.deploy-manifest.retry-task", false))
```
so yes, this feature only works when enabled.

xibz: Okay, can we call that out in documentation somewhere? I know it should be obvious when looking at the code, but may be a quick and easy determination if it is documented at the top of the class.

Mostly cause I was like "Oooo, can we use this feature?" But we dont use dynamic configuration ><

dbyron-sf: Hmmm...I feel like the dynamic config bits are an implementation detail.  The PR description does have:
> The following configuration flags (and their defaults) control the behavior of this new feature:
```
tasks.monitor-kato-task.kubernetes.deploy-manifest:
  retry-task.enabled: false
  maximum-period-inactivity-ms: 300000L
  maximum-forced-retries: 3
  account: "" # the name of the account where clouddriver runs, required if enabled is true
  namespace: spinnaker
```
and I figured I'd add the same to the release notes.  Is that enough?

xibz: Yep! Works for me

xibz (approved):

PR #4648, opened 2024-02-20
fix(queue): Fix `SqlQueue#doContainsMessage` to handle queues with more than 100 items

kbatalin:
`SqlQueue#doContainsMessage` doesn't process more than 1 batch because of an incorrect loop inside. When the last element in the batch is processed (`ResultSet#next` returns `false`), the following invocation of `ResultSet#getRow` will return 0. No matter how many rows were processed before.

Basically, this PR is just a copy of https://github.com/spinnaker/orca/pull/4184 with addressed comments. But https://github.com/spinnaker/orca/pull/4184 is abandoned so opened this one. Kudos to Ivor for the original PR


jasonmcintosh (approved):
NICE. And thanks for the tests!!

jasonmcintosh:
@Mergifyio backport release-1.33.x release-1.32.x release-1.31.x

PR #4649, opened 2024-02-21
feat(servergroup): Allow users to opt-out of the target desired size check when verifying if the instances scaled up or down successfully

ovidiupopa07:
Some of our users prefer to use the actual instance size list to verify if everything was scaled up or down accordingly instead of relying on the target percentage. They noticed this is causing instability on their end

### DIFF
--- a/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ResizeServerGroupProperties.java
+++ b/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ResizeServerGroupProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 Netflix, Inc.
+ * Copyright 2024 Harness, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


jasonmcintosh (commented):
Kinda want to figure out a better properties organization strategy, e.g. a `config.sg.use-resize-server-group` option.  I'd personally prefer to have "resize-server-group" to also match the variable "config.sg.use-target-desired-size" option.

Other question: IS there a need to allow this PER operation.  E.g. "context.get("useTargetGroupDesiredSize") to figure out which path to use?

kkotula (approved):

ovidiupopa07:
> IS there a need to allow this PER operation. E.g. "context.get("useTargetGroupDesiredSize") to figure out which path to use?
I don't think there's a need to extract this to stage/operation level. This is just for the match capacity task to use real number of instances instead of relying on the percentage stuff

jasonmcintosh (approved):

ovidiupopa07:
@Mergifyio backport release-1.28.x

ovidiupopa07:
@Mergifyio backport release-1.33.x release 1.32.x release-1.31.x

PR #4655, opened 2024-02-23
fix(core): handle null when invoking pipelines with stages that are missing a "type"

dbyron-sf:
so callers see 400 with a helpful mesage instead of 500.

There's also code here in PipelineBuilder.withStages to throw an IllegalArgumentException for null stages.  Due to [this change](https://github.com/spinnaker/orca/pull/4102/files#diff-88a2bb8f8eddf44a88866f3d27f096d67577c6c5f651acd88fdb3da32ed88d0aR211) in https://github.com/spinnaker/orca/pull/4102, PipelineBuilder.withStages isn't called with null even if stages is null in the incoming pipeline.  https://github.com/spinnaker/orca/pull/4182 (Nov 30, 2021, version 1.28.0) was a behavior change from a NullPointerException/500 to succeed (but do nothing) for a pipeline with null stages.  https://github.com/spinnaker/orca/pull/4102 (Jan 3, 2022) was also part of version 1.28.0.  Changing the behavior to explictly reject null stages / respond with 400 instead of accepting them and doing nothing feels better, but I'll leave that for a separate PR as it's potentially more controversial.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/PipelineBuilder.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/PipelineBuilder.java
@@ -84,14 +84,30 @@ public class PipelineBuilder {
   }
 
   public PipelineBuilder withStages(List<Map<String, Object>> stages) {
-    if (stages != null) {
-      stages.forEach(
-          it -> {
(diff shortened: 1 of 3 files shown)


xibz (approved):
:shipit:

PR #4656, opened 2024-02-23
feat(execution): remove spinnaker accounts from execution context

dbyron-sf:
Add the `executions.includeAllowedAccounts` flag which defaults to true. If set to true, `authentication.allowedAccounts` will continue to be populated by accounts in the pipeline execution json. If set to false, the pipeline execution json will show `authentication.allowedAccounts = []`

Testing notes:
After setting the feature flag to false,

1. Pipelines triggered manually (by clicking play) and automatically (using a service account) continue to run. Pipelines that deploy resources into a kubernetes cluster are still able to do so
2. If the user or service account is unauthorized to run the pipeline, the pipeline will not run

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/config/ExecutionConfigurationProperties.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/config/ExecutionConfigurationProperties.java
@@ -42,6 +42,9 @@ public class ExecutionConfigurationProperties {
    */
   private Set<OrchestrationExecution> allowedOrchestrationExecutions = Set.of();
 
+  /** flag to include/exclude the set of allowed accounts in an execution context */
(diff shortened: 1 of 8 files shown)


xibz (approved):
LGTM! One thing I was wondering, but don't have enough context is how useful it would to at least log the amount of allowed accounts. I know fiat, clouddriver, etc do some logging of accounts, but I am unsure if orca does. It would allow for us to ensure that the expected amount is there at least rather than completely missing. But that can be a follow up PR since it is really minor

dbyron-sf:
Yeah, if someone wants that, let's add it in another PR.  This seems like a thing where we don't actually need the info, so I'm not sure how helpful it'll be, but I won't stand in the way of logging it if someone wants to.

PR #4675, opened 2024-03-12
fix(retrofit): use SpinnakerServerException.getHttpMethod() instead of reflection in SpinnakerServerExceptionHandler

abe21412:
`BaseRetrofitExceptionHandler`'s retry behavior uses a method called `findHttpMethodAnnotation` to retrieve an exception's HTTP method. For retrofit2 exceptions, this method no longer works due to things being annotated differently in retrofit2. However, Spinnaker*Exceptions can use `getHttpMethod()` to retrieve this value, so this PR updates the exception handler to default to using the getter and only use `findHttpMethodAnnotation` if the getter returns null.

### DIFF
--- a/orca-retrofit/orca-retrofit.gradle
+++ b/orca-retrofit/orca-retrofit.gradle
@@ -37,6 +37,8 @@ dependencies {
   testImplementation("org.junit.jupiter:junit-jupiter-params")
   testImplementation("org.mockito:mockito-core")
   testImplementation("org.springframework.boot:spring-boot-test")
+  testImplementation("com.squareup.okhttp3:mockwebserver")
+  testImplementation("com.squareup.retrofit2:converter-jackson")
   testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine")
   testRuntimeOnly("org.springframework:spring-test")
 
--- a/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/BaseRetrofitExceptionHandler.groovy
+++ b/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/BaseRetrofitExceptionHandler.groovy
@@ -24,6 +24,10 @@ import static java.net.HttpURLConnection.*
 
 abstract class BaseRetrofitExceptionHandler implements ExceptionHandler {
   boolean shouldRetry(Exception e, String kind, Integer responseCode) {
+    return shouldRetry(e, kind, null, responseCode)
+  }
+
+  boolean shouldRetry(Exception e, String kind, String httpMethod, Integer responseCode) {
     if (isMalformedRequest(kind, e.getMessage())) {
       return false
     }
@@ -33,7 +37,11 @@ abstract class BaseRetrofitExceptionHandler implements ExceptionHandler {
       return true
     }
 
-    return isIdempotentRequest(e) && (isNetworkError(kind) || isGatewayErrorCode(kind, responseCode) || isThrottle(kind, responseCode))
+    if(httpMethod == null) {
+      httpMethod = findHttpMethodAnnotation(e)
+    }
+
+    return isIdempotentRequest(httpMethod) && (isNetworkError(kind) || isGatewayErrorCode(kind, responseCode) || isThrottle(kind, responseCode))
   }
 
   private boolean isGatewayErrorCode(String kind, Integer responseCode) {
@@ -55,8 +63,8 @@ abstract class BaseRetrofitExceptionHandler implements ExceptionHandler {
     return "UNEXPECTED".equals(kind) && exceptionMessage?.contains("Path parameter")
   }
 
-  private static boolean isIdempotentRequest(Exception e) {
-    findHttpMethodAnnotation(e) in ["GET", "HEAD", "DELETE", "PUT"]
+  private static boolean isIdempotentRequest(String httpMethod) {
+    httpMethod in ["GET", "HEAD", "DELETE", "PUT"]
   }
 
   private static String findHttpMethodAnnotation(Exception exception) {
--- a/orca-retrofit/src/test/java/com/netflix/spinnaker/orca/retrofit/exceptions/SpinnakerServerExceptionHandlerTest.java
+++ b/orca-retrofit/src/test/java/com/netflix/spinnaker/orca/retrofit/exceptions/SpinnakerServerExceptionHandlerTest.java
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.orca.retrofit.exceptions;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowableOfType;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
@@ -24,6 +25,7 @@ import static org.mockito.Mockito.when;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 import com.google.gson.Gson;
+import com.netflix.spinnaker.kork.retrofit.ErrorHandlingExecutorCallAdapterFactory;
 import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException;
 import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerNetworkException;
 import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerRetrofitErrorHandler;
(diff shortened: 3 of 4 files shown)


orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/BaseRetrofitExceptionHandler.groovy
```diff
@@ -36,6 +36,23 @@ abstract class BaseRetrofitExceptionHandler implements ExceptionHandler {
     return isIdempotentRequest(e) && (isNetworkError(kind) || isGatewayErrorCode(kind, responseCode) || isThrottle(kind, responseCode))
   }
 
+  boolean shouldRetry(Exception e, String kind, String httpMethod, Integer responseCode) {
```
dbyron-sf, line 39: I suspect there's a way to reduce some duplication here by keeping only one shouldRetry method with the additional `httpMethod` argument (or maybe keeping the existing one without and calling the new one with `httpMethod` as null).  Then, if `httpMethod` is null, call `findHttpMethodAnnotation` to populate it, and then we end up with only one `isIdempotentRequest` method as well.

abe21412: I think keeping the existing one makes sense for now but I can definitely reduce some duplication inside the two methods

dbyron-sf: LGTM, thanks!

orca-retrofit/src/test/java/com/netflix/spinnaker/orca/retrofit/exceptions/SpinnakerServerExceptionHandlerTest.java
```diff
@@ -153,12 +159,82 @@ private static Stream<Arguments> exceptionsForHandlesTest() {
...
+        makeRetrofit2Response(throttledNonIdempotentRequest, HttpStatus.TOO_MANY_REQUESTS.value());
+    SpinnakerServerException throttledNonIdempotent =
+        new SpinnakerHttpException(throttledNonIdempotentResponse, retrofit2Service);
+
+    return Stream.of(
+        Arguments.of(retryable, true),
+        Arguments.of(notRetryable, false),
+        Arguments.of(notRetryableIdempotent, false),
+        Arguments.of(idempotentNetworkException, true),
+        Arguments.of(throttledIdempotent, true),
```
dbyron-sf, line 234: This test fails before the code change.  As in, `throttledIdempotent` is not considered retryable before the code change.  Same with `idempotentNetworkException`.

I think we need the exception to happen as the result of an actual call to a retrofit2 class to see, instead of building the exception class manually.  I bet something the Retrofit2Service class from [here](https://github.com/spinnaker/kork/blob/e823562f7dd449b980708c1562b718c1e9da70e1/kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofit2ErrorHandleTest.java#L69) and [here](https://github.com/spinnaker/kork/blob/e823562f7dd449b980708c1562b718c1e9da70e1/kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofit2ErrorHandleTest.java#L187) would do it?

abe21412: yeah, I was trying to avoid the extra setup but looks like it can't be helped. Should I go ahead and do this a new branch with only the tests or keep it here?

dbyron-sf: I'd like to keep it in this PR so all the comments/context stay together.  I'm happy to help with the git juggling if you'd like.

abe21412: I think those two exceptions are not being considered retryable because `findHttpMethodAnnotation` is looking for this [`RestMethod`](https://github.com/spinnaker/orca/blob/74ba9caa785f817e7f475d1994add5b9c5c5c4e4/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/BaseRetrofitExceptionHandler.groovy#L74) annotation which is present in [retrofit](https://github.com/square/retrofit/blob/36eab5b67dbb4bf37c71a7d07ac571992b4916b2/retrofit/src/main/java/retrofit/http/GET.java#L29) but not [retrofit2](https://github.com/square/retrofit/blob/trunk/retrofit/src/main/java/retrofit2/http/GET.java), so it's getting null for the HTTP method. It seems that without any code changes all retrofit2 exceptions will be marked not retryable.

dbyron-sf: Good point!  So then maybe this isn't a refactor, but a fix?  It's maybe a nit since I don't think orca actually uses retrofit2 anywhere yet.

I think still it would be better if the tests generated exceptions from "real" retrofit2 calls.  You agree?

abe21412: yeah, I agree on the test exceptions. Do you think it's worth updating `findHttpMethodAnnotation` to also work for retrofit2 exceptions on the off chance that `getHttpMethod` returns null?

dbyron-sf: Nah, I think the way we're handing retrofit2 exceptions is sufficient.  From my reading of [the code](https://github.com/spinnaker/kork/blob/e823562f7dd449b980708c1562b718c1e9da70e1/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerException.java), we set `httpMethod` every time from an okhttp3 Request object, and the [kotlin code for that](https://github.com/square/okhttp/blob/parent-4.9.3/okhttp/src/main/kotlin/okhttp3/Request.kt#L30) looks to me like it can't be null.

orca-retrofit/src/test/java/com/netflix/spinnaker/orca/retrofit/exceptions/SpinnakerServerExceptionHandlerTest.java
```diff
@@ -33,10 +35,18 @@
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
+import okhttp3.*;
```
dbyron-sf, line 40: please avoid star imports.

dbyron-sf:
I'm seeing a number of failures with only [the commit that adds the tests](https://github.com/spinnaker/orca/pull/4675/commits/8a828b3f7c3178a91a4fa7e614f555cf5046eb0b), but not the code changes.  Ideally there wouldn't be any failures, and the code change commit would adjust the tests so the behavior change is clear.  I realize this is a lot to ask. I am quite curious what we're actually fixing in this PR though.
image

orca-retrofit/orca-retrofit.gradle
```diff
@@ -30,13 +30,15 @@ dependencies {
   implementation("io.reactivex:rxjava")
   implementation("io.spinnaker.kork:kork-retrofit")
   implementation("com.jakewharton.retrofit:retrofit1-okhttp3-client:1.1.0")
+  implementation("com.squareup.retrofit2:converter-jackson")
```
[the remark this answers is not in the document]
abe21412: it's part of the test setup [here](https://github.com/spinnaker/orca/blob/5a58ab932b7b59aab7da6e064b3e97e61faa1412/orca-retrofit/src/test/java/com/netflix/spinnaker/orca/retrofit/exceptions/SpinnakerServerExceptionHandlerTest.java#L166). I guess maybe it should be a `testImplementation` for now since it's currently only needed to generate the retrofit2 exceptions for testing.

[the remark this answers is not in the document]
dbyron-sf: Yeah, it does mean a bit more juggling, but it does make things easier to follow.

dbyron-sf:
It would be nice to see the behavior changes in detail, but I'll get over it.  Would you mind changing the PR title / commit message to be `fix` instead of refactor though, and say that what we're fixing is retry behavior for retrofit2?

orca-retrofit/orca-retrofit.gradle
```diff
@@ -37,6 +37,8 @@ dependencies {
   testImplementation("org.junit.jupiter:junit-jupiter-params")
   testImplementation("org.mockito:mockito-core")
   testImplementation("org.springframework.boot:spring-boot-test")
+  testImplementation("com.squareup.okhttp3:mockwebserver")
+  implementation("com.squareup.retrofit2:converter-jackson")
```
dbyron-sf, line 41:
```suggestion
  testImplementation("com.squareup.retrofit2:converter-jackson")
```
?

abe21412: can't believe I missed that, I'll update it

dbyron-sf (approved):

PR #4679, opened 2024-03-21
refactor(mine): use SpinnakerRetrofitErrorHandler with MineService

Pranav-b-7:
This PR lays the foundational work for upgrading the retrofit version to 2.x, specifically focusing on refactoring the exception handling for MineService

There is no behaviour changes involved. Couple of tests added to verify the same for only RegisterCanaryTask in the previous commits.

### DIFF
--- a/orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/RegisterCanaryTask.groovy
+++ b/orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/RegisterCanaryTask.groovy
@@ -16,6 +16,10 @@
 
 package com.netflix.spinnaker.orca.mine.tasks
 
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerConversionException
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerNetworkException
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
 import com.netflix.spinnaker.orca.api.pipeline.Task
 import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
@@ -25,7 +29,6 @@ import com.netflix.spinnaker.orca.mine.pipeline.DeployCanaryStage
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
-import retrofit.RetrofitError
 import retrofit.client.Response
 
 import javax.annotation.Nonnull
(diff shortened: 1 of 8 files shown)


orca-mine/src/test/java/com/netflix/spinnaker/orca/mine/tasks/RegisterCanaryTaskTest.java
```diff
@@ -0,0 +1,211 @@
...
+            HttpStatus.NOT_ACCEPTABLE.getReasonPhrase(),
+            List.of(),
+            new TypedString("canaryId"));
+    String errorResponseBody = objectMapper.writeValueAsString(response);
+
+    // simulate error HTTP 406
+    simulateFault("/registerCanary", errorResponseBody, HttpStatus.NOT_ACCEPTABLE);
+
+    // Format as per the error response thrown by register canary task.
+    errorResponseBody =
```
dbyron-sf, line 127: Is using a Response really the easiest way to build up an response body?  Seems like it's be clearer to write out the string directly.

dbyron-sf, line 127: This doesn't get used til lower down, so perhaps move it there?

orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/RegisterCanaryTask.groovy
```diff
@@ -59,21 +61,30 @@ class RegisterCanaryTask implements Task {
...
 
       throw new IllegalStateException(
         "Unable to register canary (executionId: ${stage.execution.id}, stageId: ${stage.id} canary: ${c}), response: ${response}"
       )
+    } catch(SpinnakerServerException e){
+      if(e instanceOf SpinnakerNetworkException) {
+        def response = [:]
+        response.errorKind = "NETWORK"
+        response.status = null
+        throw new IllegalStateException(
```
dbyron-sf, line 84: Doesn't the old code do this for all RetrofitErrors, not only network errors?

orca-mine/src/test/java/com/netflix/spinnaker/orca/mine/tasks/RegisterCanaryTaskTest.java
Pranav-b-7: This formatting part is removed, because the expected order of the response details printed in the logs may not match with the actual order. 
So, the alternative used here to assert is "hasMessageContaining()".

orca-mine/src/main/groovy/com/netflix/spinnaker/orca/mine/tasks/RegisterCanaryTask.groovy
Pranav-b-7: The wiremock tests could only replicate the network and http errors, So i implemented in this way.

However, this part of the code is modified in the recent commits. Default case is also now added.

PR #4712, opened 2024-04-26
refactor(retrofit): Cleanup unused RetrofitError occurrences

Pranav-b-7:
As part of the foundational work paving the way towards migrating to retrofit 2.x, this PR aims at cleaning up unused occurrences of RetrofitError throughout ORCA service.

### DIFF
--- a/orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/api/BakeryServiceSpec.groovy
+++ b/orca-bakery/src/test/groovy/com/netflix/spinnaker/orca/bakery/api/BakeryServiceSpec.groovy
@@ -21,7 +21,6 @@ import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException
 import com.netflix.spinnaker.orca.bakery.config.BakeryConfiguration
 import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper
 import retrofit.RequestInterceptor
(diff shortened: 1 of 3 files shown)


orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/RetrofitConfiguration.groovy
```diff
@@ -99,18 +98,6 @@ class RetrofitConfiguration {
     return LogLevel.valueOf(retrofitLogLevel)
   }
 
-  /**
```
dbyron-sf, line 102: This is a big enough change to warrant it's own separate PR.  I'm pretty sure we've made enough changes so that nothing is throwing RetrofitError anymore, but I'd like to stare at the codebase a fair bit before we remove it.  Removed the unused imports that this PR helps to do that.  And to clarify, retrofit still throws RetrofitError, but since we use SpinnakerRetrofitErrorHandler everywhere, it transforms RetrofitError to Spinnaker*Exception.

Pranav-b-7: That makes sense. I am only keeping the changes to remove the unused import statement in this PR. Rest of the changes will be moved to the upcoming PR.

dbyron-sf (approved):

PR #4727, opened 2024-05-22
fix(liquibase): fix checkSum errors occurring with spinnaker upgrade

kirangodishala:
Added new integration test to validate the postgres migration scenario which uses latest 1.32.x image(us-docker.pkg.dev/spinnaker-community/docker/orca:8.36.3-dev-release-1.32.x-3f8965d03-202406101625-unvalidated) as base image. 

This PR fixes checkSum errors caused after liquibase upgrade(to 4.24.0) as the checksum calculations in liquibase underwent some changes causing this issue.

This PR fixes the checksum errors by adding validChecksum tags to the failing changesets.

Similar clouddriver PR for reference: https://github.com/spinnaker/clouddriver/pull/6217

### DIFF
--- a/orca-integration/src/test/java/com/netflix/spinnaker/orca/PostgresMigrationContainerTest.java
+++ b/orca-integration/src/test/java/com/netflix/spinnaker/orca/PostgresMigrationContainerTest.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright 2024 OpsMx, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.orca;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.time.Duration;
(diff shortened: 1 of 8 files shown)


dbyron-sf:
@mergifyio update

orca-integration/src/test/java/com/netflix/spinnaker/orca/PostgresMigrationContainerTest.java
```diff
@@ -0,0 +1,154 @@
+/*
+ * Copyright 2024 Netflix, Inc.
```
dbyron-sf, line 2:
```suggestion
 * Copyright 2024 OpsMx, Inc.
```
?

kirangodishala: Done

orca-sql/src/main/resources/db/changelog/20200327-deleted-executions-table.yml
```diff
@@ -1,10 +1,8 @@
 databaseChangeLog:
   - changeSet:
+      validCheckSum: 8:10687930e669629d8c370cd5a2348585
       id: create-deleted-executions-table
       author: mvulfson
-      validCheckSum:
```
dbyron-sf, line 5: Is the idea that this was never used because it's a list instead of a string?  I'd rather keep this around if possible.

kirangodishala: We can't have two tags with the same name

dbyron-sf: You mean two `validCheckSum` properties?  Yeah, I figure we can have only one.  My question was whether it can be a list, but from https://docs.liquibase.com/concepts/changelogs/changeset-checksums.html#yaml_example apparently not.  So folks must have migrated enough to not need the one we're removing.

dbyron-sf (approved):

dbyron-sf:
@mergifyio backport release-1.33.x release-1.34.x

PR #4728, opened 2024-05-22
fix(blueGreen): Scaling replicaSets should not be considered for deletion

christosarvanitis:
During Blue/Green deployment if the old replicaSet happens to be autoscaling or for any reason some replica's are not up and running the previous implementation would consider them for Deletion.

This has led to a weird situation where healthy replicaSets were deleted and making the rollback impossible.


kkotula (approved):
LGMT 🚀

jasonmcintosh (approved):

ovidiupopa07:
@Mergifyio backport release-1.34.x release-1.33.x release-1.32.x

PR #4748, opened 2024-06-12
chore(gradle): Fix gradle test memory allocation

jasonmcintosh:


gradle.properties
```diff
@@ -2,7 +2,7 @@ fiatVersion=1.45.0
 korkVersion=7.224.0
 kotlinVersion=1.6.21
 org.gradle.parallel=true
-org.gradle.jvmargs=-Xmx6g
+org.gradle.jvmargs=-Xmx4g
```
dbyron-sf, line 5: github actions still uses 6g, e.g. [here](https://github.com/spinnaker/orca/blob/f6d23180112ad5b9c46f84d8cf6409781f261c11/.github/workflows/pr.yml#L6).

jasonmcintosh: THINK i've caught em all now...

dbyron-sf: One more in Dockerfile.compile...

dbyron-sf:
@mergifyio update

dbyron-sf (approved):

PR #4749, opened 2024-06-12
perf(sql): Store child pipeline execution in trigger as reference

edgarulg:
### Purpose

When a new pipeline is saved, pipeline triggers are stored as references instead of the full execution. This applies only to PipelineTrigger types. The main advantage is that it allows for smaller pipelines to be stored in the database, while the complete execution context remains accessible through an additional SQL query that retrieves the execution context from the parent execution.

The improvement allows deserialization of PipelineRef entities in the database even if the feature flag is disabled, ensuring that executions stored as PipelineRef are correctly displayed in the UI.

### How to enable the feature

This performance improve is behind a feature flag. Here is the configuration you need to include to enable it.
```
executionRepository:
  sql:
    enabled: true
    pipelineRef:
      enabled: true
```

### Features related

This feature flag can work together with: https://github.com/spinnaker/orca/pull/4620
Also it is related to partially fix: https://github.com/spinnaker/spinnaker/issues/6159
And it is based on the following implementation that was reverted: https://github.com/spinnaker/orca/pull/4014

### SQL performance statistics 

Here is the performance statistics: https://github.com/spinnaker/orca/pull/4749#issuecomment-2165448266

### Next Steps

While this change reduces the size of executions in the database, further modifications are needed to minimize the execution payload sent to the Spinnaker UI. I will work on it later but smaller changes are less risky :)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/jackson.kt
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright 2024 Netflix, Inc.
```
jasonmcintosh, line 2: Don't think this is netflix :)

orca-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
```diff
@@ -161,4 +170,13 @@ class SqlConfiguration {
   fun lockProvider(datasource: DataSource): LockProvider {
     return JdbcTemplateLockProvider(datasource)
   }
+
+  @Bean
+  @ConditionalOnProperty("execution-repository.sql.pipeline-ref.enabled")
+  fun pipelineRefTriggerDeserializer(@Qualifier("mapper") mapper: ObjectMapper): CustomTriggerDeserializerSupplier {
+    val customTrigger = PipelineRefTriggerDeserializerSupplier()
+    TriggerDeserializer.customTriggerSuppliers.add(customTrigger)
```
edgarulg, line 178: This is required because CustomTriggerSupplier is a list in a companion block. Everything it is initialized with empty and this ensure Orca will know how to deserialize the PipelineRefTrigger

dbyron-sf:
This is super exciting!  Did you find specific bugs with [the original implementation](https://github.com/spinnaker/orca/pull/3986)?  Or maybe you took a different approach?  I'm curious what's different.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
```diff
@@ -22,15 +22,7 @@ import com.fasterxml.jackson.databind.JsonNode
...
-import com.netflix.spinnaker.orca.pipeline.model.ArtifactoryTrigger
-import com.netflix.spinnaker.orca.pipeline.model.ConcourseTrigger
-import com.netflix.spinnaker.orca.pipeline.model.DefaultTrigger
-import com.netflix.spinnaker.orca.pipeline.model.DockerTrigger
-import com.netflix.spinnaker.orca.pipeline.model.GitTrigger
-import com.netflix.spinnaker.orca.pipeline.model.JenkinsTrigger
-import com.netflix.spinnaker.orca.pipeline.model.NexusTrigger
-import com.netflix.spinnaker.orca.pipeline.model.PipelineTrigger
-import com.netflix.spinnaker.orca.pipeline.model.PluginTrigger
+import com.netflix.spinnaker.orca.pipeline.model.*
```
dbyron-sf, line 25: please avoid star imports...minimizes the diffs too.

orca-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
```diff
@@ -73,7 +73,8 @@ class SqlConfiguration {
     orcaSqlProperties: OrcaSqlProperties,
     interlink: Optional<Interlink>,
     executionRepositoryListeners: Collection<ExecutionRepositoryListener>,
-    compressionProperties: ExecutionCompressionProperties
+    compressionProperties: ExecutionCompressionProperties,
+    @Value("\${execution-repository.sql.pipeline-ref.enabled:false}") pipelineRefEnabled: Boolean
```
dbyron-sf, line 82: I'm not a fan of `@Value` since they can be hard to find. Can you add a new config class (or a new member to an existing class) for this please?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/ExecutionMapper.kt
```diff
@@ -132,4 +139,31 @@ class ExecutionMapper(
...
+    if (pipelineRefEnabled && trigger is PipelineRefTrigger) {
+      val parentExecution = context
+        .select(listOf(
+          field("id"),
+          field("body"),
+          field(name("partition"))
+        ))
+        .from(ExecutionType.PIPELINE.tableName)
+        .where(field("id").eq(trigger.parentExecutionId))
+        .fetchExecutions(mapper, 200, compressionProperties, context, pipelineRefEnabled)
```
dbyron-sf, line 148: What's 200?

dbyron-sf:
Trying to understand more about [this commit](https://github.com/spinnaker/orca/pull/4749/commits/1673c7570b4d710210af1a2c1066bc6325200996)
```
perf(pipelineRef): define deserializer as CustomTriggerDeserializerSupplier and move changes in orca-core to orca-sql.
```
What's the motivation for this change?  On the surface it looks like more of a refactor than a perf change.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/jooq.kt
```diff
@@ -20,11 +20,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.kork.core.RetrySupport
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType
 import de.huxhorn.sulky.ulid.ULID
-import org.jooq.DSLContext
-import org.jooq.Field
-import org.jooq.Record
-import org.jooq.SelectForUpdateStep
-import org.jooq.Table
+import org.jooq.*
```
dbyron-sf, line 23: please avoid star imports

christosarvanitis:
This is potentially a huge improvement/win for Orca! Thanks for tackling this @edgarulg 🚀 


Adding some details based on a testbed. The nested pipeline looks like this:

Parent pipeline:
TriggerPipeline1 -> TriggerPipeline2 -> TriggerPipeline3 -> TriggerPipeline4 -> TriggerPipeline5

Each TriggerPipeline:
DeployManifest -> TriggerNestedPipeline

Each NestedPipeline: 
DeployManifest 

With this feature `disabled` the stored data in SQL is: 
```
MySQL [orca]> select id,json_extract(`body`,'$.name'), sum(char_length(body))/1024  from pipelines where application='orca-pipeline-ref'  group by id order by start_time desc; 
+----------------------------+-------------------------------+-----------------------------+
| id                         | json_extract(`body`,'$.name') | sum(char_length(body))/1024 |
+----------------------------+-------------------------------+-----------------------------+
| 01J08AQPH1ZYJRCTWJH1REEH5C | "child3-trigger-5"            |                   2129.2207 |
| 01J08AQNCVDGDF48JGSRK15HDJ | "child2-trigger-5"            |                   1553.7979 |
| 01J08AQB9M7AAVVMKHWT7PMC3Y | "child1-trigger-5"            |                    835.6719 |
| 01J08AP95DZ5VJ0YFJZD1F5HR7 | "child3-trigger-4"            |                   1714.0029 |
| 01J08AP85M577CVNGYNNAGSXHY | "child2-trigger-4"            |                   1242.7734 |
| 01J08ANYA3KX5ZK1M59QT303BZ | "child1-trigger-4"            |                    629.0137 |
| 01J08AMW50196MQR1HJZD8F7TX | "child3-trigger-3"            |                   1299.4766 |
| 01J08AMV4WWSHGMNTYMJD8DKC1 | "child2-trigger-3"            |                    932.2676 |
| 01J08AMHBBG8XJ77H1KN0QDRAX | "child1-trigger-3"            |                    422.7012 |
| 01J08AKEV24R0XNJK46FFVVEVP | "child3-trigger-2"            |                    885.6416 |
| 01J08AKDWWDT3WN6G6E7JXZ7K0 | "child2-trigger-2"            |                    622.2803 |
| 01J08AK4H09TW9SYKPH3R9N4S4 | "child1-trigger-2"            |                    216.7344 |
| 01J08AJ1R6DDAFWDV8Y04J2H2Z | "child3-trigger-1"            |                    470.7793 |
| 01J08AJ0Y6JQXMP77SWME09DWP | "child2-trigger-1"            |                    311.6084 |
| 01J08AHQNVE5THQ30BT1QP9QSN | "child1-trigger-1"            |                     10.7695 |
| 01J08AHQ1K8JT7KRZT1FRSF5RV | "parent"                      |                      1.5371 |
+----------------------------+-------------------------------+-----------------------------+
16 rows in set (0.143 sec)
```

With this feature `enabled`:
```
MySQL [orca]> select id,json_extract(`body`,'$.name'), sum(char_length(body))/1024  from pipelines where application='orca-pipeline-ref'  group by id order by start_time desc; 
+----------------------------+-------------------------------+-----------------------------+
| id                         | json_extract(`body`,'$.name') | sum(char_length(body))/1024 |
+----------------------------+-------------------------------+-----------------------------+
| 01J08D5J2GP8JD1M9DCN0E310X | "child3-trigger-5"            |                      2.3477 |
| 01J08D5H4SMPRZ19ANP66W6JM7 | "child2-trigger-5"            |                      2.3223 |
| 01J08D56VZ049NDDMZMPEZBXK2 | "child1-trigger-5"            |                      2.3477 |
| 01J08D44X7KE9K67XFP3JJT6H2 | "child3-trigger-4"            |                      2.1748 |
| 01J08D43Z1SB4H48WTES76Q9GE | "child2-trigger-4"            |                      2.1494 |
| 01J08D3SW6P6BJ86K9QYXTSKF4 | "child1-trigger-4"            |                      2.1748 |
| 01J08D2R247TPGM3G13S12WSF0 | "child3-trigger-3"            |                      2.0020 |
| 01J08D2Q3Q190JYZ2BPD7HXM99 | "child2-trigger-3"            |                      1.9766 |
| 01J08D2CWTY0W936Y1JHXF7MNC | "child1-trigger-3"            |                      2.0020 |
| 01J08D1AXYD5KJ3ZDZ00MQCC9W | "child3-trigger-2"            |                      1.8291 |
| 01J08D1A0FGXCHW40WS64X439S | "child2-trigger-2"            |                      1.8037 |
| 01J08D0ZY2WK7D5KJN81CXC010 | "child1-trigger-2"            |                      1.8291 |
| 01J08CZY1F6S039YNYT5JWV5M9 | "child3-trigger-1"            |                      1.6563 |
| 01J08CZX26J5W9C136NM8EVFH2 | "child2-trigger-1"            |                      1.6309 |
| 01J08CZJSAEZ15MQ4C3W7S3KCN | "child1-trigger-1"            |                      1.4844 |
| 01J08CZHSBBDYHVVX3J7YTCSWS | "parent"                      |                      1.5371 |
+----------------------------+-------------------------------+-----------------------------+
16 rows in set (0.002 sec)
```

Clearly the benefit is that we store way less data per nested pipeline level. We are roughly storing the execution context of the nested pipeline as if it is a standalone pipeline + some small data for the pipelineRef.


@edgarulg based on the testing i see that more data are returned/deserialised by the Orca than needed after the feature enabled. I will share some examples internally but for the above example see the returned size from the executions API: 

With the feature disabled:
```
TimeConnect:0.229900
TimeStartTransfer:1.206076
TimeTotal:1.210140
SizeDownload:4759
```

With the feature enabled:
```
TimeConnect:0.428730
TimeStartTransfer:1.738250
TimeTotal:2.510231
SizeDownload:282096
```

I have the contexts available and i will share them with you and we can sync internally.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/ExecutionMapper.kt
```diff
@@ -132,4 +137,27 @@ class ExecutionMapper(
           }
       )
   }
+
+  @VisibleForTesting
+  fun convertPipelineRefTrigger(execution: PipelineExecution, context: DSLContext) {
```
christosarvanitis, line 142: This is tied to the comment that although we are storing less data we are returning more data with this feature enabled. 

Maybe return a data object that matches the same data with enabled/disabled the feature?
In the future this will help us trim down the data returned only to what is necessary gaining more improvements.

christosarvanitis: Basically the `stages` needs to be dropped from each parentExecution and this will match the previous returned object.

orca-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
edgarulg: sure, I doubted on creating a config class because it is only a single boolean but I will add it to an existing config class

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/ExecutionMapper.kt
edgarulg: it is the stageReadSize. I set it to 200 because it is the same value we have in `SqlExecutionRepository` but now that you mentioned it, it is not clear what is that 200 use for. I will name the parameter to give more context.

edgarulg:
> Trying to understand more about [this commit](https://github.com/spinnaker/orca/pull/4749/commits/1673c7570b4d710210af1a2c1066bc6325200996)
> 
> ```
> perf(pipelineRef): define deserializer as CustomTriggerDeserializerSupplier and move changes in orca-core to orca-sql.
> ```
> 
> What's the motivation for this change? On the surface it looks like more of a refactor than a perf change.

That is my mistake in the commit message. It it a refactor I did from my initial version. I defined custom deserializer in orca-core but then I figured it out how to move it into orca-sql module.

edgarulg:
> This is super exciting! Did you find specific bugs with [the original implementation](https://github.com/spinnaker/orca/pull/3986)? Or maybe you took a different approach? I'm curious what's different.

@dbyron-sf 
I am using the original implementation and the bug I found on it was related on the deserialized. The `PipelineRefTriggerDeserializerSupplier` is created as a Spring bean but it wasn't added as part of the `customTriggerSuppliers` in orca-core. Currently the `customTriggerSuppliers` are part of a companion object and they are initialized empty and there isn't a mechanism in Orca that will add `CustomTriggerDeserializerSupplier` beans in the companion object.

So the original implementation store executions as reference  properly but on retrieve executions orca-core did not know about `PipelineRefTriggerDeserializerSupplier` and it was deserializing the custom trigger as a DefaultTrigger. So it was very bad because the the execution context was not retrieved correctly and it broke Spels evaluation, and everything in his path.

Fixing the deserializer issue did the trick.

orca-core-tck/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/PipelineExecutionRepositoryTck.groovy
```diff
@@ -596,8 +596,12 @@ abstract class PipelineExecutionRepositoryTck<T extends ExecutionRepository> ext
 
   def "parses the parent execution of a pipeline trigger"() {
     given:
+    def parent = pipeline()
```
dbyron-sf, line 599: It makes me slightly nervous that an existing test is changing as part of this.  Does this mean the feature is enabled by default, or?  What happens if this test doesn't change?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/CustomTriggerDeserializerSupplier.kt
```diff
@@ -29,5 +30,5 @@ import com.netflix.spinnaker.orca.api.pipeline.models.Trigger
  */
 interface CustomTriggerDeserializerSupplier {
   val predicate: (node: JsonNode) -> Boolean
-  val deserializer: (node: JsonNode) -> Trigger
+  val deserializer: (node: JsonNode, parser: JsonParser) -> Trigger
```
dbyron-sf, line 33: Update the javadoc to mention JsonParser?

nemesisOsorio:
Can you please validate if this works using Pipeline stages?

edgarulg:
> Can you please validate if this works using Pipeline stages?

I tested it using the pipeline stage in Spinnaker but based on the unit test it is hard to write a unit test for that case. It should be a e2e test but we do not have such that thing in OSS as long as I know. Please let me know if we are talking about the same pipeline stage to be sure.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/ExecutionMapper.kt
edgarulg: I tested the behavior of orca in master and every time the SqlExecutionRepository retrieves an execution it contains all the stage's objects.
SqlExecutionRepository store the pipeline execution without stages because it split up the store for stages. In a few words the store is done in chunks (pipelines first, then stages) and for every update it is responsible on update pipeline executions & stage executions. But the retrieve always returns the full execution context including stages and all their data related.

If I drop the stage's data from a parentExecution it will lead to issues in other places in Orca because the SpEL expressions might fail because the data of the stages is not present and even in other places that I still don't know

I do think we need to improve more all the data retrieved for executions but that should be in another performance improvement.

orca-core-tck/src/main/groovy/com/netflix/spinnaker/orca/pipeline/persistence/PipelineExecutionRepositoryTck.groovy
edgarulg: This is wrong. I got this change from the original implementation but my changes are behind the feature flag and everything should work as expected. I am deleting this update because it is not needed.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/CustomTriggerDeserializerSupplier.kt
edgarulg: How the javadoc is updated? I am not sure how to do that.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
```diff
@@ -0,0 +1,59 @@
...
+              notifications = get("notifications")?.listValue(parser) ?: mutableListOf(),
+              isRebake = get("rebake")?.booleanValue() == true,
+              isDryRun = get("dryRun")?.booleanValue() == true,
+              isStrategy = get("strategy")?.booleanValue() == true,
+              parentExecutionId = get("parentExecutionId").textValue(),
+              parentPipelineStageId = get("parentPipelineStageId")?.textValue()
+            )
+          }
+        }
+        else -> throw SystemException("Trigger does not look like a pipelineRef trigger")
```
xibz, line 44: I wonder if we should be throwing an exception here? How does the code get here?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/PipelineRefTrigger.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.netflix.spinnaker.orca.api.pipeline.models.PipelineExecution
+import com.netflix.spinnaker.orca.api.pipeline.models.Trigger
+import com.netflix.spinnaker.orca.pipeline.model.PipelineTrigger
+
+/**
+ * A trigger type used by SQL implementations to store a [PipelineTrigger] without embedding the entire parent
+ * [PipelineExecution]. Instead of embedding, only the execution ID is stored, and then eagerly loaded at
+ * deserialization time.
+ */
+data class PipelineRefTrigger(
```
xibz, line 20: One thing I really dont like is if our pipeline structure changed, we would have to remember to update this as well, right? We should re-think about this if that's the case, which I think it is. This is one area I can see a bug happening

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -869,7 +872,36 @@ class SqlExecutionRepository(
...
       // Restore original object state.
       execution.stages.addAll(stages)
+      pipelineTrigger?.let {
+        execution.trigger = it
+      }
+    }
+  }
+
+  /**
+   * Converts a [PipelineTrigger] into a [PipelineRefTrigger] for storage.
```
xibz, line 882: I believe this only handling state of pipeline executions. I wonder if this is the best place for this. I dont want to overload the class more than needed, and have it be focused for its one particular job.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -1198,34 +1230,8 @@ class SqlExecutionRepository(
       .let { seek(it) }
   }
 
-  private fun DSLContext.selectExecution(type: ExecutionType, fields: List<Field<Any>> = selectFields()): SelectJoinStep<Record> {
```
xibz, line 1201: Why did we move these two methods?

xibz (changes requested):
@edgarulg - Thanks for the PR! I had some questions and suggestions. Please let me know what you think.

orca-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
```diff
@@ -161,4 +170,13 @@ class SqlConfiguration {
   fun lockProvider(datasource: DataSource): LockProvider {
     return JdbcTemplateLockProvider(datasource)
   }
+
+  @Bean
+  @ConditionalOnProperty("execution-repository.sql.pipeline-ref.enabled")
+  fun pipelineRefTriggerDeserializer(@Qualifier("mapper") mapper: ObjectMapper): CustomTriggerDeserializerSupplier {
```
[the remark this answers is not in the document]
edgarulg: good catch. I am not using the mapper object, I left it from a previous version I was testing. my bad.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
edgarulg: I throw an exception here because I left the default case but it isn't necessary because the `predicate` function is ensuring that the trigger is a PipelineRef. There is no way the code will reach out the exception.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/PipelineRefTrigger.kt
edgarulg: I see what you mean but the PipelineRefTrigger extends of the main Trigger class so any change in the base trigger class shouldn't break PipelineRefTrigger. An option that comes to my mind is I can try to extend from PipelineTrigger class but I am not seeing a real benefit on doing that.

I would say the best option will be update the structure of the PipelineTrigger to have the `parentExecutionId` but that will be a big change on the behavior and the feature flag would be useless.

I am open to option you have in mind as well :)

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
edgarulg: I moved to `jooq.kt` class because I required these functions in the `ExecutionMapper` class and I want to reuse them.
The ExecutionMapper needs these function because as the PipelineRef feature can work together with the Compression feature the mapper needs to know if it will get the executionParent from the `pipeline` table or the `compressedPipeline` table

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
```diff
@@ -0,0 +1,59 @@
...
+class PipelineRefTriggerDeserializerSupplier : CustomTriggerDeserializerSupplier {
+  override val predicate: (node: JsonNode) -> Boolean
+    get() = { node ->
+      node.isPipelineRefTrigger()
+    }
+
+  override val deserializer: (node: JsonNode, parser: JsonParser) -> Trigger
+    get() = { node, parser ->
+      when {
+        node.isPipelineRefTrigger() -> {
```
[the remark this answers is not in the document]
edgarulg: ok, I will try it out and see if I can use the ObjectMapper

edgarulg:
@xibz thanks for the feedback! I respond most of your comments and something I will try to change is using the ObjectMapper to handle the deserialization part. I will give it a shot and will let you know

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/PipelineRefTrigger.kt
xibz: Hmm, let me think about this some and Ill get back to you on this. I want to look into the various options. I definitely dont want to block this PR if a major change is needed. However, I want to first exhaust a few options before moving forward. I may need a couple days, if that's okay, or if some other reviewer could do it, then Id be okay with that as well.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
xibz: Okay, can we add a comment to note that this exception should never be thrown, and if it is something clearly is wrong

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/PipelineRefTrigger.kt
xibz: Okay, after looking I think the best approach may be to wrap the `PipelineTrigger` class so we don't need to maintain two separate pipeline models.

My thoughts were, mind you I did not spend too much thinking about how valid this would be
```java
data class PipelineRefTrigger(
    private val pipelineTrigger: PipelineTrigger // The change I care about most is here
    // then whatever relevant references here
```

This should simplify this class tremendously and allow us to only need to manage one model. Further with the object mapper change, this becomes pretty easy to handle as well :)

edgarulg:
@xibz I've pushed more changes, and this time I'm using objectMapper to modify the PipelineTrigger.

I tried to adjust how the trigger is modified in SqlExecutionRepository, but after spending a lot of time analyzing the execution flows in Orca, I discovered the following:

1. All triggers have an other object, which I assume is used to include extra fields during serialization. However, this causes an issue because once the trigger is deserialized, all the fields in the trigger are duplicated in the other object, leading to duplicated fields in the database.
2. The deserialization step happens multiple times. - I’m not sure why, but in orca-core, there are multiple calls to the objectMapper, resulting in the same trigger being (de)serialized multiple times before it's stored in the database. This means it's deserializing something that has already been deserialized.
3. Deserialization only occurs on insert. - I had to modify the execution again in SqlExecutionRepository because when we insert the execution for the first time, the mapper handles the trigger correctly. However, during updates, we retrieve the execution from the database and then update it. Based on the new flow I’m proposing, when a PipelineTrigger is stored, I change it to a PipelineRef, but upon retrieval, if the trigger is a PipelineRef, I change it back to a PipelineTrigger. Since we are retrieving the PipelineTrigger from PipelineRefTrigger, Orca updates the executions with the full execution context again.

I still need to update my tests, but I would like to hear your thoughts on this approach.

xibz:
> The deserialization step happens multiple times. - I’m not sure why, but in orca-core, there are multiple calls to the objectMapper, resulting in the same trigger being (de)serialized multiple times before it's stored in the database. This means it's deserializing something that has already been deserialized.

So this tells me someone probably didn't understand how the visitor pattern works, and created a cyclic reference to various deserializers. That's definitely a bug, but not introduced by this change, right? If it wasn't introduced in this change, I would suggest creating an issue and keep this PR more focus on the implementation of the feature than trying to fix everything

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/CustomTriggerDeserializerSupplier.kt
```diff
@@ -29,6 +29,9 @@ import com.netflix.spinnaker.orca.api.pipeline.models.Trigger
  * TODO(rz): Refactor so that JsonNode is not needed (for orca-api compat)
```
xibz, line 29: lol at this comment

but yea, hence why I wanted to use something that was higher level than `JsonNode`.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
```diff
@@ -42,6 +42,12 @@ class TriggerDeserializer :
   override fun deserialize(parser: JsonParser, context: DeserializationContext): Trigger =
     parser.codec.readTree<JsonNode>(parser).run {
       return when {
+        looksLikeCustom() -> {
```
xibz, line 45: Why did we move this? For priority?

Hmm, that could cause a lot of issues that may not be obvious. I imagine custom should be one of the least priority-wise deserializers. Why did we make it first?

I think we would need to run this in a more robust environment. Unfortunately Im not comfortable in our current tests that might ensure this will not be a breaking change. Can you give me reasons on why this needed to be moved?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
```diff
@@ -209,6 +219,6 @@ class TriggerDeserializer :
     hasNonNull("pluginId")
 
   private fun JsonNode.looksLikeCustom() =
-    customTriggerSuppliers.any { it.predicate.invoke(this) }
+    customTriggerSuppliers.any { it.predicateByNode.invoke(this) }
```
xibz, line 222: Why did we change it? We just are changing the name? Yea, let's not do that for git history reasons.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
```diff
@@ -9,42 +9,70 @@ package com.netflix.spinnaker.orca.sql
...
+              isRebake = get("rebake")?.booleanValue() == false,
+              isDryRun = get("dryRun")?.booleanValue() == false,
+              isStrategy = get("strategy")?.booleanValue() == false,
+              parentExecutionId = get("parentExecution").get("id").textValue(),
+              parentPipelineStageId = get("parentPipelineStageId").textValue()
+            )
+          }
+        }
+        else -> {
+          with(node) {
```
xibz, line 54: Why do we have an else that does the same thing?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -783,7 +783,7 @@ class SqlExecutionRepository(
   private fun storeExecutionInternal(ctx: DSLContext, execution: PipelineExecution, storeStages: Boolean = false) {
     validateHandledPartitionOrThrow(execution)
 
-    val pipelineTrigger = mutatePipelineTrigger(execution)
+    execution.trigger = mapper.convertValue(execution.trigger, Trigger::class.java)
```
xibz, line 786: :+1: on utilizing converters. This is the perfect use case for it. However, Id register it on the mapper instead of calling it directly. That way this code doesn't change.

What needs to happen here is a bean needs to be specific for the sql execution repository. This bean must register **two** converters.

1st converter will convert `PipelineTrigger` -> `PipelineTriggerRef`
2nd converter goes the other way `PipelineTriggerRef` -> `PipelineTrigger`

If we go with this approach we may not even need to touch that crazy trigger deserializer and do things much betteer.

edgarulg:
>> Can you expand on this some? This seems... odd? Can you point us to some code and give some examples? I think this further illustrates the design needing to be reworked some (out of scope of this PR).
> 
> By default, when a Trigger is deserialized, all fields from the JsonParser are added to the other object. You can see this here: https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt#L183. Additionally, the TriggerMixin used for serializing each trigger employs the @JsonAnyGetter annotation, which includes all fields in the other object during serialization. This can be seen here: https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/jackson/mixin/TriggerMixin.java#L37. As a result, every Trigger stored in the database ends up with duplicated fields.
```
    "trigger": {
        "type": "manual",
        "user": "[anonymous]",
        "parameters": {
            "root-name": "root"
        },
        "artifacts": [],
        "notifications": [],
        "rebake": false,
        "dryRun": false,
        "strategy": false,
        "resolvedExpectedArtifacts": [],
        "type": "manual",
        "user": "[anonymous]",
        "parameters": {
            "root-name": "root"
        },
        "artifacts": [],
        "notifications": [],
        "rebake": false,
        "dryRun": false,
        "strategy": false,
        "resolvedExpectedArtifacts": [],
        "expectedArtifacts": []
    },
```

>>So this tells me someone probably didn't understand how the visitor pattern works, and created a cyclic reference to various deserializers. That's definitely a bug, but not introduced by this change, right? If it wasn't introduced in this change, I would suggest creating an issue and keep this PR more focus on the implementation of the feature rather than trying to fix everything
> 
> Yes, that is a bug that I believe it's been around but it was "failing successfully".  The (de)serialization is done twice before the execution is store because in the following line: https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncher.java#L91
The ExecutionLancher initializes the store of the execution but it parse the config object. The config object already was processed by the mapper and it has all the datatypes from the (de)serialization. So the following line in the ExecutionLauncher is useless because we already have everything mapped in the config object: https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncher.java#L217
>> Yes, this is similar to how the artifact store works and makes things completely backwards compatible, so would 100% support that approach.
> Glad to see I am going to the right way

xibz (changes requested):

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
edgarulg: I moved it at the beginning for priority. As I am defining a custom deserializer that will override the PipelineTrigger If it is not at the beginning the deserializer won't convert a PipelineTrigger into a PipelineRefTrigger.

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/support/TriggerDeserializer.kt
edgarulg: I changed it because I defined two predicates.

- predicateByNode will receive a JsonNode and will return a boolean if the customDeserializer needs to handle the trigger.
- predicateByTrigger will receive a Trigger and will return a boolean if the customDeserializer needs to handle the trigger.

I use the definition to workaround the issue with the other object in the Trigger.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
edgarulg: I must to return the trigger always. If I do not return a trigger object the Trigger will be deserializer as a Default Trigger and that is wrong.
My code is ugly here. I might the same object and conditionally populate the parentExecutionRef.

orca-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
dbyron-sf: @edgarulg I still see `@Value` here...  By the way, is the config knob here completely independent from compression?

edgarulg: Yes, the pipelineRef config is independent to the compression feature. Let me remove the `@Value`

orca-sql/src/main/kotlin/com/netflix/spinnaker/config/PipelineRefProperties.kt
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2024 Harness Inc. All rights reserved.
+ * Use of this source code is governed by the PolyForm Free Trial 1.0.0 license
```
dbyron-sf, line 3: This is a different license than the Apache license we use elsewhere.  Mind changing it?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2024 Harness Inc. All rights reserved.
+ * Use of this source code is governed by the PolyForm Free Trial 1.0.0 license
```
dbyron-sf, line 3: Same license question here.

dbyron-sf:
For the record, I'd say this 100% closes https://github.com/spinnaker/spinnaker/issues/6159.

orca-sql/src/main/kotlin/com/netflix/spinnaker/config/PipelineRefProperties.kt
edgarulg: I updated license on the new files

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
edgarulg: updated

dbyron-sf:
@edgarulg This looks good to me.  I have one request though.  There are some snippets of this in comments above, but I don't see the particular thing I'm looking for.  I'd like to see a before/after of a pipeline execution context with this feature enabled to see exactly which bits disappear and what the reference looks like.  If I understand correctly, the in-memory representation doesn't change, it's only the storage in the database.  Do I have that right?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/PipelineRefTrigger.kt
xibz: For context, this was closed cause it turned out that if we add the `PipelineTrigger` as a field, it causes the deserializer to not work properly. Rather than trying to find out all the key details, Im fine with having it as is, and fixing this later with some jackson magic to keep things backwards compatible.

xibz (approved):
@edgarulg Thanks for all the updates Edgar! This looks really good. Thanks for putting up with my strict reviews :p, but I think we both learned a lot about how triggers are serialized and deserialized. I dont have any further comments, but please make sure any other comments from @dbyron-sf is also addressed

dbyron-sf:
Excellent.  Please write a release note!

dbyron-sf (approved):

jasonmcintosh:
We DO need to fix the license though!  Polyform isn't applicable for spinnaker.  I'd like to get some updated metrics as well.

christosarvanitis:
> For the record, I'd say this 100% closes [spinnaker/spinnaker#6159](https://github.com/spinnaker/spinnaker/issues/6159).

@dbyron-sf I would argue that this is partially true :) We do store way less data in the database that makes Orca faster in writes and reads to the database which is a huge win! 
But since the returning object to Gate and the end-user remains the same it wont solve the problem that for the end-user the application view load will be slow on heavy nested pipelines. I could see a next round of enhancements on this feature to instead Orca substituting in-memory the `PipelineTriger` and returning the full object to the end-user to return the `PipelineRefTrigger`  and Deck to handle that.

dbyron-sf:
> > For the record, I'd say this 100% closes [spinnaker/spinnaker#6159](https://github.com/spinnaker/spinnaker/issues/6159).
> 
> @dbyron-sf I would argue that this is partially true :) We do store way less data in the database that makes Orca faster in writes and reads to the database which is a huge win! But since the returning object to Gate and the end-user remains the same it wont solve the problem that for the end-user the application view load will be slow on heavy nested pipelines. I could see a next round of enhancements on this feature to instead Orca substituting in-memory the `PipelineTriger` and returning the full object to the end-user to return the `PipelineRefTrigger` and Deck to handle that.

Good point @christosarvanitis .  And yes, I agree on the next round.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
```diff
@@ -24,13 +24,19 @@ import com.netflix.spinnaker.orca.pipeline.model.support.mapValue
...
+  private val pipelineRefEnabled: Boolean
+) : CustomTriggerDeserializerSupplier {
 
   override val type: String = "pipelineRef"
 
   override val predicate: (node: JsonNode) -> Boolean
     get() = { node ->
-      node.looksLikePipeline() || node.isPipelineRefTrigger()
+      if (pipelineRefEnabled) {
+        node.looksLikePipeline() || node.isPipelineRefTrigger() //if pipelineRef enabled we deserialize PipelineTrigger as PipelineRefTrigger
```
dbyron-sf, line 36: Not quite sure I follow this.  Don't we always want the same deserialize logic, and the flag controls what happens during serialization?

edgarulg: This condition guarantees that the triggers saved under pipelineRef can still be processed even when the feature flag is turned off. Therefore, if the flag is disabled, Deck will display executions that include pipelineRef without issues

dbyron-sf: I still don't understand why the deserialization logic needs to be different depending on pipelineRefEnabled.

edgarulg:
@dbyron-sf I wrote the release notes for this feature flag: https://github.com/spinnaker/spinnaker.io/pull/439
Can you take a look when you have a chance please?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/PipelineRefTriggerDeserializerSupplier.kt
edgarulg: When the feature flag is enabled:

The deserializer is set up to handle both `PipelineTrigger` and `PipelineRefTrigger`.
This means that whenever Orca encounters a trigger that resembles either a `PipelineTrigger` or `PipelineRefTrigger`, it treats it as a `PipelineRefTrigger`.
Storing triggers as `PipelineRefTrigger` ensures that every execution containing a `PipelineTrigger` will store its trigger data in a format that allows the complete execution context to be retrieved later.

When the feature flag is disabled:

Orca's database may contain executions that were stored using `PipelineRefTrigger`.
Without the feature flag enabled, Orca cannot automatically deserialize a `PipelineRefTrigger`.
To maintain compatibility with older executions that used `pipelineRefTrigger` when the feature flag was enabled, the deserializer exclusively interprets `PipelineRefTrigger`.
This approach enables Orca to continue retrieving the full execution context for executions that were stored while the feature flag was active.
Additionally, new `PipelineTriggers` will no longer be stored as `PipelineRefTriggers`, ensuring consistent behavior going forward.

Therefore, in the CustomDeserializer, if the feature flag is enabled, both `PipelineTrigger` and `PipelineRefTrigger` are processed.

jasonmcintosh (approved):

edgarulg:
@dbyron-sf I updated my tests to prove my latest changes regarding deserializer. Please let me know if you have questions about it.

PR #4753, opened 2024-06-22
fix(build): remove dokka plugin since it doesn't work with java 17

dbyron-sf:
From e.g. https://github.com/spinnaker/orca/actions/runs/9507661391/job/26207523287:
```
Caused by: java.lang.NoSuchMethodError: 'org.jetbrains.dokka.model.DInterface org.jetbrains.dokka.model.DInterface.copy$default(org.jetbrains.dokka.model.DInterface, org.jetbrains.dokka.links.DRI, java.lang.String, java.util.Map, org.jetbrains.dokka.DokkaConfiguration$DokkaSourceSet, java.util.Map, java.util.List, java.util.List, java.util.List, java.util.Map, org.jetbrains.dokka.model.DObject, java.util.List, java.util.Map, java.util.Set, boolean, org.jetbrains.dokka.model.properties.PropertyContainer, int, java.lang.Object)'
```
Note that this github action didn't do anything when it succeeded.  From e.g. https://github.com/spinnaker/orca/actions/runs/9505579020/job/26200789091:
```
Branch 'create-pull-request/patch' is not ahead of base 'master' and will not be created
```


jasonmcintosh (approved):

jasonmcintosh:
SHOULD work on java 17 - but don't think we're using the output of it soo :P

PR #4760, opened 2024-07-04
test(sql): Add extra test  for pipelineRef feature flag

edgarulg:
I added more tests in executionLaucherTest regarding PipelineRef feature include as part of https://github.com/spinnaker/orca/pull/4749

### DIFF
--- a/orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncherTest.java
+++ b/orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncherTest.java
@@ -27,7 +27,12 @@ import com.netflix.spinnaker.kork.common.Header;
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType;
 import com.netflix.spinnaker.orca.api.pipeline.models.PipelineExecution;
 import com.netflix.spinnaker.orca.config.ExecutionConfigurationProperties;
+import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper;
+import com.netflix.spinnaker.orca.pipeline.model.PipelineTrigger;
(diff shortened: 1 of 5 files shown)


orca-core/src/test/java/com/netflix/spinnaker/orca/pipeline/ExecutionLauncherTest.java
```diff
@@ -332,6 +337,120 @@ public void testExcludeSpinnakerAccountsFromPipeline() throws Exception {
     assertThat(pipelineExecution.getAuthentication().getAllowedAccounts()).isEqualTo(Set.of());
   }
 
+  @DisplayName(
+      "when pipelineRef.enabled: true, the object mapper can deserialize the PipelineTrigger into PipelineRefTrigger")
```
dbyron-sf, line 341: Isn't the operation here about converting from pipeline config to pipeline execution?  Maybe this isn't 100% of what's happening since triggers come in from manual execution, and from parent pipeline executing children.  But, these triggers never have full on execution contexts, do they?  I mean, the pipeline in question hasn't run yet?  If that's true, wouldn't we expect the same behavior whether pipelineRef.enabled is true or not?

edgarulg: In the ExecutionLauncher the pipeline hasn't run yet. The purpose of the ExecutionLauncher is take a pipeline config and parse it into a execution context then store it using SqlExecutionRepository.

The main idea of my new tests is that even with the customTrigger configured in the objectMapper the ExecutionLauncher can start a new execution. I just include an assert to validate that the customTrigger is being use as well but the same behavior is kept.
I will change the description on the test to avoid implementation details

dbyron-sf (approved):

PR #4778, opened 2024-08-28
chore(runJobStage): Fix logs annotation processing on manifest artifact

christosarvanitis:
In the RunJob Stage when the annotation:
```
job.spinnaker.io/logs: "someExternalUrl"
```
is present the expectation is that the context will be amended with:
```
"execution": {
"logs": "someExternalUrl"
}
```
which is rendered by Deck to an external URL for the console logs.

This is working as expected when the source is text (inline manifest) but when the manifest is coming from an artifact store the process needs to happen afterStages because only then the manifest content is present in the context

This PR addresses this missing use case.


dbyron-sf:
What's involved in adding a test for this?

christosarvanitis:
@dbyron-sf pushed some tests for the stage

jasonmcintosh (approved):

PR #4779, opened 2024-09-06
fix(orca-sql): Add migration for mariadb support

juangod-wise:
This contribution was discussed in this slack thread: 

https://spinnakerteam.slack.com/archives/C7J1CU4LA/p1722343931881969?thread_ts=1710447204.964799&cid=C7J1CU4LA

In summary: 

The migration in `20201214-create-compressed-executions-table.yaml` that is adding the column `compression_type` is not being ran for MariaDB because it contains `dbms: mysql`. This causes Orca to throw the following error on startup if using MariaDB:

```
liquibase.exception.DatabaseException: Unknown data type: 'COMPRESSION_TYPE_ENUM' [Failed SQL: (4161) ALTER TABLE spinnaker_orca.pipelines_compressed_executions ADD compression_type COMPRESSION_TYPE_ENUM DEFAULT 'ZLIB' NOT NULL
```

This happens because, since the column is not being created on that migration, the Postgres migration is ran because of this conditional:

https://github.com/spinnaker/orca/blob/ad53c1762f8a63dea8fea7b80517b5063d4109d7/orca-sql/src/main/resources/db/changelog/20220427-add-compression-type-column-to-compressed-executions-tables-postgres.yaml#L13-L18

And running the Postgres migration on MariaDB causes the error to appear. The solution for this is to create a new migration for MariaDB that is executed before the Postgres migration that adds the column only if using MariaDB. We can't change previous migrations because the changeset hash would change and would start failing for everyone who already applied it.

Spinnaker should be able to support MariaDB as there are mentions in the documentation:

https://spinnaker.io/docs/setup/productionize/persistence/orca-sql/#mariadb

### DIFF
--- a/orca-sql/src/main/resources/db/changelog/20240731-add-compression-type-column-mariadb.yaml
+++ b/orca-sql/src/main/resources/db/changelog/20240731-add-compression-type-column-mariadb.yaml
@@ -1,0 +1,29 @@
+databaseChangeLog:
+  - changeSet:
+      id: create-pipelines-compressed-executions-mariadb
+      author: juangod-wise
+      changes:
+        - sql:
+            dbms: mariadb
+            sql: ALTER TABLE `pipelines_compressed_executions` ADD COLUMN `compression_type` ENUM("GZIP", "ZLIB") NOT NULL DEFAULT "ZLIB"
+  - changeSet:
+      id: create-pipeline-stages-compressed-executions-mariadb
+      author: juangod-wise
+      changes:
+        - sql:
+            dbms: mariadb
+            sql: ALTER TABLE `pipeline_stages_compressed_executions` ADD COLUMN `compression_type` ENUM("GZIP", "ZLIB") NOT NULL DEFAULT "ZLIB"
+  - changeSet:
+      id: create-orchestrations-compressed-executions-mariadb
+      author: juangod-wise
+      changes:
+        - sql:
+            dbms: mariadb
+            sql: ALTER TABLE `orchestrations_compressed_executions` ADD COLUMN `compression_type` ENUM("GZIP", "ZLIB") NOT NULL DEFAULT "ZLIB"
+  - changeSet:
+      id: create-orchestrations-stages-compressed-executions-mariadb
+      author: juangod-wise
+      changes:
+        - sql:
+            dbms: mariadb
+            sql: ALTER TABLE `orchestration_stages_compressed_executions` ADD COLUMN `compression_type` ENUM("GZIP", "ZLIB") NOT NULL DEFAULT "ZLIB"
--- a/orca-sql/src/main/resources/db/changelog/20240731-add-compression-type-column-mariadb.yaml
+++ b/orca-sql/src/main/resources/db/changelog/20240731-add-compression-type-column-mariadb.yaml
@@ -1,27 +1,27 @@
 databaseChangeLog:
   - changeSet:
-      id: create-pipelines-compressed-executions-mariadb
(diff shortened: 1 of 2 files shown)


orca-sql/src/main/resources/db/changelog/20240731-add-compression-type-column-mariadb.yaml
```diff
@@ -0,0 +1,29 @@
+databaseChangeLog:
+  - changeSet:
+      id: create-pipelines-compressed-executions-mariadb
```
dbyron-sf, line 3:
```suggestion
      id: add-compression-type-column-to-pipelines-compressed-executions-mariadb
```
and similar changes below.

orca-sql/src/main/resources/db/changelog/20240731-add-compression-type-column-mariadb.yaml
```diff
@@ -0,0 +1,29 @@
...
+      author: juangod-wise
+      changes:
```
dbyron-sf, line 5: Do we need something like:
```
      preConditions:
        onFail: MARK_RAN
        not:
          columnExists:
            tableName: pipelines_compressed_executions
            columnName: compression_type
```
like the [postgres change](https://github.com/spinnaker/orca/blob/eb61f0e25330b2c2a34d543e63e59f6be95de859/orca-sql/src/main/resources/db/changelog/20220427-add-compression-type-column-to-compressed-executions-tables-postgres.yaml#L13-L18) has?

juangod-wise: I don't believe this is necessary because in MariaDB with the current given migrations no column would have been created. The migration won't be ran twice because of the changeset checksums. In Wise we have this migration in Prod for a couple of months now and we had no issue. The only scenario I can think of where it may be useful is if someone pulls this migration in, has MariaDB, and already has another migration that created the same column. But it would have been amended when the OSS changes are merged in.. I could add it if you want, to be extra careful. As you wish :) .

dbyron-sf: OK, I think I see it.  The [postgres changeset](https://github.com/spinnaker/orca/blob/eb61f0e25330b2c2a34d543e63e59f6be95de859/orca-sql/src/main/resources/db/changelog/20220427-add-compression-type-column-to-compressed-executions-tables-postgres.yaml#L10) isn't actually specific to postgres. It doesn't have something like
```
      preConditions:
        onFail: CONTINUE
        dbms:
          type: postgres
```
I think we need something like that here.  There could easily be a better way to express it.

dbyron-sf: Yes please.  The important thing is to add a precondition that this changeset only runs for mariadb, and then if the precondition fails, to continue.

juangod-wise: But by having `dbms: mariadb` on all changesets, it already only runs for mariadb. How do we know this? Because this bug happened because `20201214-create-compressed-executions-table.yaml` did not run the changesets that contained 

```
    - sql:
          dbms: mysql
```

dbyron-sf: fair enough.

dbyron-sf (approved):

PR #4783, opened 2024-09-24
feat(clouddriver): make fetching properties file more resilient for k8s jobs

kirangodishala:
Spinnaker doesn't handle retrieving property file contents properly when a job has multiple pods. Under the covers, it runs the kubectl logs jobs/ command, which doesn't give the right answer if there are two pods running at the same time for the same job. In such cases, kubernetes always defaults to the first pod, but in certain scenarios, we would want it to query the second pod for logs as the first one may have failed but the second one may have succeeded.

This PR enables getting logs from a successful pod directly in case the original kubectl logs jobs call failed.

This PR relies on https://github.com/spinnaker/clouddriver/pull/5184 & https://github.com/spinnaker/clouddriver/pull/5778 to enable it to get logs from a pod directly.


dbyron-sf:
@mergifyio update

dbyron-sf (approved):

PR #4803, opened 2024-11-20
feat(sql): use a connection pool named "read" for some read operations in SqlExecutionRepository

kirangodishala:
Introduce a read pool for some read-only database queries in SqlExecutionRepository.  Enable this by configuring an additional connection pool named `"read"`, like this:
```
sql:
  connectionPools:
    default:
      <...>
    read:
      jdbcUrl: jdbc:...
      user: orca_service
      password: 
      connectionTimeoutMs: 
      validationTimeoutMs: 
      maxPoolSize:
      minIdle:
      maxLifetimeMs:
      idleTimeoutMs:
```

Additional read-only database queries happen in places that expect to see "just-written" contents.  They're not safe to convert to use the read replica until there's logic in place to be aware of replication lag, and wait until the read replica is up-to-date.

### DIFF
--- a/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
+++ b/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
@@ -1275,14 +1275,10 @@ class SqlExecutionRepository(
   private fun selectExecution(
     ctx: DSLContext,
     type: ExecutionType,
-    id: String,
-    forUpdate: Boolean = false
+    id: String
   ): PipelineExecution? {
     withPool(poolName) {
       val select = ctx.selectExecution(type, compressionProperties).where(id.toWhereCondition())
-      if (forUpdate) {
-        select.forUpdate()
-      }
       return select.fetchExecution()
     }
   }
--- a/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
+++ b/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
@@ -76,12 +76,14 @@ import org.jooq.impl.DSL.table
 import org.jooq.impl.DSL.timestampSub
 import org.jooq.impl.DSL.value
 import org.slf4j.LoggerFactory
+import org.springframework.jdbc.datasource.lookup.AbstractRoutingDataSource
 import rx.Observable
 import java.io.ByteArrayOutputStream
 import java.lang.System.currentTimeMillis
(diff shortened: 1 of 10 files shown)


dzhengg (commented):
Could we include an example (maybe in one of the commit messages) for how to configure orca to use a read pool?

dzhengg (approved):

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -112,18 +114,20 @@ import kotlin.collections.toTypedArray
  * There is a small amount of MySQL/PostgreSQL specific commands inside, but
  * all of which safely fallback to common SQL if the dialect does not match.
  */
-class SqlExecutionRepository(
+  class SqlExecutionRepository(
```
dbyron-sf, line 117:
```suggestion
class SqlExecutionRepository(
```

dbyron-sf (approved):

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -677,21 +685,27 @@ class SqlExecutionRepository(
...
         .fetchExecution()
 
-      if (execution != null) {
-        if (!execution.status.isComplete) {
-          return execution
-        }
-        jooq.transactional {
-          it.deleteFrom(table("correlation_ids")).where(field("id").eq(correlationId)).execute()
-        }
+      if (execution == null) {
```
jasonmcintosh, line 688: A POTENTIAL concern since we ARE querying now from the read pool:  Can the orchestration be on a delayed sync, so it'd actually exist in the RW & not be sync'd yet?  Wondering if a fallback is needed in such a case.  OR this could be handled upstream when it doesn't find it to retry?

jasonmcintosh: Yeah it's a concern - I know of some workflows that use orchestration stuff that could get impacted by this, and read-sync issues are the "biggest" headaches of RO queries based on past.  This seems to be the biggest spot that's a POTENTIAL risk as I THINK (and still tracing) the others do retries on an operation that throws an error or doesn't respond.  It's possible these upstream may ALSO retry, but not 100% sure YET if that exception is encountered...

dbyron-sf: OK, reverted back to `withPool(poolName)`

christosarvanitis (approved):

jasonmcintosh (approved):

PR #4804, opened 2024-11-24
feat(sql): Optimize executions

kirangodishala:
This PR improves performance when using sql as the backend by optimizing the way the results are obtained for the api call `/applications/{application}/pipelines?expand=false&limit=2` frequently initiated by deck.

Deck makes that call to Gate. Gate further forwards that call to Orca, specifically to `/v2/applications/{application}/pipelines` endpoint.

Orca calls front50 to get the pipeline and strategy config ids and then merges them. Then for each id in the merged list, it queries its own db (SQL) to handle the query parameters, such as limit=2 and filter by statuses. It is observed that these db operations are taking longer times and this PR fixes the slowness by optimizing the queries to the db.

Refactored TaskController to support externalized config properties

Optimize getPipelinesForApplication() by reducing the number of relevant pipeline config ids to query

Further optimize getPipelinesForApplication() by processing multiple pipeline config ids at a time and multi-threading each config id batch 

#### Breaking changes:
These configuration properties
```yaml
tasks:
   days-of-execution-history:  
   number-of-old-pipeline-executions-to-include: 
```
are replaced by the below configuration along with few more newly added ones:
```yaml
tasks:
   controller:
      days-of-execution-history: 
      number-of-old-pipeline-executions-to-include: 

      optimize-execution-retrieval: <boolean>
      max-execution-retrieval-threads:
      max-number-of-pipeline-executions-to-process: 
      execution-retrieval-timeout-seconds: 
```

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
@@ -221,10 +221,12 @@ class DualExecutionRepository(
 
   override fun retrievePipelineExecutionDetailsForApplication(
     @Nonnull application: String,
-    pipelineConfigIds: List<String>): Collection<PipelineExecution> {
+    pipelineConfigIds: List<String>,
+    queryTimeoutSeconds: Int
+  ): Collection<PipelineExecution> {
     return (
-      primary.retrievePipelineExecutionDetailsForApplication(application, pipelineConfigIds) +
-      previous.retrievePipelineExecutionDetailsForApplication(application, pipelineConfigIds)
+      primary.retrievePipelineExecutionDetailsForApplication(application, pipelineConfigIds, queryTimeoutSeconds) +
+      previous.retrievePipelineExecutionDetailsForApplication(application, pipelineConfigIds, queryTimeoutSeconds)
       ).distinctBy { it.id }
   }
 
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
@@ -192,6 +192,13 @@ class DualExecutionRepository(
     ).distinct { it.id }
   }
 
+  override fun retrievePipelineConfigIdsForApplication(application: String): List<String> {
+    return (
+      primary.retrievePipelineConfigIdsForApplication(application) +
+      previous.retrievePipelineConfigIdsForApplication(application)
+      ).distinct()
+  }
+
   override fun retrievePipelinesForPipelineConfigId(
     pipelineConfigId: String,
     criteria: ExecutionCriteria
@@ -200,6 +207,17 @@ class DualExecutionRepository(
       primary.retrievePipelinesForPipelineConfigId(pipelineConfigId, criteria),
       previous.retrievePipelinesForPipelineConfigId(pipelineConfigId, criteria)
     ).distinct { it.id }
+  }
+
+  override fun retrievePipelineExecutionsForApplication(
+    application: String,
+    pipelineConfigIds: List<String>,
+    criteria: ExecutionCriteria
+  ): Collection<PipelineExecution> {
+    return (
+      primary.retrievePipelineExecutionsForApplication(application, pipelineConfigIds, criteria) +
+      previous.retrievePipelineExecutionsForApplication(application, pipelineConfigIds, criteria)
+      ).distinctBy { it.id }
   }
 
   override fun retrievePipelinesForPipelineConfigIdsBetweenBuildTimeBoundary(
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
@@ -30,6 +30,7 @@ import org.springframework.context.ApplicationContext
 import org.springframework.context.annotation.Primary
 import org.springframework.stereotype.Component
 import rx.Observable
+import javax.annotation.Nonnull
 
 /**
  * Intended for performing red/black Orca deployments which do not share the
@@ -209,14 +210,19 @@ class DualExecutionRepository(
     ).distinct { it.id }
   }
 
-  override fun retrievePipelineExecutionsForApplication(
-    application: String,
-    pipelineConfigIds: List<String>,
-    criteria: ExecutionCriteria
-  ): Collection<PipelineExecution> {
+  override fun filterPipelineExecutionsForApplication(@Nonnull application: String,
+                                             @Nonnull pipelineConfigIds: List<String>,
+                                             @Nonnull criteria: ExecutionCriteria): List<String>{
+    return primary.filterPipelineExecutionsForApplication(application, pipelineConfigIds, criteria) +
+      previous.filterPipelineExecutionsForApplication(application,pipelineConfigIds, criteria)
+  }
+
+  override fun retrievePipelineExecutionsDetailsForApplication(
+    @Nonnull application: String,
+    pipelineConfigIds: List<String>): Collection<PipelineExecution> {
     return (
-      primary.retrievePipelineExecutionsForApplication(application, pipelineConfigIds, criteria) +
-      previous.retrievePipelineExecutionsForApplication(application, pipelineConfigIds, criteria)
+      primary.retrievePipelineExecutionsDetailsForApplication(application, pipelineConfigIds) +
+      previous.retrievePipelineExecutionsDetailsForApplication(application,pipelineConfigIds)
       ).distinctBy { it.id }
   }
 
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
@@ -210,19 +210,21 @@ class DualExecutionRepository(
     ).distinct { it.id }
   }
 
-  override fun filterPipelineExecutionsForApplication(@Nonnull application: String,
-                                             @Nonnull pipelineConfigIds: List<String>,
-                                             @Nonnull criteria: ExecutionCriteria): List<String>{
-    return primary.filterPipelineExecutionsForApplication(application, pipelineConfigIds, criteria) +
-      previous.filterPipelineExecutionsForApplication(application,pipelineConfigIds, criteria)
+  override fun retrieveAndFilterPipelineExecutionIdsForApplication(
+    @Nonnull application: String,
+    @Nonnull pipelineConfigIds: List<String>,
+    @Nonnull criteria: ExecutionCriteria
+  ): List<String> {
+    return primary.retrieveAndFilterPipelineExecutionIdsForApplication(application, pipelineConfigIds, criteria) +
+      previous.retrieveAndFilterPipelineExecutionIdsForApplication(application, pipelineConfigIds, criteria)
   }
 
-  override fun retrievePipelineExecutionsDetailsForApplication(
+  override fun retrievePipelineExecutionDetailsForApplication(
     @Nonnull application: String,
     pipelineConfigIds: List<String>): Collection<PipelineExecution> {
     return (
-      primary.retrievePipelineExecutionsDetailsForApplication(application, pipelineConfigIds) +
-      previous.retrievePipelineExecutionsDetailsForApplication(application,pipelineConfigIds)
+      primary.retrievePipelineExecutionDetailsForApplication(application, pipelineConfigIds) +
+      previous.retrievePipelineExecutionDetailsForApplication(application, pipelineConfigIds)
       ).distinctBy { it.id }
   }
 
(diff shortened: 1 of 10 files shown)


orca-web/orca-web.gradle
```diff
@@ -94,6 +94,16 @@ dependencies {
   testImplementation("io.strikt:strikt-core")
   testImplementation("io.mockk:mockk")
   testImplementation("org.apache.groovy:groovy-json")
+  testImplementation("com.nhaarman:mockito-kotlin:1.6.0")
```
dbyron-sf, line 97:
```suggestion
  testImplementation("com.nhaarman:mockito-kotlin")
```
since [kork](https://github.com/spinnaker/kork/blob/6d13c519bb3163fc09d7838df06fead7647fe97b/spinnaker-dependencies/spinnaker-dependencies.gradle#L135) specifies the version.

kirangodishala: done

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/DualExecutionRepository.kt
```diff
@@ -202,6 +210,26 @@ class DualExecutionRepository(
     ).distinct { it.id }
   }
 
+  override fun retrieveAndFilterPipelineExecutionIdsForApplication(
+    @Nonnull application: String,
+    @Nonnull pipelineConfigIds: List<String>,
+    @Nonnull criteria: ExecutionCriteria
+  ): List<String> {
+    return primary.retrieveAndFilterPipelineExecutionIdsForApplication(application, pipelineConfigIds, criteria) +
```
jasonmcintosh, line 218: MINOR:  BUT ... this ordering DOES matter on the selection criteria and don't think we have a test for this (aka primary MUST come first or it'd select previous data potentially).  Just an observation that this is missing a test .. but not sure how "critical" that test really is.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -427,6 +451,130 @@ class SqlExecutionRepository(
...
+    application: String,
+    pipelineConfigIds: List<String>,
+    criteria: ExecutionCriteria
+  ): List<String> {
+
+    // baseQueryPredicate for the flow where there are no statuses in the execution criteria
+    var baseQueryPredicate = field("application").eq(application)
+      .and(field("config_id").`in`(*pipelineConfigIds.toTypedArray()))
+
+    var table = if (jooq.dialect() == SQLDialect.MYSQL) PIPELINE.tableName.forceIndex("pipeline_application_idx")
```
jasonmcintosh, line 501: Question:  DO we really want to force an index, or is another index on the combination needed?  OR perhaps merging a couple of these indexes.  NOT sure best solution but... we have a number of indexes with some of this data and wondering if we couldn't do this a touch cleaner.  I'd EXPECT the optimizer to handle MOST of this without the need to force an index.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -427,6 +451,130 @@ class SqlExecutionRepository(
...
+      baseQueryPredicate = baseQueryPredicate
+        .and(field("status").`in`(*statusStrings.toTypedArray()))
+
+      table = if (jooq.dialect() == SQLDialect.MYSQL) PIPELINE.tableName.forceIndex("pipeline_application_status_starttime_idx")
+      else PIPELINE.tableName
+    }
+
+    val finalResult: MutableList<String> = mutableListOf()
+
+    log.info("getting execution ids")
```
jasonmcintosh, line 515: NIT:  Log.info which isn't terribly useful here.  If anything maybe a log.debug level possibly with something more useful?  OR remove entirely as this I'd think would get called a LOT

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -427,6 +451,130 @@ class SqlExecutionRepository(
...
+   *    SELECT config_id, id
+        FROM pipelines force index (`pipeline_application_status_starttime_idx`)
+        WHERE (
+          application = "myapp" and
+          status in ("status1", "status2)
+        )
+        ORDER BY
+        config_id;
+
+   * It then applies the limit execution criteria on the result set obtained above. We observed load issues in the db
```
jasonmcintosh, line 487: Looking at this - and looking at the indexes, that's I SUSPECT because there's no index with config_id, id application and status.  I think a better index here would help a LOT on narrowing this down & performing better.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -583,38 +589,51 @@ class TaskController {
...
-
     statuses = statuses ?: ExecutionStatus.values()*.toString().join(",")
     def executionCriteria = new ExecutionCriteria(
       pageSize: limit,
       statuses: (statuses.split(",") as Collection)
     )
 
+    // get all relevant pipeline and strategy configs from front50
     def pipelineConfigIds = front50Service.getPipelines(application, false)*.id as List<String>
+    log.info("received ${pipelineConfigIds.size()} pipelines for application: $application from front50")
```
jasonmcintosh, line 610: Same nit on log level - this seems more useful as a debug level log.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -850,6 +870,113 @@ class TaskController {
...
+        executionCriteria
+    )
+    if (filteredPipelineExecutionIds.size() == 0) {
+      log.info("no pipeline executions found")
+      return finalResult
+    }
+
+    // need to define a new executor service since we want a dedicated set of threads to be made available for every
+    // new request for performance reasons
+    ExecutorService executorService = Executors.newFixedThreadPool(
```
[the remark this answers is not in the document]
dbyron-sf: It's not quite a thread-per-request since the number of threads is set by a config prop (default 10).

https://github.com/spinnaker/orca/pull/4803 comes next, with more to follow as we move requests to a read-only database replica.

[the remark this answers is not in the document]
jasonmcintosh: Yeah saw the properties but generating a thread pool on the request even WITH limits is still backed by a queue ... It's still shutdown but this could absolutely hammer a database badly because of a potential exponential call graph. 
 E.g. a browser can repeat calls which then would make new calls on timeouts that'd overload this badly. The RO PR should help A LOT with that however, and IN THEORY this should improve performance.  It's just a concern based upon how I've seen these API's get called.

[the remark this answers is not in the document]
dbyron-sf: Yup.  We also ended up making changes in deck to improve things.  Those are gonna be harder for us to upstream, but we'll try.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
jasonmcintosh: Can we get performance benchmarks on this?  PARTICULARLY if we re-ran those benchmarks with different indexes created and NOT having the force index section?

orca-web/src/main/java/com/netflix/spinnaker/config/TaskControllerConfigurationProperties.java
```diff
@@ -0,0 +1,80 @@
...
+   * only applicable if optimizeExecutionRetrieval = true. It specifies how many pipeline executions
+   * should be processed at a time. 150 pipeline executions was selected as the default after
+   * testing this number against an orca sql db that contained lots of pipelines and executions for
+   * a single application (about 1200 pipelines and 1500 executions). Each execution was 1 MB or
+   * more in size.
+   *
+   * <p>It can be further tuned, depending on your setup, since 150 executions work well for some
+   * applications but a higher number may be appropriate for others.
+   */
+  int maxNumberOfPipelineExecutionsToProcess = 150;
```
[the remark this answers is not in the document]
kirangodishala: This is a configurable value and matters only when `optimizeExecutionRetrieval = true` which by default is false. The idea is that this feature is enabled only when the pipelines and/executions are high.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
kirangodishala: changed to debug

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
dbyron-sf: I don't think so.  This is one of the great struggles of taking so long to get stuff pushed upstream.  Any data we may have had has vanished.  Plus also, we've since moved on to use the read replica for this, so collecting new data is gonna be tough.

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
christosarvanitis: Agree with @jasonmcintosh that especially in the TaskController there are a lot of log.info that will fill out the console :) would prefer to keep in info only what is absolutely necessary since TaskController is heavily called

christosarvanitis:
The whole idea seems like a good optimisation to me 👍 @dbyron-sf @kirangodishala how long have you been running this in your forks? 

Also could we fit the same concept of optimisations in `retrievePipelinesForPipelineConfigIdsBetweenBuildTimeBoundary` in this PR?

orca-web/src/main/java/com/netflix/spinnaker/config/TaskControllerConfigurationProperties.java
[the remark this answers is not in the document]
dbyron-sf: It would still be nice if the defaults are useful for more folks.  What do you suggest here @jasonmcintosh ?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -427,6 +451,129 @@ class SqlExecutionRepository(
...
+        )
+        .leftOuterJoin(PIPELINE.tableName.compressedExecTable).using(field("id"))
+        .where(
+          field("application").eq(application)
+            .and(field("id").`in`(*pipelineExecutions.toTypedArray()))
+        )
+        .queryTimeout(queryTimeoutSeconds) // add an explicit timeout so that the query doesn't run forever
+        .fetch()
+
+      log.info("getting stage information for all the executions found so far")
```
dbyron-sf, line 572: This doesn't seem valuable enough to leave at info.
```suggestion
      log.debug("getting stage information for all the executions found so far")
```

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -583,38 +589,51 @@ class TaskController {
...
+      allPipelineExecutions = rx.Observable.merge(allFront50PipelineConfigIds.collect {
+        log.debug("processing pipeline config id: $it")
+        executionRepository.retrievePipelinesForPipelineConfigId(it, executionCriteria)
+      }).subscribeOn(Schedulers.io()).toList().toBlocking().single()
+    }
+
+    allPipelineExecutions.sort(startTimeOrId)
     if (!expand) {
-      unexpandPipelineExecutions(allPipelines)
+      log.info("unexpanding pipeline executions")
```
dbyron-sf, line 631:
```suggestion
      log.debug("unexpanding pipeline executions")
```

orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -583,38 +589,51 @@ class TaskController {
...
+      unexpandPipelineExecutions(allPipelineExecutions)
     }
 
-    return filterPipelinesByHistoryCutoff(allPipelines, limit)
+    log.info("filtering pipelines by history")
```
dbyron-sf, line 635:
```suggestion
    log.debug("filtering pipelines by history")
```

orca-web/src/main/java/com/netflix/spinnaker/config/TaskControllerConfigurationProperties.java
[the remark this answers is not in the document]
jasonmcintosh: *ponder* so NORMAL recommendations on pool/concurrency size is tied to CPU's.  I think most people run on smaller instances with limited cores, and with other background processes... I'd just go to smaller defaults?  Probably maxExecutionRetrievalThreads set to 4 maybe.  *shrug* it's a hard one to judge - m5.2xlarge has only 8 cpus (which... i'll ignore hyperthreading) so you could hit some concurrency issues.  It's not IMPOSSIBLE... I just tend to vote "smaller, more of them" kinda stuff on this.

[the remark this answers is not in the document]
kirangodishala: Updated the defaults now

PR #4808, opened 2024-12-06
feat(pipeline/configuration): Updated pipeline level configuration that introduces a new property named metadata.

tonyg4864:
SpEL processing pipeline.metadata so that it is resolved/ready when referenced.

### DIFF
--- a/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/PipelineExecution.java
+++ b/orca-api/src/main/java/com/netflix/spinnaker/orca/api/pipeline/models/PipelineExecution.java
@@ -34,6 +34,10 @@ public interface PipelineExecution {
   @Nonnull
   ExecutionType getType();
 
+  Map<String, Object> getMetadata();
+
+  void setMetadata(Map<String, Object> metadata);
+
   String getId();
 
   void setId(String id);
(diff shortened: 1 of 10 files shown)


dbyron-sf:
What's the motivation for this change?

tonyg4864:
"I'm waffling on the need for a feature flag in case people already have metadata in their config and don't need/want/expect it to make it the execution." If you think this is necessary/mandatory I can add it , let me know. 

"It seems harmless, but increasing the size of executions can cause pain." I understand and I was also concerned about this. So far no issues in our workloads.

"The part I'm a little stuck on is the change to EvaluateVariablesStage. If you're saying that the idea is to be able to reference pipeline.metadata without an EvaluateVariablesStage, then why do we need a code change there?": It's the other way around. It is needed for the case where pipeline.metadata has SpEL expressions and pipeline.metadata is referenced in a EvaluateVariable stage. The changes in EvaluateVariableStage run pipeline.metadata through the SpEL processor so that it is resolved in case it is referenced somewhere in the stage.

dbyron-sf:
> "I'm waffling on the need for a feature flag in case people already have metadata in their config and don't need/want/expect it to make it the execution." If you think this is necessary/mandatory I can add it , let me know.

Yes please.

> "The part I'm a little stuck on is the change to EvaluateVariablesStage. If you're saying that the idea is to be able to reference pipeline.metadata without an EvaluateVariablesStage, then why do we need a code change there?": It's the other way around. It is needed for the case where pipeline.metadata has SpEL expressions and pipeline.metadata is referenced in a EvaluateVariable stage. The changes in EvaluateVariableStage run pipeline.metadata through the SpEL processor so that it is resolved in case it is referenced somewhere in the stage.

Aaah, so what something else references pipeline.metadata?  Is this where we get back to needed nested SpEL expression evaluation?  If so I'd rather pursue that than treating EvaluateVariablesStage as a special case.

PR #4820, opened 2025-01-06
Lazy loading with pipelineRef

edgarulg:
pipeline triggers contains the whole execution of the parent execution.  When a Spinnaker user defines multiple pipelines that are chained by pipeline triggers Spinnaker's UI is slowed down because it loads all nested executions.

Orca will return any execution to outsider services with PipelineRef trigger but for Orca itself it will use the whole execution context to evaluate SPeLs.

it will provide a new endpoint to retrieve the execution context including nested execution.


edgarulg:
closed in favor of: https://github.com/spinnaker/orca/pull/4842

PR #4824, opened 2025-01-16
perf(pipeline): Improve execution times for dense pipeline graphs

dzhengg:
There's a number of performance improvements here:
1. Memoize the anyUpstreamStagesFailed extension function to improve time complexity from exponential to linear
2. Optimize getAncestorsImpl to reduce time complexity by a factor of N, where N is the number of stages in a pipeline
3. Optimize StartStageHandler to only call withAuth (which calls getAncestorsImpl) when needed

Overall, these improvements reduce pipeline execution times across the board, with the biggest gains seen from very dense pipeline graphs such as the contrived example in ComplexPipeline.kt. After the optimizations, the timing of the test added to StartStageHandler improves from never completing to finishing in ~160ms on my machine.


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionInternals.java
```diff
@@ -41,15 +42,17 @@ static List<StageExecution> getAncestorsImpl(
...
+      Collection<String> requisiteStageRefIds = stage.getRequisiteStageRefIds();
       List<StageExecution> previousStages =
           stage.getExecution().getStages().stream()
-              .filter(it -> stage.getRequisiteStageRefIds().contains(it.getRefId()))
               .filter(it -> !visited.contains(it.getRefId()))
+              .filter(it -> requisiteStageRefIds.contains(it.getRefId()))
               .collect(toList());
       List<StageExecution> syntheticStages =
           stage.getExecution().getStages().stream()
+              .filter(s -> s.getSyntheticStageOwner() != null)
```
christosarvanitis, line 55: Was it  really going through all the stages before and not only the STAGE_BEFORE/AFTER?

dzhengg: Yes :(

christosarvanitis: That is a good catch 🚀 👏

christosarvanitis (approved):

PR #4833, opened 2025-02-11
feat(webhook): add http method + url-based allow list to webhook stages

dbyron-sf:
Configure with the following properties:
```
webhook.allowedRequestsEnabled (default: false)
webhook.allowedRequests: (default: empty list)
  - httpMethods: list of strings, e.g. [ "GET" ]
    urlPrefix: "https://my-url:port/path/starts/with"
```
Note, if allowRequestsEnabled is true, and allowedRequests is empty, no requests are allowed.

### DIFF
--- a/orca-webhook/orca-webhook.gradle
+++ b/orca-webhook/orca-webhook.gradle
@@ -32,8 +32,10 @@ dependencies {
   implementation("com.squareup.okhttp3:okhttp")
   implementation("com.jakewharton.retrofit:retrofit1-okhttp3-client:1.1.0")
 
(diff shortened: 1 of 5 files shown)


xibz (commented):
Overall this looks good, but I do see an interesting question of any method or any URI. An example may be we dont care about the uri prefix, we only care about all requests are GET. This change forces them both to be populated.

Im wondering if we need to handle for the individual use cases as well. Im fine approving, but wanted to at least see if it is something we want to do now, or until someone asks for it

xibz (approved):
Went ahead and approved. As I don't think the previous question is blocking. There's ways to do it with how it is designed right now.

PR #4837, opened 2025-02-15
feat(webhook): add request/response size limits

dbyron-sf:
via new config properties webhook.maxRequestBytes and webhook.maxResponseBytes. The default value for both is 0.
When greater than 0, only requests/responses whose header + body size is less than or equal to the max are allowed.

### DIFF
--- a/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
+++ b/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
@@ -40,8 +40,12 @@ import java.util.Map;
 import java.util.Optional;
 import javax.net.ssl.*;
 import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
 import okhttp3.Response;
 import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
@@ -68,6 +72,8 @@ import org.springframework.web.client.RestTemplate;
 public class WebhookConfiguration {
   private final WebhookProperties webhookProperties;
 
+  private final Logger log = LoggerFactory.getLogger(getClass());
+
   @Autowired
   public WebhookConfiguration(WebhookProperties webhookProperties) {
     this.webhookProperties = webhookProperties;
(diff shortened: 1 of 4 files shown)


orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
```diff
@@ -124,6 +138,127 @@ public ClientHttpRequestFactory webhookRequestFactory(
...
+    RequestBody requestBody = request.body();
+    if (requestBody == null) {
+      return;
+    }
+
+    long contentLength = 0L;
+    try {
+      contentLength = requestBody.contentLength();
+      if (contentLength == -1) {
+        // Given that OkHttp3ClientHttpRequestFactory.buildRequest (called by
```
xibz, line 168: I believe this is possible. I think this occurs for streaming, e.g. trying to send a request with a massive file returns content length == -1. See chunked encoding.

orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
```diff
@@ -124,6 +138,127 @@ public ClientHttpRequestFactory webhookRequestFactory(
...
+  /** Return if the request size is valid. Throw an exception otherwise. */
+  private void validateRequestSize(Request request, long maxRequestBytes) throws IOException {
+    if (maxRequestBytes <= 0L) {
+      return;
+    }
+
+    long headerSize = request.headers().byteCount();
+
+    // If the headers are already too big, no need to deal with the body size
+    if (headerSize > maxRequestBytes) {
```
xibz, line 150: I do like max payload size, but having a limitation on headers seems like that would already be handled by *MOST* http servers. Albeit they all default to some non-consistent value.

Is this more from a security stand point?

orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
xibz: Now, if webhooks can't do chunked encoding, then we can assume `-1` wont be the value.

orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
dbyron-sf: Yes, security, It's more from a quality of service / denial of service standpoint.  Like, we didn't want pipeline authors to be able to generate massive requests and bog down Spinnaker.

orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
dbyron-sf: I'll admit I haven't tried that hard to send a streamed request via a webhook stage.  I don't *think* it's possible though since the request is defined in the pipeline config.  Of course with SpEL expressions, etc. it can be dynamic and huge, but I still think it ends up being a String with a known length by the time we get here.

xibz (approved):

PR #4838, opened 2025-02-16
feat(webhook): add new webhook.followRedirects property

dbyron-sf:
Default to true so there's no behavior change. When false, don't follow redirects in webhook stages.

### DIFF
--- a/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
+++ b/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
@@ -114,11 +114,13 @@ public class WebhookConfiguration {
 
                   validateResponseSize(response, webhookProperties.getMaxResponseBytes());
 
-                  if (webhookProperties.isVerifyRedirects() && response.isRedirect()) {
-                    // verify that we are not redirecting to a restricted url
+                  if (response.isRedirect()) {
                     String redirectLocation = response.header("Location");
-                    if (redirectLocation != null && !redirectLocation.trim().startsWith("/")) {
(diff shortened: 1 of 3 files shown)


orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
```diff
@@ -114,11 +114,18 @@ public ClientHttpRequestFactory webhookRequestFactory(
 
                   validateResponseSize(response, webhookProperties.getMaxResponseBytes());
 
-                  if (webhookProperties.isVerifyRedirects() && response.isRedirect()) {
-                    // verify that we are not redirecting to a restricted url
+                  if (response.isRedirect()) {
```
xibz, line 117: So I see we have this `validateRequestSize` and `validateResponseSize`, would it make sense to move this to a method called `validateAnyRedirects`?

xibz (approved):
Very minor comment. Mostly for consistency

orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
dbyron-sf: We didn't decide the `isRedirect` method name here.  It's a method on the okhttp3.Response class.

We do get to decide property names though, and I figured `followRedirects` was consistent with the existing `verifyRedirects`.

dbyron-sf: Aaah, I see...We could move all this logic to a separate method for handling redirect validation...yes, that makes sense.

PR #4841, opened 2025-02-17
feat(queue): provide a way for pipeline stages to specify a backoff period

dbyron-sf:
via a new backoffPeriodMs stage configuration property. Before this, pipeline authors had no control over the backoff period. It came from either spinnaker configuration properties or implementations of RetryableTask.getDynamicBackoffPeriod.

This makes it possible for pipeline authors to, for example, control the delay between attempts when webhook stages retry. The actual backoff used is the largest of:

- backoffPeriodMs in the stage
- tasks.global.backOffPeriod
- tasks.<cloud provider>.backOffPeriod
- tasks.<cloud provider>.<account name>.backOffPeriod

### DIFF
--- a/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
+++ b/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
@@ -333,6 +333,9 @@ class RunTaskHandler(
       )
     )
 
+    val stageOverrideBackoffPeriod = stage.getBackoffPeriod()
+    stageOverrideBackoffPeriod.ifPresent { backOffs.add(it) }
+
     if (this is CloudProviderAware && hasCloudProvider(stage)) {
       backOffs.add(
         dynamicConfigService.getConfig(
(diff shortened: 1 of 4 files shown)


orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
```diff
@@ -316,6 +316,9 @@ class RunTaskHandler(
    * `tasks.aws.backOffPeriod = 80000`
    * `tasks.aws.someAccount.backoffPeriod = 60000`
    * `tasks.aws.backoffPeriod` will be used (given the criteria matches and unless the default dynamicBackOffPeriod is greater).
+   *
+   * This function also considers the `backoffPeriodMs` property in a stage
+   * configuration.  If it's larger than the above, it's used.
```
xibz, line 321: Im not following this comment. Where is the logic to determine what is used by magnitude?

xibz (approved):

orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt
xibz: Nvm. I see it in the code below that needed expanding in gh

PR #4842, opened 2025-02-18
feat(pipelineRef): Replace PipelineTrigger with PipelineRef for Spinnaker UI

edgarulg:
**Feature: Performance Optimization for Nested Pipelines in Spinnaker Orca**
In Spinnaker, pipelines with many nested child pipelines (e.g., Pipeline → Pipeline → Pipeline → Pipeline) can cause significant delays in loading times, particularly for large instances. This is because the Spinnaker UI often needs to load parent executions for any pipeline that was triggered by another, which unnecessarily increases the load on the Orca execution repository.

**Solution:**
To address this issue, I’ve introduced an optimization using the `PipelineRef` feature that was previously delivered. The idea is to propagate the `PipelineRef trigger` instead of the full `PipelineTrigger` for nested pipelines. This reduces the load on Orca and improves UI performance.

**How It Works:**

   **ExecutionRepository Update:** 
I added a new flag, `includeNestedExecutions`, to the retrieve method of the ExecutionRepository interface.

   **Default Behavior:** 
Existing execution repository implementations don’t need to change. 

-    The default behavior remains to use the normal retrieve method.

   **SqlExecutionRepository Logic:**
The only repository that requires an update is the `SqlExecutionRepository`, where the business logic for `includeNestedExecutions` is implemented.

- If includeNestedExecutions is true, Orca will convert any `PipelineRefTrigger` into a `PipelineTrigger` to ensure nested executions are returned.

- If the flag is false, Orca will return executions with the `PipelineRefTrigger`.

**SPeL Compatibility:**
To maintain backward compatibility with Spinnaker’s SPeL, the `OrcaMessageHandler` ensures that nested executions are included when evaluating any expressions. This guarantees that SPeL expressions still resolve correctly.

**SPeLAutoComplete Compatibility:**
The `SPeLAutoComplete` feature continues to work as expected. The endpoint that retrieves previous executions to feed the autocomplete is still retrieving executions with the full execution context, including nested executions. This ensures that the autocomplete functionality is unaffected by the performance improvement.

**Summary:**
The main idea is to minimize unnecessary load on Orca by returning a PipelineRefTrigger for external requests, while keeping full execution context for internal modules that need it. This change significantly reduces the pressure on the Orca execution repository and improves UI performance, especially in large Spinnaker instances.

This performance improvement is backward compatible, and SPeLAutoComplete continues to function as expected, with no disruption to existing Spinnaker functionality.

Here is an example on how the Spinnaker UI looks like when a pipeline is triggered by another pipeline:
Captura de pantalla 2025-02-17 a la(s) 8 26 28 p m

Here is another example on how the execution looks like if we inspect it in the Spinnaker UI:
Captura de pantalla 2025-02-17 a la(s) 8 26 50 p m

Finally here is another example that probes the SPeLAutoComplete still works by converting PipelineRefTrigger to PipelineTrigger and existing SPeLs still works as expected:
Captura de pantalla 2025-02-17 a la(s) 8 31 20 p m

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
@@ -72,6 +72,14 @@ public interface ExecutionRepository {
   PipelineExecution retrieve(@Nonnull ExecutionType type, @Nonnull String id)
       throws ExecutionNotFoundException;
 
+  @Nonnull
+  default PipelineExecution retrieve(
+      @Nonnull ExecutionType type, @Nonnull String id, boolean includeNestedExecutions)
+      throws ExecutionNotFoundException {
+    // Default behavior: ignore `includeNestedExecutions`
+    return retrieve(type, id);
+  }
+
   void delete(@Nonnull ExecutionType type, @Nonnull String id);
 
   void delete(@Nonnull ExecutionType type, @Nonnull List<String> idsToDelete);
@@ -91,6 +99,15 @@ public interface ExecutionRepository {
   @Nonnull
   Observable<PipelineExecution> retrievePipelinesForPipelineConfigId(
       @Nonnull String pipelineConfigId, @Nonnull ExecutionCriteria criteria);
+
+  @Nonnull
+  default Observable<PipelineExecution> retrievePipelinesForPipelineConfigId(
+      @Nonnull String pipelineConfigId,
+      @Nonnull ExecutionCriteria criteria,
+      Boolean includeNestedExecutions) {
+    // Default behavior: ignore `includeNestedExecutions`
+    return retrievePipelinesForPipelineConfigId(pipelineConfigId, criteria);
+  }
 
   @Nonnull
   Collection<String> retrievePipelineConfigIdsForApplication(@Nonnull String application);
--- a/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/ExecutionMapper.kt
+++ b/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/ExecutionMapper.kt
@@ -69,7 +69,7 @@ class ExecutionMapper(
     }
   }
 
-  fun map(rs: ResultSet, context: DSLContext): Collection<PipelineExecution> {
+  fun map(rs: ResultSet, context: DSLContext, includeNestedExecutions: Boolean): Collection<PipelineExecution> {
     val results = mutableListOf<PipelineExecution>()
     val executionMap = mutableMapOf<String, PipelineExecution>()
     val legacyMap = mutableMapOf<String, String>()
@@ -80,7 +80,7 @@ class ExecutionMapper(
         mapper.readValue<PipelineExecution>(body)
           .also {
             execution ->
-            convertPipelineRefTrigger(execution, context)
+            convertPipelineRefTrigger(execution, context, includeNestedExecutions)
             execution.setSize(body.length.toLong())
             results.add(execution)
             execution.partition = rs.getString("partition")
@@ -138,9 +138,9 @@ class ExecutionMapper(
   }
 
   @VisibleForTesting
-  fun convertPipelineRefTrigger(execution: PipelineExecution, context: DSLContext) {
+  fun convertPipelineRefTrigger(execution: PipelineExecution, context: DSLContext, includeNestedExecutions: Boolean) {
     val trigger = execution.trigger
-    if (trigger is PipelineRefTrigger) {
+    if (trigger is PipelineRefTrigger && (!pipelineRefEnabled || includeNestedExecutions)) {
       val parentExecution = fetchParentExecution(execution.type, trigger, context)
 
       if (parentExecution == null) {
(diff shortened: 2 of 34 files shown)


edgarulg:
I will add more tests around TaskController and endpoints.

edgarulg:
I added all tests I have in mind for this feature. This is ready now.

christosarvanitis:
@dbyron-sf Since the 1.37.x has been released do you think we can start the review process for this?

orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java
```diff
@@ -92,6 +100,15 @@ Observable<PipelineExecution> retrieve(
   Observable<PipelineExecution> retrievePipelinesForPipelineConfigId(
       @Nonnull String pipelineConfigId, @Nonnull ExecutionCriteria criteria);
 
+  @Nonnull
+  default Observable<PipelineExecution> retrievePipelinesForPipelineConfigId(
+      @Nonnull String pipelineConfigId,
+      @Nonnull ExecutionCriteria criteria,
+      Boolean includeNestedExecutions) {
```
christosarvanitis, line 107: Possible `boolean` instead of `Boolean`?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/ExecutionMapper.kt
```diff
@@ -69,7 +69,7 @@ class ExecutionMapper(
     }
   }
 
-  fun map(rs: ResultSet, context: DSLContext): Collection<PipelineExecution> {
+  fun map(rs: ResultSet, context: DSLContext, includeNestedExecutions: Boolean): Collection<PipelineExecution> {
```
christosarvanitis, line 72: Same here:  `boolean` since `Boolean` doesnt seem to be needed? Unless there is a specific reason im missing

christosarvanitis (commented):
Thanks @edgarulg I have tested this PR and it looks good functionality wise! This will boost the heavy nested pipelines performance by a lot 🚀 

What is your view on adding an additional configuration/feature flag on the pipelineRef feature to control the includeNestedPipelines flag? 

Im thinking that a user might want to keep the reconstruction of the PipelineRef in Orca as it was first implemented (for whatever reason). Others might want to have the Ref which is done via this PR. 
Or maybe we discover a bug in the feature and instead of rolling back the release we just disable this added functionality

edgarulg:
@christosarvanitis Thanks for the review. I tried to put this improvement behind a feature flag to allow for easy disabling or rollback in case of issues. However, since my changes update functionality in orca-core, fully encapsulating them is challenging but not impossible. Doing so would require additional changes to my PR that are not directly related to this improvement. Additionally, my changes are already substantial, making the review process more complex. :( 

However, I remember I hear there is a initiative to make more easy to write experimental code behind a feature flag but not sure if we can use that for this or what is the state for that work

edgarulg:
@dbyron-sf @jasonmcintosh Can I get a review in my PR please?

dbyron-sf:
I'm still struggling to get 1.37.x working...

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -419,8 +423,8 @@ class SqlExecutionRepository(
     return retrieve(type, criteria, null)
   }
 
-  private fun retrieve(type: ExecutionType, criteria: ExecutionCriteria, partition: String?): Observable<PipelineExecution> {
-    withPool(readPoolName) {
+  private fun retrieve(type: ExecutionType, criteria: ExecutionCriteria, partition: String?, includeNestedExecutions: Boolean = false): Observable<PipelineExecution> {
+    withPool(poolName) {
```
mattgogerly, line 427: What's the rationale for this change? I would have thought we would want to be using the read pool here?

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
```diff
@@ -605,13 +609,21 @@ class SqlExecutionRepository(
...
-      return ExecutionMapper(mapper, stageReadSize,compressionProperties, pipelineRefEnabled).map(baseQuery.intoResultSet(), jooq)
+      return ExecutionMapper(mapper, stageReadSize,compressionProperties, pipelineRefEnabled).map(baseQuery.intoResultSet(), jooq, false)
     }
   }
 
   override fun retrievePipelinesForPipelineConfigId(
     pipelineConfigId: String,
     criteria: ExecutionCriteria
+  ): Observable<PipelineExecution> {
+  return retrievePipelinesForPipelineConfigId(pipelineConfigId, criteria, false)
```
mattgogerly, line 620: nit: indentation

mattgogerly (commented):
Might just be me, but I'm finding it hard to follow the passing of `true`/`false` everywhere. I wonder if it would be easier to read if we used dedicated methods for includingNestedExecutions.

orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
jasonmcintosh: Yeah I SUSPECT this was a confusion on this.  There's a risk possibly on chained executions on concurrency but... still think should be readPoolName on this one...

edgarulg: This is a mistake I did in the code, the right pool should be readPoolName.

PR #4846, opened 2025-02-22
fix(artifacts): fix NPE when expected artifact is null 

kirangodishala:
This is to address the issue : https://github.com/spinnaker/spinnaker/issues/7004

While resolving expected artifact NullPointerException is thrown if the artifact is null. 

Added a test to demonstrate the issue and provided the fix which identifies null artifact and throws InvalidRequestException which is a SpinnakerException.

### DIFF
--- a/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
+++ b/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
@@ -120,6 +120,9 @@ public final class ArtifactResolver {
     ImmutableList.Builder<ExpectedArtifact> boundExpectedArtifacts = ImmutableList.builder();
 
     for (ExpectedArtifact expectedArtifact : expectedArtifacts) {
+      if (expectedArtifact == null) {
+        throw new InvalidRequestException("expected artifact cannot be null.");
+      }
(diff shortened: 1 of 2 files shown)


orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java
```diff
@@ -120,6 +120,9 @@ public ResolveResult resolveExpectedArtifacts(Iterable<ExpectedArtifact> expecte
     ImmutableList.Builder<ExpectedArtifact> boundExpectedArtifacts = ImmutableList.builder();
 
     for (ExpectedArtifact expectedArtifact : expectedArtifacts) {
+      if (expectedArtifact == null) {
```
xibz, line 123: eh? Why would an expected artifact be null? Is this due to clouddriver failing image resolution and deciding to return null instead of an error?

kirangodishala: It is not supposed to be null but looks like someone got NPE somehow(ref: https://github.com/spinnaker/spinnaker/issues/7004). We will investigate the reason but I believe allowing NPE in any case is bad, so this PR is to address those edge cases.

PR #4773, opened 2024-08-12
feat(batchUpdate): Pipeline config batch update

richard-timpson:
Change is dependent on https://github.com/spinnaker/front50/pull/1483

Added code to save multiple pipelines at once to sql database.

This is part of: spinnaker/spinnaker#6147.

Enhanced SavePipelineTask.java to

Added code to ensure that the SavePipelineTask.java also accepts list of pipelines json.
This method will validate all the pipelines.
This method will call the front50 service to save the pipelines list.

Enhanced Front50Service.groovy to

Added new rest api which accepts list of pipelines json.

Also contains some changes to make the timeout values in the front50 client configurable.  They all default to 10000:
* front50.okhttp.readTimeoutMs
* front50.okhttp.writeTimeoutMs
* front50.okhttp.connectTimeoutMs

PR #4849, opened 2025-04-02
fix(okhttpconfig): increase the defaul timeout due to low number

shlomodaari:


dbyron-sf:
There's enough stuff going on with retrofit2 upgrades that I'd love to leave this code alone.  Isn't changing your local config sufficient?

jasonmcintosh:
Sooo what we hit recently:
* OKHTTP timeouts are normally set via kork's common okhttp configurations.
* The bulk upload stopped using the service locator lookups which USES those okhttp for it's OWN configuration by service
* This then caused failures where the okhttp global timeouts which were high... suddenly no longer applied and these front50 operations timed out with lower thresholds.

SO the bulk PR ... broke some things rather badly b/c of this.  I'm not sure changing these defaults high is the answer... vs. adding a bit more logic.  A few ideas:

* "OK if these are SET, use the set values, otherwise default to okhttp".  
* I could also see "use whichever is higher... MAYBE.  
* Alternative would be use Integer, no defaults set, and check if the okhttps are set, and if not, use the 10 seconds instead via a spel expression to allow an override for a specific service.

dbyron-sf:
Which bean is bulk upload?  Can you provide a link to the code that creates it?  From there perhaps @kirangodishala can take a look and see if there's some way to restore the ability to configure it independently?

shlomodaari:
I will close this one the go ahead with Jason idea! 
"OK if these are SET, use the set values, otherwise default to okhttp".
If you both okay with it

jasonmcintosh:
https://github.com/spinnaker/orca/pull/4773 was what broke things this way as it overloaded the okhttp settings for front50 calls WITHOUT considering okhttp global defaults.

jasonmcintosh:
NOTE right now kinda thinking of a "int to Integer" then doing
```@Value("${front50.okhttp.readTimeoutMs != null ? front50.okhttp.readTimeoutMs : (ok-http-client.readTimeoutMs != null ? ok-http-client.readTimeoutMs : 10)}")``` or similar kinda thing

shlomodaari:
https://github.com/spinnaker/orca/pull/4851
I created this new one based on Jason input

PR #4851, opened 2025-04-02
Fix(orca-front50): front50 timeout fallback

shlomodaari:
**Issue**

PR #4773 (Pipeline config batch update) introduced intermittent timeout failures in pipeline operations. After investigation, we discovered that the PR implemented its own timeout handling that bypassed the global OkHttp client configuration:

It used hardcoded and relatively short timeout values (10 seconds) for read/write/connect operations
These timeouts were insufficient for batch pipeline operations under load
Authentication failures occurred when timeout was too short for authorization checks to complete

**Solution**
This PR implements a robust timeout fallback chain that maintains backward compatibility while fixing the timeout issues:

**First priority**: Use explicit Front50-specific timeouts when configured in YAML
**Second priority**: Fall back to global OkHttp client timeouts (from Kork)
**Last resort**: Use conservative default values (60s read/write, 10s connect)

**Implementation Details**

Added clear fallback chain logic in Front50Configuration.groovy
Enhanced Front50ConfigurationProperties.java with better defaults and documentation
Added hasCustomTimeouts() helper method to detect explicit configuration
Added detailed DEBUG logging for which timeout source is being used
Added comprehensive unit tests for the fallback chain logic

**Testing**

Verified all unit tests pass, including batch pipeline operations
Confirmed the fallback chain correctly selects timeouts based on priority
Ensured changes don't break existing functionality


dbyron-sf:
I'd like @kirangodishala to take a look at this before merging.

kirangodishala:
@shlomodaari - the timeout values are configurable at present as well. 10s is the default value but using the following configuration, you can set them to any desired value. Have you tried this?

```yaml
front50:
   okhttp:
      readTimeoutMs: <> 
      writeTimeoutMs: <>
      connectTimeoutMs: <>
```

shlomodaari:
Hi @kirangodishala, 
Yes, we're aware of the configuration options. The issue we discovered is more subtle:

After investigating intermittent auth failures with batch pipeline updates, we identified that the configuration wasn't being reliably applied in all code paths. This PR fixes that by:

1. **Creating a clear fallback chain**:
   - First tries user-specified Front50 config
   - Then falls back to global OkHttp settings 
   - Finally uses safe defaults (60s read/write, 10s connect)

2. **Properly detecting explicit configuration** vs default values to prevent scenarios where default values accidentally override global settings

3. **Making behavior consistent** across all Front50 client operations rather than having different behaviors depending on code path

The 10-second default was especially problematic for batch operations with multiple pipelines. This implementation maintains backward compatibility while ensuring explicitly configured timeouts are always honored.

I've added detailed Javadoc comments to better document this behavior, which should have been documented from the beginning. This PR ensures that existing configuration works as expected and behaves predictably.

shlomodaari:
Yes, we're aware of the configuration options, but we weren't aware of how they were implemented until we had an incident.

The change in PR #4773 wasn't documented as a breaking change in the release notes. We only discovered these configuration options after experiencing the incident, and upon investigation, we found that the 10-second defaults were actually overriding our global OkHttp client timeouts (which were higher), causing intermittent auth failures with batch pipeline updates.


This implementation maintains your changes but makes them more suitable and efficient. It ensures backward compatibility while preventing timeouts from accidentally overriding longer global timeout values.

I've added detailed Javadoc comments to better document this behavior.

dbyron-sf:
> The change in PR https://github.com/spinnaker/orca/pull/4773 wasn't documented as a breaking change in the release notes.

@shlomodaari Can you make a PR to the release notes to add something about this please?

shlomodaari:
@dbyron-sf I definitely will create a PR tomorrow during my morning :)

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/config/Front50Configuration.groovy
```diff
@@ -75,23 +84,91 @@ class Front50Configuration {
...
+   * 3. Default value
+   */
+  private long getEffectiveTimeout(Integer explicitTimeout, long globalTimeout, long defaultTimeout, String timeoutType) {
+    // First check if the explicit timeout is non-null and has been explicitly configured
+    // to something other than the default
+    if (explicitTimeout != null) {
+      // Check if the value differs from default configuration values
+      boolean isDefaultReadValue = explicitTimeout == 60000 && timeoutType == "read"
+      boolean isDefaultWriteValue = explicitTimeout == 60000 && timeoutType == "write"
+      boolean isDefaultConnectValue = explicitTimeout == 10000 && timeoutType == "connect"
```
kirangodishala, line 155: Instead of hardcoded values here, using the constants defined (i.e. DEFAULT_READ_TIMEOUT_MS etc) would be more meaningful I guess.

shlomodaari: I address this issue :) 
Let me know if anything else should be addressed @kirangodishala

dbyron-sf: I still don't understand enough about the issue.  I need to see that release note.  What was the behavior before?  What is the new behavior?  Would we be better off reverting something?  This PR is filled with so many hard-coded magic numbers and special logic that I'd really love to find a different way.

shlomodaari: The PR #4773 introduced batch pipeline updates which is valuable functionality. Our changes preserve this functionality while ensuring reliable timeout handling.

What our implementation does:
- Creates a clear, documented fallback chain for timeout values
- Properly respects both explicit and global configuration
- Maintains backward compatibility for all users
- Adds proper logging to help debugging

The approach we've taken uses the existing configuration capabilities while making the behavior more consistent and predictable. The special logic is necessary to handle different configuration scenarios while ensuring optimal performance.

With our implementation, users have flexibility to:
1. Configure Front50-specific timeouts when needed
2. Rely on global configuration when appropriate  
3. Benefit from sensible defaults otherwise

The tests we've added verify all these scenarios work correctly. We believe this implementation strikes the right balance between functionality, reliability, and backward compatibility.

shlomodaari: I don't see any issue with not going forward with this PR and let's make sure next if any breaking change let's at least add some tests! Thank you! 
Working on the release notes in the next hour.

dbyron-sf: That may be true, but this code is super brittle.  I still need to understand more.  Again, what was it like before? What did https://github.com/spinnaker/orca/pull/4773 change?  Why is it valuable? I'd like to consider some other alternatives before going ahead with this.

jasonmcintosh: SO the previous code used a locator in kork which loaded default okhttp configurations.  Which had timeouts of 60 seconds.  4773 overloaded those defaults to be 10 seconds and ignore kork set defaults.

jasonmcintosh: this led to breakage where SUDDENLY ... front50 timeouts did NOT use the settings from kork's okhttp config, and now use some custom front50 specific config... with defaults SIGNIFICANTLY lower.  AND... if you set the okhttp configs... those configs wouldn't apply here.  This is why the argument of the timeout logic should be something like:
* IF per service timeouts are expicitly set, per service timeouts should have a higher precedence.
* IF Not explicitly set, default to the okhttp timeouts

shlomodaari: @dbyron-sf please review this change in the doc: https://github.com/spinnaker/spinnaker.io/pull/504
If it looks good I will do the same for the rest of the versions that are impacted by the change! 

@jasonmcintosh  @kirangodishala please let me know if you think that I need to change anything else in the PR! 🙏🏽

shlomodaari: @kirangodishala @dbyron-sf do you have any update? 🙏🏽

shlomodaari: @dbyron-sf @kirangodishala I wonder if you had any time to review this?

dbyron-sf: @shlomodaari Not sure why spinnakerbot closed https://github.com/spinnaker/spinnaker.io/pull/504.  I re-opened it.  Can you take a look at https://github.com/spinnaker/spinnaker.io/pull/504#discussion_r2038326708?

shlomodaari: Yes, I will take a look at this. Do we have an ETA for reviewing this PR as well?

dbyron-sf: OK, thanks for updating https://github.com/spinnaker/spinnaker.io/pull/504.  I think I understand it now.  One benefit of that PR is that it lets people configure different timeouts when orca communicates with front50 than the "generic" ok-http-client timeouts, but I agree it was a breaking change, and it'd be great to make it less breaking, or not breaking at all.

We could help one class of users (the ones that didn't specify ok-http-client timeouts / used the defaults there), but making the defaults in Front50ConfigurationProperties.OkHttpConfigurationProperties match the defaults in OkHttpClientConfigurationProperties.

If we did that, users that specified something for ok-http-client still see a change in behavior.  But isn't it relatively straightforward to fix that?  By adding an OkHttpClientConfigurationProperties argument to Front50Configuration.front50Service and doing something like https://github.com/spinnaker/orca/pull/4839/commits/f12cd1fbc008d803c2bbc161ce0815c7b60b38f1, I think we'd get there.

shlomodaari: Thanks for your feedback on spinnaker/spinnaker.io#504!

In PR #4851, we've implemented a solution that handles all three user scenarios with clear priority order:

1. If a user has default OkHttp settings (no custom configuration):
   - The default OkHttp client settings (60s connect, 300s read) are applied to Front50
   - This maintains the behavior for users who didn't customize anything

2. If a user has custom global OkHttp settings:
   - These custom global settings are applied to Front50
   - For example, if they configured ok-http-client.readTimeoutMs=90000, Front50 will use that value
   - This fixes the regression for users who relied on global settings

3. If a user has Front50-specific settings configured:
   - The Front50-specific settings override any global settings (default or custom)
   - For example, if they set front50.okHttp.readTimeoutMs=120000, that takes precedence
   - This preserves the new functionality 

The priority order is clear:
1. Front50-specific settings (highest priority)
2. Custom global OkHttp settings (middle priority)
3. Default OkHttp settings (lowest priority)

This approach ensures backward compatibility while preserving the ability to have service-specific configuration when needed.

dbyron-sf: This PR is too complicated/brittle.  I'd like to see a PR that has no additional "magic numbers" in it.  It still seems to be that we could pull this off with two commits:
* change the defaults in Front50ConfigurationProperties.OkHttpConfigurationProperties to match the defaults in OkHttpClientConfigurationProperties
* add an OkHttpClientConfigurationProperties argument to Front50Configuration.front50Service and do something like https://github.com/spinnaker/orca/commit/f12cd1fbc008d803c2bbc161ce0815c7b60b38f1.  As in, if the front50-specific properties are configured, use them.  Otherwise, fall back to ok-http-client.

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/config/Front50ConfigurationSpec.groovy
```diff
@@ -0,0 +1,113 @@
...
+import com.netflix.spinnaker.okhttp.OkHttpClientConfigurationProperties
+import retrofit.Endpoint
+import spock.lang.Specification
+import spock.lang.Subject
+import java.util.concurrent.TimeUnit
+
+/**
+ * Tests for Front50 timeout configuration
+ */
+class Front50ConfigurationSpec extends Specification {
```
dbyron-sf, line 31: Would you mind converting this to java please.  The less groovy code we have in the world, the better.

shlomodaari: Yes, let me know if it looks better now :)

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/config/Front50ConfigurationProperties.java
```diff
@@ -19,6 +19,30 @@
 import lombok.Data;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
+/**
```
dbyron-sf, line 22: I can't decide whether I think this comment is helpful enough vs. the chances of it getting stale.  Since these are the higher-precedence properties, I'm pretty tempted to ditch the comment.

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/config/Front50ConfigurationProperties.java
```diff
@@ -34,14 +58,33 @@ public class Front50ConfigurationProperties {
...
+  /**
+   * Configuration properties for the OkHttp client connecting to Front50. These will only be used
+   * if explicitly set in the configuration. Otherwise, global client timeouts will be used as
+   * fallback.
+   */
   @Data
   public static class OkHttpConfigurationProperties {
-    int readTimeoutMs = 10000;
+    /** Read timeout in milliseconds. Default is 120 seconds (120000ms) */
+    private Long readTimeoutMs = 120000L;
```
dbyron-sf, line 72:
```suggestion
    private long readTimeoutMs = 120000L;
```

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/config/Front50ConfigurationProperties.java
```diff
@@ -34,14 +58,33 @@ public class Front50ConfigurationProperties {
...
+
+    /** Write timeout in milliseconds. Default is 60 seconds (60000ms) */
+    private Long writeTimeoutMs = 60000L;
 
-    int writeTimeoutMs = 10000;
+    /** Connection timeout in milliseconds. Default is 5 seconds (5000ms) */
+    private Long connectTimeoutMs = 5000L;
```
dbyron-sf, line 78:
```suggestion
    private long connectTimeoutMs = 5000L;
```

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/config/Front50ConfigurationProperties.java
```diff
@@ -34,14 +58,33 @@ public class Front50ConfigurationProperties {
...
 
-    int connectTimeoutMs = 10000;
+    /**
+     * Checks if this instance has any custom timeout configuration.
+     *
+     * @return true if any timeout is non-default, false otherwise
+     */
+    public boolean hasCustomTimeouts() {
```
dbyron-sf, line 85: Don't think we need this method.

orca-front50/src/test/groovy/com/netflix/spinnaker/orca/front50/config/Front50ConfigurationSpec.groovy
dbyron-sf: Still looks like groovy to me...

orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/config/Front50Configuration.groovy
```diff
@@ -74,17 +79,41 @@ class Front50Configuration {
...
+    // Get base client with global configuration
+    OkHttpClient baseClient = clientProvider.getClient(new DefaultServiceEndpoint("front50", front50Endpoint.getUrl()))
+    OkHttpClient.Builder builder = baseClient.newBuilder()
+    
+    // Apply global timeouts first
+    builder.connectTimeout(okHttpClientConfigurationProperties.getConnectTimeoutMs(), TimeUnit.MILLISECONDS)
+    builder.readTimeout(okHttpClientConfigurationProperties.getReadTimeoutMs(), TimeUnit.MILLISECONDS)
+    
+    // Override with Front50-specific timeouts if explicitly defined
+    if (front50ConfigurationProperties.okhttp?.connectTimeoutMs != null) {
```
dbyron-sf, line 97: Even with capital-L long, I don't think this way of checking works, since the properties have default values.  I think we gotta do the `environment.containsProperty` thing.  There's a failing test, but I'm not sure it's failing because of this...so perhaps we need some more test coverage like [WebhookConfigurationTest](https://github.com/spinnaker/orca/commit/f12cd1fbc008d803c2bbc161ce0815c7b60b38f1#diff-37791a04535b606ac591093e90a70595f5eae99e9fe54e72e348280264d8bc0f).

For me the code is easier to follow using local variables and then calling the builder once, like:
```
    long readTimeoutMs =
         (environment.containsProperty("webhook.readTimeoutMs")
                 || environment.containsProperty("webhook.read-timeout-ms"))
             ? webhookProperties.getReadTimeoutMs()
             : okHttpClientConfigurationProperties.getReadTimeoutMs();
    requestFactory.setReadTimeout(Math.toIntExact(readTimeoutMs));
  ```

PR #4864, opened 2025-05-04
feat(Openshift): Add group owner to deploy on openshift

yanjunding:
https://docs.redhat.com/en/documentation/openshift_container_platform/4.16/html/images/creating-images#use-uid_create-images

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -4,7 +4,7 @@ RUN apk --no-cache add --update bash curl openjdk17-jre
 RUN addgroup -S -g 10111 spinnaker
 RUN adduser -S -G spinnaker -u 10111 spinnaker
 COPY orca-web/build/install/orca /opt/orca
-RUN mkdir -p /opt/orca/plugins && chown -R spinnaker:nogroup /opt/orca/plugins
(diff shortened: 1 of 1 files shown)


Dockerfile.slim
```diff
@@ -4,7 +4,7 @@ RUN apk --no-cache add --update bash curl openjdk17-jre
 RUN addgroup -S -g 10111 spinnaker
 RUN adduser -S -G spinnaker -u 10111 spinnaker
 COPY orca-web/build/install/orca /opt/orca
-RUN mkdir -p /opt/orca/plugins && chown -R spinnaker:nogroup /opt/orca/plugins
+RUN mkdir -p /opt/orca/plugins && chown -R spinnaker:0 /opt/orca/plugins
```
jasonmcintosh, line 7: Group 0 is the wheel group normally and probably not ideal to be added to.  The intent here was not thave any groups on plugins other than the spinnaker owner itself... I don't have an OS cluster handy to test, but... I'd wonder about this change...

jasonmcintosh: To note, these PR's should probably go now to https://github.com/spinnaker/spinnaker

